PR #202, opened 2014-12-02
added openmp for the MC Loop

mklauser:
This PR provides an openmp implementation for tardis

TODO

- [ ] making sure that the code is thread-safe


wkerzendorf:
@mklauser What is missing here?

PR #211, opened 2014-12-11
[WIP] restructure of the model Framework

wkerzendorf:
This is my idea of a restructure of the model framework. Currently it showcases an easy way to attach different attributes to models and link them (`density0` will be scaled to density when accessing `.density`). This will make it easy to store the models in dataformats like HDF5, make them more modular. This also includes some documentation (to see how they work) and tests

**TODO**
- [x] preliminary overview with some documentation and testing
- [ ] implement docstrings for the attributes
- [ ] implement HDF5 reader writer


wkerzendorf:
Will continue on this when #215 is merged

unoebauer:
@wkerzendorf: why exactly was this closed? Is the idea to have a completely new go at a model restructure in the future (at some point) or have we abandoned the idea entirely?

wkerzendorf:
@unoebauer - currently this PR has some good suggestions, but is not currently being worked on. As closing only means that it disappears from the list of active PRs and can be reopened, I closed it for now. But the idea is to built a better model in the future, just not right now.

PR #212, opened 2014-12-12
Testing the C-part of tardis

wkerzendorf:
I'm in the process of making a framework that will test the currently untested C-part of TARDIS (the search sorted incident, was a little bit of a wake-up call). 

So @ssim, @mklauser and @orbitfold I need your help - in a couple of ways:

After playing around with C unit testing frameworks and realizing that this is more difficult to integrate into `py.test` (our current testing framework), I think we should wrap each function (this takes a minute) and then test them individually. Now there are still some problems, that maybe either @mklauser or @orbitfold can help with:

We want to make one or more mock rpackets and the question is how do we translate them from python to c and back. Maybe something like this: http://markflorisson.wordpress.com/2011/08/23/memoryviews/?


tardis/montecarlo/tests/test_packet.py
```diff
@@ -0,0 +1,27 @@
...
+     'next_line_id': 1,
+     'next_shell_id': 1,
+     'nu': 1.0,
+     'nu_line': 1.0,
+     'r': 1.0,
+     'recently_crossed_boundary': 0,
+     'tau_event': 1.0,
+     'virtual_packet': 1,
+     'virtual_packet_flag': 1}
+    return packet_test_wrapper.RPacket(rpacket_dict)
```
kaushik94: @wkerzendorf  just my two cents, could be related to [this](http://www.psychicorigami.com/2007/06/21/python-ctypes-illegal-instruction-and-garbage-collected-callbacks/)

wkerzendorf: Just looked at this - maybe that's an interesting solution.

wkerzendorf:
Part of GSOC 2015. Closing for now

PR #222, opened 2015-03-04
Creates a webapp for creating config files

utkbansal:
* Creates a basic form to create the Supernova config


utkbansal:
Why does it say that it could not get flask-wtf? It works fine on my  system.

wkerzendorf:
it should work with anaconda, that is what is currently used for travis. See what anacondas package for flask is.

utkbansal:
it is 'flask' as mentioned here - http://docs.continuum.io/anaconda/pkg-docs.html

wkerzendorf:
then you need to change the pip-requirements file (better yet rename it to `conda_requirements.txt` and change the `.travis.yml` file

conda-requirements.txt
```diff
@@ -0,0 +1,41 @@
+# This file may be used to create an environment using:
```
wkerzendorf: just use the old `pip-requirements` file for this and put the things that don't exist in anaconda into the `pip-requirements` file

wkerzendorf: i  see you changed this. one last thing get rid of any py27_0

PR #223, opened 2015-03-04
Update git_workflow.rst

ujjwalwahi:
Change directory name from astropy to tardis in workflow document


wkerzendorf:
that looks great thanks! Will merge once it builds

PR #224, opened 2015-03-04
Web Interface Mockup

mavidser:
This is a basic mockup of a web-interface for generating YAML config files.

The main purpose is to identify the appropriate input methods for each field, verify the inputs which are conditional (the inputs with the type container-property), and figure out an easy way to gather the user inputs.

~~A live version of the mockup is available [here](https://aqueous-chamber-6869.herokuapp.com/).~~

The current version of the web app is available [here](https://whispering-atoll-6188.herokuapp.com/).


wkerzendorf:
@mavidser it looks nice, but is not in python. We'd prefer something like this, but in `tornado` or `flask`

wkerzendorf:
in the end it would be great if it can read the schema file and autogenerate this.

mavidser:
Yes, it's kinda like the expected output for the schema I hope to implement.
I'll try to create a schema in Python to generate a similar interface .

mavidser:
Hi.
In the latest commit, I tried to generate the form using the YAML Schema [given here](http://tardis.readthedocs.org/en/latest/configuration/configuration.html). 

The form isn't complete yet, and I still have to parse `container-property` (hence, Model and MonteCarlo hasn't been implemented), but I am facing some problems understanding its properties (see #239).

You can check out the form currently generated [here](https://whispering-atoll-6188.herokuapp.com/).

mavidser:
@wkerzendorf @orbitfold 

I think the form works fairly good for now, given a few kinks, which I'll sort out in further updates. ([Preview](https://whispering-atoll-6188.herokuapp.com/)). Please take a look and see if there's something I'm doing wrong, or any suggestions you might have regarding the form.

wkerzendorf:
@mklauser is one of our mentors and so I'm messaging him here as well.

web/lib/gallifrey.py
```diff
@@ -0,0 +1,92 @@
+import json
```
wkerzendorf: nice file name :wink:

wkerzendorf:
Can you rebase onto the newest master. It looks very nice! Please also start writing up the application.

mavidser:
It's done. I'll start with the application now too. Thanks.  :smile:

web/lib/gallifrey.py
mavidser: Just following the pattern. :smile:

Issue #228, opened 2015-03-05
Error while running pytest coverage


vidit23:
When in the tardis directory, running the following command:
py.test --cov tardis tardis/tests/      gives the following error:

Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/_pytest/config.py", line 543, in importconftest
    mod = conftestpath.pyimport()
  File "/usr/local/lib/python2.7/dist-packages/py/_path/local.py", line 629, in pyimport
    __import__(pkgpath.basename)
  File "/home/vidit/tardis/tardis/__init__.py", line 7, in <module>
    from ._astropy_init import *
  File "/home/vidit/tardis/tardis/_astropy_init.py", line 125, in <module>
    config.configuration.update_default_config(__package__, config_dir)
  File "/usr/local/lib/python2.7/dist-packages/astropy/config/configuration.py", line 844, in update_default_config
    version = mod.__version__
AttributeError: 'module' object has no attribute '__version__'
ERROR: could not load /home/vidit/tardis/tardis/conftest.py

wkerzendorf:
@vidit23 Can you please run it with anaconda, that will mean that you will have the right packages installed. This is described in the manual.

wkerzendorf:
let me know how it goes.

vidit23:
I ran it using anaconda but I am still getting the same error.

I used py.test --traceconfig, I get errors like:
/usr/local/lib/python2.7/dist-packages/py/_path/common.py:334: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
/usr/local/lib/python2.7/dist-packages/py/_path/common.py:370: in gen
    if p.check(dir=1) and (rec is None or rec(p))])
/usr/local/lib/python2.7/dist-packages/_pytest/main.py:628: in _recurse
    ihook.pytest_collect_directory(path=path, parent=self)
/usr/local/lib/python2.7/dist-packages/_pytest/main.py:166: in call_matching_hooks
    plugins = self.config._getmatchingplugins(self.fspath)
/usr/local/lib/python2.7/dist-packages/_pytest/config.py:688: in _getmatchingplugins
    plugins += self._conftest.getconftestmodules(fspath)
/usr/local/lib/python2.7/dist-packages/_pytest/config.py:521: in getconftestmodules
    mod = self.importconftest(conftestpath)
/usr/local/lib/python2.7/dist-packages/_pytest/config.py:545: in importconftest
[...]

PR #232, opened 2015-03-06
py.test panics about not finding the version

kaushik94:
Please take a look at #228 . This fixes it.

### DIFF
--- a/tardis/_astropy_init.py
+++ b/tardis/_astropy_init.py
@@ -122,7 +122,7 @@ if not _ASTROPY_SETUP_:
     if not os.environ.get('ASTROPY_SKIP_CONFIG_UPDATE', False):
         config_dir = os.path.dirname(__file__)
         try:
-            config.configuration.update_default_config(__package__, config_dir)
+            config.configuration.update_default_config(__package__, config_dir, version=__version__)
         except config.configuration.ConfigurationDefaultMissingError as e:
             wmsg = (e.args[0] + " Cannot install default profile. If you are "
                     "importing from source, this is expected.")


wkerzendorf:
@embray is this sensible? if so should it go to package-template

@kaushik94 as an explanation TARDIS uses the package-template of astropy and if this is a good fix then it should also be fixed there.

kaushik94:
@wkerzendorf  IMHO the problem might not be due to package-template, not quite sure why but when we don't supply the version number to astropy it tries to fetch it from the config file but in this case it isn't able to. Moreover the PR I submitted is merely a hack and not a solution, we should still dig into whats happening. The reason I came up with a quick fix is because `py.test` isn't working and its not possible to test code locally without travis. I think this could act as a temporary solution till we find the actual culprit.

wkerzendorf:
run the tests using `python setup.py test`. That's how they're designed to be run. let me know if the problem still persists.

tardis/_astropy_init.py
```diff
@@ -122,7 +122,7 @@ def test(package=None, test_path=None, args=None, plugins=None,
     if not os.environ.get('ASTROPY_SKIP_CONFIG_UPDATE', False):
         config_dir = os.path.dirname(__file__)
         try:
-            config.configuration.update_default_config(__package__, config_dir)
+            config.configuration.update_default_config(__package__, config_dir, version=__version__)
```
embray: Weird. I don't think this should be necessary, but I haven't really looked closely at the problem yet.  You should be getting a `version.py` generated when you run `setup.py`.

wkerzendorf:
@embray this is likely the problem. I think @kaushik94 is running `pytest` without `setup.py`

kaushik94:
@wkerzendorf  I had run both `python setup.py install` and `python setup.py test` and I also got the version.py but still get the same error.
```
==================== 98 passed, 12 skipped in 5.36 seconds =====================
kaushik@ghost:~/Desktop/soc/tardis$ cd tardis
kaushik@ghost:~/Desktop/soc/tardis/tardis$ cd tests/
kaushik@ghost:~/Desktop/soc/tardis/tardis/tests$ py.test
  File "/usr/local/lib/python2.7/dist-packages/_pytest/config.py", line 543, in importconftest
    mod = conftestpath.pyimport()
  File "/usr/local/lib/python2.7/dist-packages/py/_path/local.py", line 629, in pyimport
    __import__(pkgpath.basename)
  File "/home/kaushik/Desktop/soc/tardis/tardis/__init__.py", line 7, in <module>
    from ._astropy_init import *
  File "/home/kaushik/Desktop/soc/tardis/tardis/_astropy_init.py", line 125, in <module>
    config.configuration.update_default_config(__package__, config_dir)
  File "/usr/local/lib/python2.7/dist-packages/astropy/config/configuration.py", line 844, in update_default_config
    version = mod.__version__
AttributeError: 'module' object has no attribute '__version__'
ERROR: could not load /home/kaushik/Desktop/soc/tardis/tardis/conftest.py
```

wkerzendorf:
you do not need to run the install. I don't know how the astropy version is. I suggest using anaconda as described in the manual.

wkerzendorf:
@kaushik94 i fixed it I think

wkerzendorf:
@kaushik94 if it's fixed - close this (but still put this PR in your application :wink: )

kaushik94:
@wkrzendorf are you talking about PR #250 ? That looks like a cleaner fix than mine, will pull your changes and close this now. Will let you know if I run into trouble.
>(but still put this PR in your application :wink: )

Haha sure :laughing:

kaushik94:
also take a look at #234 when you find some time

Issue #233, opened 2015-03-06
Wrong usage of yield in test_utils.py


kaushik94:
I just figured out that yeild does not run the function, it simply returns a generator object, and hence the assertions are not really tested there. For example if we make this change
```
--- a/tardis/tests/test_util.py
+++ b/tardis/tests/test_util.py
@@ -49,7 +49,7 @@ def test_element_symbol_reformatter():
 def test_species_string_to_species():
     atom_data = atomic.AtomData.from_hdf5(atomic.default_atom_h5_path)
     def _test_species_string_to_species_tuple(species_string, species_tuple):
-        assert species_string_to_tuple(species_string) == species_tuple
+        assert species_string_to_tuple(species_string) != species_tuple
 
     data = [('si ii', (14, 1) ),
             ('si 2', (14, 1)),
```
It still passes the tests without an error. When I did a little research I found this SO question http://stackoverflow.com/questions/231767/what-does-the-yield-keyword-do-in-python
Please read it carefully.
I have submitted a PR #234 removing instances of yeild usage and now things seem to work fine

orbitfold:
No, yield does not disable all statements in a function it is in. Unless you mean something different by a statement then what it is - a smallest standalone element in an imperative programming language.

orbitfold:
Try running this for example:

def a():
    i = 0
    while True:
        if i < 100:
            yield i
            i += 1
            print i, i, i
        else:
            raise StopIteration

if __name__ == '__main__':
    for x in a():
        pass

kaushik94:
Sorry my bad, what I meant to say is that yield doesn't run the code in the function, it lazily returns a generator object that has to be iterated. Take a look at my PR, when we are yielding the function, we are not really running them hence not running assertions in the tests

kaushik94:
@orbitfold  thanks for pointing out, I just updated the comment. @wkerzendorf  take a look and let me know what you think.

orbitfold:
I don't know why whoever wrote those used yield there, but you should not assume it was by mistake. I think it might have something to with this http://pytest.org/latest/yieldfixture.html, but I don't know for sure.

wkerzendorf:
I wrote them. If you don't use yield, then pytest will just show this as one test and not as multiple tests.

wkerzendorf:
it is a way to parametrize a test

wkerzendorf:
[...]

PR #234, opened 2015-03-06
Removing usage of yield in test_utils.py

kaushik94:
Please take a look at #233. Also added a test for `species_tuple_to_string`. @wkerzendorf  please take a look.

### DIFF
--- a/tardis/tests/test_util.py
+++ b/tardis/tests/test_util.py
@@ -30,6 +30,7 @@ def test_atomic_number2element_symbol():
 def test_element_symbol2atomic_number(element_symbol, atomic_number):
     assert element_symbol2atomic_number(element_symbol) == atomic_number
 
+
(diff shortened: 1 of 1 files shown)


kaushik94:
Works perfect now( with the fix also in #232 )
```
kaushik@ghost:~/Desktop/soc/tardis/tardis/tests$ py.test -q test_util.py
..................
18 passed in 0.69 seconds
```

kaushik94:
@wkerzendorf  The test_utils.py doesn't work currently even if we plug in the wrong values, because of the usage of yield and I think this PR solves it. I think we should merge it before we make any major changes to the files. What do you think ?

wkerzendorf:
@kaushik94 well done!

kaushik94:
Thanks @wkerzendorf .

PR #248, opened 2015-03-06
Added steps to start the gui in linux

gautam1168:
These steps work when starting the gui from linux using anaconda setup.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -55,39 +55,39 @@ This model can then be used for inspecting the run as described
 Graphical User Interface
 ========================
 
-To run the GUI follow these steps:
+**To run the GUI follow these steps:**
 
-1. To be done only once for running GUI for the first time:
-    
-    .. code-block:: none
+**1**. To be done only once when running GUI for the first time
+	
+.. code-block:: none
 
-        source activate tardis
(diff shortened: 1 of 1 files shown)


wkerzendorf:
have you made sure that these documents build with sphinx?

gautam1168:
I have now tested that the changes I made build with sphinx. For doing this
test
I cloned the repo at github and added the changes to running.rst in docs
folder.
Then I ran `make html`, and checked that I am getting the right output by
opening
the running.html that was generated. Is that what you wanted?

Also there are a few packages which need to be installed to run the gui
that don't
get installed when the conda create is passed the file

https://raw.githubusercontent.com/tardis-sn/tardis/master/conda-requirements

These are

ipython 3.0.0
pyside 1.2.1
shiboken 1.2.1

Should I include instructions to install these in the running.rst, or
would these be added to the conda-requirements file?

wkerzendorf:
just add instructions to running rst

gautam1168:
I have now created another pull request which in addition to an updated running.rst has an updated gui.py, which can work with both PySide and PyQt4. So I am closing this request. I will redit the running.rst on the other pull request to include instructions on installation of the packages.

PR #251, opened 2015-03-07
gui.py and running.rst edited for PyQt4 and PySide

gautam1168:
gui.py will now import PyQt4 or PySide depending on which is available and chosen (see documentation). So it is compatible with both bindings now. 

Details on how to choose a binding is added in running.rst.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -55,6 +55,34 @@ This model can then be used for inspecting the run as described
 Graphical User Interface
 ========================
 
-A graphical user interface is being developed for this code using QT.
-We envision to have this available in the next few minor releases.
+**To run the GUI(under development) follow these steps:**
 
+The gui can use one of two python bindings for qt, namely PyQt4
+and PySide. You can choose which binding is used by setting the
+environment variable QT_API in your bash.
+
+**1**. Choosing between PySide and PyQt4 (optional)
+
+.. code-block:: none
+
+	#To choose PySide
+	export QT_API=pyside
+	
+	#To go back to PyQt4
+	unset QT_API
+
+**2**. An example of creating a model and GUI
+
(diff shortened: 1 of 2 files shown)


gautam1168:
I don't understand what the coverage decreased means. Can someone help me?

imjalpreet:
@gautam1168, Coverage decreased means that the percentage of lines of code that are currently being tested/covered by the tests has decreased

wkerzendorf:
because you added additional lines do the code, that are not tested. The GUI is difficult to test, so do not worry about it (i think we can handle 0.04%).

wkerzendorf:
@gautam1168 so it works for pyqt4 and pyside?

gautam1168:
Yes as I have written in the running.rst step 2, one can choose between PySide and PyQt4 by setting/unsetting the environment variable QT_API in bash before launching the ipython shell.

wkerzendorf:
I understand, did you try it with an environment with pyside and one with pyqt4

gautam1168:
No, one doesn't need to keep only one package in the environment. One can have both PySide and PyQt4 and choose between the two. Then after launching the ipython shell one can check which one
was imported using the function loaded_api() after importing it as 

`from IPython.external.qt_loaders import loaded_api`

But I will try it out with environments that have only one of them installed and let you know.

wkerzendorf:
@gautam1168 good work!

gautam1168:
Hold on, don't merge it yet. I think something has broken. I cannot execute run_tardis. Let me fix this and check those environments. I'll try to sort all of this out by midnight today.

gautam1168:
It all works. It is ready for merging.

I created two clones of my environment and removed pyqt from one and pyside from the other. Then I ran the gui in each after setting/unsetting QT_API as required. And it works in both cases as well as in the environment that has both the bindings.

PR #238, opened 2015-03-06
Test added in test_atomic.py file

vaibhav4595:
test function for read_lines_data() in atomic.py added to test_atomic.py

PR #252, opened 2015-03-07
Tests added and updated for test_atomic

imjalpreet:
1. New tests added for function atomic.read_lines_data()
2. Some previous tests updated.


imjalpreet:
@wkerzendorf I have added some tests for atomic.py and also updated some previous tests. Do give me a feedback whenever you have time. :smile:

imjalpreet:
I checked later that this pr is similar (not exactly but similar) to #

imjalpreet:
I checked later that this pr is similar (not exactly but similar) to #238. Should should I keep it closed?

PR #253, opened 2015-03-07
test for read_collision_data added for the atomic.py file

vaibhav4595:
Needs a collision data file with which the return value can be tested, only after the collision file is available. Will make this test complete after that.


vaibhav4595:
@wkerzendorf Do have a look at this. Though this needs a data file which I had asked after which it would be a complete test for that function

PR #255, opened 2015-03-07
More tests added for test_atomic.py

imjalpreet:
1. Tests added for the function read_macro_atom_data.


imjalpreet:
@wkerzendorf I have added the test for read_macro_atom_data but it is without the use of fixtures. I will push the one with the fixtures soon. Till then whenever you are free, can you review this one?

Issue #256, opened 2015-03-08
Table headers not showing in gui


gautam1168:
The gui doesn't show horizontal or vertical headers due to a bug.

gautam1168:
This was fixed long ago.

PR #257, opened 2015-03-08
Table headers now showing

gautam1168:
The table headers are now showing. The problem was in the method
headerdata(). The returned data type was not the right type.


gautam1168:
@wkerzendorf I fixed the table headers. This fixes #256

PR #261, opened 2015-03-09
Added tests for atomic.py

vidit23:
I had to update my forked repository so, I combined all the seperate PR into one.
There are tests forthe following functions in atomic.py 
read_lines_data
read_zeta_data
read_collision_data


vidit23:
I tried using Cython wrappers to make some tests for cmontecarlo.c but it didnt work out. I am working on using SWIG or manual wrapping of C function.

PR #262, opened 2015-03-10
Ctest

vaibhav4595:
Temporarily, conftest has been added in this particular src directory(Works well this way"
Right now """""make test"""" runs the tests from the src directory.
A way to incorporate this in the main conftest.py in the tardis directory is needed
Then the next step is to incorporate this with setup.py test


vaibhav4595:
@wkerzendorf  Do have a look at this. I have temporarily added files inside the src directory itself. 
There is a function which checks the constant value of C from the cmontecarlo.h file .
Right now I have added the conftest.py inside the src itself. The command 'make test' from the src directory right now calls py.test to test the particular function.

wkerzendorf:
@vaibhav4595 it's fine to have multiple conftests in the structure. I think however that a call to `python setup.py test` should call the tests that you created. This is not the case yet :wink:

vaibhav4595:
@wkerzendorf I have got to work it now. Soon will be creating another PR.

wkerzendorf:
@vaibhav4595 if you don't need this PR anymore - can we close it?

vaibhav4595:
@wkerzendorf Yes sure we can. I am closing this.

PR #265, opened 2015-03-12
C test function

vaibhav4595:
This works on the local machine if the absolute path of the executable is provided in the conftest.py . (This needs to be worked out) Also, how to execute the makefile once the setup.py test begins needs to be worked out


vaibhav4595:
@wkerzendorf setup.py test now detects it. But the problem is that in the contest.py if I provide the absolute path to the executable then it works fine, but if I don't it gives an error.

wkerzendorf:
I'm going to be mean: As long as TRAVIS fails it's not tested :wink: It's a hard problem and your persistence is admirable: Well done! So how I tried to get around the problem is by not trying to convince setup.py to build the c-tests, but by wrapping each function in Cython and then using normal pytest to solve that problem (I got stuck there again as I had trouble converting a dict to a struct). Both things are possible. Good luck!

PR #266, opened 2015-03-13
Tests added for Montecarlo

imjalpreet:
- [x] Tests have been added for some basic functions in **cmontecarlo.c**- 
- [x] This is still incomplete as I am still trying to get the **absolute path** of the file **cmontecarlo.so** in the **tardis/montecarlo/src/** directory. **(DONE)**
- [x] For, now it works with **py.test -v test_montecarlo.py** but it fails with **python setup.py test** as the path that I have used is not from the **build directory**, so it results in **"No such fiile or directory"**
**(DONE)**


imjalpreet:
@wkerzendorf Can you have a look at this and tell if it's fine? I know it will not pass the tests till I find a way to write the absolute path but can you check if it's fine otherwise?

wkerzendorf:
This looks like a good first effort. Please also think about your application.

imjalpreet:
@wkerzendorf Thanks and I am also planning about my application.

imjalpreet:
@wkerzendorf Please have a look at this whenever you have time :smile:

wkerzendorf:
@imjalpreet an interesting solution using ctypes. I presume you fixed the path problems. If that is the case, please update the description of your PR (For other PRs you also make checkboxes which you check when certain things are done, that might have been useful here).

imjalpreet:
@wkerzendorf Done!

PR #267, opened 2015-03-13
Prliminary test using wrappers

vaibhav4595:
@wkerzendorf  I tried just some preliminary tests(for a check). But I was unable to build extensions for the files inside the wrappers directory like the way in which extensions get built for the montecarlo files using montecarlo.pyx and setup_package.py.  Though I made a setup_package.py for the wrappers extension as well but this did not have built the extensions when I built tardis again.

### DIFF
--- a/tardis/montecarlo/wrappers/get_test_wrapper.pyx
+++ b/tardis/montecarlo/wrappers/get_test_wrapper.pyx
@@ -1,0 +1,34 @@
+import numpy as np
+cimport numpy as np
+
+ctypedef np.int64_t int_type_t
+
+cdef extern from "../src/cmontecarlo.h":
+    ctypedef enum rpacket_status_t:
+        TARDIS_PACKET_STATUS_IN_PROCESS = 0
+        TARDIS_PACKET_STATUS_EMITTED = 1
+        TARDIS_PACKET_STATUS_REABSORBED = 2
+
+    ctypedef struct rpacket_t:
+        double nu
+        double mu
+        double energy
+        double r
+        double tau_event
+        double nu_line
+        int_type_t current_shell_id
+        int_type_t next_line_id
+        int_type_t last_line
+        int_type_t close_line
+        int_type_t recently_crossed_boundary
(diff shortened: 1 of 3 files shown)


vaibhav4595:
@wkerzendorf Could you have a look at this ?

wkerzendorf:
@vaibhav4595 a little bit more information would be helpful - tracebacks, building console output, etc. 
The first thing that I see that wrappers doesn't have an `__init__.py`

vaibhav4595:
@wkerzendorf After I initialised __init__.py file ....then I ran python setup.py develop....then I had the following error......
..
...
...
writing manifest file 'tardis_sn.egg-info/SOURCES.txt'
running build_ext
error: [Errno 2] Could not find C file tardis/montecarlo/wrappers/get_test_wrapper.c for Cython file tardis/montecarlo/wrappers/get_test_wrapper.pyx when building extension tardis.montecarlo.wrappers.get_test_wrapper. Cython must be installed to build from a git checkout.: 'tardis/montecarlo/wrappers/get_test_wrapper.c'
.....
.....
This is the error in the end that I have
But the problem is that I have Python already installed

wkerzendorf:
a couple of comments: 
 1. try to use pastebin for longer debug logs
 2. I presume that you are in the wrong anaconda environment or something and it doesn't see `cython`

let me know how it goes.

vaibhav4595:
https://gist.github.com/vaibhav4595/087bcfb941ab3b524ad7

wkerzendorf:
Yes there is an astropy issue - but the solution is also posted in there: make sure cython is installed!

vaibhav4595:
@wkerzendorf  I have got it fixed now. Cython was already installed but the path was different from what it was using. Installed it again under the anaconda environment by extracting the source files and then running the install. Thank you for your patience.

vaibhav4595:
@wkerzendorf  Do have a look at the latest commit...I have just tried a basic method which seems to be running well...though this is just a very simple way of handling things

PR #272, opened 2015-03-17
[GSoC 2015 WIP] Tests for C code in montecarlo

kaushik94:
@wkerzendorf  please take a look at the framework. This is very basic and will start adding tests from today. But I'm afraid I don't have through knowledge of how to use the functions in `cmontecarlo.c`. Could you help me out by throwing some example function so I can get a basic familiarity.

Focus of the PR is :
- [x] construct a make file to compile all the C files properly to get a test object
- [x] write a `conftest.py` file to make `py.test` play well
- [x] write off some basic get/set tests to see if the above is working  
- [x] configure the `setup.py` to make it know that `montecarlo/tests` needs to be tested
- [ ] make the setup.py know how to build the tests in C for montecarlo( so that even py.test command plays well with it ). Currently running into problems with this.
- [ ] complete the getter/setter tests for cmontecarlo
- [ ] complete distance tests for montecarlo
- [ ] complete end-to-end tests for montecarlo
- [ ] re factor the code in cmontecarlo a bit into utilities/helpers etc ( probably )


kaushik94:
The right way to test is
`make`
`py.test`

wkerzendorf:
Test the getters and setters for `rpacket`. It needs to work with `python setup.py test`

kaushik94:
@wkerzendorf  how do I get this to work with `setup.py test`. The way it is done now is going into the src directory where the tests are, defined the `conftest.py` there so have to run `py.test` in there itself.

wkerzendorf:
@kaushik94 I don't know. You can have multiple `conftest.py` in subdirectories. I think you somehow need to convince setup.py to build these C-files. I went the cython wrapper way, as that seemed easier.

wkerzendorf:
maybe move the test package one up to `montecarlo/tests/` maybe that will work.

kaushik94:
sure, will do that now. In the meanwhile, could you please take a look at my proposal ? I have shared it with you.

wkerzendorf:
I did look at it and gave you comments on gitter. So the coverage decreased by `0.18 %` - what happened there :wink:

wkerzendorf:
@kaushik94 does it work now with `setup.py test`?

kaushik94:
@wkerzendorf  sorry I was having my exams so couldn't give much time to this PR. Yes its working with `setup.py test` finally, and its detecting `test_*.c` files in the `tests` folder when I did as you said. But the problem is that its unable to detect the C object file that corresponds to `test_cmontecarlo.c` i.e `test_cmontecarlo`. And I was not able to understand why, any suggestions ?
And the decrease in test coverage, Im not sure. Could be because I had a dummy file in my previous commit called `test_me.py` and later removed, could that be the reason ?

kaushik94:
I think the problem of not finding the executable is because `shutil.copytree` not copying them to the `tmp` directory. Thats why they are running with `py.test`. I don't think its a good idea to leave it like this. @wkerzendorf  what do you think ?

wkerzendorf:
nevermind the decrease in coverage (it was not meant that serious :wink: ) Sorry can you elaborate: Does it run the Ctests or not?

kaushik94:
> nevermind the decrease in coverage (it was not meant that serious :wink: ) 

Haha I never understood what `Coverage decreased x to x-delta when pulling` meant.
> Does it run the Ctests or not?

Yes it does run them, But it can't find the corresponding executable. For example if I run `setup.py test`

```
>           raise child_exception
E           OSError: [Errno 2] No such file or directory
------------------------------- Captured stdout --------------------------------
/tmp/tardis-test-cCxXUT/lib.linux-x86_64-2.7/tardis/montecarlo/tests/test_cmontecarlo
================ 96 passed, 12 skipped, 1 error in 6.61 seconds ================
```

Its unable to find the `test_cmontecarlo` in the `/tmp/tardis-test-.....` directory. Any guesses ?

wkerzendorf:
look at the `setup_package.py` files. you need to specify any C file that should be copied over as far as I remember.

kaushik94:
Is there a way to run the C tests in-place or someway to get the C executables into the tmp directory ?

kaushik94:
> look at the setup_package.py files. you need to specify any C file that should be copied over as far as I remember.

No I tried it, Its able to detect the `.c` files actually when I run `isfile` command, but it throws error when I check the same for `test_cmontecarlo` executable.

wkerzendorf:
You need to add your files  to the  'setup_package.py' in montecarlo dir

kaushik94:
Still the same error

```
_________ ERROR collecting tardis/montecarlo/tests/test_cmontecarlo.c __________
_pytest.runner:137: in __init__
>   ???
_pytest.main:418: in _memocollect
>   ???
_pytest.main:295: in _memoizedcall
>   ???
_pytest.main:418: in <lambda>
>   ???
tardis/montecarlo/tests/conftest.py:24: in collect
>       test_output = subprocess.check_output(test_exe)
/usr/lib/python2.7/subprocess.py:566: in check_output
>       process = Popen(stdout=PIPE, *popenargs, **kwargs)
/usr/lib/python2.7/subprocess.py:710: in __init__
>                               errread, errwrite)
/usr/lib/python2.7/subprocess.py:1327: in _execute_child
>           raise child_exception
E           OSError: [Errno 2] No such file or directory
------------------------------- Captured stdout --------------------------------
/tmp/tardis-test-0s5rUj/lib.linux-x86_64-2.7/tardis/montecarlo/tests/test_cmontecarlo
================ 96 passed, 12 skipped, 1 error in 6.72 seconds ================
```

Please take a look at my conftest.py file inside `montecarlo/tests` and let me know if im doing something terribly wrong ?

kaushik94:
But when I run `py.test` inside the directory, everything runs fine !

wkerzendorf:
Maybe @embray has an idea - the astropy setup structure was developed partly by him.

tardis/montecarlo/tests/Makefile
```diff
@@ -0,0 +1,22 @@
+all: test_cmontecarlo
+
+rk_isaac.o: ../src/randomkit/rk_isaac.c
+	g++ -fpermissive -c ../src/randomkit/rk_isaac.c
```
wkerzendorf: try to get the compiler from setup.py. g++ is not always available (Macs mostly use clang)

vaibhav4595:
@kaushik94 I think the probable reason for this is that the .so files for the C test files have not been created in the build (similar to the way in which montecarlo.so etc. files get created once you have developed). Have you tried searching for the .so files in the build directory for the C test files that you wrote ?

embray:
Although this page doesn't give enough details, start here: http://docs.astropy.org/en/stable/development/ccython.html  The Makefile should be removed entirely--instead use the distutils-style C module declarations so that setup.py knows how to build your test modules for testing.

kaushik94:
@vaibhav4595  yes thats what I was talking about. Basically these binary files are created, but the way setup.py works is it does `copytree` to the whole directory into the tmp directory which gets erased right after the execution of tests. Take a look at astropy test_helper. So I don't understand why this happens. @embray  solution seems to be sensible, have to try it out. Thanks for chipping in and helping out guys.

kaushik94:
I could somewhat get things to work. @embray  did it your way and it worked pretty well. In the sense that I could compile the `.c` files and can see them in the build directory. But there is a small problem to fix here. In the setup_package.py I added lines

```
sources_test = #all those files you need to compile test_cmontecarlo.c
Extension('tardis.montecarlo.tests.test_cmontecarlo', sources_test,
                 include_dirs=['tardis/montecarlo/src',
                 'tardis/montecarlo/src/randomkit',
                 'tardis/montecarlo/tests',
                 np.get_include()])]
```

so its creating a shared library `.so` which can't be executed and throws an error `returned non-zero exit status -11`. Is there a way to directly create an executable. But the `.so` file is surely present and I checked it with the line `print os.path.isfile(test_exe+'.so')` and returns True. @wkerzendorf  take a look and let me know if you have an idea.

tardis/montecarlo/tests/Makefile
kaushik94: The make file is only for linux. Take a look at my `setup_package.py` file. I directly built them as an extension.

wkerzendorf: it needs to work  on both

kaushik94:
@wkerzendorf  can this be reopened ? I might need it in future, if that's not a problem.

tardis/montecarlo/setup_package.py
```diff
@@ -13,7 +13,19 @@ def get_extensions():
...
+        os.path.join(os.path.dirname(__file__), 'src', '*.c'))]
+    sources_test += [os.path.relpath(fname) for fname in glob(
+        os.path.join(os.path.dirname(__file__), 'src/randomkit', '*.c'))]
+
     return [Extension('tardis.montecarlo.montecarlo', sources,
                       include_dirs=['tardis/montecarlo/src',
                                     'tardis/montecarlo/src/randomkit',
+                                    'tardis/montecarlo/tests',
+                                    np.get_include()]),
+	    Extension('tardis.montecarlo.tests.test_cmontecarlo', sources_test,
```
wkerzendorf: so in the end you should be able to do `import tardis.montecarlo.tests.test_cmontecarlo`

tardis/montecarlo/tests/conftest.py
```diff
@@ -0,0 +1,96 @@
+import subprocess
```
wkerzendorf: get rid of this!

PR #277, opened 2015-03-22
Update test_atomic.py with test_lines_h5_readin()

spark1729:


wkerzendorf:
good - we don't merge PRs, but use them to evaluate applications. So do not wonder if things don't get merged.

PR #281, opened 2015-03-26
Testing the C code

alexmherrmann:
Added a small test case for rpacket_{get,set}_nu using cython and assert statements
Other test cases to follow, this is just an example of one way to do it.
Even without GSOC it would be cool to keep writing the C test cases, and increasing coverage

### DIFF
--- a/tardis/tests/test_montecarlo.py
+++ b/tardis/tests/test_montecarlo.py
@@ -1,8 +1,13 @@
 import numpy as np
-from tardis import montecarlo
+from tardis.montecarlo import montecarlo
+import tardis
 import pytest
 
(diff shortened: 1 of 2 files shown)


tardis/tests/test_montecarlo.py
```diff
@@ -1,9 +1,14 @@
...
-from tardis import montecarlo
+from tardis.montecarlo import montecarlo
+import tardis
 import pytest
 
 test_line_list = np.array([10, 9, 8, 7, 6, 5, 5, 4, 3, 2, 1]).astype(np.float64)
 
+def test_rpacket():
+    print dir(tardis.montecarlo.montecarlo)
+    assert montecarlo.rpacket_test() == True
```
wkerzendorf: just run it. If an assertion error is being thrown pytest should already catch it.

Issue #287, opened 2015-04-07
Missmatch between model.py and tardis_config_definition.yml


unoebauer:
Referring to commit 35a0119dfc29d4bcceee859800c3bc77f0ec3d96

In `tardis_config_definition.yml` the parameter `hold-iterations` (for consistency should be called `hold_iterations`) is defined in the `convergence_strategy` block. However, `model.py` still expects the parameter `hold` on line 297:
```python
                    self.iterations_remaining = self.global_convergence_parameters['hold']
```
This should be changed.

unoebauer:
Solved by #288

PR #288, opened 2015-04-07
-fixed discrepancy concerning hold_iterations

unoebauer:
Fix for #287


wkerzendorf:
@unoebauer @mklauser looks good and is passing

PR #292, opened 2015-04-24
One character added: fix to artis read of density?

ssim:
Fixes #290. 

Have been checking that the file reads work in the current version - looking for correct reading of artis files and consistency with reading of ascii files. 

It seems to me that the mean_density read for artis files is current mis-indexed by 1. This pull request is a suggested fix. 

@wkerzendorf What do you think? We should put in a test for this, but I forgot how to do that. Can you help with that sometime when you have a mo.

PR #299, opened 2015-05-05
Test of model_reader.py

aoifeboyle:
@wkerzendorf Continuing on from PRs #292 and #297. Just added another few lines to the test to check that the zone parameters were being matched correctly between the model and abundance files. This was an issue with a previous version so it's a good thing to check.


wkerzendorf:
@aoifeboyle well done! I'm merging this now!

Issue #293, opened 2015-05-02
GSoC 2015 - Re-factor Montecarlo 


kaushik94:
- [x] Remove all the getter and setter functions and all the endpoints(ones which are accessible to user) to a separate file `montecarlo.c`
- [x] Functions like computing distances and other helper routines should be moved into separate file like `geometry.c`
- [x] Functions like binary search, linear search should be a separate file like `search_routines.c`

@wkerzendorf, @mklauser , @orbitfold  Please let me know what you guys think. Update the checklist with other ways you think appropriate.

wkerzendorf:
I think that is a good start. The computing distances part is more a geometry task and should maybe go into `geometry.c`. I think a PR is the best way to proceed as one can much easier comment.

PR #301, opened 2015-05-08
[GSoC 2015 MRG] Restructuring Montecarlo

kaushik94:
This is still unfinished and in an embarrassing state. TODO - split the header files accordingly and all other usages of the functions in python files have to he changes I guess. #293.


orbitfold:
I think the best way to structure the code is to let the structure and the methods that operate on that structure to be in the same file. Think how you would structure a large C++ program. It really does not make sense for some functions that operate on an rpacket to sit in one file and others in another. I propose to make a file called rpacket.c to hold the methods and rpacket.h to hold the data structure. storage.h should hold the storage structure and utilities.c and utilities.h should hold the search functions and other stuff. It is really important that getter and setter methods go in the same file as other functions that operate on an rpacket. Basically think of it as a C++ class and put all the "methods" in the same file. Now it is arbitrary and confusing IMO.

kaushik94:
@orbitfold  It was very dumb on my part. I have to clearly know first what these functions do, so will do it in a while. About a week or so for the finished one. Check out the new one, better ?

tardis/montecarlo/src/search_routines.c
```diff
@@ -0,0 +1,112 @@
+#include "search_routines.h"
```
wkerzendorf: maybe put this into helpers or call it `utils.c`. @orbitfold what do you think?

wkerzendorf:
@kaushik94 no worries. It's all a learning process :wink:

kaushik94:
Hey @wkerzendorf  and @orbitfold  can you please take a look at it and let me know if thats in the lines of what you are thinking. Now we can create `geometry.c` and `utils.h` and I think we'll be done with the restructuring ?

wkerzendorf:
@kaushik94 it looks like a good start. why did you change the access from 644->755?

kaushik94:
@wkerzendorf  even I am wondering. I took a new mac, its my first time working on OSX. I thing I messed up with the configuration or something. Anyway to revert the permissions back ?

wkerzendorf:
hmm - use chmod. that should work.

kaushik94:
Yes, I think that worked !

wkerzendorf:
looks good!

kaushik94:
I think we have to change the name of `cmontecarlo` to `utils` or `geometry` or both. But just waiting for @orbitfold to review, know what he thinks. Anything more we should do ?

wkerzendorf:
@kaushik94 - not really except maybe makeing a new PR from that that tries to test the rpacket stuff. That should be fairly simple.

kaushik94:
Yes I'll get to that by Tuesday probably. I'll get done with this one, write documentation, remove unused imports etc. Shall we go with the testing once this is merged or you want me make it in parallel?

wkerzendorf:
@orbitfold can you have a quick look and tell us if it is ready to be merged. @ssim if you get a chance please have a look. I will merge by monday if there are no serious issues.

wkerzendorf:
@mklauser please have a look as well.

PR #302, opened 2015-05-12
Bug fix

chvogl:
Bug fix for the atomic.base.py and for the atomic.plugins.py. We found a few syntax errors. 
From @mklauser this is WP, however we could merge it to atomic/ restructure.

### DIFF
--- a/tardis/atomic/base.py
+++ b/tardis/atomic/base.py
@@ -9,9 +9,9 @@ class AtomicData(object):
     def from_hdf(cls, fname):
         return cls(pd.HDFStore(fname, mode='r'))
 
-    def __init__(self, hdf_store, data_types):
+    def __init__(self, hdf_store, data_types=self.all_data_types):
         self.data_types = [data_type()
-                        for data_type in self.all_data_types]
+                           for data_type in data_types]
 
         self.name2data_type = {item.name:item for item in self.data_types}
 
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@chvogl congratulations on your first PR! I'll have some comments inline for you.

tardis/atomic/base.py
```diff
@@ -9,9 +9,9 @@ class AtomicData(object):
     def from_hdf(cls, fname):
         return cls(pd.HDFStore(fname, mode='r'))
 
-    def __init__(self, hdf_store, data_types):
+    def __init__(self, hdf_store, data_types=self.all_data_types):
```
wkerzendorf: this won't quite work. you can't access `self` in the initializer. the `data_types` should also be given by the user. so for now, I suggest leaving it as is.

tardis/atomic/base.py
```diff
@@ -9,9 +9,9 @@ class AtomicData(object):
...
         self.data_types = [data_type()
```
wkerzendorf: I see the confusion. hmm - yes that was definitely not right what I wrote. I think we need to think a little more how we want this object to be initialized. has @mklauser given you an HDF5 file?

tardis/atomic/base.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, hdf_store, data_types):
 
     def __getattr__(self, item):
         if item in self.name2data_type:
-            return self.name2date_type[item].data
+            return self.name2date_type[item].value
```
wkerzendorf: very good catch! I missed that.

tardis/atomic/base.py
chvogl: @mklauser hasn't given me an HDF5 file yet, but he wants to do so today.

wkerzendorf:
@chvogl is this still relevant?

chvogl:
@wkerzendorf
I'm sorry for taking so long to reply, but I just came back from my holidays.
To answer your question:
The last two items seem like they are still relevant, since there haven't been any changes in the atomic/restructure since then. However I'm not sure what the current plan for the atomic restructure is.

wkerzendorf:
@chvogl is this still relevant. Can we close?

chvogl:
To be honest, I have no idea what the current status regarding the atomic/restructure is and if we plan
on using any of it in the future.
I would assume that this PR has become irrelevant and that it can be closed but i cannot tell
you with absolute certainty.

wkerzendorf:
@chvogl I will close it - we can reopen if necessary.

PR #308, opened 2015-05-25
Added atomic properties tests for plasma.

aoifeboyle:
Also added model_reader.py tests (already part of tardis master) and commented out an incomplete test.


tardis/io/tests/test_model_reader.py
```diff
@@ -0,0 +1,35 @@
...
+data_path = os.path.join(tardis.__path__[0], 'io', 'tests', 'data')
+
+@pytest.fixture
+def artis_density_fname():
+    return os.path.join(data_path, 'artis_model.dat')
+@pytest.fixture
+def artis_abundances_fname():
+    return os.path.join(data_path, 'artis_abundances.dat')
+
+def test_simple_read_artis_density(artis_density_fname):
```
wkerzendorf: Have a look how I use the fixtures here and use them in a similar manner in your tests (in the call-sign of the function)

aoifeboyle: Ah, yes, I see now! Is that better? Any other suggestions for improvements?

wkerzendorf:
@kaushik94 can you have a look at this

tardis/plasma/tests/test_property_atomic.py
```diff
@@ -0,0 +1,49 @@
...
+def test_lines_lower_level_index_property(included_he_atomic_data):
+    
+    selected_atoms = [2]
+    levels_module = Levels(None)
+    lines_module = Lines(None)
+    levels = levels_module.calculate(included_he_atomic_data, selected_atoms)
+    lines = lines_module.calculate(included_he_atomic_data, selected_atoms)
+    lines_lower_module = LinesLowerLevelIndex(None)
+    lines_lower_level_index = lines_lower_module.calculate(levels, lines)
+    assert lines_lower_level_index[9]==0
```
wkerzendorf: being picky (we are trying to follow the coding standard PEP 8; https://www.python.org/dev/peps/pep-0008/): there should be a space `[9] == 0` :wink:

wkerzendorf:
This looks very good! I'll try to get @kaushik94 to have a look for PEP8 as well as to show him the test implementation. But I think we'll merge very soon.

tardis/plasma/tests/test_property_atomic.py
orbitfold: There also shouldn't be an empty line after the function header. If we're being picky :) Empty lines are discouraged generally in Python, since they make code much harder to read. They should only go between classes, methods, etc.

aoifeboyle: Thanks @wkerzendorf @orbitfold. Being picky is good! I want to adopt the right habits now. I'll have a look through the PEP 8 webpage. And I'll push and updated version when Kaushik has had a look :)

tardis/io/tests/test_model_reader.py
```diff
@@ -0,0 +1,35 @@
+import tardis
+from tardis.io.model_reader import (read_artis_density,
+read_simple_ascii_abundances)
+
```
kaushik94: I'm not sure but if there is no necessity for brackets, remove them and put in a single line.

tardis/io/tests/test_model_reader.py
```diff
@@ -0,0 +1,35 @@
...
+    (time_of_model, index, v_inner, v_outer,
+     mean_density) = read_artis_density(artis_density_fname)
+
+    assert np.isclose(0.00114661 * u.day, time_of_model, atol=1e-7 * u.day)
+    assert np.isclose(mean_density[23], 0.2250048 * u.g / u.cm**3, atol=1.e-6 * 
+        u.g / u.cm**3)
+    assert len(mean_density) == 69
+    assert len(mean_density) == len(v_inner)
+
+#Artis files are currently read with read ascii files function
```
kaushik94: You should leave a space - "# Artis files are....."

tardis/plasma/tests/test_property_atomic.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import numpy as np
+import os
+import tardis
+
+from tardis.plasma.properties.atomic import Levels, Lines, LinesLowerLevelIndex, LinesUpperLevelIndex, AtomicMass, IonizationData
+
+def test_levels_property(included_he_atomic_data):
+
```
kaushik94: Remove the newline.

tardis/io/tests/test_model_reader.py
```diff
@@ -0,0 +1,35 @@
...
+from tardis.io.model_reader import (read_artis_density,
+read_simple_ascii_abundances)
+
+from astropy import units as u
+import numpy as np
+
+import pytest
+
+import os
+
```
kaushik94: System imports like os, sys should go to the top. Then leave a line followed by our tardis imports. Same for other files.
Read this https://github.com/Khan/style-guides/blob/master/style/python.md

tardis/io/tests/test_model_reader.py
```diff
@@ -0,0 +1,35 @@
...
+data_path = os.path.join(tardis.__path__[0], 'io', 'tests', 'data')
+
+@pytest.fixture
+def artis_density_fname():
+    return os.path.join(data_path, 'artis_model.dat')
```
kaushik94: Leave a new line here

tardis/plasma/tests/test_property_atomic.py
```diff
@@ -0,0 +1,49 @@
...
+    selected_atoms = [2]
+    levels = levels_module.calculate(included_he_atomic_data, selected_atoms)
+    assert np.isclose(levels.ix[2].ix[0].ix[1]['energy'], 3.17545416e-11)
+
+def test_lines_property(included_he_atomic_data):
+
+    lines_module = Lines(None)
+    selected_atoms = [2]
+    lines = lines_module.calculate(included_he_atomic_data, selected_atoms)
+    assert np.isclose(lines.ix[564954]['wavelength'], 10833.307)
```
kaushik94: Its better and more clear to write it like this `assert np.isclose(lines.ix[564954]['wavelength'], 10833.307) is True`

tardis/io/tests/test_model_reader.py
aoifeboyle: @kaushik94 I just checked, and I think it is necessary because the maximum line length is supposed to be 79 characters I think?

tardis/plasma/tests/test_partition_function.py
```diff
@@ -3,16 +3,16 @@
...
 
-def test_level_boltzmann_factor1(standard_lte_plasma_he_db):
-    level_boltzmann_factor_module = LevelBoltzmannFactor(None)
-    levels = standard_lte_plasma_he_db.levels
-    factor = level_boltzmann_factor_module.calculate(levels, standard_lte_plasma_he_db.beta_rad)
-    assert np.allclose(factor.loc[2, 0, 0], levels.loc[2, 0, 0].g)
-    assert np.allclose(factor.loc[2, 1, 0], levels.loc[2, 1, 0].g)
-    assert np.allclose(factor.loc[2, 0, 10], 7.6218092841240705e-12)
+#def test_level_boltzmann_factor1(standard_lte_plasma_he_db):
+#    level_boltzmann_factor_module = LevelBoltzmannFactor(None)
```
kaushik94: It is not suggested to comment the functions for not passing. Rather use markers like this https://pytest.org/latest/skipping.html. something like `@pytest.mark.xfail` will make travis know that it should fail and will not be counted.

tardis/io/tests/test_model_reader.py
kaushik94: Ohh okay then its good, but I see an other import exceeding that, thats why

tardis/plasma/tests/test_property_atomic.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import numpy as np
+import os
+import tardis
+
+from tardis.plasma.properties.atomic import Levels, Lines, LinesLowerLevelIndex, LinesUpperLevelIndex, AtomicMass, IonizationData
```
kaushik94: Import is too long

tardis/plasma/tests/test_property_atomic.py
```diff
@@ -0,0 +1,49 @@
...
+    lines_upper_module = LinesUpperLevelIndex(None)
+    lines_upper_level_index = lines_upper_module.calculate(levels, lines)
+    assert lines_upper_level_index[9]==30
+
+def test_ionization_data_property(included_he_atomic_data):
+    
+    ionization_data_module = IonizationData(None)
+    selected_atoms = [2]
+    ionization_data = ionization_data_module.calculate(included_he_atomic_data, selected_atoms)
+    assert np.isclose(float(ionization_data.ix[2].ix[1]),  3.9393336e-11)
```
kaushik94: 2 spaces after a comma ?

tardis/plasma/tests/test_property_atomic.py
wkerzendorf: I would actually disagree with that. I think the `isclose` already describes it will

tardis/io/tests/test_model_reader.py
wkerzendorf: that's good. you can use pycharm to do that for you using optimize imports.

kaushik94:
Apart from the few comments I made rest seems good to me. The most important part is not to comment functions and use the markers( I sent a link )

tardis/io/tests/test_model_reader.py
wkerzendorf: both of you are right the way that this is broken is wrong, but it needs to be broken onto two lines. Pycharm should do it right.

aoifeboyle: Yeah you're right! I just changed it. Going through to make sure all the lines are within the limit now.

tardis/plasma/tests/test_property_atomic.py
kaushik94: I just checked, you are right @wkerzendorf and @aoifeboyle . No need to change this.

tardis/io/tests/test_model_reader.py
kaushik94: @aoifeboyle  I think its like this https://www.python.org/dev/peps/pep-0328/. @wkerzendorf ?

aoifeboyle: @kaushik94 Isn't the final suggestion under 'Rationale for Parentheses' the same as what we have?

tardis/io/tests/test_model_reader.py
aoifeboyle: @kaushik94 Do astropy, pytest etc. get grouped with the system imports or the tardis imports?

tardis/io/tests/test_model_reader.py
kaushik94: sorry I commented at the wrong place. Supposed to be at the place where there was a long import statement !

tardis/plasma/tests/test_property_atomic.py
kaushik94: @aoifeboyle here actually https://www.python.org/dev/peps/pep-0328/

tardis/io/tests/test_model_reader.py
kaushik94: They are 3rd party imports. Only imports from core python are system imports. So the preference is like this
system imports like `import os, sys`
*new line*
3rd party imports like `import astropy, pytest` or anything else
*new line*
our own imports like `from tardis import xyz`

tardis/io/tests/test_model_reader.py
aoifeboyle: Ah ok, no worries! I've updated that line too.

tardis/io/tests/test_model_reader.py
aoifeboyle: Thanks! :)

tardis/io/tests/test_model_reader.py
```diff
@@ -0,0 +1,35 @@
...
+@pytest.fixture
+def artis_abundances_fname():
+    return os.path.join(data_path, 'artis_abundances.dat')
+
+def test_simple_read_artis_density(artis_density_fname):
+    (time_of_model, index, v_inner, v_outer,
+     mean_density) = read_artis_density(artis_density_fname)
```
aoifeboyle: @kaushik94 Do you know whether the mean_density line here should have one more indent? I read in the guide that when you continue onto the next line with parentheses you should indent, but it doesn't specify whether that's true when the parentheses are before the equals sign.

tardis/plasma/tests/test_partition_function.py
wkerzendorf: that's probably better.

wkerzendorf:
@aoifeboyle @kaushik94 - great work!!

aoifeboyle:
@wkerzendorf @kaushik94 I've made all the changes you suggested. Thanks for your help. I'm just configuring pycharm at the moment. Once I get going with that I'll update this PR :)

tardis/io/tests/test_model_reader.py
kaushik94: @aoifeboyle not sure but seems okay to me

kaushik94:
@aoifeboyle  okay but what about the commented python code. I think its not a good idea to leave it commented, rather use the pytest markers. Take a look at this https://pytest.org/latest/skipping.html.

aoifeboyle:
@kaushik94 I have changed that part as you suggested (added @pytest.mark.xfail to each function, and it worked fine when tested). For any of your suggestions that I didn't comment on, I did make those changes. I only commented where I had questions.

kaushik94:
@aoifeboyle ohh I misunderstood, I din't see the changes reflected here so I kept commenting. I din't realise you din't push them yet. Please update the changes so we can take a look. Thanks :)

wkerzendorf:
@aoifeboyle @kaushik94 - should we merge now?

kaushik94:
+1 for merge

aoifeboyle:
Yes please! :)

wkerzendorf:
Well done both of you!!

PR #309, opened 2015-05-25
[ WIP ] C testing framework

kaushik94:
@wkerzendorf  how to go about from here. Should we use C types to call the function `test_rpacket_get_nu()` from a python test ?


tardis/montecarlo/tests/test_rpacket.c
```diff
@@ -0,0 +1,16 @@
+#include "../src/rpacket.h"
+#include <assert.h>
+#include <stdlib.h>
+
+void test_rpacket_get_nu(){
+	rpacket_t * rp = (rpacket_t *) malloc(sizeof(rpacket_t));
+	double value = 10.2;
+	rpacket_set_nu(rp, value);
+	assert(value == rpacket_get_nu(rp));
```
wkerzendorf: you need to write a code that raises a python exception when this test fails:

```python
if value != rpacket_get_nu(rp){ python assert}
``` - I don't know how maybe @juliantaylor knows.

Comment (juliantaylor) [inline tardis/montecarlo/tests/test_rpacket.c]: assert will end the process, so you wan't get good error messages or other tests running

raising a Python exception from C looks like this:
```c
PyErr_Format(PyExc_AssertionError, "value %g not %g", value, rpacket_get_nu(rp)))
// cleanup
return NULL;
```

tardis/montecarlo/tests/test_rpacket.c
```diff
@@ -0,0 +1,19 @@
...
+#include <Python.h>
+
+void test_rpacket_get_nu(){
+	rpacket_t * rp = (rpacket_t *) malloc(sizeof(rpacket_t));
+	double value = 10.2;
+	rpacket_set_nu(rp, value);
+	if( value != rpacket_get_nu(rp) ){
+		PyErr_Format(PyExc_AssertionError, "value %g not %g", value, rpacket_get_nu(rp));
+	}
+}
```
kaushik94: @juliantaylor does this seem good ? And what is the best way to run this test file `test_rpacket.c` using py.test, should I use `ctypes` ? @wkerzendorf please take a look.

wkerzendorf: just use `python setup.py test` - it should automatically detect it. test if you can import it from python?

kaushik94: No it doesn't test it. Thats the problem. when we run that command, the `setup.py` stores them in the build directory in the following manner
```
build
  |
  scripts
  temp -> where all the .o files are compiled
  lib -> where all the .so files are compiled
```
Now the conftest.py tries to find `test_rpacket.o` in the `lib` directory and fails to find it. This is happening due to the astropy_helpers. We need the conftest to somehow detect the `test_rpacket.o` file. You can reproduce it by `python setup.py build` and see the build directory.

kaushik94: No we can't directly import them into python. we need to use ctypes for that.

wkerzendorf: yes you can directly import them. have a look at http://nedbatchelder.com/text/whirlext.html

wkerzendorf: Do you see how you can import (you need to add a few more python specific C functions - described in the link above)? Let me know if you are stuck.

kaushik94: Will need some time, probably in 3 hrs will let you know what I'm upto ?

wkerzendorf: okay

kaushik94: okay @wkerzendorf  I just watched it. Even that uses wrappers but with pytest we have the luxury of directly building C files and testing them, only problem is that they are not being build in the right place where we want. Do you want me to go with the method proposed in the video ?

wkerzendorf:
So I actually didn't look at the video. I just looked at the slides and explanations below. I suggest we make sure that we can do `from tardis.montecarlo.tests.test_rpacket import test_rpacket_get_nu` and then go from there.

kaushik94:
Ohh okay I'll get done with that by tomorrow then.

kaushik94:
@wkerzendorf  do you know why these tests are failing ? Its running fine on my local machine.

wkerzendorf:
@kaushik94 have you checked that you can import and run the function? The coverage has decreased for now it's still working.

wkerzendorf:
very interesting: `IOError: [Errno 28] No space left on device` - never encountered that before.

wkerzendorf:
@kaushik94 I also get that error:

```python
In [1]: from tardis.montecarlo.tests import test_cmontecarlo
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-6be5d926c07e> in <module>()
----> 1 from tardis.montecarlo.tests import test_cmontecarlo

ImportError: dynamic module does not define init function (inittest_cmontecarlo)
```

wkerzendorf:
@kaushik94 how are you getting along?

tardis/montecarlo/tests/test_rpacket.c
```diff
@@ -0,0 +1,20 @@
...
+
+#include "../src/rpacket.h"
+
+int testing_rpacket_get_nu();
+
+int
+testing_rpacket_get_nu(){
+	rpacket_t * rp = (rpacket_t *) malloc(sizeof(rpacket_t));
+	double value = 10.2;
+	/*rpacket_set_nu(rp, value);
```
kaushik94: @wkerzendorf if I uncomment this particular line, the compiler is throwing an error. How do I call the function `rpacket_set_nu` from `rpacket.h` ?

wkerzendorf: I don't know. can you put the error that it is throwing here? maybe we can figure out why. (there will be a lot more C python stuff so it might be useful to learn these things).

kaushik94:
@wkerzendorf isn't this taking a little too long to run ?

wkerzendorf:
Huh?

kaushik94:
@wkerzendorf the coverall tests. Its almost been a day and a half I guess. They are still running ?

wkerzendorf:
Well coveralls is just being triggered from Travis. I think it's struggling with the C wrapping. Which is fine - I guess. As long as TRAVIS passes - that's the important part

kaushik94:
@wkerzendorf  Ohh yes I totally forgot that. You are right. If we randomise we might not be able to reproduce. Just let me know what should we do about that while I finish writing the other tests.

PR #310, opened 2015-05-26
A more detailed description of the montecarlo process

wkerzendorf:
After talking to Frederik this morning I whipped this up to give a more detailed description of the montecarlo process to the users. @ssim please confirm that this is correct

@fredRos


ssim:
Hello:

Mostly ok - but we don’t (shouldn’t at least!) assume emission is “isotropic in angle for a half-sphere” - we should be using mu = sqrt(z) [as in Abbott & Lucy 1985], which is not isotropic.

Typically, packets do work on the ejecta and so we expect on average that they lose energy during interactions.

wkerzendorf:
@ssim well interestingly enough the escaping packets seem to gain energy. 
So I agree that mu=sqrt(z), but presumed that means they are isotropically distributed in theta (mu = cos(theta)) - but maybe that is wrong.

ssim:
No - isotropic would not be sqrt! Isotropic would be mu = z (for positive hemisphere).

Overall, work should be done on the ejecta, but it may be that that mainly shows up in the backscattered packets. Also may depend on which frame you calculate the energies in.

wkerzendorf:
@ssim I just changed the wording. Let me know if we are ready to merge this.

PR #317, opened 2015-05-29
Finished plasma property tests for now.

aoifeboyle:
This PR includes a test for the calculation of tau sobolev values. It also finishes off incomplete tests that were previously marked to be skipped, and updates some old tests to allow them to make use of the new pytest fixtures in conftest. I'm happy for this PR to be merged if it passes the Travis tests.

### DIFF
--- a/tardis/plasma/properties/radiative_properties.py
+++ b/tardis/plasma/properties/radiative_properties.py
@@ -44,7 +44,7 @@ class TauSobolev(ProcessingPlasmaProperty):
 
     def get_g_lower(self, levels, lines_lower_level_index):
         if self._g_lower is None:
-            g_lower = np.array(levels.g.iloc[lines_lower_level_index],
+            g_lower = np.array(levels.g.ix[lines_lower_level_index],
                                      dtype=np.float64)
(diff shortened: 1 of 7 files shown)


wkerzendorf:
@aoifeboyle travis fails :wink:

aoifeboyle:
@wkerzendorf I know :\ Currently puzzling over why. I'll let you know when I manage to get it to pass!

tardis/plasma/properties/radiative_properties.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, plasma_parent):
 
     def get_g_lower(self, levels, lines_lower_level_index):
         if self._g_lower is None:
-            g_lower = np.array(levels.g.iloc[lines_lower_level_index],
+            g_lower = np.array(levels.g.ix[lines_lower_level_index],
```
aoifeboyle: @wkerzendorf This was the problem. This originally used .iloc rather than .ix. It turns out that .iloc causes errors in Pandas 0.12 that were fixed in version 0.13. But .ix works fine.

PR #318, opened 2015-05-31
[GSoC 2015 MRG] Testing getters and setters in montecarlo

kaushik94:


wkerzendorf:
@kaushik94 this looks very good! Well done! So I would just make sure that we try to avoid random numbers and maybe use linspace or something. I we want to use random numbers - let's make sure that we set the seed.

wkerzendorf:
@kaushik - after we resolved the random numbers issue, this is ready to be merged, right?

kaushik94:
@wkerzendorf  have to add a few more tests for getters and setters and then yes. And about the random function, I don't think its a good idea to set a few test cases and always run the tests on them. I think we should randomise them as much as possible but sure we should have a good seed function.

wkerzendorf:
I agree that we should test a large variety of floats. BUT if you use
random numbers - it does not become reproducible. So we need to set the
seed for the RNG 'np.random.seed'

wkerzendorf:
@kaushik94 so for a fixed seed number a random number generator will always produce the same sequence of "random" numbers. you can set the seed number using: `http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.seed.html` common seeds that we use in tardis can be found in the yaml file.

kaushik94:
@wkerzendorf take a look at the `test_rpacket.py`. Look at the first function. Is this what was in your mind ? How do we select the seed ?

wkerzendorf:
That looks good.

kaushik94:
@wkerzendorf  please review and merge.

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,146 @@
...
+
+# Testing functions without any parameters
+def test_rpacket_get_last_line():
+	assert tests.test_rpacket_get_last_line()
+
+def test_rpacket_get_close_line():
+	assert tests.test_rpacket_get_close_line()
+
+def test_rpacket_get_status():
+	assert tests.test_rpacket_get_status()
```
kaushik94: This part is a little tricky, I'll write the tests for this along with the ones in `cmontecarlo.c`.

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,146 @@
...
+MAX_FLOAT, MIN_FLOAT = floats[0], floats[3]
+MAX_INT = 10000
+MIN_INT = -10000
+
+tests = CDLL(test_path)
+SIZE = 10
+
+
+# Testing functions with float(C double) valued parameters
+np.random.seed(1)
```
wkerzendorf: I think you should set one random seed at the top and then be done with it. The sequence of random numbers will then always be the same for all tests.

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,146 @@
...
+from ctypes import CDLL, c_double
+
+import pytest
+import numpy as np
+
+from tardis import __path__ as path
+
+test_path = os.path.join(path[0], 'montecarlo', 'montecarlo.so')
+MAX_FLOAT, MIN_FLOAT = floats[0], floats[3]
+MAX_INT = 10000
```
wkerzendorf: use `pytest` fixtures for this, then we can parameterize in the future. global variables are a bad idea :wink:

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,146 @@
...
+	)
+def test_rpacket_get_nu_line(double_value):
+	assert tests.test_rpacket_get_d_electron(double_value)
+
+np.random.seed(9)
+@pytest.mark.parametrize(
+	"double_value", map(c_double, np.random.uniform(
+						MIN_FLOAT, MAX_FLOAT, size=SIZE))
+	)
+def test_rpacket_get_nu_line(double_value):
```
wkerzendorf: `test_rpacket_get_nu_line` was defined before two times. Have a look some of these are not quite correct.

wkerzendorf:
@kaushik94 we should definitely chat about how we want to proceed.

kaushik94:
@wkerzendorf sure. Let me know the time. I'll make the changes in a few hours.

kaushik94:
@wkerzendorf I think now we don't need fixtures right ? I am not using any global variables.

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,126 @@
...
+
+test_path = os.path.join(path[0], 'montecarlo', 'montecarlo.so')
+
+tests = CDLL(test_path)
+SIZE = 10
+
+np.random.seed(1)
+doubles = map(c_double, np.random.uniform(
+			floats[3], floats[0], size=SIZE))
+unsigned_integers = np.random.randint(0, 10000, size=SIZE)
```
wkerzendorf: these are global variables.

wkerzendorf:
@kaushik94 I think we are close. Let's do the last thing with fixtures and then we can merge this. What's with the other tests for the montecarlo part?

kaushik94:
@wkerzendorf how should I move from here. This test will fail. It can't find `MIN_INT` and `MAX_INT`. How can I let parametrize know their values ? How should I move from here. Regarding the other tests in montecarlo I have a few things to discuss with you, mostly they'll will be done in a week.

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,148 @@
...
+
+from tardis import __path__ as path
+
+test_path = os.path.join(path[0], 'montecarlo', 'montecarlo.so')
+
+tests = CDLL(test_path)
+
+np.random.seed(1)
+@pytest.fixture
+def doubles():
```
wkerzendorf: you need to return something. I'll put an example down a few lines down

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,148 @@
...
+	from sys import float_info as floats
+	MAX_INT, MIN_INT = floats[0], floats[3]
+	SIZE = 10
+
+
```
wkerzendorf: here's an example of fixtures:

```python

@pytest.yield_fixture
def double_value():
    for double_value in np.random.uniform(MIN_INT, MAX_INT, size=10):
        yield double_value
```

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,148 @@
...
+	MAX_INT, MIN_INT = floats[0], floats[3]
+	SIZE = 10
+
+	
+
+unsigned_integers = np.random.randint(0, 10000, size=10)
+integers = np.random.randint(-10000, 10000, size=10)
+
+# Testing functions with float(C double) valued parameters
+@pytest.mark.parametrize(
```
wkerzendorf: then you can get rid of this parametrize

tardis/tests/test_rpacket.py
kaushik94: Tried that but fixtures allow only a single `yield`. https://pytest.org/latest/yieldfixture.html

kaushik94: But why are global variables bad ?

wkerzendorf: Yes - you are absolutely right with the yield. So we should parameterize our fixtures using the `request.param` method. Have a look here: https://pytest.org/latest/example/parametrize.html

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,112 @@
...
+
+tests = CDLL(test_path)
+
+np.random.seed(1)
+
+def generate_testcases_doubles(metafunc):
+	from sys import float_info as floats
+	MAX_FLOAT, MIN_FLOAT = floats[0], floats[3]
+	for each in metafunc.fixturenames:
+		if 'double_value' in inspect.getargspec(each):
```
kaushik94: This is supposed to see if `double_value` is in any of the func's parameters, but it doesn't work.

tardis/tests/test_rpacket.py
```diff
@@ -0,0 +1,112 @@
...
+			metafunc.parametrize("double_value", map(c_double, 
+				np.random.uniform(MIN_FLOAT, MAX_FLOAT, size=10)))
+
+def generate_testcases_integers(metafunc):
+	MAX_INT, MIN_INT = 10000, -10000
+	if 'int_value' in metafunc.fixturenames:
```
kaushik94: even this doesn't work.

kaushik94:
@wkerzendorf can you suggest me how to move from here. I couldn't get the test cases run.

tardis/tests/test_rpacket.py
wkerzendorf: you're not using fixtures quite right. a fixture is an input value to a test suite. I'll make a PR to your PR.

wkerzendorf:
can you remove the `__author__` from the `__init__.py` file - sorry forgot about that.

wkerzendorf:
@kaushik94 Let's what else needs to be done on this PR?

kaushik94:
@wkerzendorf I don't think much. The rest of the functions go into another PR and we will be done with the straight forward tests for montecarlo. I would like to add a few examples like the ones mentioned in your paper as tests(end-to-end). What do you think ?

kaushik94:
@wkerzendorf I have to resolve the merge conflicts and then its ready for merge I guess

kaushik94:
closing in favour of #326

PR #319, opened 2015-06-01
Work in progress: Replacing plasma_array with LegacyPlasma.

aoifeboyle:
*Don't Merge*
I'm trying to replace the plasma_array file with the new legacy plasma in the model file, but I'm encountering an error related to the montecarlo stuff. It looks to me like one of the plasma properties is in the wrong format for the montecarlo part of the code, but I'm not sure which one or what the correct format should be. It might be the electron density property that I have just added that is causing this. I don't know much about the C part of Tardis so if anyone knows how to fix this that would be great. Here's the error:
error
@wkerzendorf

### DIFF
--- a/tardis/plasma/properties/ion_population.py
+++ b/tardis/plasma/properties/ion_population.py
@@ -8,7 +8,7 @@ from tardis.plasma.exceptions import PlasmaIonizationError
 logger = logging.getLogger(__name__)
 
 __all__ = ['PhiSahaLTE', 'RadiationFieldCorrection',
-           'IonNumberDensity', ]
+           'IonNumberDensity', 'ElectronDensity']
 
 class PhiSahaNebular(ProcessingPlasmaProperty):
     """
@@ -243,3 +243,12 @@ class IonNumberDensity(ProcessingPlasmaProperty):
             n_electron = 0.5 * (new_n_electron + n_electron)
 
         return ion_number_density
+
+class ElectronDensity(ProcessingPlasmaProperty):
+    name = 'electron_densities'
+
+    def calculate(self, ion_number_density):
+        ion_numbers = ion_number_density.index.get_level_values(1).values
+        ion_numbers = ion_numbers.reshape((ion_numbers.shape[0], 1))
+        n_electron = (ion_number_density.values * ion_numbers).sum(axis=0)
+        return n_electron
--- a/tardis/plasma/properties/ion_population.py
+++ b/tardis/plasma/properties/ion_population.py
@@ -222,7 +222,6 @@ class IonNumberDensity(ProcessingPlasmaProperty):
         n_e_convergence_threshold = 0.05
         n_electron = number_density.sum(axis=0)
         n_electron_iterations = 0
-        new_n_electron = np.zeros_like(n_electron)
         while True:
             ion_number_density = self.calculate_with_n_electron(
                 phi, partition_function, number_density, n_electron)
--- a/tardis/plasma/properties/ion_population.py
+++ b/tardis/plasma/properties/ion_population.py
@@ -219,10 +219,11 @@ class IonNumberDensity(ProcessingPlasmaProperty):
         return ion_populations
 
     def calculate(self, phi, partition_function, number_density):
+        n_e_convergence_threshold = 0.05
         n_electron = number_density.sum(axis=0)
         n_electron_iterations = 0
         new_n_electron = np.zeros_like(n_electron)
-        while not np.allclose(new_n_electron, n_electron):
+        while True:
             ion_number_density = self.calculate_with_n_electron(
                 phi, partition_function, number_density, n_electron)
             ion_numbers = ion_number_density.index.get_level_values(1).values
@@ -240,6 +241,10 @@ class IonNumberDensity(ProcessingPlasmaProperty):
                             ' something is probably wrong'.format(
                     n_electron_iterations))
 
+            if np.all(np.abs(new_n_electron - n_electron)
+                / n_electron < n_e_convergence_threshold):
+                break
+
             n_electron = 0.5 * (new_n_electron + n_electron)
 
         return ion_number_density
@@ -247,8 +252,18 @@ class IonNumberDensity(ProcessingPlasmaProperty):
 class ElectronDensity(ProcessingPlasmaProperty):
     name = 'electron_densities'
 
-    def calculate(self, ion_number_density):
-        ion_numbers = ion_number_density.index.get_level_values(1).values
-        ion_numbers = ion_numbers.reshape((ion_numbers.shape[0], 1))
-        n_electron = (ion_number_density.values * ion_numbers).sum(axis=0)
+    def calculate(self, ion_number_density, phi):
+        # Could check electron density for any element/ions in each zone, but
+        # if number density of element/ions was zero, it would not work.
+        # So setting this to calculate from the dominant ion in each zone.
+        dominant_ions = ion_number_density.sum(level=(0,1)).idxmax()
+        upper_ions = []
+        for ion in dominant_ions.values:
+            upper_ions.append((ion[0], ion[1]+1))
+        n_electron = []
+        for zone in range(len(ion_number_density.columns)):
+            n_electron.append(
+                (ion_number_density[zone].ix[dominant_ions[zone]] /
+                ion_number_density[zone].ix[upper_ions[zone]]) *
+                phi.ix[upper_ions[zone]][zone])
         return pd.Series(n_electron, index=np.arange(0, len(n_electron)))
--- a/tardis/plasma/properties/ion_population.py
+++ b/tardis/plasma/properties/ion_population.py
@@ -251,4 +251,4 @@ class ElectronDensity(ProcessingPlasmaProperty):
         ion_numbers = ion_number_density.index.get_level_values(1).values
         ion_numbers = ion_numbers.reshape((ion_numbers.shape[0], 1))
         n_electron = (ion_number_density.values * ion_numbers).sum(axis=0)
-        return n_electron
+        return pd.Series(n_electron, index=np.arange(0, len(n_electron)))
--- a/tardis/plasma/standard_plasmas.py
+++ b/tardis/plasma/standard_plasmas.py
@@ -1,25 +1,71 @@
 import logging
+import pandas as pd
+import numpy as np
 
 from tardis.plasma import BasePlasma
 from tardis.plasma.properties.property_collections import (basic_inputs,
-    lte_processing_properties)
+    basic_properties, lte_excitation_properties, lte_ionization_properties)
 
 logger = logging.getLogger(__name__)
 
 class LTEPlasma(BasePlasma):
 
     def __init__(self, t_rad, abundance, density, time_explosion, atomic_data):
-        plasma_modules = basic_inputs + lte_processing_properties
+        plasma_modules = basic_inputs + basic_properties + \
+            lte_excitation_properties + lte_ionization_properties
 
         super(LTEPlasma, self).__init__(plasma_modules=plasma_modules,
-                                        t_rad=t_rad, abundance=abundance,
-                                        atomic_data=atomic_data,
-                                        density=density,
-                                        time_explosion=time_explosion)
+            t_rad=t_rad, abundance=abundance, atomic_data=atomic_data,
+            density=density, time_explosion=time_explosion)
 
-class LegacyPlasma(BasePlasma):
+class LegacyPlasmaArray(BasePlasma):
 
-    def __init__(self, number_densities, atom_data, time_explosion,
-                 delta_treatment=None, nlte_config=None, ionization_mode='lte',
-                 excitation_mode='lte'):
-        pass
+    def from_number_densities(self, number_densities, atomic_data):
+        atomic_mass = atomic_data.atom_data.ix[number_densities.index].mass
+        elemental_density = number_densities.multiply(atomic_mass,
+            axis='index')
+        density = elemental_density.sum()
+        abundance = pd.DataFrame(elemental_density/density,
+            index=number_densities.index, columns=number_densities.columns,
+            dtype=np.float64)
+        return abundance, density
+
+    def initial_t_rad(self, number_densities):
+        return np.ones(len(number_densities.columns)) * 10000
+
+    def update_radiationfield(self, t_rad, ws=None, j_blues=None,
+        t_electrons=None, n_e_convergence_threshold=0.05,
+        initialize_nlte=False):
+        self.update(t_rad=t_rad)
+
+    def __init__(self, number_densities, atomic_data, time_explosion,
+        t_rad=None, delta_treatment=None, nlte_config=None,
+        ionization_mode='lte', excitation_mode='lte'):
+
+        plasma_modules = basic_inputs + basic_properties
+        if excitation_mode == 'lte':
+            plasma_modules += lte_excitation_properties
+        elif excitation_mode == 'dilute-lte':
+            raise NotImplementedError('Sorry ' + excitation_mode +
+                ' not implemented yet.')
+        if ionization_mode == 'lte':
+            plasma_modules += lte_ionization_properties
+        elif ionization_mode == 'nebular':
+            raise NotImplementedError('Sorry ' + ionization_mode +
+                ' not implemented yet.')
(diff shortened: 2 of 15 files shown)


wkerzendorf:
It is very likely an issue that you are returning a numpy array - but tardis expects a pandas dataframe

tardis/plasma/properties/ion_population.py
```diff
@@ -243,3 +243,12 @@ def calculate(self, phi, partition_function, number_density):
...
 
         return ion_number_density
+
+class ElectronDensity(ProcessingPlasmaProperty):
+    name = 'electron_densities'
+
+    def calculate(self, ion_number_density):
+        ion_numbers = ion_number_density.index.get_level_values(1).values
+        ion_numbers = ion_numbers.reshape((ion_numbers.shape[0], 1))
+        n_electron = (ion_number_density.values * ion_numbers).sum(axis=0)
```
kaushik94: @aoifeboyle it could be because of this, though I am not sure. Just check the dimensions of `ion_number_density` ?

aoifeboyle:
@wkerzendorf @kaushik94 I tried returning a pandas dataframe for the electron density already but it didn't like that either. It gave me a different error. Also in the old plasma the electron density was returned as an array, not a dataframe. Are there any other alternative formats for the data that might work?

aoifeboyle:
@wkerzendorf @kaushik94 I think a pandas Series might just have worked. I didn't think there was a difference between a Series and DataFrame apart from dimensions, but it looks like there is.

aoifeboyle:
@wkerzendorf @kaushik94 That was the last little glitch - Tardis is now working with the new plasma structure (for LTE). :) I've updated it here but don't merge yet - I want to add some tests etc. first. Thanks for your help!

wkerzendorf:
@aoifeboyle well done!! @ssim have a look. I am also surprised that `Series` vs `DataFrame`. @aoifeboyle Do we get the same result?

wkerzendorf:
@aoifeboyle this suggests that we should make sure that we document somewhere what needs to be what. let me know what ideas you have.

tardis/plasma/standard_plasmas.py
```diff
@@ -1,25 +1,71 @@
...
+            raise NotImplementedError('Sorry, delta treatment not implemented \
+                yet')
+
+        if t_rad is None:
+            t_rad = self.initial_t_rad(number_densities)
+
+        abundance, density = self.from_number_densities(number_densities,
+            atomic_data)
+
+        super(LegacyPlasmaArray, self).__init__(plasma_modules=plasma_modules,
```
wkerzendorf: very good!

aoifeboyle:
@wkerzendorf Just about to run a comparison simulation now (my computer's really slow). I'll let you know when I've compared the results! I just got excited when it ran so I put it up. I'll put documenting the correct format for each of the plasma properties on my to-do list :)

ssim:
@aoifeboyle @wkerzendorf Making progress - good stuff! But seems that Travis is not super happy?!

aoifeboyle:
@ssim I've had to finish up for today but I'll figure out what's wrong there tomorrow morning. It builds fine on my computer so it's probably just something minor!

aoifeboyle:
@wkerzendorf @ssim The current travis problem seems to just be that I haven't added transition probabilities for the legacy plasma yet. I don't need the transition probabilities for scatter so I will add it soon but I just have a few other things I'm concerned about first. I noticed that the montecarlo module is giving me some weird results for the new legacy plasma in some cases. Is there an easy way to debug the Cython stuff Wolfgang? Something like the ipython debugger? I want to check that the values are being read in from the legacy plasma ok. It's weird because the code runs without any errors but I get different answers for the two plasmas :(

wkerzendorf:
@aoifeboyle well, I think it would be great if we can make sure that the transition probabilities are not needed. The code is supposed to be modular and should work with only scattering. This will be even more important when we add more atomic data that might not be in every dataset. Can you look into that and tell me about any ideas. 

The Cython/C problem - well that is exactley what @kaushik94 is doing. We should make sure that we have checks and balances in place that will test that the plasma properties are transferred in a good way. It's not trivial to debug this, but @kaushik94 and you should work on this - because his job is to make a testing framework. Keep me in the loop at what problems you are encountering - now begins the difficult part of the journey :wink:

aoifeboyle:
Thanks @wkerzendorf. I will have a look at the transition probabilities soon. Still working on figuring out this montecarlo thing, but I think I just made some progress. I'll contact you later @kaushik94 if I don't get any futher, if that's ok?

aoifeboyle:
Hi @kaushik94 and @wkerzendorf. Sorry this is going to be long but if either of you could give me any suggestions/solutions to this when you get time I'd be delighted. I've spent the last two days trying to figure this out so I think there's some gap in my programming knowledge that's preventing me from understanding it.

When I use my LegacyPlasma in place of the old plasma class (BasePlasmaArray), I get different new values for t_inner after the first iteration, because the montecarlo code returns very different values of j_estimators and nubar_estimators. In both cases, the values of plasma_array.tau_sobolevs and plasma_array.electron_densities (which as far as I can tell are the only attributes used in the montecarlo stuff) are identical. If I create a BasePlasma object and update its tau_sobolevs and electron_densities properties to match those of the LegacyPlasma, I get the right answer, but if I use the LegacyPlasma itself I get totally different results. I *think* it has something to do with the structure of the tau_sobolevs data. I've noticed that if I create a BasePlasma object and update the tau_sobolevs property like:

self.plasma_array.tau_sobolevs.update(legacy_plasma.tau_sobolevs)

and then pass it through the montecarlo code I get the expected answers. But if I update it as:

self.plasma_array.tau_sobolevs = legacy_plasma.tau_sobolevs

and then call the montecarlo code I get the wrong answers. That makes me think that it's got something to do with the structure of the tau_sobolevs dataframe in LegacyPlasma and how it is interpreted by the Cython code. But as far as I can see the two dataframes from LegacyPlasma and BasePlasma are the same. Both have the same dimensions, same data types, index, values, etc. So basically my question is whether either of you can tell why this is happening? The code runs fine and I don't get any errors from the montecarlo module but I get different answers.

wkerzendorf:
> Hi @kaushik94 and @wkerzendorf. Sorry this is going to be long but if either of you could give me any suggestions/solutions to this when you get time I'd be delighted. I've spent the last two days trying to figure this out so I think there's some gap in my programming knowledge that's preventing me from understanding it.
> 
> When I use my LegacyPlasma in place of the old plasma class (BasePlasmaArray), I get different new values for t_inner after the first iteration, because the montecarlo code returns very different values of j_estimators and nubar_estimators. In both cases, the values of plasma_array.tau_sobolevs and plasma_array.electron_densities (which as far as I can tell are the only attributes used in the montecarlo stuff) are identical. If I create a BasePlasma object and update its tau_sobolevs and electron_densities properties to match those of the LegacyPlasma, I get the right answer, but if I use the LegacyPlasma itself I get totally different results. I think it has something to do with the structure of the tau_sobolevs data. I've noticed that if I create a BasePlasma object and update the tau_sobolevs property like:
>

That is strange. 

> self.plasma_array.tau_sobolevs.update(legacy_plasma.tau_sobolevs)
> 
> and then pass it through the montecarlo code I get the expected answers. But if I update it as:
> 
> self.plasma_array.tau_sobolevs = legacy_plasma.tau_sobolevs
> 

to make sure that we are not dealing with a view. 

> and then call the montecarlo code I get the wrong answers. That makes me think that it's got something to do with the structure of the tau_sobolevs dataframe in LegacyPlasma and how it is interpreted by the Cython code. But as far as I can see the two dataframes from LegacyPlasma and BasePlasma are the same. Both have the same dimensions, same data types, index, values, etc. So basically my question is whether either of you can tell why this is happening? The code runs fine and I don't get any errors from the montecarlo module but I get different answers.

I think one of them might be a view: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy

I think this problem, shows very clearly that we need a good API for sending things between Python-C. Let me know if you have any ideas.

orbitfold:
What is all the data I would need to recreate this problem? How to run TARDIS etc.

orbitfold:
@aoifeboyle If I try to run tardis from this pull request I get the following error:

```
/usr/lib/python2.7/dist-packages/pkg_resources.py:1031: UserWarning: /home/vytas/.python-eggs is writable by group/others and vulnerable to attack when used with get_resource_filename. Consider a more secure location (set with .set_extraction_path or the PYTHON_EGG_CACHE environment variable).
  warnings.warn(msg, UserWarning)
tardis.io.config_reader - INFO - Reading Atomic Data from kurucz_atom_pure_simple.h5
tardis.atomic - INFO - Read Atom Data with UUID=ca2684c53cc511e39f0ec8bcc8a04795 and MD5=4cccdc9b4c09faf3e348a2c710fb1715
tardis.io.config_reader - INFO - "initial_t_inner" is not specified in the plasma section - initializing to 9933.95199592 K with given luminosity
tardis.io.config_reader - WARNING - No "species" given - ignoring other NLTE options given:
{   'classical_nebular': False, 'coronal_approximation': False}
tardis.io.config_reader - WARNING - No convergence criteria selected - just damping by 0.5 for w, t_rad and t_inner
tardis.simulation - INFO - Remaining run 20
tardis.model - INFO - Calculating J_blues for radiative_rates_type=dilute-blackbody
  File "/usr/local/bin/tardis", line 5, in <module>
    pkg_resources.run_script('tardis-sn==1.0.1', 'tardis')
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 528, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 1394, in run_script
    execfile(script_filename, namespace, namespace)
  File "/usr/local/lib/python2.7/dist-packages/tardis_sn-1.0.1-py2.7-linux-x86_64.egg/EGG-INFO/scripts/tardis", line 70, in <module>
    simulation.run_radial1d(radial1d_mdl)
  File "/usr/local/lib/python2.7/dist-packages/tardis_sn-1.0.1-py2.7-linux-x86_64.egg/tardis/simulation.py", line 27, in run_radial1d
    initialize_j_blues=initialize_j_blues)
  File "/usr/local/lib/python2.7/dist-packages/tardis_sn-1.0.1-py2.7-linux-x86_64.egg/tardis/model.py", line 334, in simulate
    self.update_plasmas(initialize_nlte=initialize_nlte)
  File "/usr/local/lib/python2.7/dist-packages/tardis_sn-1.0.1-py2.7-linux-x86_64.egg/tardis/model.py", line 241, in update_plasmas
    self.transition_probabilities = self.plasma_array.calculate_transition_probabilities()
  File "/usr/local/lib/python2.7/dist-packages/tardis_sn-1.0.1-py2.7-linux-x86_64.egg/tardis/plasma/base.py", line 26, in __getattr__
    super(BasePlasma, self).__getattribute__(item)
AttributeError: 'LegacyPlasmaArray' object has no attribute 'calculate_transition_probabilities'
```

aoifeboyle:
Hi @orbitfold, thanks for looking at it for me! Set your ionization and excitation modes to LTE and your line interaction type to scatter in your config. file. I've only developed the LegacyPlasma for those settings so far. When I was trying to figure out the problem, I changed the code at line 129 of model.py to create self.plasma_array with the old plasma class (BasePlasma from plasma_array), and then I set the code to break into the debugger just after the montecarlo_radial1d process that happens at line 352. Then in the debugger I created a second plasma by calling LegacyPlasma, and compared it to the first. When I set self.plasma_array equal to the LegacyPlasma (or if I called the LegacyPlasma in the first instance at line 129), and ran the montecarlo module stuff, I got very different results. Most of the packets emitted for LegacyPlasma had negative energies (which I think means they were re-absorbed), but most had positive energies for BasePlasma. 

Just reading through that stuff on views now @wkerzendorf, thanks!

aoifeboyle:
Sorry @orbitfold, should have said to set radiative rates type to 'blackbody'. I changed the name in the last commit.

aoifeboyle:
@wkerzendorf @orbitfold It's working! You were right about the view/copy stuff Wolfgang. The tau_sobolevs in LegacyPlasma was a view when it should have been a copy. Thanks so much for your help. Wish I'd asked for help sooner since I was never going to get that myself. So relieved I can actually go and work on something else now...

wkerzendorf:
I couldn't have given the help of you hadn't debugged the problem !! This
is great - we're moving very quick on this very important aspect of TARDIS.
Well done. Let's try to chat tomorrow if there's a need for a better Python
c API.

kaushik94:
Hi. Sorry I just saw this discussion. Btw great work @aoifeboyle , I will try to look into this stuff today. And @wkerzendorf I would also like to chip in to the discussion about a better C-Python API as I think I know a little bit about it, please let me know the time.

aoifeboyle:
Latest commit:
Added j_blues as plasma input.
Added montecarlo capabilities (new plasma properties added: beta_sobolev, transition_probabilities, j_blues).
Made stimulated emission factor a separate property (as it contributes to both tau_sobolev and transition probabilities).
Seems to run fine and converge on the right values, but I'd like to add some more tests and documentation before merging. I also need to see how the Travis tests turn out (this update *should* fix the transition_probabilities error we were having, but there could be other issues).

wkerzendorf:
@aoifeboyle it all looks very nice and goes in the right direction. I think we should have a quick chat next week.

aoifeboyle:
@wkerzendorf Sure! Let me know when suits you.

aoifeboyle:
@wkerzendorf, I'm getting an error from Travis because the spectral luminosity does not exactly match what it used to be. I've been investigating and I think that's to be expected for the following reason: In the old and new plasma structures, the ion number density convergence process is slightly different. In the old plasma, the initial electron density values tried for each iteration after the first are those that converged on the last iteration. In the new plasma, the initial electron density values tried in each iteration are just taken as a sum over the number density (as taking the converged electron densities from the previous iteration would create a property 'cycle', which is not allowed in the new structure). So although with both methods, the values of t_rads, for example, converge towards the same value over time, they don't reach exactly the same value, and therefore there is a slightly different output spectrum, which is causing the tests to fail. Subjectively, when I look at the output spectra, they seem the same, and I'm running with more packets for both plasma structures now just to compare them and be certain. I was just wondering if that all makes sense?

ssim:
In LTE the partition functions shouldn't have the dilution factor. It would be a good thing if the partition function calculation always really uses the same bits of code that actually calculate level populations - is that what's done?

aoifeboyle:
Yeah that's what I thought! Pretty much. In both the old and new plasma, a property is defined (called level population proportionalities in the old plasma and level boltzmann factor in the new), which is g*exp(-energy/kT). That property is then used to calculate the level populations from the ion populations, and summed over to calculate the partition functions. Is that the way that you mean?

ssim:
Yes, I think that's what I mean - that's a good way to do it: so that the same "proportionalities" are embedded in the partition functions and the level population calculations - that's how it should be really.

aoifeboyle:
Now that I think about it - the dilution factor could probably be included in the proportionalities property to prevent this kind of inconsistency from occurring.

ssim:
I think that's a good idea!

aoifeboyle:
@wkerzendorf Here's a comparison of the two spectra (virtual) after 30 iterations with 1.e+6 packets - one with the LegacyPlasma and one with the old plasma structure (can't remember which is which though). They seem pretty similar to me but let me know what you think! 
comparison_log.

aoifeboyle:
@wkerzendorf I skipped the test of the spectra that was causing the Travis failure above (because of the slightly different convergence patterns between the new and old plasma) and now Travis seems happy. I'm going to add some more comments and tests, and then I'd quite like to merge this.

ssim:
I think these seems sensible - the spectra above seem to agree to the level we would care about. Are you going to replace the spectrum test with a comparison to this one now, or what's the plan?

aoifeboyle:
Great! That's my plan, but I'll add it in the next PR along with some other tests I'm working on now. Would you be able to merge this PR @ssim? I want to get it merged and start a new PR because this one's got so long.

ssim:
I'm happy to merge. @wkerzendorf can you confirm that you're also happy for me to go ahead and merge this one in?

wkerzendorf:
@aoifeboyle I just talked to @ssim and I think it looks ready to be merged. Good work!

PR #320, opened 2015-06-02
virtual packets frequency cut

unoebauer:
Added possibility to emit virtual packets only in specified wavelength range


wkerzendorf:
@ssim This is the virtual packet cuts you suggested - builtin by @unoebauer 

Have a look at it.

wkerzendorf:
looks good - @ssim if there are no further comments - I'll merge

ssim:
Looks good to me!

PR #323, opened 2015-06-05
Gui package (WIP)

gautam1168:
Splitting the gui and making it into a package. The running instructions are changed as follows:

```python
    from tardis.gui import Tardis
    win = Tardis()
    win.show_model(mdl)
```


gautam1168:
Sorry! Wrong branch!

PR #324, opened 2015-06-06
Gui package (WIP)

gautam1168:
Convert the gui to a subpackage inside the tardis package. The running instructions will change to the following:

```python
    from tardis.gui import interface
    interface.show(model)
```

Now you can also run the gui from outside ipython. To do this navigate to the gui subpackage since the interface.py is there (this will be fixed soon so that it can be started from anywhere):

```bash
    $python interface.py yamlconfig.yml atomdata.hd5
```


gautam1168:
@wkerzendorf 
I have split the gui into three files and put them inside the subpackage. There is one bug due to which the line info tables are not showing up but its easy to fix and I will probably be able to sort it out by tomorrow. One good side-effect of this  splitting is that it made it really easy to enable the gui to be launched from outside the ipython environment. I have left instructions about how to do this at the top. So aside from fixing that bug, what else do you want me to change as a part of this PR?

gautam1168:
@wkerzendorf 
I fixed the bug. Everything is working now.

wkerzendorf:
unfortunately this is intertwined with the GUI stuff. Let's chat how to make this PR only about the package.

gautam1168:
@wkerzendorf  I have sorted out the conflicts. My local branch and this are now in sync. I have also added docstrings to the interface file and maintained the PEP specifications to the best of my knowledge. Please take a look. The running.rst and gui.rst will require yet another change after this is merged. Or should I update the instructions in this PR.

wkerzendorf:
Either way is fine.

tardis/gui/datahandler.py
```diff
@@ -0,0 +1,562 @@
...
+        No error is raised if the data list index is exceeded. A None is
+        returned in such cases.
+
+        """
+        try:
+            return self.data[i]
+        except IndexError:
+            return None
+
+    def getParent(self):
```
wkerzendorf: hi this should be `get_parent` as this is the PEP8 style.

tardis/gui/datahandler.py
```diff
@@ -0,0 +1,562 @@
...
+        """Return the parent of this node."""
+        return self.parent
+
+    def getIndexOfSelf(self):
```
wkerzendorf: same here `get_index_of_self`

gautam1168: Hi Wolfgang,

I'll make the changes. I'm sorry I'm just a little busy now. I'll try to
finish up this PR and prepare a demo of the new layout by the weekend.

Yours sincerely,
Gaurav Gautam

gautam1168:
@wkerzendorf I am changing the function names in the class I defined to be all small case and underscores for separation. But the other classes whitch are inheriting from qt classes like the `TreeModel` or the `SimpleTableModel` have to implement the functions that are defined in the super class. For example see this:

http://doc.qt.io/qt-5/qabstractitemmodel.html

So I cannot change the function names to the PEP specification in those classes. Considering this, would you say that I should change the function names in the classes that I have defined and do not inherit from qt classes, and leave the others unchanged, or do you think I should just leave the function names as they are, i.e. capitalized letters for separating words? Let me know so I can commit the changes.

wkerzendorf:
@gautam1168 I reopened it.

wkerzendorf:
@gautam1168 when you introduce new functions definitely try to stick with PEP8. But for `QT` stuff definitely try to stick with their camel case. I think for `QT` it's a legacy problem as most of this is written in Camel case in C.

gautam1168:
I have changed the function names and updated the documentation.

tardis/gui/datahandler.py
```diff
@@ -0,0 +1,562 @@
...
+        """Call constructor of the QAbstractTableModel and set parameters
+        given by user.
+        """
+        super(SimpleTableModel, self).__init__(parent, *args)
+        self.headerdata = headerdata
+        self.arraydata = []
+        self.iterate_header = iterate_header
+        self.index_info = index_info
+
+    def addData(self, datain):
```
wkerzendorf: method name still camelcase.

wkerzendorf: disregard - thanks for the explanation

gautam1168:
@wkerzendorf I have renamed all the functions that I could in handler.py and widgets.py. Interface.py has no functions so this covers all the files.

wkerzendorf:
@gautam1168 this is ready to be merged?

gautam1168:
Yes it is.

PR #332, opened 2015-06-15
Added dilute-lte and nebular ionisation to the new plasma.

aoifeboyle:
This works fine (I've compared the results from the new and old plasmas) but needs to be tidied up and have a few more properties added before merging.

Modules to change:
- [x] Split `LevelBoltzmannFactor` module into separate LTE and Dilute LTE functions (dilution factor now included in the Boltzmann factor rather than being added separately in both the partition function and level population calculations).
- [x] Replace `LTEPartitionFunction` module with general `PartitionFunction`.
- [x] Replace `LevelPopulationLTE` and `LevelPopulationDiluteLTE` with general `Level Population` module.

Modules to add:
- [x] `PhiSahaNebular`
- [x] `DilutionFactor`
- [x] `ZetaData`
- [x] `DeltaTreatment`
- [x] `TElectron`

Other tasks:
- [x] Add tests for all new modules.
- [x] Create an option to set `delta` from config. file.
- [x] Make `ZetaData` a `BaseAtomicDataProperty`.
- [x] Make `IonisationData` a `BaseAtomicDataProperty`.
- [x] Add warning for missing `ZetaData`.
- [x] Raise error for missing `IonizationData`.

### DIFF
--- a/tardis/plasma/properties/atomic.py
+++ b/tardis/plasma/properties/atomic.py
@@ -1,6 +1,7 @@
 from abc import ABCMeta, abstractmethod
 import numpy as np
 import pandas as pd
+from collections import Counter as counter
 
 from tardis.plasma.properties.base import ProcessingPlasmaProperty
 from tardis.plasma.exceptions import IncompleteAtomicData
@@ -30,17 +31,14 @@ class BaseAtomicDataProperty(ProcessingPlasmaProperty):
         if self.value is not None:
             return self.value
         else:
-            if not getattr(atomic_data, 'has_{0}'.format(
-                    self.name)):
-                raise IncompleteAtomicData(self.name)
-            else:
-                try:
-                    raw_atomic_property = getattr(atomic_data, '_' + self.name)
-                except:
-                    raw_atomic_property = getattr(atomic_data, self.name)
+            try:
+                raw_atomic_property = getattr(atomic_data, '_' + self.name)
                 return self._set_index(self._filter_atomic_property(
                     raw_atomic_property, selected_atoms), atomic_data)
-
+            except:
+                raw_atomic_property = getattr(atomic_data, self.name)
+                return self._set_index(self._filter_atomic_property(
+                    raw_atomic_property, selected_atoms), atomic_data)
 
 
 class Levels(BaseAtomicDataProperty):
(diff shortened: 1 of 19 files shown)


tardis/plasma/properties/ion_population.py
```diff
@@ -136,36 +137,31 @@ class RadiationFieldCorrection():
...
-            departure_coefficient = 1. / self.ws
-
-        ionization_data = self.atom_data.ionization_data
-
+    @staticmethod
+    def calculate(w, ionization_data, beta_rad, t_electron, t_rad,
+        beta_electron, levels):
+        # factor delta ML 1993
+        departure_coefficient = 1. / w
+        chi_0_species=(20, 2)
```
aoifeboyle: @wkerzendorf In the old plasma, `departure_coefficient` and `chi_0_species` were both arguments for this function, with these default values. But there didn't seem to be anywhere where they could actually be altered to be anything other than these values. Should I make them their own `ProcessingPlasmaProperty` objects, or should I just leave their values constant?

tardis/plasma/properties/ion_population.py
```diff
@@ -136,36 +137,31 @@ class RadiationFieldCorrection():
...
-            np.outer(ionization_data.ionization_energy.values[less_than_chi_0], self.beta_rads) -
-             chi_0*self.beta_electrons)
-
-        return pd.DataFrame(radiation_field_correction, columns=np.arange(len(self.t_rads)),
+            np.outer(ionization_data.ionization_energy.values[less_than_chi_0],
+                     beta_rad) - chi_0 * beta_electron)
+        delta = pd.DataFrame(radiation_field_correction,
+                            columns=np.arange(len(t_rad)),
                             index=ionization_data.index)
+        delta.drop([(1,1), (4,3)], inplace=True)
```
aoifeboyle: Something weird going on with indexing here. The `delta` dataframe is generated with two extra ions compared to the `phi` ones. I understand the (1,1) one because the ionization data isn't being filtered (something to change), but I don't understand where the (4,3) comes from.

tardis/plasma/properties/ion_population.py
wkerzendorf: Okay for these kinds of things, we want them to be parameters of the object, so put them in the `__init__` and them access them with `self`. Does this make sense?

tardis/plasma/properties/ion_population.py
wkerzendorf: huh. not sure.

tardis/plasma/properties/ion_population.py
aoifeboyle: Yes, will do.

tardis/plasma/properties/atomic.py
```diff
@@ -108,8 +109,17 @@ def calculate(self, atomic_data, selected_atoms):
...
 
-    def calculate(self, atomic_data, selected_atoms):
+    def calculate(self, atomic_data, levels):
         if self.value is not None:
             return self.value
         else:
             return atomic_data.ionization_data
+
+class ZetaData(ProcessingPlasmaProperty):
+    name = 'zeta_data'
```
aoifeboyle: @wkerzendorf I'd like to make `ZetaData` a `BaseAtomicDataProperty` so that it could be filtered in the same way as the other atomic properties. That would require explicitly naming the indices in the atomic datafiles as `atomic_number` and `ion_number`. Could we change that in current/future datasets? Also, in some datasets, e.g. the `included_he_atomic_data` included in `conftest.py`, some atomic properties (`lines`, `levels`) are named with an underscore in front. What's the reason for that? I've noticed it's true in some datasets but not in others I use. It's just important to know which one is the standard one so the plasma can find it.

wkerzendorf: @aoifeboyle yes, I think that sounds like a very good plan to make `ZetaData` standards compliant. 
they shouldn't have underscores, we need to check this out.

aoifeboyle: There aren't any underscores in the dataset I used for my paper, but there are in the testing one. Just for `_lines` and `_levels` from what I can see. Would you like me to change that and make a PR to the Tardis master branch? I think I know how (I modified an atomic dataset before to update the helium data). I can try to update the `ZetaData` stuff too?

wkerzendorf: I think they shouldn't be there, but we are in the process of changing the dataset anyways - so for now I would leave that. I think the `ZetaData` stuff is more sensible now.

aoifeboyle:
@wkerzendorf, what's the best complete atomic dataset to use? At the moment I'm testing using one called `kurucz_chianti_full_he.h5` but the number of zeta values provided for each element doesn't seem to be consistent with the number of ions. Also, Be only has 3 ions instead of 4, when all of the other elements have a number of ions equal to their atomic number. I've made `zeta_data` and `ionization_data` into `BaseAtomicPlasmaProperties` so they could be filtered according to the `selected_atoms` like the other atomic properties for consistency, but when I use this dataset the indices don't match up for the `zeta_data` and `ionization_data`, which means I can't multiply them.

wkerzendorf:
@ssim maybe this is a question for you.

ssim:
I'd suggest you try with the general release one:

https://tardis.readthedocs.org/en/latest/atomic/atomic_data.html

although that may have similar issues? Does it? @aoifeboyle

aoifeboyle:
@ssim, I'm afraid it does. I had a good look and here's a summary of issues:
* The final (most ionised) Be atom is missing, including in ionization_data, levels, lines and zeta data.
* All of the zeta data is missing for atomic numbers 29 and 30.
* There zeta data is missing for the final ions of atomic numbers 3, 4, 5 and 9+.
* Also, @wkerzendorf, the `_levels` and `_lines` data have underscores here too.

I don't think any of this is crucial to the old plasma, but I really need consistent data to ensure I get the indexing right in the new plasma because of the filtering process. I'm pretty sure I know how to update these files but could someone point me in the right direction for finding the missing data? I can just fill in 1s for the zeta values if it's not too important.

ssim:
Pragmatically, I would say that setting all the zetas to 1 where we don't have more information would be fine. Is that ok with you too @wkerzendorf ?

Strange that Be III is missing. @wkerzendorf and I should look into that in case it suggests that something is wrong. That ion is completely unimportant for us though...but for consistency I guess we should sort it out.

aoifeboyle:
Ok, well I've added all the extra zetas as ones, and I've added the Be III ion with just its ground state for a level. I think that'll do fine for my purposes anyway. This means that when we merge the plasma/restructure branch with the master branch it won't work with the current atomic dataset and we'll have to replace it with the one I've modified. Is that ok @wkerzendorf?

wkerzendorf:
@aoifeboyle hmm - not sure what's going on. I need to look at this.

wkerzendorf:
if Be is not specified in the tardis yaml file it should not be loaded, right?

aoifeboyle:
@wkerzendorf Well, at present, the `ionization_data` for Be is loaded even if it is not a specified atom in the yaml file, but the other pieces of data for Be are not. That's what I'm hoping to change. The reason I want to change it is that I'm having difficulty implementing the `radiation_field_correction` because that particular property is based on the `ionization_data`, and then it ends up with different dimensions to the `phi` data so I can't multiply them. So I want to make the `ionization_data` a `BaseAtomicDataProperty` to allow it to be filtered in the same way as the other atomic data. But there's a missing ion in that data, which leads us to the current problem. And I just noticed when I was investigating that line of the code that the `zeta_data` had the wrong dimensions too. I've been trying to modify the atomic dataset the same way as I did last year, which is sort of manually rather than using one of the scripts. But the code isn't happy with it so I'm downloading tardisatomic now and I'll try again with it.

wkerzendorf:
@aoifeboyle I think it is important that we can deal with missing data (maybe raise an exception when we request to do something, for which the atomic data is missing). This can definitely all be embedded in an `BaseAtomicDataProperty`

aoifeboyle:
Ok, so to deal with the Be issue I'll implement a test to check that all of the ions are there for the `selected_atoms`. And for the `zeta_data`, I can just add something to the function to check for missing zetas and then add them in as ones (and raise a warning). Would that be alright? I do like the idea of not being so dependent on the right atomic data.

wkerzendorf:
Thats great!!

wkerzendorf:
very likely because we never include copper or zinc in the runs. I think in the end, we should raise an exception. We can always later make that possible to be switched off.

aoifeboyle:
@wkerzendorf, merge this if you're happy with it all.

wkerzendorf:
@aoifeboyle Looks good - merge after TRAVIS passes.

PR #333, opened 2015-06-16
Plasma/montecarlo

aoifeboyle:
Added MontecarloRunner class.


wkerzendorf:
@aoifeboyle yes, the montecarlo runner class is somewhat what I imagine, but let's get this in a different PR.

aoifeboyle:
@wkerzendorf Ok - will do once I figure out why Travis doesn't like it!

PR #330, opened 2015-06-12
[GSoC 2015 WIP] Testing other montecarlo functions

kaushik94:
@wkerzendorf is this how it should be done ?

PR #334, opened 2015-06-16
[GSoC 2015 WIP] test_montecarlo_line_scatter

kaushik94:
@aoifeboyle, @wkerzendorf  ping !

### DIFF
--- a/tardis/montecarlo/src/test_montecarlo_line_scatter.c
+++ b/tardis/montecarlo/src/test_montecarlo_line_scatter.c
@@ -71,7 +71,7 @@ init_storage_model(void){
 double
 test_compute_distance2boundary(){
 	/*
-	* distance to boundary = -0.0109999775
+	* this returns nan
 	*/
 	double D_BOUNDARY = compute_distance2boundary(rp, sm);
 	rpacket_set_d_boundary(rp, D_BOUNDARY);
--- a/tardis/montecarlo/src/test_montecarlo_line_scatter.c
+++ b/tardis/montecarlo/src/test_montecarlo_line_scatter.c
@@ -8,46 +8,53 @@
 rpacket_t * rp;
 storage_model_t * sm;
 
-bool test_montecarlo_line_scatter(void);
+double TIME_EXPLOSION = 51840000.0;
+double R_INNER_VALUE = 622080000.0;
+double R_OUTER_VALUE = 7776600000.0;
+
+void init_rpacket(void);
+void init_storage_model(void);
+double test_compute_distance2boundary(void);
 
 /* initialise RPacket */
 void
 init_rpacket(void){
 	rp = (rpacket_t *) malloc(sizeof(rpacket_t));
 	
-	double MU = 1.5*pow(10, 5);
-	double R = 3.3*pow(10, 3);
-	double TAU_EVENT = 3.2*pow(10, 4);
-	double NU_LINE = 1.5;
-	double ENERGY = 10.0;
-	double D_BOUNDARY = 5.8;
-	double D_ELECTRON = 4.3;
-	double D_LINE = 3.9;
-	unsigned int NEXT_LINE_ID = 1;
+	double MU = 0.3;
+	double R = 13*pow(10, 3)*TIME_EXPLOSION;
+	double ENERGY = 0.9;
 
+	rpacket_set_current_shell_id(rp, 0);
 	rpacket_set_mu(rp, MU);
 	rpacket_set_r(rp, R);
+	rpacket_set_recently_crossed_boundary(rp, 1);
+
+	/*
 	rpacket_set_virtual_packet(rp, 1);
 	rpacket_set_current_shell_id(rp, 1);
 	rpacket_set_tau_event(rp, TAU_EVENT);
-	rpacket_set_recently_crossed_boundary(rp, 1);
 	rpacket_set_nu_line(rp, NU_LINE);
 	rpacket_set_last_line(rp, true);
 	rpacket_set_energy(rp, ENERGY);
 	rpacket_set_status(rp, TARDIS_PACKET_STATUS_IN_PROCESS);
-	rpacket_set_d_boundary(rp, D_BOUNDARY);
 	rpacket_set_d_electron(rp, D_ELECTRON);
 	rpacket_set_d_line(rp, D_LINE);
 	rpacket_set_next_line_id(rp, NEXT_LINE_ID);
+	*/
+
 }
 
 /* initialise storage model */
 void
 init_storage_model(void){
 	sm = (storage_model_t *) malloc(sizeof(storage_model_t));
-	double R_OUTER[2] = {1.2, 2.6};
-	double R_INNER[2] = {2.3, 3.4};
-	double TIME_EXPLOSION = 36.0;
+	double R_OUTER[2] = {R_OUTER_VALUE, R_OUTER_VALUE};
+	double R_INNER[2] = {R_INNER_VALUE, R_INNER_VALUE};
+	sm->r_outer = R_OUTER;
+	sm->r_inner = R_INNER;
+	
+	/*
 	double LINE_LIST_NU[2] = {1.5, 3.2};
 	double LINE_LISTS_J_BLUES[2] = {5.6, 4.4};
 
@@ -58,13 +65,15 @@ init_storage_model(void){
 	sm->r_inner = R_INNER;
 	sm->line_list_nu = LINE_LIST_NU;
 	sm->line_lists_j_blues = LINE_LISTS_J_BLUES;
+	*/
 }
 
-bool
-test_montecarlo_line_scatter(){
-	double DISTANCE = 10.5;
-	montecarlo_line_scatter(rp, sm, DISTANCE);
-
-	/* Return true if the test passes */
-	return true;
+double
+test_compute_distance2boundary(){
+	/*
+	* distance to boundary = -0.0109999775
+	*/
+	double D_BOUNDARY = compute_distance2boundary(rp, sm);
+	rpacket_set_d_boundary(rp, D_BOUNDARY);
+	return D_BOUNDARY;
 }
--- a/tardis/montecarlo/src/test_montecarlo_line_scatter.c
+++ b/tardis/montecarlo/src/test_montecarlo_line_scatter.c
@@ -8,9 +8,9 @@
 rpacket_t * rp;
 storage_model_t * sm;
 
-double TIME_EXPLOSION = 51840000.0;
-double R_INNER_VALUE = 622080000.0;
-double R_OUTER_VALUE = 7776600000.0;
+double TIME_EXPLOSION =     51840000.0;
+double R_INNER_VALUE =  622080000000.0; /* 12,000xTIME_EXPLOSION */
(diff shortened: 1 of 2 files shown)


tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,70 @@
...
+	rpacket_set_next_line_id(rp, NEXT_LINE_ID);
+}
+
+/* initialise storage model */
+void
+init_storage_model(void){
+	sm = (storage_model_t *) malloc(sizeof(storage_model_t));
+	double R_OUTER[2] = {1.2, 2.6};
+	double R_INNER[2] = {2.3, 3.4};
+	double TIME_EXPLOSION = 36.0;
```
wkerzendorf: okay let's try to set sensible value: `TIME_EXPLOSION` = 10 days in seconds

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,70 @@
...
+	rpacket_set_d_line(rp, D_LINE);
+	rpacket_set_next_line_id(rp, NEXT_LINE_ID);
+}
+
+/* initialise storage model */
+void
+init_storage_model(void){
+	sm = (storage_model_t *) malloc(sizeof(storage_model_t));
+	double R_OUTER[2] = {1.2, 2.6};
+	double R_INNER[2] = {2.3, 3.4};
```
wkerzendorf: `R_INNER` = 10000, 13000 km / s * `TIME_EXPLOSION
`R_INNER` = 13000, 16000 km / s * `TIME_EXPLOSION

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,70 @@
...
+storage_model_t * sm;
+
+bool test_montecarlo_line_scatter(void);
+
+/* initialise RPacket */
+void
+init_rpacket(void){
+	rp = (rpacket_t *) malloc(sizeof(rpacket_t));
+	
+	double MU = 1.5*pow(10, 5);
```
wkerzendorf: `MU` is between -1, 1

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,70 @@
...
+	double R = 3.3*pow(10, 3);
```
wkerzendorf: `R` is between one `R_INNER` and `R_OUTER`

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,70 @@
...
+	double TAU_EVENT = 3.2*pow(10, 4);
+	double NU_LINE = 1.5;
+	double ENERGY = 10.0;
```
wkerzendorf: ENERGY between 0.8 and 1.1

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,70 @@
...
+	double D_BOUNDARY = 5.8;
```
wkerzendorf: `D_BOUNDARY` you should calculate with the c function

kaushik94:
@wkerzendorf  done, what about the storage model ?

kaushik94:
@wkerzendorf  the `compute_distance2boundary` function is returning `nan` with the given values, please check if I did it the right way ?

wkerzendorf:
@kaushik94 great this should also be done in the other tests. I think we are getting dangerously close to the C-Python API that I was talking about :wink:

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,79 @@
...
+	sm->r_inner = R_INNER;
+	sm->line_list_nu = LINE_LIST_NU;
+	sm->line_lists_j_blues = LINE_LISTS_J_BLUES;
+	*/
+}
+
+double
+test_compute_distance2boundary(){
+	/*
+	* distance to boundary = -0.0109999775
```
wkerzendorf: isn't it returning -0.0109999 ?

kaushik94:
> @kaushik94 great this should also be done in the other tests. I think we are getting dangerously close to the C-Python API that I was talking about

Yes but first will have to figure what should be done to this function. If I can figure that out, others will be relatively easy. This is returning `nan`, can you figure out why ?

tardis/montecarlo/src/test_montecarlo_line_scatter.c
kaushik94: Sorry forgot to remove that comment. It was for the previous PR. Now its not, pull my branch into your local, its returning `nan`

wkerzendorf:
Well have a look what the code actually does. It's relatively simple math. Make sure that the values used are within their parameters. Mu only goes from -1,1 the current r needs to be within the rinner and router.

wkerzendorf:
@aoifeboyle thanks. @kaushik94 and @aoifeboyle  have a look at http://tardis.readthedocs.org/en/latest/physics/montecarlo.html#montecarlo-geometry

wkerzendorf:
that describes what it should do.

kaushik94:
Figured out the issue. http://www.tutorialspoint.com/cprogramming/c_data_types.htm.
```
double d_outer =
    sqrt (r_outer * r_outer + ((mu * mu - 1.0) * r * r)) - (r * mu);
```
We are multiplying 12 decimal doubles with 12 decimal doubles, I don't think c offers enough space to do that. Might have to switch to `long long double`s

kaushik94:
I could be wrong, just guessing.

wkerzendorf:
@kaushik94 there's no need for longs: this works easily with doubles.

wkerzendorf:
@kaushik94 the largest numbers we can store as doubles is `10 ^ 308` this is a 1 with 308 zeros.

kaushik94:
@wkerzendorf  still the same error.

kaushik94:
@aoifeboyle thanks. See if `compute_distance2boundary` is getting all the values required. It could also be because of that.

aoifeboyle:
Yeah that's what I'm looking at now! My knowledge of C isn't so great so give me a while :P

aoifeboyle:
@kaushik94 I have to leave this for tonight because it's quite late, but I'll get back on it tomorrow. The problem is that the `r_outer` and `r_inner` are both 0 for some reason. That means that `d_outer` ends up involving the square root of a negative number, creating the `nan` value. So it's something to do with how `r_outer` and `r_inner` are being set.

tardis/montecarlo/tests/test_montecarlo_line_scatter.py
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+
+from tardis import __path__ as path
+
+test_path = os.path.join(path[0], 'montecarlo', 'montecarlo.so')
+
+tests = CDLL(test_path)
+
+def test_compute_distance2boundary():
+	tests.test_compute_distance2boundary.restype = c_longdouble
```
aoifeboyle: This should be `c_double` instead of `c_longdouble`.

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,79 @@
...
+
+/* initialise storage model */
+void
+init_storage_model(void){
+	sm = (storage_model_t *) malloc(sizeof(storage_model_t));
+	double R_OUTER[2] = {R_OUTER_VALUE, R_OUTER_VALUE};
+	double R_INNER[2] = {R_INNER_VALUE, R_INNER_VALUE};
+	sm->r_outer = R_OUTER;
+	sm->r_inner = R_INNER;
+
```
aoifeboyle: @kaushik94 I think the reason that `r_inner` and `r_outer` in the `calculate_distance2boundary` function are being assigned a value of 0 is that for some reason `sm->r_outer` is actually a pointer. When I change it so that `*sm->r_outer = R_OUTER_VALUE`, etc., the correct value can be obtained in `calculate_distance2boundary` like `double r_outer = *storage->r_outer`, and the function returns a value rather than nan, which looks correct on first glance. But there must be a proper solution somewhere that doesn't involve altering the function being tested itself.

tardis/montecarlo/tests/test_montecarlo_line_scatter.py
```diff
@@ -0,0 +1,20 @@
+import os
+from ctypes import CDLL, c_double, c_longdouble
```
wkerzendorf: reiterated from @aoifeboyle we do not need a c_longdouble. As I mentioned, double is more than big enough to store our values.

tardis/montecarlo/src/test_montecarlo_line_scatter.c
kaushik94: No I think we shouldn't do that. I think we'll give some time and figure this out tonight ! If that doesn't work, we'll go about changing the actual function.

tardis/montecarlo/tests/test_montecarlo_line_scatter.py
kaushik94: Ohh yes, sorry I updated in my local, just forgot to push those changes. Trying out some stuff, will push in a while @wkerzendorf .

kaushik94:
@wkerzendorf , @aoifeboyle solved. We din't assign space for the `sm->r_inner` and `sm->r_outer`. Now the tests pass.

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,83 @@
...
+/* initialise storage model */
+void
+init_storage_model(void){
+	sm = (storage_model_t *) malloc(sizeof(storage_model_t));
+	double R_OUTER[2] = {R_OUTER_VALUE, R_OUTER_VALUE};
+	double R_INNER[2] = {R_INNER_VALUE, R_INNER_VALUE};
+	int NUMBER_OF_PACKETS = 2;
+	sm->r_outer = (double *) malloc(sizeof(double)*NUMBER_OF_PACKETS);
+	sm->r_outer[0] = R_OUTER_VALUE;
```
kaushik94: We have to assign values like this.

tardis/montecarlo/src/test_montecarlo_line_scatter.c
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/* initialise storage model */
+void
+init_storage_model(void){
+	sm = (storage_model_t *) malloc(sizeof(storage_model_t));
+	double R_OUTER[2] = {R_OUTER_VALUE, R_OUTER_VALUE};
+	double R_INNER[2] = {R_INNER_VALUE, R_INNER_VALUE};
+	int NUMBER_OF_PACKETS = 2;
+	sm->r_outer = (double *) malloc(sizeof(double)*NUMBER_OF_PACKETS);
```
wkerzendorf: I think, we might want to skype on this.

kaushik94: Okay ping me.

kaushik94:
test added to #330

PR #337, opened 2015-06-18
MontecarloRunner

aoifeboyle:
Don't merge yet. 

Added MontecarloRunner in separate PR as requested. Works fine for scatter but I get a 'segmentation fault' error when I run it for macroatom. I think it has something to do with the way I'm passing in the `transition_probabilities` but I can't see what. Suggestions welcome!


aoifeboyle:
@wkerzendorf No idea at all what's going on with this segmentation fault / Travis failure but I assume they're both the same issue. Let me know if you have any idea what the problem could be.

wkerzendorf:
@aoifeboyle - found it.

wkerzendorf:
@aoifeboyle yes I agree with your assessment - no idea, why this is happening.

tardis/montecarlo/base.py
```diff
@@ -0,0 +1,31 @@
+from tardis.montecarlo import montecarlo
+
+class MontecarloRunner():
```
wkerzendorf: one should always inherit from `object` for classes

aoifeboyle: Ok, thanks! You don't happen to see any glaring errors that might be causing my `transition_probabilities` issue, do you? :(

aoifeboyle:
@orbitfold @kaushik94 I think you both know more about this C/Cython stuff than me. I'd be really grateful if either of you could have a look at this! I've narrowed the issue down to one line. Something about the way I'm assigning the `transition_probabilities` is causing a segmentation error. Wolfgang doesn't know what it is either.

kaushik94:
@aoifeboyle can you tell me how to pull this on my local so I can take a look ?

aoifeboyle:
@kaushik94 Sure - thanks! Go to the group chat on our google doc.

PR #339, opened 2015-06-26
Added NLTE (with issues)

aoifeboyle:
Having an issue with property 'cycles' now that I've added the module `LevelBoltzmannFactorNLTE`. I've looked into it and there seems to actually be two cycles among the properties (loops of property dependencies, which the new plasma structure does not allow).
* `level_boltzmann_factor_nlte` requires as input `beta_sobolev`, which is itself calculated from the `tau_sobolevs`, which are determined using the `level_number_density` property obtained using `level_boltzmann_factor_nlte`.
* `level_boltzmann_factor_nlte` also requires as input the `electron_densities`, which are dependent on the `ion_number_densities`, which themselves contribute to `level_boltzmann_factor_nlte`.

Not really sure what the correct approach to dealing with this is. @wkerzendorf @ssim I'd appreciate if either of you could have a look at this because I'm not very familiar with the NLTE physics so I might not have coded this right.


aoifeboyle:
Sorry - made PR to wrong branch!

wkerzendorf:
@aoifeboyle you can rebase

aoifeboyle:
@wkerzendorf I get this weird error now when I run the tests: 

```
  File "<string>", line 1, in <module>
  File "tardis/_astropy_init.py", line 111, in test
    coverage=coverage, open_files=open_files, **kwargs)
TypeError: run_tests() got an unexpected keyword argument 'repeat'
```

wkerzendorf:
try doing `git clean -dfx` to delete everything in the repo that does not belong

aoifeboyle:
@wkerzendorf I did but I still get the same error. I also tried just pulling a clean version of `plasma_restructure` onto a new branch and when I test it I get the same error...?

wkerzendorf:
New version of Astropy?

PR #340, opened 2015-06-28
Added NLTE (with issues).

aoifeboyle:
Having an issue with property 'cycles' now that I've added the module `LevelBoltzmannFactorNLTE`. I've looked into it and there seems to actually be two cycles among the properties (loops of property dependencies, which the new plasma structure does not allow).
* `level_boltzmann_factor_nlte` requires as input `beta_sobolev`, which is itself calculated from the `tau_sobolevs`, which are determined using the `level_number_density` property obtained using `level_boltzmann_factor_nlte`.
* `level_boltzmann_factor_nlte` also requires as input the `electron_densities`, which are dependent on the `ion_number_densities`, which themselves contribute to `level_boltzmann_factor_nlte`.

Not really sure what the correct approach to dealing with this is. @wkerzendorf @ssim I'd appreciate if either of you could have a look at this because I'm not very familiar with the NLTE physics so I might not have coded this right.


wkerzendorf:
So I think if we do NLTE we need to do this all in one and make one NLTE calculation. So any loop needs to be contained to one property. But we should talk about this.

aoifeboyle:
Yeah that's what I thought. Planning it out now. I'm free to talk whenever but try to give me about 15 minutes notice.

ssim:
I’m not sure I understand this very well, but I’d be keen to be involved in this discussion to understand better… Are you free to do that sometime today, Wolfgang?

aoifeboyle:
@wkerzendorf I don't understand why, in the old plasma structure, the NLTE calculation wasn't included in the loop to calculate the ion_number_density and the electron_density? Am I right in saying that the NLTE calculation both requires the electron_density and also influences it, so it should be included in this loop?

ssim:
@aoifeboyle In the old NLTE calculation, the electron_density would be required but the calculation would not directly affect it, I think. The effect of the old NLTE on the electron_density would only be via the partition functions, which enter the ionization calculation. Was the old NLTE calculation called inside the calculation of the ionization state, or was it done sequentially (i.e. after the ionization state was known?).

aoifeboyle:
@ssim Yes, that's what I mean. The NLTE calculation changes the partition function, which alters the phis and therefore the ionisation state. In the old plasma, the NLTE calculation was performed after the ionisation state / electron density loop had converged, which I find strange.

ssim:
@aoifeboyle I think that was done (quasi-)deliberately: the idea was to keep the excitation and ionization steps separate in the code. Over the whole cycle of multiple iterations, the result should still converge (because the partition functions from the previous cycle should be being used in the new ionisation calculation - so once it gets near to convergence it should be all ok). However, it's worth discussing whether it would be more elegant to deal with these things simultaneously rather than sequentially. 

In the end, it's unlikely to matter too much since for the ions we usually are worrying about in the NLTE calculations so far, the partition function is unlikely to be very different in NLTE compare to LTE. But it might matter more for some other ions and so it is worth thinking about what we want to do.

ssim:
To integrate them, I guess there are two choices - either one calls the NLTE solver every time a new electron density is tried inside the ionization solver or else one finds a converged ionization state then calls the NLTE solver to get new partition functions and then goes back and calls the ionization solver again...and repeat until both electron density and NLTE populations are simultaneously converged. 

Or we leave it as it is...

Or maybe there's another possible approach? Any suggestions?

wkerzendorf:
@ssim in the new structure it is very difficult to deal with loops (for now). So they would have to be done internal to the properties. My suggestion is to make a property that calculates both NLTE and LTE level populations at the same time, but maybe that is a bad idea. I wonder if we want to skype on that.

ssim:
Might be good to discuss. Are you free sometime today? @aoifeboyle said she would be, given 15min-ish notice. I'm free for the next ~hour or else in the afternoon.

wkerzendorf:
@ssim @aoifeboyle I'm free for ~30 minutes. Should we give this a try.

aoifeboyle:
@ssim Thanks for that clarification. I see what the difference is now. So in the old plasma, the partition functions used were an iteration behind? In the new plasma, whenever a single property is changed, all of the dependent properties (including those that are indirectly related) are updated automatically. 

Both of your suggestions had occurred to me already, but there are restrictions in the new plasma structure that didn't apply in the old plasma structure that makes those solutions difficult. As far as your second suggestion is concerned, the new plasma structure does not allow loops between properties. The first suggestion is what I originally tried, but things get very messy quickly then. So each plasma property class can only set the values of one property. In the case of the `IonNumberDensity` module, it only sets the ion populations through the usual method of iterating with various electron density values, and the separate `ElectronDensity` class then calculates the electron density by taking the ion number densities that were output and working backwards through the Saha equations. The issue with adding the NLTE stuff to the ionisation property is that the `ElectronDensity` module would then have to work backwards the whole way through the NLTE calculations as well.

aoifeboyle:
Yeah sure! I just need about 2 minutes.

ssim:
Ok - I'll to go Skype or g+ ...?

aoifeboyle:
Skype is fine!

wkerzendorf:
@aoifeboyle I'm available to chat now.

aoifeboyle:
@wkerzendorf @ssim Still working on getting this NLTE stuff going. I've run into a difficulty with the atomic data I think. I get an error from tardis/atomic saying that my NLTE data doesn't have a property called the `C_ul_interpolator`. I'm using the dataset: kurucz_cd23_chianti_H_He.h5. Is there another atomic dataset I could use that does have this?

ssim:
@aoifeboyle That error means that the data set doesn't have collisional data in it. Probably we need to get your a better dataset to test with. @wkerzendorf can you give Aoife one of the NLTE sets used in our paper that has collisional data in it?

wkerzendorf:
@aoifeboyle NLTE does work even if the collision coefficients are not provided (as far as I remember). A lot of this machinery currently lives at https://github.com/tardis-sn/tardis/blob/master/tardis/atomic.py#L558 and should be ported into the plasma properties. Maybe giving a new module for these rather large properties is a good idea. So if we are missing collisional data we should generate zeros.  I'll hunt for the data.

aoifeboyle:
@wkerzendorf Ok thanks! I'll make it a new atomic plasma property.

aoifeboyle:
@wkerzendorf, I can see that the later `c_ul_matrix` values are set to zero if they're missing, but I still have an issue at line: https://github.com/tardis-sn/tardis/blob/master/tardis/atomic.py#L610 and further on because there is no data at all (rather than just missing data). I'm just going to comment out this part of the module for now and come back to it.

aoifeboyle:
@wkerzendorf Ok, I thought I could just comment out the collision matrix stuff but when I do that and say `rates_matrix = r_lu_matrix + r_ul_matrix` I get an error saying it's a singular matrix. Let me know if there's a specific way you want me to deal with this, in case someone uses atomic data without collision data in future.

wkerzendorf:
@aoifeboyle okay - I have now linked the documentation pages for you to see what the NLTE does. That will definitely help. For now you need to make these matrices for the collisional data and fill them with 0. The rates matrix can not be 0.

wkerzendorf:
@aoifeboyle have a look under the physics part of the documentation.

aoifeboyle:
Ok, sorry to be annoying but I'm really confused. You said the rates matrix cannot be zero, but:
`rates_matrix = r_lu_matrix + r_ul_matrix + collision_matrix`
and those three components are all set equal to zero. Also `r_ul_matrix_reshaped`/`r_lu_matrix_reshaped` don't seem to be used?

```
            r_ul_matrix = np.zeros((number_of_levels, number_of_levels,
                len(t_electron)), dtype=np.float64)
            r_ul_matrix_reshaped = r_ul_matrix.reshape((number_of_levels**2,
                len(t_electron)))
            r_ul_matrix_reshaped[r_ul_index] = A_uls[np.newaxis].T + \
                B_uls[np.newaxis].T * j_blues_nlte[np.newaxis].T
            r_ul_matrix_reshaped[r_ul_index] *= beta_sobolev_nlte[lines_index]

            r_lu_matrix = np.zeros_like(r_ul_matrix)
            r_lu_matrix_reshaped = r_lu_matrix.reshape((number_of_levels**2,
                len(t_electron)))
            r_lu_matrix_reshaped[r_lu_index] = B_lus[np.newaxis].T * \
                j_blues_nlte[np.newaxis].T * \
                beta_sobolev_nlte[lines_index]

#            collision_matrix = nlte_data.get_collision_matrix(species,
#                t_electron) * electron_densities.values

            collision_matrix = r_ul_matrix.copy()
            collision_matrix.fill(0.0)

            rates_matrix = r_lu_matrix + r_ul_matrix + collision_matrix

            level_boltzmann_factor = general_level_boltzmann_factor

            for i in xrange(number_of_levels):
                rates_matrix[i, i] = -rates_matrix[:, i].sum(axis=0)

            rates_matrix[0, :, :] = 1.0

            x = np.zeros(rates_matrix.shape[0])
            x[0] = 1.0
            for i in xrange(len(t_electron)):
                nlte_level_boltzmann_factor = \
                    np.linalg.solve(rates_matrix[:, :, i], x)
```

Still getting the singular matrix error with this code.

aoifeboyle:
Tests will fail, but I've just pushed my current version to show the current status of the NLTE implementation.

wkerzendorf:
@aoifeboyle have a Look at the updated documentation. I'll be with you shortly

wkerzendorf:
@aoifeboyle skype?

wkerzendorf:
@aoifeboyle skype again?

aoifeboyle:
@wkerzendorf I think I actually might be getting somewhere with working out the issue so leave it with me for now! That ok? I compared it with the old plasma structure as you suggested and it works so I'm just comparing the two now.

wkerzendorf:
@aoifeboyle sounds good!

aoifeboyle:
Ok @wkerzendorf, so with both the old and new plasmas, setting the entire `collisional_matrix` equal to 0 and having all the `j_blues` be 0 as well (i.e. the coronal approximation) leads to a singular matrix error with the `rates_matrix`. The problem doesn't occur if either the `collisional_matrix` or the `j_blues` are not equal to 0. I think there might be some kind of physical contradiction there? With the new plasma setup, the plasma is initialised with the `j_blues` as 0 and then they're updated with the `update_radiationfield` function, but I can change that if you think that's the right way to get around this?

wkerzendorf:
@aoifeboyle okay we should have a class that is called `LTEJblues` that takes as an input `betarad` and lines and as an output gives 


```
def intensity_black_body(nu, T):
    """
        Calculate the intensity of a black-body according to the following formula
        .. math::
            I(\\nu, T) = \\frac{2h\\nu^3}{c^2}\frac{1}{e^{h\\nu \\beta_\\textrm{rad}} - 1}
    """
    beta_rad = 1 / (k_B_cgs * T)

    return (2 * (h_cgs * nu ** 3) / (c_cgs ** 2)) / (
        np.exp(h_cgs * nu * beta_rad) - 1)
```

tardis/plasma/properties/ion_population.py
```diff
@@ -239,29 +241,198 @@ def calculate(self, phi, partition_function, number_density):
...
+                     r'\times \Phi_{i, j+1}/N_e + \dots)$')
+
+    outputs = ('ion_number_density', 'electron_densities',
+               'level_boltzmann_factor')
+
+    def __init__(self, plasma_parent, ion_zero_threshold=1e-20):
+        super(IonNumberDensityNLTE, self).__init__(plasma_parent)
+        self.ion_zero_threshold = ion_zero_threshold
+
+    def calculate_nlte(self, nlte_input, beta_sobolev, j_blues, levels,
```
wkerzendorf: why is this under `ion_number_density`?

tardis/plasma/properties/level_population.py
```diff
@@ -32,64 +33,65 @@ def calculate(level_population_fraction, ion_number_density):
             level_population_fraction.index.droplevel(2)].values
         return level_population_fraction * ion_number_density_broadcast
 
-class LevelPopulationNLTE(ProcessingPlasmaProperty):
+class LevelBoltzmannFactorNLTE(ProcessingPlasmaProperty):
```
wkerzendorf: Let's refactor this a bit.

tardis/plasma/properties/level_population.py
```diff
@@ -32,64 +33,65 @@ def calculate(level_population_fraction, ion_number_density):
...
+    name = 'level_boltzmann_factor_nlte'
+
     @staticmethod
-    def calculate(self):
+    def calculate(nlte_input, beta_sobolev, j_blues, levels, nlte_data,
```
wkerzendorf: `nlte_input` is a dictionary and does not belong here.

tardis/plasma/properties/ion_population.py
aoifeboyle: There's a loop of property dependencies that goes: 
`level_boltzmann_factor` (from nlte level population calculation) -> `partition_function` -> `ion_number_density` -> `level_number_density` -> `tau_sobolevs` -> `beta_sobolevs` -> `level_boltzmann_factor` (nlte level population calculation again). So I thought we needed that loop to be inside one plasma property for consistency.

tardis/plasma/properties/level_population.py
aoifeboyle: Ok - I've been using it as a property. What would you suggest instead? Separate properties for the coronal approximation, classical nebular approximation, etc.?

tardis/plasma/properties/ion_population.py
aoifeboyle: @wkerzendorf Have I got the wrong idea? Is there a better way?

aoifeboyle:
Closing for now - will create new PR for this later.

PR #341, opened 2015-06-29
Initial steps.

aoifeboyle:
Just creating a PR so we can work on this together @wkerzendorf. Been trying to figure out where to start using the `IonNumberDensity` module as an example. I'm not sure if we want to keep the output properties from the same function grouped together or separate? If you can show me how to get started when you have time that would be great.


wkerzendorf:
@aoifeboyle so there are now two PRs with the new `outputs` structure, right?

aoifeboyle:
@wkerzendorf No, I haven't added the new `outputs` stuff to PR #340. Is that the other one you're referring to? You said to create a new branch for this stuff. I was planning to go back to #340 when I get this working.

wkerzendorf:
@aoifeboyle let's quickly skype if you have time.

PR #344, opened 2015-07-01
Plasma/multiple outputs

aoifeboyle:
Added the `electron_density` to the `IonNumberDensity` class as a second output. Seems to work great! Thanks @wkerzendorf. I ran a quick simulation to compare the results to when the separate `ElectronDensity` class was used, and they were identical. Next step is to pick up where I left off with the NLTE stuff.


wkerzendorf:
@aoifeboyle that's great - let me know how NLTE goes.

aoifeboyle:
@wkerzendorf How's this? Travis is fine but the coverage test fails. I'll be adding more tests later. Should I merge?

wkerzendorf:
Sure

PR #345, opened 2015-07-02
New layout with poppable widgets (WIP)

gautam1168:
I have added the new classes that create the container that can
house a widget and pop it out of the main window. The next step is
to use these classes in the layout of the main window.

Once that is done I will take some time to edit the stylesheets to improve the appearance of the widgets.


gautam1168:
@wkerzendorf So I have put in the popping widgets in and I am working on the colorscheme now. I was wondering if you would like the shell plot to have fewer colors. The jet colormap is like a rainbow and I feel its a bit much. I also want to ask if anybody else has any ideas regarding the colorscheme or layout. Any suggestions will be helpful. Here is a snapshot of how it looks now, but it will change before this is merged.
layout

wkerzendorf:
Hi @gautam1168 that looks very nice!!

wkerzendorf:
sure the logo would be great!

PR #347, opened 2015-07-07
NLTE

aoifeboyle:
New approach to NLTE. This time the `beta_sobolevs` from the previous iteration of the plasma are used to calculate the NLTE level populations, in a similar way to what was done in the old plasma. Seems to be working well, but I'm a bit confused about the atomic data, specifically whether the `lines_idx` in the NLTE data should be the same as `lines.index` in the atomic data. It seems like that should be the case for consistency, but it doesn't seem to be the case here, and it's creating some confusion when calculating the NLTE level populations. Here, `lines_index` is taken from the NLTE data:
missing atomic data


aoifeboyle:
@wkerzendorf So I think I've got a good system here with the NLTE now. Could you have a look and see what you think? It works on a similar principle to the NLTE in the old plasma. Quick summary:

So the property cycling issue consists of two property cycles. The NLTE calculation indirectly affects the `beta_sobolev` and `electron_densities` values, both of which are also inputs for it. In the old plasma, the `ion_number_density` convergence loop was performed making use of the NLTE calculation of the previous run. In the case of the new plasma, it's the reverse. The `NLTE` calculation is performed making use of the `ion_number_density` outputs of the previous iteration.

Old plasma calculation sequence:
NLTE (iteration 0) --> beta_sobolev, electron_densities (iteration 0) --> ion_number_density convergence (iteration 1) --> beta_sobolev, electron_densities (iteration 1) --> NLTE (iteration 1).
New plasma calculation sequence:
NLTE (iteration 0) --> beta_sobolev, electron_densities (iteration 0) --> ion_number_density convergence (iteration 0) --> beta_sobolev, electron_densities (iteration 1) --> NLTE (iteration 1).

I hope that makes some sense! If you're happy with this, I'd like to merge it although the coverage test has failed. I plan to open a new PR for adding the tests.

wkerzendorf:
@aoifeboyle do they converge to the same values?

aoifeboyle:
@wkerzendorf Yep! Better accuracy than I expected in fact.

tardis/plasma/properties/nlte.py
```diff
@@ -0,0 +1,17 @@
+from tardis.plasma.properties.base import BasePlasmaProperty
```
aoifeboyle: @wkerzendorf Think I've found a better way of doing this now, without having to output and read in values from previous iterations from model.py. Created this new property class to store values from previous iterations. The values are set just prior to all of the properties being updated. I think it's a lot more straightforward!

tardis/plasma/properties/partition_function.py
```diff
@@ -51,6 +53,243 @@ def calculate(levels, beta_rad, w):
...
+            for i in xrange(len(t_electron)):
+                level_boltzmann_factor = \
+                    np.linalg.solve(rates_matrix[:, :, i], x)
+                general_level_boltzmann_factor[i].ix[species] = \
+                    level_boltzmann_factor
+        return general_level_boltzmann_factor
+
+class LevelBoltzmannFactorNLTEGeneral(ProcessingPlasmaProperty):
+    outputs = ('level_boltzmann_factor',)
+
```
wkerzendorf: I think this is the only function we should have. This is my suggestion:

```
class LBFNLTEG(PPP):
   outputs = ...

   def __init__(self, classical_nebular=False):
       if classical_nebular = True:
           self.inputs.pop('j_blues')
....
```

tardis/plasma/properties/partition_function.py
```diff
@@ -51,6 +50,260 @@ def calculate(levels, beta_rad, w):
...
+    def _calculate_classical_nebular(self, general_level_boltzmann_factor):
+        return general_level_boltzmann_factor
+    def _calculate_coronal_approximation(self, general_level_boltzmann_factor):
+        return general_level_boltzmann_factor
+    def _calculate_general(self, general_level_boltzmann_factor):
+        return general_level_boltzmann_factor
+    def _main_nlte_calculation(self):
+        pass
+'''
```
aoifeboyle: Hi @wkerzendorf. Just trying to get the basic structure here to work like you suggested earlier. When I run this I get the error, 'Can't instantiate abstract class LevelBoltzmannFactorNLTE with abstract methods calculate.' I assume that's just because there's no explicit `def calculate()`? How do I make this work?

wkerzendorf: you define one: 

```python
def calculate(self, *args, **kwargs):
    pass
```

then overwrite it

tardis/plasma/properties/partition_function.py
```diff
@@ -51,6 +50,124 @@ def calculate(levels, beta_rad, w):
...
+    def __init__(self, plasma_parent, classical_nebular=False,
+        coronal_approximation=False):
+        super(LevelBoltzmannFactorNLTE, self).__init__(plasma_parent)
+        if classical_nebular == True:
+            self.calculate = self._calculate_classical_nebular
+        elif coronal_approximation == True:
+            self.calculate = self._calculate_coronal_approximation
+        else:
+            self.calculate = self._calculate_general
+        self._update_inputs()
```
wkerzendorf: that is smart!

tardis/plasma/properties/partition_function.py
```diff
@@ -51,6 +50,124 @@ def calculate(levels, beta_rad, w):
...
+class LevelBoltzmannFactorNLTE(ProcessingPlasmaProperty):
+    outputs = ('level_boltzmann_factor',)
+
+    def calculate(self):
+        pass
+
+    def __init__(self, plasma_parent, classical_nebular=False,
+        coronal_approximation=False):
+        super(LevelBoltzmannFactorNLTE, self).__init__(plasma_parent)
+        if classical_nebular == True:
```
wkerzendorf: maybe making sure that only one can be True is not unimportant

PR #363, opened 2015-08-07
Plasma/non equilibrium ionisation solver

aoifeboyle:
@wkerzendorf @ssim Just starting this PR now so we have somewhere to talk about it easily.


aoifeboyle:
PDF version of maths docs: https://www.dropbox.com/s/974isma89ppra9d/nlte_ionization.pdf?dl=0

ssim:
@aoifeboyle Good stuff!

You're missing the diagonal matrix elements at the top: e.g. the 1,1 element of the matrix should be (I think) r_{12}. And the indices of the other elements might be flipped - so e.g. the first line of the equation should really read

N_1 r_{12}  + N_2 r_{21} = 0 [since assumes r_13 and r_31 are both zero]

Also - eqn 2 - I'd argue that the T here should be T_e not T_r.

aoifeboyle:
Oh I see, that does make more sense! We should discuss the Te/Tr thing. I could use some clarification on it in some other areas as well.

ssim:
would it be useful to e.g. Skype later this week? I could make time one afternoon (not Tuesday) probably...let me know what works for you

aoifeboyle:
Sure! I can't do this afternoon so maybe Wednesday afternoon?

aoifeboyle:
Updated PDF: https://www.dropbox.com/s/974isma89ppra9d/nlte_ionization.pdf?dl=0

aoifeboyle:
This seems to be working! Some notes:
 * Until we get the atomic data, this just assumes that the cross-section = (nu_i / nu) ^ 3, where nu_i is the ionisation frequency.
 * The integrations are performed with sampling of 1000 points, up to a limit of 10 times the ionization frequency (I did some manual calculations and that's high enough to give an accurate result).
 * To prevent a property loop, this module uses the value of `electron_densities` from the previous code iteration each time. This is the same as what is done in the NLTE excitation module.

aoifeboyle:
Updated maths pdf https://www.dropbox.com/s/974isma89ppra9d/nlte_ionization.pdf?dl=0

@ssim, could you please let me know if I'm calculating J_{nu} correctly? I have written down that you said integrate W * B_{nu}(T_{r}) over frequency but I'm not sure.

ssim:
@aoifeboyle - I'll check! will get back to you...

ssim:
@aoifeboyle There's no integral needed there: it's just

J_nu = W * B_nu(T_r)

Also typo in equation (4) - we'd call that gamma or something (not alpha^sp)

ssim:
Thought about the note above - might be a worry using the last electron density here if the ionization is quite different from the last iteration - not sure how good that will be...but I guess we can test!

aoifeboyle:
Ok - I originally had it without the integral but then I confused myself so I can just revert it :) can't open the dropbox link on my phone but you're probably right about the typo. With the electron density I was hoping that with enough iterations it would converge pretty well ( it does with the NLTE excitation) but it's something we can test.

aoifeboyle:
Corrected maths: https://db.tt/viAs8R0r Going to put this in a .rst file soon.

wkerzendorf:
@aoifeboyle can you rebase this onto the newest master. There are seemingly many outdated changes

unoebauer:
@wkerzendorf - what's with this one? Should this be still merged (2 years after the last changes!)?

unoebauer:
@chvogl has raised serious concerns about merging this PR.

unoebauer:
Closing due to persistent inactivity

PR #373, opened 2015-08-12
Removing test_plasma_nlte.py.

aoifeboyle:
Will re-implement in a different way.


wkerzendorf:
so destructive - deleting all that old code :wink:

aoifeboyle:
I know, but it just doesn't make sense there! It'll be back soon I promise :P

PR #374, opened 2015-08-14
WIP: Helium approx.

aoifeboyle:
Still working on this.


aoifeboyle:
@wkerzendorf Could you take a look at this and see if you're happy with how it's coded? I still have to add tests etc.

tardis/plasma/properties/nlte.py
```diff
@@ -14,4 +21,44 @@ class PreviousElectronDensities(PreviousIterationProperty):
...
+
+class HeliumNLTE(ProcessingPlasmaProperty):
+    outputs = ('helium_population',)
+
+    def calculate(self, level_boltzmann_factor, electron_densities,
+        ionization_data, beta_rad, g, g_electron, w, t_rad, t_electrons,
+        delta, zeta_data, number_density):
+        helium_population = deepcopy(level_boltzmann_factor.ix[2])
+        # He I excited states
+        he_one_population = level_boltzmann_factor.ix[2].ix[0] * \
```
wkerzendorf: this calculation is done somewhere else as well. Let's just make one function for it.

tardis/plasma/properties/nlte.py
```diff
@@ -14,4 +21,44 @@ class PreviousElectronDensities(PreviousIterationProperty):
...
+        he_two_population = level_boltzmann_factor.ix[2,1].mul(w * \
+            (g.ix[2,1].ix[0]**(-1)) * (t_electrons / t_rad)**0.5)
+        helium_population.ix[1].update(he_two_population)
+        #He II ground state
+        helium_population.ix[1].ix[0] = 1.0
+        #He III states
+        zeta = interpolate.interp1d(zeta_data.columns.values,
+            zeta_data.ix[2,2].values)(t_rad)
+        delta = delta.ix[2].ix[2]
+        he_three_population = 2 * ((electron_densities)**(-1)) * g_electron * \
```
wkerzendorf: same here I think.

wkerzendorf:
@aoifeboyle looks like a good first effort, but I feel that we should make sure that code is not repeated. Let me know if that explains it.

aoifeboyle:
@wkerzendorf Ok, I see what you mean. Leave it with me.

tardis/plasma/properties/nlte.py
aoifeboyle: @wkerzendorf So the He I level population calculation here is different to what the code normally uses because it's calculated from the GS of the He II ion rather than the He I ion. You're right with the He II and He III level populations though, but it's easier for me to keep all of these calculations in the same function so that I can normalise the populations easily using the number density. But I'm having a think now about ways to simplify it.

aoifeboyle: Ok so I was thinking that I could just use some kind of scaling method to change the He II and He III populations after the He II ground state population is set to 1.0, but that would require using the level_number_density property as an input property here, which would create a loop. I think this is the only way to do it simply.

tardis/plasma/properties/nlte.py
```diff
@@ -14,4 +21,44 @@ class PreviousElectronDensities(PreviousIterationProperty):
...
\ No newline at end of file
+    outputs = ('previous_beta_sobolevs',)
+
+class HeliumNLTE(ProcessingPlasmaProperty):
+    outputs = ('helium_population',)
+
+    def calculate(self, level_boltzmann_factor, electron_densities,
+        ionization_data, beta_rad, g, g_electron, w, t_rad, t_electrons,
+        delta, zeta_data, number_density):
+        helium_population = deepcopy(level_boltzmann_factor.ix[2])
```
orbitfold: Is there a reason to use deepcopy? It can be incredibly slow, could be a problem if this is called often.

aoifeboyle: @orbitfold Not really! I just use it from habit. I didn't know it was slow. I'll replace it with copy.

orbitfold: Well you might need it depending on what you want to do. But if you can use copy do that.

tardis/plasma/properties/nlte.py
wkerzendorf: @aoifeboyle What I meant was. Add functions (maybe static ones) to the other properties in the code that do similar calculations and then both properties use their functions. One such example is in level_populations in line 26 and 47. they both use the same calculation, but it is in the code twice. Use a static function there. We can skype on this if you want.

aoifeboyle: @wkerzendorf Ok, so I can leave the HeliumNLTE property alone? I'll merge the two level_population properties into one now.

aoifeboyle: @wkerzendorf Let me know if you think what I have now for the level_populations is better. I had another look and all of the level population calculations in the `HeliumNLTE` property are different to the level population calculations in `LevelNumberDensity` because they are all calculated relative to ground state populations rather than using the ion populations. So being able to call the same calculations would not be helpful. Let me know if that makes sense to you. I think it's tidier now anyway.

wkerzendorf:
hi @aoifeboyle I can't see your comments for some reason on github (am getting emails about them). I can't look at it now in detail. But while we wait - have a look at line 26 and 47 in level_populations. you can make it one function callable from the other one (use static methods - let me know if that's unclear).

aoifeboyle:
@wkerzendorf So I've changed the Helium NLTE stuff so it makes use of the functions that already exist as static methods (the Phis, etc.), and I've removed repetition from other parts of the code too. Could you have a look through and tell me what you think when you have time?

tardis/plasma/properties/nlte.py
```diff
@@ -14,4 +21,85 @@ class PreviousElectronDensities(PreviousIterationProperty):
...
+
+    def calculate_helium_one(self, g_electron, beta_rad, partition_function,
+            ionization_data, level_boltzmann_factor, electron_densities, g,
+            w, t_rad, t_electron):
+        (partition_function_index, ionization_data_index, partition_function,
+            ionization_data) = self.filter_with_helium_index(2, 1,
+            partition_function, ionization_data)
+        phis = (1 / PhiSahaLTE.calculate(g_electron, beta_rad,
+            partition_function, ionization_data)) * electron_densities * \
+            (1.0/g.ix[2].ix[1].ix[0]) * (1/w) * (t_rad/t_electron)**(0.5)
```
wkerzendorf: you can do .ix[2, 1, 0]. Maybe try not to us ix but iloc.

aoifeboyle: Changed to .ix[2,1,0]. I used to use .iloc but it doesn't work with one of the Pandas versions on here so I always use .ix now.

aoifeboyle:
@wkerzendorf Can I merge this?

aoifeboyle:
@wkerzendorf Got the error 245 for one of the Travis checks here but nothing happens when I click the button to re-run it for some reason? It's definitely not the code because the previous commit worked fine and this one was just a small change to an .rst file.

wkerzendorf:
@aoifeboyle I think that looks very good. As I'm quite busy this week, I'll let @orbitfold have another look for any code style issues and then we can merge (do this at your own discretion).

orbitfold:
Well, one obvious thing is that there are no docstrings. And the chances of anyone adding them later if you don't now are slim :) And there should really be docstrings for each method explaining the arguments and what the method does and what it returns. I don't believe we are very consistent with out docstring conventions but any docstring is better than none. This is as good as anything for a style guide I guess:

http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html

Or look at our code that does contain them.

wkerzendorf:
We use numpydoc docstring format (if we do)

orbitfold:
In that case:
https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

aoifeboyle:
Ok, I'll start on that now!

aoifeboyle:
@orbitfold Let me know if there are any other issues, because I might merge this and then create a new PR with docstrings for the whole plasma.

orbitfold:
I just looked through it and I can't think of anything nasty to say about it, I think it's fine. Why do you need a separate PR for docstrings though?

aoifeboyle:
I don't have to. It's just I haven't done proper docstrings for all the other plasma stuff I've added, so I'd like to do the whole lot at once.

orbitfold:
Either way is fine, just curious :)

PR #384, opened 2015-08-19
Minor changes in Monte Carlo and model part for Kromer plots

unoebauer:


orbitfold:
Can't see anything wrong with this.

wkerzendorf:
@orbitfold can you take a quick look if that is okay? I think it's a sensible addition, but just quickly skimmed it.

PR #385, opened 2015-08-20
Plasma documentation PR (in progress)

aoifeboyle:
Still working on this but making a PR so people can comment etc.

### DIFF
--- a/docs/physics/plasma.rst
+++ b/docs/physics/plasma.rst
@@ -1,98 +1,41 @@
-.. _plasmas:
-
 ******
 Plasma
 ******
 
-.. currentmodule:: tardis.plasma_array
+The role of the plasma module is to determine the ionisation and excitation states of the elements of the 
+supernova ejecta, given the basic structure, including the elemental abundances, densities and radiation temperature.
+After the calculation of the plasma state, the :math:`\tau_{\textrm{sobolev}}` values can be calculated.
 
+The Tardis plasma structure inherits from the `BasePlasma class`. The code currently uses the `LegacyPlasmaArray`
+for generating a plasma from the information provided by `model`. A variety of different plasmas can be generated
+depending on the options selected in the plasma section of the Tardis config. file. The options currently considered
+by the Legacy Plasma when creating the plasma calculation structure include:
 
-.. warning::
-    This information (particularly code examples) is not up-to-date.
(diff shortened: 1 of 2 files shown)


aoifeboyle:
@orbitfold, question about docstrings. So I already had some docstrings written for some of the plasma properties, but with my own made-up format, e.g.:

    class ZetaData(BaseAtomicDataProperty):
        """
        Outputs:
        zeta_data : Pandas DataFrame
            Zeta data for the elements used
            Required for nebular ionization scheme.
            The zeta value represents the fraction of recombination events from the ionized state that go
            directly to the ground state.
        """
        outputs = ('zeta_data',)
        ...

Do you think it would be ok to stick with that rather than using the same docstring format as the rest of Tardis? I think this format makes more sense for the setup we have. Is there any more information you would include? It may seem like a trivial question but I want to check it with you guys before I go and write docstrings for all of the other plasma classes ;)

wkerzendorf:
Have a look here https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#class-docstring. I agree we have these outputs that are kind of special. Let's try to at least stick to some of the conventions. Maybe instead of outputs call them attributes and then numpydoc will be able to deal with them.

orbitfold:
I'd just stick with the numpy format:

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-classes

It should also have a short description at the very start of what the method does in broad terms as well as what it returns at the end of the docstring. Did you document a lot already? It shouldn't be hard to change. These are also not only for us but for automated documentation generating tools. These expect a certain format.

aoifeboyle:
Ok, I see what you mean. It won't be too hard to change at all. I just thought it made sense to put things in terms of inputs and outputs since that's kind of how the plasma works, but I didn't know that there were documentation generating tools. I'll change 'outputs' to 'attributes', and add a description for the entire class at the start too. Thanks both of you!

wkerzendorf:
@aoifeboyle that looks very nice and detailed.

aoifeboyle:
Glad you like it! :)

aoifeboyle:
Closing PR temporarily as adding docs to current master is easier than merging conflicts.

PR #386, opened 2015-08-20
improve inlining specifications and fix a few problems

mreineck:
This pull request should enable all small rpacket-related functions to be inlinable, by moving them from rpacket.c to rpacket.h. I also fixed a few other inline-related problems, added missing prototypes to the header files and created an extra header for mt_state and line_search(), to avoid recursive header inclusion.


wkerzendorf:
@mreineck this is great!! let's see if it passes the tests.

wkerzendorf:
@mreineck Okay it seems to be stumbling at `rpacket_get_status` at https://github.com/tardis-sn/tardis/blob/master/tardis/montecarlo/src/test_cmontecarlo.c#L274. I think the status is an `enum` but maybe I'm wrong.

wkerzendorf:
@mreineck there is always the possibility to mark this test as `xfail` - expected to fail.

unoebauer:
@mreineck, @wkerzendorf 
I can confirm the speed-up of about a factor 2 when using the tardis_example setup
Tardis Master:
`Running with OpenMP - 1 threadstardis.simulation - INFO - Finished in 20 iterations and took 95.94 s`
PR 386:
`Running with OpenMP - 1 threadstardis.simulation - INFO - Finished in 20 iterations and took 55.12 s`

wkerzendorf:
@unoebauer this is very good!!

wkerzendorf:
I think we are now roughly a factor of 10 below my initial cython implementation :wink:

unoebauer:
@mreinecke @wkerzendorf 
The virtual spectra obtain with Tardis Master and PR 386 are also identical

tardis/montecarlo/src/test_cmontecarlo.c
```diff
@@ -274,7 +278,7 @@ bool
 test_montecarlo_bound_free_scatter(){
 	double DISTANCE = 1e13;
 	montecarlo_bound_free_scatter(rp, sm, DISTANCE);
-	return rpacket_get_status(rp);
+	return (rpacket_get_status(rp)!=0);
```
wkerzendorf: @mreineck no mark it as xfail in the python file. Here's how: https://github.com/tardis-sn/tardis/blob/master/tardis/montecarlo/tests/test_cmontecarlo.py#L74 add `@pytest.mark.xfail` above that. that will show us continually that this is an xfail.

mreineck:
I did that now as well. But the condition in the C code was also incorrect: thet status can go from 0 to some positive value>1, but the return type of bool indicates that only 0 and 1 are acceptable.

wkerzendorf:
Well it seems all checks pass. @orbitfold can you have a quick view as well? If we are all happy - let's merge.

PR #392, opened 2015-08-21
Consistency between old virtual packet arrays and new ones introduced in PR 384

unoebauer:
All arrays holding properties of virtual packets now begin with
```
virt_packet_
````


unoebauer:
@wkerzendorf, @orbitfold Could you have a quick look? It was really only a renaming exercise.

orbitfold:
Looks good to me.

PR #396, opened 2015-08-24
added virt_packet properties to to_hdf5 list (Kromer plot)

unoebauer:
All required packet properties are written to the hdf5 file when calling `mode.to_hdf5` to produce a Kromer plot.

Note:
Additionally, the line info has to be available, e.g. by doing
```model.atom_data.lines.to_hdf("lines.hdf5", "lines")```


unoebauer:
Now it is possible to create plots like

kromer_plot

from hdf5 files

PR #397, opened 2015-08-24
Montecarlo/memoryview

wkerzendorf:
This was prompted by trying to avoid costly copies of large arrays, but has evolved into throwing away quite a bit of code that is not necessary anymore.

### DIFF
--- a/tardis/montecarlo/montecarlo.pyx
+++ b/tardis/montecarlo/montecarlo.pyx
@@ -92,10 +92,10 @@ cdef extern from "src/cmontecarlo.h":
 
 cdef initialize_storage_model(model, runner, storage_model_t *storage):
 
-    storage.no_of_packets = model.packet_nus.size
-    storage.packet_nus = <double*> PyArray_DATA(model.packet_nus)
-    storage.packet_mus = <double*> PyArray_DATA(model.packet_mus)
-    storage.packet_energies = <double*> PyArray_DATA(model.packet_energies)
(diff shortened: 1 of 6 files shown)


tardis/montecarlo/montecarlo.pyx
```diff
@@ -86,61 +87,31 @@ cdef extern from "src/cmontecarlo.h":
...
     # Setup of structure
     structure = model.tardis_config.structure
     storage.no_of_shells = structure.no_of_shells
-    cdef np.ndarray[double, ndim=1] r_inner = structure.r_inner.to('cm').value
-    storage.r_inner = <double*> r_inner.data
-    cdef np.ndarray[double, ndim=1] r_outer = structure.r_outer.to('cm').value
-    storage.r_outer = <double*> r_outer.data
-    cdef np.ndarray[double, ndim=1] v_inner = structure.v_inner.to('cm/s').value
-    storage.v_inner = <double*> v_inner.data
+    storage.r_inner = <double*> PyArray_DATA(structure.r_inner.to('cm').value)
```
wkerzendorf: @orbitfold this is the main thing I'm doing - a much easier way to access the arrays.

wkerzendorf:
@orbitfold can you check this and merge? @ssim @chvogl @mreinecke this is the slightly updated C-structure. We now use c-contiguous arrays everywhere (instead of Fortran ones) and adding new data to the code should be much simpler now.

PR #401, opened 2015-08-26
Alternative fix for issue #393

mreineck:
These changes remove the global mt_state variable. The RNG state is created whenever needed and passed to all functions that generate random numbers. This solves the multi-threading issue.
Related to this, I also reworked the cmontecarlo tests so that they need no longer be executed in order. Each test now creates and tears down its own environment and should no longer leak resources.

### DIFF
--- a/tardis/montecarlo/src/cmontecarlo.c
+++ b/tardis/montecarlo/src/cmontecarlo.c
@@ -4,14 +4,6 @@
 #endif
 #include "cmontecarlo.h"
 
-rk_state mt_state;
-
-static void
-initialize_random_kit (unsigned long seed)
-{
-  rk_seed (seed, &mt_state);
-}
-
 /** Look for a place to insert a value in an inversely sorted float array.
  *
  * @param x an inversely (largest to lowest) sorted float array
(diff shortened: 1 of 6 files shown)


mreineck:
This PR is just a suggestion to judge the advantages and drawbacks of different solutions a bit more easily. It is mutually exclusive with #395.

wkerzendorf:
@orbitfold can you see if this is good?

orbitfold:
I don't really like polluting procedure argument lists with another argument which frankly does not belong there. Why not just make it part of storage?

orbitfold:
Let's just go with this, I'll merge.

wkerzendorf:
if it's not a big deal , from my side is fine. I'm trying anyways to keep out of the C-part :wink:

wkerzendorf:
@orbitfold merge away - when you are happy.

mreineck:
Making it part of storage does not address the problem, because storage is not thread-private. We would still have to explicitly allocate enough instances of mt_state for the individual threads, and every thread would have to make sure it was accessing its "own" copy.
Yes, the interface is ugly. If we were writing C++, I could provide something much cleaner :)

Issue #402, opened 2015-08-26
Problems with pandas slicing (again)


ssim:
I'm seeing problems (again) when trying to use the macro atom mode - it complains:

```
/Users/stuartsim/Tardis/tardis/build/lib.macosx-10.10-x86_64-2.7/tardis/atomic.py:532: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  'count_total'].values[:-1])))
/Users/stuartsim/Tardis/tardis/build/lib.macosx-10.10-x86_64-2.7/tardis/atomic.py:535: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  self.macro_atom_references['references_idx'] = np.arange(len(self.macro_atom_references))
/Users/stuartsim/Tardis/tardis/build/lib.macosx-10.10-x86_64-2.7/tardis/atomic.py:537: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  self.macro_atom_data['lines_idx'] = self.lines_index.ix[self.macro_atom_data['transition_line_id']].values
/Users/stuartsim/Tardis/tardis/build/lib.macosx-10.10-x86_64-2.7/tardis/atomic.py:553: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  tmp_macro_destination_level_idx].values.astype(np.int64)
```

Here's the yml file I'm trying to use

```
tardis_config_version: v1.0

supernova:
    luminosity_requested: 9.44 log_lsun
    time_explosion: 13 day

atom_data: kurucz_cd23_chianti_H_He.h5

model:
    structure:
        type: specific

[...]

PR #403, opened 2015-08-27
fixes at least some of the pandas loc errors

wkerzendorf:
This is a fix for #402 - but doesn't deal with all of them. Two of them I do not know how to get rid of.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
         - COMPILER=gcc
         - CYTHON_VERSION=0.21
         - NUMPY_VERSION=1.8
-        - PANDAS_VERSION=0.12
+        - PANDAS_VERSION=0.16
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@ssim @aoifeboyle merge if happy.

ssim:
@aoifeboyle any thoughts on this? I don't understand well enough what it's doing to really judge right now.

aoifeboyle:
I'm out today but if you could just leave it open til later I'll take a look this evening? I'll see if I can work out the other errors too.

ssim:
@aoifeboyle Thanks!

@wkerzendorf If it's fresh in your mind, could you add some comments to the code to explain what these lines you've changed are actually doing? This stuff seems very opaque to me - would be good to explain what's going on

PR #404, opened 2015-08-27
Montecarlo/new packet source

wkerzendorf:
This is the implementation of the algorithm suggested by @unoebauer. it works fine. Still adding tests and then it's ready to merge.


wkerzendorf:
@tardis-sn/tardis-core can anyone cross check this and merge if happy.

PR #405, opened 2015-08-28
Pandas macroatom issue fix

aoifeboyle:
Alternative fix to issue #402 that seems to remove all of the errors.

@ssim This removes all of the errors for me with pandas 0.16.2 (using the config. file you provided). Could you checkout this PR and check it works for you too?

The problem seems to arise when a column is added to a dataframe with an unsorted index. I think the warning is intended to highlight that the index is unsorted and therefore that the data being added may not match up with the index correctly. But the use of .insert() with the data in Pandas series form seems to work, probably because the index of the new data is then specified. That's my understanding of it anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
         - COMPILER=gcc
         - CYTHON_VERSION=0.21
         - NUMPY_VERSION=1.8
-        - PANDAS_VERSION=0.12
+        - PANDAS_VERSION=0.16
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@aoifeboyle thanks! good catch

ssim:
@aoifeboyle Yes - this seems to clear it up perfectly for me! Good work!

@wkerzendorf - if you are also happy with this, let's merge it?

PR #409, opened 2015-09-07
A new structure to also take care of isotopes

wkerzendorf:
This is the beginning of a new structure that can hold isotope abundances and then decay them using nuclear networks. The structure inherits from a `DataFrame` and has a multi-index with `atomic_number` and `mass_number`. The columns are shells like in all our other TARDIS frames. 

Here is an example. This is the input 
```python
>>> simple_abundance_model
atomic_number mass_number
28            56           1  1
```
and after being decayed:
```python
simple_abundance_model.decay(100)
atomic_number mass_number
27            56             0.442379    0.442379
26            56              0.55752     0.55752
28            56           1.1086e-05  1.1086e-05
```

This is mostly to read Stephan Blondin's files but I'm sure can be used for other things. Currently, don't have the time to finish it.

 - [ ] finish tests
 - [ ] make it possible to write out elemental abundances
 - [ ] allow for elemental abundances with unknown isotope distribution
 - [ ] replace the current ARTIS file decay mechanism
 - [ ] read CMFgen files

### DIFF
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -1,0 +1,87 @@
+import pandas as pd
+from pyne import nucname, material
+from astropy import units as u
+
+class IsotopeAbundances(pd.DataFrame):
+
+    @property
+    def _constructor(self):
+        return IsotopeAbundances
+
+    def _update_material(self):
+        self.comp_dicts = [{}] * len(self.columns)
+        for (atomic_number, mass_number), abundances in self.iterrows():
+            nuclear_symbol = '%s%d'.format(nucname.name(atomic_number),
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@tardis-sn/tardis-core comments welcome.

rishivivek1:
@wkerzendorf would yoy please help me understand this project . 
Is there any decay algorithm(any documentation ) so that i can study over that and implement a structure which inherits the raw input and gives the decayed fraction output (if i understood it well enough).

wkerzendorf:
Have a look at pyne- we use their underlying architecture.

Issue #411, opened 2015-09-24
Deal with missing zeta data in a more sensible way


wkerzendorf:
https://github.com/tardis-sn/tardis/blob/master/tardis/plasma/properties/ion_population.py#L67

extend the interpolation.

aoifeboyle:
In the old plasma I made it so that the zeta values for the highest temperature available were used even for temperatures exceeding that value. Do you want me to just implement that again here?

wkerzendorf:
@aoifeboyle I think such an extrapolation would be good. The second small issue is that when complaining about missing zeta values that it doesn't specify which ones. It would be good to at least say which atoms. What do you think?

aoifeboyle:
Sure! I can try to do that. Shouldn't be a problem.

wkerzendorf:
@aoifeboyle don't be alarmed but currently the master is broken. There is an issue with pytest. I should have fixed it by now.

aoifeboyle:
@wkerzendorf Ok, I assume that's why my Travis build isn't working? I'm just relieved it's not something I've done!

wkerzendorf:
@aoifeboyle yes it was a problem with new versions of pytest and coverage. It should all be fixed now.

wkerzendorf:
@aoifeboyle I think this issue is fixed - right? If you agree, just close the issue.

PR #412, opened 2015-09-25
Fix for zeta issues.

aoifeboyle:
Response to issue #411

### DIFF
--- a/tardis/plasma/properties/atomic.py
+++ b/tardis/plasma/properties/atomic.py
@@ -153,8 +153,6 @@ class ZetaData(BaseAtomicDataProperty):
     outputs = ('zeta_data',)
 
     def _filter_atomic_property(self, zeta_data, selected_atoms):
-        for temperature in range(42000, 82000, 2000):
-            zeta_data[temperature] = zeta_data[40000]
         zeta_data['atomic_number'] = zeta_data.index.labels[0] + 1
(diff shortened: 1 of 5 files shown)


tardis/plasma/properties/atomic.py
```diff
@@ -153,6 +153,8 @@ class ZetaData(BaseAtomicDataProperty):
     outputs = ('zeta_data',)
 
     def _filter_atomic_property(self, zeta_data, selected_atoms):
+        for temperature in range(42000, 82000, 2000):
```
wkerzendorf: We should extrapolate and not increase the storage. I mean in the end how far it is outside -> outside means use the last valid value

wkerzendorf:
@aoifeboyle instead of using `interpolate.interp1d` use this: `interpolate.UnivariateSpline(x, y, k=1, s=0, ext=3)` that should work to use the last values

aoifeboyle:
@wkerzendorf Ok, I've had a look at that and it only works with 1-D arrays for the x and y values, so it will only work if I create a loop and interpolate individually for each ion. Is that alright?

wkerzendorf:
@aoifeboyle - that's right. I'm just fixing this - I'll make a PR to your PR. I should be done momentarily. We can also skype if you want.

wkerzendorf:
@aoifeboyle I made a PR to your PR. have a look and merge if happy.

PR #420, opened 2015-10-02
Memory issue

### LINKED ISSUE #419: Problem with inverse electron densities - no packets are escaping
Some weird behaviour in the new plasma calculation or/and in the new layout or the ```montecarlo.pyx``` file lead to the following severe problem:

During each montecarlo packet propagation cycle, not a single packet escaped through the outer boundary of the computational domain. It seems that the simulation thinks that the distance to the next electron scattering event is zero. Thus, all injected packets immediately scatter and are eventually lost at the inner boundary. Somehow, probably somewhere in the montecarlo.pyx file the inverse electron densities are messed up (potentially pointing to uninitialised memory).

wkerzendorf:
Fixes #419 - I think. So this is I guess a good for problems of this class: We attach a "permanent" pointer to the runner class (the runner class is exactly for this to hold data for the C-part of TARDIS). @chvogl @unoebauer can you run your toolkit to see if there are any other dropped references - so we can make sure this works there.

### DIFF
--- a/tardis/montecarlo/montecarlo.pyx
+++ b/tardis/montecarlo/montecarlo.pyx
@@ -118,8 +118,11 @@ cdef initialize_storage_model(model, runner, storage_model_t *storage):
     #electron density
     storage.electron_densities = <double*> PyArray_DATA(
         model.plasma_array.electron_densities.values)
-    storage.inverse_electron_densities = <double*> PyArray_DATA(
+
+    runner.inverse_electron_densities = (
         1.0 / model.plasma_array.electron_densities.values)
+    storage.inverse_electron_densities = <double*> PyArray_DATA(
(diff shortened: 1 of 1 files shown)


wkerzendorf:
@unoebauer @chvogl if there are no further objections - I'll merge this.

unoebauer:
In principle this looks reasonable - we are currently checking whether this could also help with the bf-integration of @chvogl

unoebauer:
Looks ok - merge at will

wkerzendorf:
@unoebauer does this solve your problem?

unoebauer:
I think it is definitely a step in the right direction - we will test the same concept with the bf-integration. But I think that the garbage collector should now not interfere any more.

PR #432, opened 2015-10-30
fix for initializing a linspace with quantities

wkerzendorf:
Small but important fix to ensure that TARDIS will work with newer versions of astropy/numpy as well as older ones.


wkerzendorf:
@yeganer @orbitfold that was my linspace fix for `numpy 1.10` - close if not applicable anymore

PR #435, opened 2015-11-11
Plasma/to hdf

wkerzendorf:
As I was working on comparing old runs to new runs, I figured that it would be useful to be able to save the whole plasma to disk via HDF5. 

@tardis-sn/tardis-core need someone to sign off on this.

### DIFF
--- a/tardis/plasma/base.py
+++ b/tardis/plasma/base.py
@@ -3,6 +3,7 @@ import tempfile
 import fileinput
 
 import networkx as nx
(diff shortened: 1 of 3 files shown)


orbitfold:
Looks good.

unoebauer:
I'd suggest that we worry about Issue #436 first - currently the model produced with the legacy run or the newly created Simulation framework doesn't seem to be able to save data to disk via ```to_hdf5```

Issue #436, opened 2015-11-13
legacy_run_simulation of Simulation restructure breaks to_hdf5 in radial1d_model


unoebauer:
Summary
------------
It seems that with the separation of the simulation and model aspects in PR #426, some inconsistencies were introduced in the radial1d_model handling, in particular when the legacy_run_simulation is used.

Details
---------
Running the tradis_example interactively in an ipython session using the run_tardis method, the simulation runs smoothly.
 
```python
from tardis import run_tardis
import yaml
config = yaml.load(open("tardis_example.yml"))
mdl = run_tardis(config)
```
 However, some features of the radial1d_model do not work properly:

```ipython
In [8]: mdl.line_interaction_type
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-f0a32236c03b> in <module>()
----> 1 mdl.line_interaction_type

/home/ulrich/python-virtualenv/tardis/lib/python2.7/site-packages/tardis_sn-1.5.dev1998-py2.7-linux-x86_64.egg/tardis/model.pyc in line_interaction_type(self)
    131     @property
    132     def line_interaction_type(self):
--> 133         return self._line_interaction_type
    134 
    135     @line_interaction_type.setter

AttributeError: 'Radial1DModel' object has no attribute '_line_interaction_type'

In [9]: 
```

For similar reasons, the to_hdf5 method is also broken:

```python
In [5]: mdl.to_hdf5("test_model.hdf5")
tardis.model - INFO - Writing to path 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-798e8648cb46> in <module>()
----> 1 mdl.to_hdf5("test_model.hdf5")

/home/ulrich/python-virtualenv/tardis/lib/python2.7/site-packages/tardis_sn-1.5.dev1998-py2.7-linux-x86_64.egg/tardis/model.pyc in to_hdf5(self, buffer_or_fname, path, close_h5)
    327         for key in include_from_model_in_hdf5:
    328             if include_from_model_in_hdf5[key] is None:
--> 329                 _save_model_property(getattr(self, key), key, path, hdf_store)
    330             elif callable(include_from_model_in_hdf5[key]):
    331                 include_from_model_in_hdf5[key](key, path, hdf_store)

AttributeError: 'Radial1DModel' object has no attribute 'last_line_interaction_in_id'
```

[...]

PR #440, opened 2015-11-27
started addressing issue #436

unoebauer:
- Currently, only the runner is passed back to the model in the legacy mode
- Todo: set remaining packet diagnostic arrays

 - [x] the runner has be accessible from inside the model
 - [x] set remaining packet diagnostic arrays
 - [x] model.to_hdf5 has to work
 - [x] the tardis gui has to work
 - [x] add tests

### DIFF
--- a/tardis/simulation/base.py
+++ b/tardis/simulation/base.py
@@ -329,7 +329,21 @@ class Simulation(object):
         #pass the runner to the model
         model.runner = self.runner
         #TODO: pass packet diagnostic arrays
+        (montecarlo_nu, montecarlo_energies, model.j_estimators,
+                model.nubar_estimators, last_line_interaction_in_id,
+                last_line_interaction_out_id, model.last_interaction_type,
+                model.last_line_interaction_shell_id) = model.runner.legacy_return()
 
+        model.montecarlo_nu = self.runner.output_nu
+        model.montecarlo_luminosity = self.runner.packet_luminosity
+
+
+        model.last_line_interaction_in_id = model.atom_data.lines_index.index.values[last_line_interaction_in_id]
+        model.last_line_interaction_in_id = model.last_line_interaction_in_id[last_line_interaction_in_id != -1]
+        model.last_line_interaction_out_id = model.atom_data.lines_index.index.values[last_line_interaction_out_id]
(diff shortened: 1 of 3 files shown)


tardis/simulation/base.py
```diff
@@ -310,6 +311,24 @@ def legacy_update_spectrum(self, model, no_of_virtual_packets):
...
+
+        Returns
+        -------
+            : None
+
+        """
+
+        #pass the runner to the model
+        model.runner = self.runner
+        #TODO: pass packet diagnostic arrays
```
wkerzendorf: the diagnostic arrays are bound to the runner (at least they should be).

wkerzendorf:
thanks!

unoebauer:
Looking at a simple tardis_example run, the model instance with is returned by run_tardis is still lacking:
 - last_line_interaction_in_id
 - last_line_interaction_out_id
 - last_line_interaction_shell_id
 - last_line_interaction_angstrom
This list is based on a quick inspection of include_from_model_in_hdf5 in model.py. One should also check the gui to make sure that nothing else is missing.

I'll take a look at that and will also add back the missing packet arrays.

tardis/simulation/base.py
unoebauer: True - but to have full legacy suppport, the model should have some of these arrays (with the appropriate cuts) directly, see my comment below

unoebauer:
After last commit, the to_hdf5 routine works again. Only irritating issue is the following warning that pops up:

```python
mdl.to_hdf5("test.hdf5")
tardis.model - INFO - Writing to path 
py.warnings - WARNING - /home/ulrich/python-virtualenv/tardis/lib/python2.7/site-packages/pandas/io/pytables.py:2577: PerformanceWarning: 
your performance may suffer as PyTables will pickle object types that it cannot
map directly to c-types [inferred_type->mixed,key->values] [items->None]

  warnings.warn(ws, PerformanceWarning)
```

unoebauer:
While running the gui, it emerged that the model is missing the following attributes:
 - converged

unoebauer:
With the latest commit, the gui is running again.

unoebauer:
Added tests for tests for runner and model attributes. Worked offline - let's see if Travis is happy...

unoebauer:
Can somebody have a proper look at this, @wkerzendorf? I'd say we are ready to merge

Issue #191, opened 2014-10-04
Tardis - segmentation fault


cradesto:
With the latest version ('tardis-sn==1.0.0.dev1328' from Fri Oct 3 12:14:33 2014)
python crashes if we will use next configs (see below).

For example with the version ('tardis-sn==1.0.0.dev1142' from Tue Jul 1 18:13:34 2014 +0300) it does not.

for convenience
https://dl.dropboxusercontent.com/u/13622198/tardis_segmentation_fault.zip

# --- BEGIN tardis_pure_hydrogen_with_datafiles_ph.yml
tardis_config_version: v1.0
supernova:
    luminosity_requested: 2.315e+8 solLum
    time_explosion: 14.806 day

atom_data: kurucz_cd23_chianti_H_He.h5

model:
    structure:
        type: file
        filename: density_ph.dat
        filetype: simple_ascii
    abundances:
        type: uniform
        H: 1.0

plasma:
    disable_electron_scattering: no
    ionization: nebular
    excitation: lte
    radiative_rates_type: dilute-blackbody
    line_interaction_type: macroatom

montecarlo:
    seed: 23111963
    no_of_packets : 4.0e+4
    iterations: 20
    black_body_sampling:
        start: 1 angstrom
        stop: 1000000 angstrom
        num: 1.e+7
    last_no_of_packets: 1.e+5
    no_of_virtual_packets: 50
    convergence_criteria:
        type: specific
        damping_constant: 1.0
        threshold: 0.05
        fraction: 0.8
        hold: 3
        t_inner:
            damping_constant: 1.0

spectrum:
    start : 500 angstrom
    stop : 12000 angstrom
    num: 10000
# --- END tardis_pure_hydrogen_with_datafiles_ph.yml

# --- BEGIN density_ph.dat
14.806 day 
# index velocity (km/s) density (g/cm^3)
0   6.840295E+03  8.710827E-14
1   6.961582E+03  1.613836E-13
2   7.099322E+03  9.241860E-14
3   7.359631E+03  3.120753E-14
4   7.712450E+03  1.430077E-14
5   8.089749E+03  8.212198E-15
6   8.476791E+03  4.915406E-15
7   8.858690E+03  3.070856E-15
8   9.217784E+03  2.027493E-15
9   9.530969E+03  1.458678E-15
10  9.784131E+03  1.147518E-15
11  9.919257E+03  1.394631E-15
12  9.960857E+03  3.003028E-15
[...]

PR #446, opened 2015-12-01
calculate virtual spectrum independently of vpacket loggin - fixes #445

### LINKED ISSUE #445: Virtual Spectrum Zero
I noticed this while investigating issue #191: using the current upstream/master 34f667a465b36b7f76f7e9229a7d93d057c38d4a, Tardis does not produce a virtual spectrum any more. The following model attributes are all arrays populated exclusively with zeros:

 - mdl.montecarlo_virtual_luminosity
 - mdl.spectrum_virtual.lumosity_density_nu
 - mdl.spectrum_virtual.lumosity_density_lambda

This is a serious issue. It is also worrisome that none of our automated tests picked this one up. @wkerzendorf: we should include a ```test_virtual_spectrum``` routine into the ```test_tardis_full``` script.

unoebauer:
- [x] test with --with-vpacket-logging
- [x] test without --with-vpacket-logging


wkerzendorf:
ready for merge?

unoebauer:
Tested it with the tardis_example setup and with the setup of issue #191, both with and without the vpacket_logging flag. The virtual packet spectrum is now properly generated again. 

@wkerzendorf, @orbitfold: quick inspection of the code from your side and then merge at will.

PR #449, opened 2015-12-01
added test to verify virtual spectrum

### LINKED ISSUE #447: Test for virtual spectrum required
Issue #445 clearly demonstrated that we need to test explicitly the correct determination of the virtual spectrum. We could simply add a  ``test_virtual_spectrum`` method to ``test_tardis_full.py``, which does essentially the same as ``test_spectrum`` but for the virtual spectrum.

This would imply that we do another bitwise spectral comparison. I guess this is our best option until the statistics project has matured enough....Other ideas, @orbitfold, @wkerzendorf, @ssim, @chvogl?

unoebauer:
fixes #447


unoebauer:
Ready to be merged

Issue #390, opened 2015-08-21
Montecarlo: Missing dependencies in the build script


mreineck:
If I run "python setup.py develop --user", then make a change to any of ther .h files in tardis/montecarlo/src/ and finally run "python setup.py develop --user" again, the C files are not recompiled as they should be. This dependency on the header files should somehow be added to the build script in question (I expect that it is tardis/montecarlo/setup_package.py), but I don't know how this can be done.

mreineck:
Same problem with the dependency between .pyx and .c files ... cython does not seem to be re-run automatically whenever a .pyx file is changed.

mreineck:
Another observation: if I delete a .c file by hand that was originally created by Cython, the compilation proces fails completely:
```
/scratch/martin/mytardis2>python setup.py develop --user
running develop
running egg_info
writing requirements to tardis_sn.egg-info/requires.txt
writing tardis_sn.egg-info/PKG-INFO
writing top-level names to tardis_sn.egg-info/top_level.txt
writing dependency_links to tardis_sn.egg-info/dependency_links.txt
writing entry points to tardis_sn.egg-info/entry_points.txt
reading manifest file 'tardis_sn.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'CHANGES.rst'
warning: no files found matching '*.c' under directory '*.pyx'
warning: no files found matching '*.pxd' under directory '*.pyx'
warning: no files found matching '*.h' under directory '*.pyx'
warning: no files found matching '*' under directory 'licenses'
warning: no files found matching '*' under directory 'cextern'
no previously-included directories found matching 'build'
no previously-included directories found matching 'docs/_build'
no previously-included directories found matching 'docs/api'
warning: no previously-included files matching '*.o' found anywhere in distribution
writing manifest file 'tardis_sn.egg-info/SOURCES.txt'
running build_ext
error: [Errno 2] Could not find C file tardis/montecarlo/montecarlo.c for Cython file tardis/montecarlo/montecarlo.pyx when building extension tardis.montecarlo.montecarlo. Cython must be installed to build from a git checkout.: 'tardis/montecarlo/montecarlo.c'
```

This happens despite the fact that Cython was working perfectly well when I first compiled the package directly after cloning from Github.

I have no idea what's happening here.

wkerzendorf:
@embray - this is using the astropy setup tools. Any idea why?

embray:
Something seems to be awry here:
```
[...]

PR #452, opened 2015-12-11
Build fix and nthreads fix - fixes #356

### LINKED ISSUE #356: omp get number of threads
We should use omp_get_num_threads in the fprintf statements in cmontecarlo.c instead of the value of the parameter "nthreads" since there is no guarantee omp will start the number of threads we request.

unoebauer:
A tiny PR fixing two things:
- issue #356: in fprintf, omp_get_num_threads is used instead of nthreads
- build system issue: by adopting the suggestion by @chvogl (see conversation of #390), cython should now be triggered automatically once C-files are modified


unoebauer:
Once Travis is happy, this one is ready to be merged, @orbitfold, @wkerzendorf

PR #453, opened 2015-12-15
Moved cythonize to the tardis/montecarlo/setup_package.py

yeganer:
Since tardis.montecarlo is the only Cython module the cythonize should be used in the setup_package.py to garantee no weird bugs in the future.


tardis/montecarlo/setup_package.py
```diff
@@ -30,14 +31,16 @@ def get_extensions():
...
 
-    return [Extension('tardis.montecarlo.montecarlo', sources,
-                      include_dirs=['tardis/montecarlo/src',
-                                    'tardis/montecarlo/src/randomkit',
-                                    np.get_include()],
-                      depends=deps,
-                      extra_compile_args=compile_args,
-                      extra_link_args=link_args,
-                      define_macros=define_macros)]
+    return cythonize(
```
wkerzendorf: @embray we are using package template but have the problem that when `c-files` change that are linked to `pyx-files` `python setup.py develop` does not recompile. @yeganer suggests adding cythonize in a `setup_package.py` to ensure this behaviour. Is this "`astropy`"-conform. Should we include such an example in `package-template`?

embray: That wouldn't do anything since this code is only run when the `setup.py` is run.  In any case adding cythonize here doesn't do anything regardless, since `astropy_helpers.commands.build_ext` does this for you automatically.

Something I've wanted to do for a while but has been low priority is to enhance [this code](https://github.com/astropy/astropy/blob/v1.1/astropy/__init__.py#L195) (there's a version of this in the affiliated package template too) so that it automatically rebuilds extension modules any time one of them has been changed.

In the meantime, if you'd doing development on extension modules you have to manually run `./setup.py build_ext --inplace`.

tardis/montecarlo/setup_package.py
```diff
@@ -30,14 +31,16 @@ def get_extensions():
...
+        Extension('tardis.montecarlo.montecarlo', sources,
+            include_dirs=['tardis/montecarlo/src',
+                'tardis/montecarlo/src/randomkit',
+                np.get_include()],
```
embray: This should use `include_dirs=['numpy']` as documented here: http://docs.astropy.org/en/latest/development/ccython.html#using-numpy-c-headers

wkerzendorf:
@yeganer this can be closed - right?

PR #456, opened 2015-12-16
Renamed virtual_packet_flag to virtual_packet_count

yeganer:
Renamed virtual_packet_flag to virtual_packet_count in the C backend to better reflect it's meaning.

I would suggest to also rename rpacket_get_virtual_packet to something like rpacket_get_type and introduce an enumerate to reflect the different types of packets ( REAL, different virtual ones).

If that's a good idea, i'll create another pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ distribute-*.tar.gz
 # Other
 .*.swp
 *~
+.ycm_extra_conf.py
 
 # Mac OSX
 .DS_Store
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -102,7 +102,7 @@ class MontecarloRunner(object):
                                  no_of_packets)
 
         montecarlo.montecarlo_radial1d(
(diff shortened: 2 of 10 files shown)


wkerzendorf:
@yeganer that looks good - but we will wait with the merge until @chvogl has put his PR up. So you can then rebase on this. I've thought about important things that the code needs. One of these important things is testing. Can you look at https://github.com/tardis-sn/tardis/blob/master/tardis/montecarlo/tests/test_cmontecarlo.py - I commented out these tests (read-up on py.test) because they produce segfaults. Can you look at this --- maybe there is also a better way to get the tests integrate in py.test.

yeganer:
What's the status with this PR? Is this still something that should be done( i think it will help when reading the code ). 
Should i update the PR? Otherwise i'd close this PR.

unoebauer:
Looks like a sensible change - since this is basically a refactoring and since @chvogl's continuum PR won't be merged in the near future, I'd be ok with merging this one now.

Issue #454, opened 2015-12-15
Open-MP number of threads not properly returned in info string


unoebauer:
I think this one was introduced by PR #452: when running with Open-MP, the information message does not provide the correct number of threads. Instead, it seems to always return 1:

```
Running with OpenMP - 1 threads
```
I've requested 16 threads in the yml file and my process manager confirms that 16 threads are indeed opened.

yeganer:
After merging #457 this can be closed.

yeganer:
Closing since #457 is merged.

unoebauer:
Re-opening this one since PR #457 is **not** yet merged!

PR #457, opened 2015-12-16
Fix Issue #454

yeganer:
A small fix for issue #454 
omp_get_num_threads() returns number of threads in the current team.
Outside a parallel region this is always 1.

See:
https://gcc.gnu.org/onlinedocs/libgomp/omp_005fget_005fnum_005fthreads.html


unoebauer:
We shouldn't use ``nthreads`` since we may not always get the requested number of threads (see Issue #356)

yeganer:
@unoebauer Issue should be fixed now. I had to rewrite history for a pretty commit.

unoebauer:
Testing it now...

unoebauer:
@yeganer 
works like a charm

@orbitfold, @wkerzendorf 
can you guys sign off on this so I can merge?

unoebauer:
@orbitfold, @wkerzendorf, @chvogl 
Please check and sign off on this so that we can merge it!

PR #458, opened 2015-12-16
Continuum Interactions

chvogl:
## Outline

Extends TARDIS by including continuum as well as collisional processes. 
The implementation is based on the scheme described in ([Lucy 2002](http://adsabs.harvard.edu/abs/2002A%26A...384..725L)) and ([Lucy 2003](http://adsabs.harvard.edu/abs/2003A%26A...403..261L)).  
The flow chart below shows all implemented radiation-matter-interactions and the resulting energy flows. Depicted in blue are the interactions which are currently treated in TARDIS:
macro_atom_flowchart


chvogl:
WARNING: 
At the moment conditional cython compilation is employed to switch the normalization of the transition probabilities on and off. This is probably not what we want to do in the future.

unoebauer:
level_emissivity_test

chvogl:
### Level emissivity test
#### Goal
Test if the extended macro atom scheme can reproduce asymptotically the emissivity of an atomic species whose level populations are in statistical equilibrium ([Lucy 2002](http://adsabs.harvard.edu/abs/2002A%26A...384..725L)).
#### Description
In this specific test case the emission derives entirely from collisional excitations, with the ambient radiation field set to zero. The only parameters of this test are the electron temperature and density. With these values specified, the statistical equilibrium problem can be solved for the level number densities from which accurate values for the level emissivities can be calculated.
These values can be compared to estimates of the level emissivity based on the Monte Carlo simulation.
#### Setup
* 20-level model of the H atom
* electron temperature of 20000 K
* electron density of order 1e11 cm^-3
* about 1e7 macro atom activations

#### Results
The plot below shows the absolute fractional errors of the MC level emissivities as a function of the energy levels:
level_emissivity_test

Issue #427, opened 2015-10-19
PEP8 Violations


souravsingh:
While working on an issue,I noticed PEP8 violations on some files.I am working on extracting the list of files with the violations.

wkerzendorf:
have a lookat #426 - there are some fixes for that. but many other files have that problem. and are not fixed.

souravsingh:
Here are some lists of files with PEP8 Violations-
```

    9    ./tardis/tests/test_util.py
    9    ./tardis/tests/test_plasma_simple.py
    9    ./tardis/tests/test_atomic.py
    9    ./tardis/plasma/properties/plasma_input.py
    9    ./tardis/io/tests/test_model_reader.py
    9    ./ah_bootstrap.py
    95    ./tardis/gui/datahandler.py
    8    ./tardis/plasma/tests/test_property_atomic.py
    8    ./tardis/plasma/tests/test_general_properties.py
    7    ./tardis/plasma/tests/test_property_radiative.py
    7    ./tardis/plasma/tests/test_property_nlte.py
    7    ./tardis/plasma/tests/test_partition_function.py
    7    ./tardis/plasma/properties/general.py
    7    ./tardis/plasma/exceptions.py
    6    ./tardis/montecarlo/setup_package.py
    6    ./tardis/io/util.py
    69    ./tardis/plasma/properties/nlte.py
    68    ./tardis/montecarlo/tests/test_cmontecarlo.py
    63    ./tardis/plasma/tests/conftest.py
    63    ./tardis/analysis.py
    5    ./tardis/tests/test_tardis_full.py
    5    ./tardis/plasma/setup_package.py
    5    ./tardis/montecarlo/tests/test_rpacket.py
    5    ./tardis/base.py
    5    ./tardis/_astropy_init.py
    5    ./asv/benchmarks/benchmarks.py
    4    ./tardis/plasma/tests/test_complete_plasmas.py
    4    ./tardis/montecarlo/tests/test_packet_source.py
    44    ./tardis/plasma/standard_plasmas.py
    42    ./tardis/io/model_reader.py
    40    ./tardis/plasma/properties/ion_population.py
    405    ./tardis/gui/widgets.py
    3    ./tardis/stats/base.py
    3    ./tardis/plasma/tests/test_property_level_population.py
    3    ./tardis/montecarlo/packet_source.py
    36    ./tardis/plasma/properties/radiative_properties.py
    36    ./tardis/plasma/properties/partition_function.py
[...]

PR #460, opened 2016-01-03
Some PEP8 Fixes done for Tardis

souravsingh:


souravsingh:
This is related to Issue #427

yeganer:
It seems like you moved the files. If this PR is still open, please move the files to their original position in the filetree and update this PR. Currently it would only create 3 new files in the root directory and modify `docs/conf.py`.

Issue #455, opened 2015-12-15
Inconsistencies between Tardis versions


unoebauer:
**Main description:**

TARDIS produces different output between 45eca24d77eacc91d0a0b875e5521bc2d3dc151a and master with this YML file. Why? and what commit introduced this change. 

The following YML files provide a minimum "working" example": [issue_455.zip](https://github.com/tardis-sn/tardis/files/81324/issue_455.zip)

-------
The diagnosis of the problem at hand is still work in progress, but it seems the most recent Tardis version, c9e1b1761a6745d44bae973d962e1be180295827, is inconsistent with a previous state of Tardis.

All this has been triggered by re-examining some detailed runs performed during the Abundance Tomography Project of Talytha. For that, we mainly used a Tardis version similar to aa7da173d0e7339bc962bafb62d34d60596fd951 (a working Open-MP implementation has just been incorporated, PR #338). Comparing the virtual spectra obtained with the recent Tardis version for the same setup, reveals huge discrepancies:

![tardis_cmp](https://cloud.githubusercontent.com/assets/6351165/11809315/4e0b5596-a326-11e5-8d46-6ce036514e7e.png)

The new Tardis version finds a much higher inner temperature. This goes hand in hand with a much higher radiation temperature in the ejecta and a higher ionisation state (and in turn electron density).

![ionisation_cmp](https://cloud.githubusercontent.com/assets/6351165/11809474/7af337d0-a327-11e5-90e5-d6f8c414032f.png)

### Setup:

* ``tardis_00173_2.yml``
```
atom_data: kurucz_cd23_chianti_H_He.h5
model:
  abundances:
    filename: abundances_00173_2.dat
    filetype: simple_ascii
    type: file
  structure:
    filename: densities_00173_2.dat
    filetype: simple_ascii
    type: file
    v_inner_boundary: 12000.000 km/s
    v_outer_boundary: 35000.000 km/s
montecarlo:
  black_body_sampling:
    num: 1000000
    start: 1 angstrom
    stop: 1000000 angstrom
  convergence_criteria:
    damping_constant: 1.0
    fraction: 0.8
    hold: 3
    t_inner:
      damping_constant: 1.0
    threshold: 0.05
    type: specific
  iterations: 20
  last_no_of_packets: 500000
  no_of_packets: 50000
  no_of_virtual_packets: 3
  nthreads: 16
  seed: 23111963
plasma:
  disable_electron_scattering: false
  excitation: dilute-lte
[...]

PR #462, opened 2016-01-15
Fixing plasma in 41d8fcd (not for merging)

aoifeboyle:
Issue #455: 41d8fcd is the commit at which the plasma_restructure branch was originally merged with the Tardis master branch. Here I have fixed a couple of things to ensure the plasma results are consistent with those of 45eca24 (last working example). All of the plasma quantities (taus, transition_probabilities, etc.) now agree between this version and 45eca24. But the nubar_estimators are still wrong. I think this may be a view/copy issue. Someone who knows more about the C/Python interface needs to look at these parameters to see which are being transferred wrong.

### DIFF
--- a/tardis/plasma/properties/ion_population.py
+++ b/tardis/plasma/properties/ion_population.py
@@ -65,7 +65,7 @@ class PhiSahaNebular(ProcessingPlasmaProperty):
                              '- requested {2}'.format(
                 zeta_data.columns.values.min(), zeta_data.columns.values.max(),
                 t_rad))
-        phis = general_phi * delta * w * (zeta + w * (1 - zeta)) * \
+        phis = general_phi * w * (delta * zeta + w * (1 - zeta)) * \
                (t_electrons/t_rad) ** .5
         return phis
 
--- a/tardis/plasma/properties/partition_function.py
+++ b/tardis/plasma/properties/partition_function.py
@@ -82,10 +82,10 @@ class LevelBoltzmannFactorNLTE(ProcessingPlasmaProperty):
         self._update_inputs()
 
     def _main_nlte_calculation(self, nlte_species, atomic_data, nlte_data,
-        t_electrons, j_blues, beta_sobolevs, general_level_boltzmann_factor,
+        t_electrons, j_blues_array, beta_sobolevs, general_level_boltzmann_factor,
         previous_electron_densities):
         for species in nlte_species:
-            j_blues = j_blues.values
+            j_blues_array = j_blues_array.values
             logger.info('Calculating rates for species %s', species)
             number_of_levels = atomic_data.levels.energy.ix[species].count()
             lnl = nlte_data.lines_level_number_lower[species]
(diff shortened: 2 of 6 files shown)


aoifeboyle:
@unoebauer I was planning to re-check my results from last night, but I messed something up with my conda environment when I was changing the pandas version to 0.16.0 to check for the error. Hopefully I'll get it fixed over the weekend...

yeganer:
When i try to mdl.to_hdf5('file') with this PR i'm getting a
```
(<type 'exceptions.AttributeError'>, AttributeError("'LegacyPlasmaArray' object has no attribute 'ion_populations'",), <traceback object at 0x7f3ab4ffca70>)
```
Error. (I had to add additional info in the error handling block to see, what's really going on)
Is this a common problem with that state of the master branch or is this related to your fix?

aoifeboyle:
@yeganer This was a problem with the original plasma/restructure merge that was fixed in later commits. I'm working on debugging the problem right now but later I can go back and add the fix to this version.

yeganer:
@aoifeboyle don't worry adding the fix to this version. Since i'm not familiar with Panda Dataframes and HDF5 yet i'll do my debugging and comparing with binary dumps in the C part anyway.

yeganer:
I did some more tests and found out that between 45eca24 and 41d8fcd the electron density changed only little. The change in electron density was fixed in d84d1f6 which brought the difference between 45eca24 and c1e1b97 below 1e-15.

@aoifeboyle : @unoebauer told me you found something which reduced the relative difference in your version to 1e-5? Can you please update your pull request so i can check if merging the PR with d84d1f6 solves the issue?

aoifeboyle:
@yeganer That actually falls right in line with what I was thinking - that the electron density disparity had been fixed in a later PR (I was getting a difference that @unoebauer was not with the current master). I did find a fix for the transition probabilities yesterday that reduced the relative errors to below 1.e-5, and that was with a relative electron density error of about the same magnitude in my current version. I sent Uli instructions for how to implement my fix, did you get a chance to try it with the master version yet @unoebauer? Basically it was to substitute the following for the `Lines` module in tardis/plasma/properties/atomic.py @yeganer (you can also try it):

    class Lines(ProcessingPlasmaProperty):
            outputs = (‘lines’, ‘nu’, ‘f_lu’, ‘wavelength_cm’)
            def calculate(self, atomic_data):
                    return atomic_data.lines, atomic_data.lines.nu, atomic_data.lines.f_lu, atomic_data.lines.wavelength_cm

The master has changed so much since my version that I think just taking this and pasting it in would be much quicker than doing a full merge. I promised myself I'd focus on other work today, and I know if I start looking at the Tardis stuff again I won't come back to it, but please let me know what you find.

yeganer:
@aoifeboyle Thanks for the code snippet. When i implemented that into d84d1f6 i got relative errors of 1e-13 which seems okay to me. However when implemented in the current master i got a relative error of 0.24.
I'm currently bisecting with your fix and trying to find the commit where the new error occured.

aoifeboyle:
@yeganer You mean relative errors in the transition probabilities or the nubar estimators of 1e-13? And were other plasma quantities (e.g. t_rad, electron_densities) the same?

unoebauer:
@yeganer: how about the transition probabilities? How big is their relative error? We always suspected that the nubar_estimator error may come from this. If we are now at the position that with @aoifeboyle 's fixes we get at least the transition probabilities right again, we made an important step into the right direction.

yeganer:
@unoebauer i just created a new PR with only the fix inside. As i wrote there, as far as i can tell all plasma properties have a relative error below 1e-12 ( the maximum rel. error i found were the transition_probabilities with 9.1e-13)

wkerzendorf:
@aoifeboyle can we close this - if yes .. please do.

unoebauer:
@wkerzendorf : this was definitely a by-product of the #455 debugging process. The changes of the PR only involve a simple code refactoring (jblues -> jblues_array) and one formula change. The latter is already in the current master. Consequently, I'll close this PR.

PR #465, opened 2016-01-19
fix for #455 (transition probabilities, tau_sobolevs)

yeganer:
With this fix the transition_probabilities, tau_sobololevs and electron densities have no significant error compared to 45eca24 (rel. error <1e-12)

However this doesn't solve the issue with the nu_bar_estimators. For that one need to look at f7ca576 which i identified in another git bisect as the root for the new discrepancy in nu_bar_estimators.


yeganer:
I'll close this issue since this version of the fix isn't based on the current master.

unoebauer:
Use PR #466 instead

Issue #464, opened 2016-01-16
Wrong nubar_estimators (potentially also inconsistencies in transition probabilities) -- follow-up of #455 


unoebauer:
This is a re-launch of #455, reviewing the current status of the debugging process and outlining the next steps. The reason for this re-launch was the increase in confusion in the vast discussion of #455.

## Important Conclusions from #455 

Here are the most important conclusions from the discussion of #455:

* for certain setups, wrong values of the ``nubar_estimators`` are produced at the end of a Monte Carlo cycle
* this leads to non-converging plasma-state iterations
* this affects the current Tardis version
* 45eca24 still produces the correct values of ``nubar_estimators``
* problems in the calculation of the transition_probabilities have been discovered during the debugging process; **however:** it is currently not clear whether this triggers the ``nubar_estimator`` behaviour

## Immediate Steps

I propose the following strategy:

### Immediate Steps:
- [ ]: confirm that PR #462 creates an identical plasma state (tau_sobolevs, electron_densities, transition_probabilities,  ion_populations, level_populations) as 45eca24 on the **python** level when using ``plasma_fail_test.yml``; @aoifeboyle found that there are no differences (based on 2e7a85f ), @unoebauer found differences in the transition_probabilities (based on 85c0c1e)
      * this may be due to problems introduced between 2e7a85f and 85c0c1e
      * another possibility is the use of different metrics to determine differences (**see guidelines below**)
- [ ]: once a version of Tardis after the plasma restructure has been found/produced, which produces the same plasma state as 45eca24 and still produces different ``nubar_estimators``, one should compare the plasma inputs on the **C** level -- this is clearly a task for @yeganer

### Ultimate Goals:
- [ ]: fix the nubar_estimator behaviour
- [ ]: understand why the issue did not manifest in any of the Travis test problems and in the tardis_example

## Guidelines

* Please only use the setups provided by PR #463 to debug this problem. For now, only ``plasma_fail_test`` setup is provided. If additional configurations are absolutely necessary, they will be added to the PR. Always refer to the setups by the name given to them in the PR to avoid confusion
* we should use a consistent metric for determining differences in plasma state quantities:
   * I am currently using the following prescription when determining differences of pandas.DataFrames (I am not saying that this is the best solution):
```python
np.fabs(((plasma_property_old - plasma_property_new) / plasma_property_old).fillna(0)).max()
```
[...]

PR #463, opened 2016-01-16
Setups for debugging #455 and (not for merging)

unoebauer:
This PR is not intended for merging with the code base. It's sole purpose is to provided the tardis config files for the calculations triggering issue #455 and 

Once these issues have been resolved this pull request may either be removed or its setups turned into proper tests.

PR #466, opened 2016-01-20
Probably fix for #455. ( Based on current master)

yeganer:


tardis/plasma/properties/atomic.py
```diff
@@ -39,7 +39,7 @@ def _set_index(self, levels):
         return (levels.index, levels['energy'], levels['metastable'],
             levels['g'])
 
-class Lines(BaseAtomicDataProperty):
```
wkerzendorf: we should continue to use the `BaseAtomicDataProperty`. If there is a problem with the ordering then we need to change the `set_index` method.

wkerzendorf:
@yeganer @aoifeboyle @unoebauer Congratulations on catching this - Well done!!

tardis/plasma/properties/atomic.py
aoifeboyle: @wkerzendorf I don't know how to fix this while leaving it as a `BaseAtomicDataProperty`. Could you recommend an alternative way of changing `set_index` so that it doesn't re-order the data?

wkerzendorf: I'm back soon. We'll talk about a fix then.
On Mon, Jan 25, 2016 at 13:19 aoifeboyle <notifications@github.com> wrote:

> In tardis/plasma/properties/atomic.py
> <https://github.com/tardis-sn/tardis/pull/466#discussion_r50675519>:
>
> > @@ -39,7 +39,7 @@ def _set_index(self, levels):
> >          return (levels.index, levels['energy'], levels['metastable'],
> >              levels['g'])
> >
> > -class Lines(BaseAtomicDataProperty):
>
> @wkerzendorf <https://github.com/wkerzendorf> I don't know how to fix
> this while leaving it as a BaseAtomicDataProperty. Could you recommend an
> alternative way of changing set_index so that it doesn't re-order the
> data?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/tardis-sn/tardis/pull/466/files#r50675519>.
>
-- 
Dr. Wolfgang E Kerzendorf
ESO Fellow
European Southern Observatory
Karl-Schwarzschild Str. 2, 85748
Garching bei Muenchen, Germany

unoebauer:
@aoifeboyle, @yeganer, @chvogl and I made some progress understanding the nature of the problem. We seemed to have revealed some deeply-rooted inconsistencies in the atomic data treatment in Tardis. I'll try to summarize:

It all seems to boil down to the presence of **two** line lists within Tardis. On the one hand there is ``model.atom_data._lines`` which initially (after ) contains the full information of the atom_data.hdf5 file. Within atomic.py, a copy of this original list is created, namely ``model.atom_data.lines`` which contains only a subset of the original data - filtering procedures are applied, which drop all the unnecessary entries from atom_data.lines, namely all information associated with atoms that are not present in the current calculation. Afterwards, this reduced line list is again sorted in wavelength space (even though the original data set, i.e. the ``hdf5`` file and ``atom_data._lines`` contained already wavelength ordered lists). Finally, the transition_probabilities are brought into the exact same order as ``atom_data.lines``. So far so good.

The problems start now in the plasma part. Here, the ``atom_data._lines`` list is read instead of the reduced and nicely prepared one. The following part in ``tardis/plasma/properties/base.py`` (lines is 146 - 155), describing the ``calculate`` method of the ``BaseAtomicDataProperty`` is responsible for that:
```python
# Atomic Data Issue: Some atomic property names in the h5 files are preceded
# by an underscore, e.g. _levels, _lines.
            try:
                raw_atomic_property = getattr(atomic_data, '_'
                                              + self.outputs[0])
            except AttributeError:
                raw_atomic_property = getattr(atomic_data, self.outputs[0])
            finally:
                return self._set_index(self._filter_atomic_property(
                    raw_atomic_property, selected_atoms))
```
Since ``atomic_data`` has a ``_lines`` attribute, the above try statement will never fail and this unfiltered original list will always be used in the ``Lines`` class of the plasma part. 

The ``Lines`` class now successfully filters the ``_lines`` list. Afterwards, the exact same lines are in ``_lines`` as in ``atom_data.lines``. But these two lists differ slightly in the ordering since the ``_lines`` list is **not** explicitly reordered according to wavelength. In this context, hyperfine transitions seem to be really problematic. The original Kurucz data base has a small number of transitions, which are at exactly the same frequency (1000 - 3000 seem to be in the ``hdf5`` file). The sorting process seems to reshuffle the order of these transitions, i.e. 165, 166, 167 in the original ``hdf5`` file will become 165, 167, 166 in ``atom_data.lines`` (but ``_lines`` will still be ordered as 165, 166, 167 since no resorting occurs). 

This now leads to inconsistencies between the ordering of the Plasma Line list (``_lines``) and the transitions_probabilities, which have been ordered in ``tardis/atomic.py`` **according** to ``atom_data.lines``. The re-indexing procedure in the does not solve this inconsistency since 
```python
        try:
            reindexed = lines.reindex(lines.index)
        except:
            reindexed = lines.reindex(lines.index)
```
produces a Line list which has exactly the same ordering.

At first sight, it seems strange that the mismatch in only a few thousand lines can cause the problem, in particular since the transitions which are mixed-up have the same frequency. However, within the macro-atom, wrongly identifying transition probabilities with the actual transitions may lead to populating levels which are actually forbidden - from there, the Macro Atom scheme (or downbranching scheme) may then take completely different roots and release r-packets in different transitions, ultimately leading to the discrepancies in the spectrum and the plasma state.

We have not been able to clearly identify why the issue did not show up in ``tardis_example`` or the Travis tests, since these "hyperfine" transitions are not restricted to certain elements. However, it seems plausible that the effect on the MA scheme may be more severe for some elements. In tardis_example and the Travis tests, only a handful of chemical species are taking into account. As a consequence, a very small line list is used in the calculation (containing only few of these problematic "hyperfine" transitions). In the setup triggering Issue #455, all elements from Z=6 to Z=30 are taken into account. Thus, almost the entire line list is used (only 2000 lines are discarded) and then the effect of the mismatch between line list and transition probabilities in these "hyperfine" transitions seems to be severe.

unoebauer:
## ...Continuation of previous comment

There are a few possibilities to fix issue #455:
* (**strongly discouraged**) still use ``_lines`` in the Plasma part, keep filtering capability and add a final sort instead of the re-indexing
*  (**better and cleaner solution**) leave all the filtering procedures in ``tardis/atomic.py`` and use the filtered line list in the plasma part, i.e. get ``atom_data.lines`` instead of ``atom_data._lines``. This way, there would be again a natural consistency between the used line list and ordering of the transition_probabilities

I would also suggest to get rid of the the "private" atomic properties (``_lines``, ``_levels``, etc) after the preparation process in ``tardis/atomic.py`` and only keep the filtered properties.

yeganer:
@unoebauer I think you mixed this part up:

> This now leads to inconsistencies between the ordering of the Plasma Line list (_lines) and the transitions_probabilities, which have been ordered in tardis/atomic.py according to atom_data.lines.

What actually happens is, the transition_probabilities are ordered according to the Plasma Line list (**filtered** `atom_data._lines`) but the C montecarlo part uses the ordering of `atom_data.lines` which is the **filtered** and **sorted** list.

I agree with @unoebauer that we should discard all unused properties after the preparation. If there is a need for `_lines` then one should rename it to something which better reflects it's purpose like `unfiltered_lines`.

Another thing to note is that for consitency it may be appropriate to sort atom_data.lines by the wavelength and then by line_id. That way no ''magic'' can happen during the sort and the ouput is always in an expected format( i.e. line 166 will always be between 165 and 167 if they all have the same wavelength)

ssim:
This sounds like a lot of progress in understand, which is great. 

It certainly sounds to me like the existence of this _lines thingy is not a good idea: if we never need it (and I don't see why we would?). I would support the idea of getting rid of it and having all the sorting done in atomic.py. I also like the suggestion of imposing a hierarchical sorting (wavelength, then line_id, as proposed) so that the format expected is concrete and can in fact be tested for an expected result (adding a test that the line indices are always what they should be would be nice to include here).

Well done everyone!

yeganer:
I updated the PR and removed `atom_data._lines` and `atom_data._levels` for testing purposes. After reflecting that changes in the plasma part there were no errors and i didn't notice a difference in the spectra ( comparing with 45eca24 ). So this seems to work and is a step in the right direction in my opinion.

I didn't change `ZetaData` nor `IonCXData` because i have no idea which format is expected by the current implementation and what is going on in the code.

### TODO: (personal opinion)
* Move all preparing or formatting of data into AtomData. ( relevant for `ZetaData` and `IonCXData` )
* If we want a safety net in the plasma part, check whether the input is in the expected format and abort otherwise. This should normally happen every time when passing data between modules (although it is unnecessary in most cases, it helps preventing nasty bugs)
* Ideally one would have Classes for AtomDate Properties that define how the data they hold is formatted and which are responsible for loading and formatting the data. ( I understand there is no urgent need for that feature but having that part of the code modular in the same way as the plasma part would improve the code)

yeganer:
@wkerzendorf @unoebauer @aoifeboyle @orbitfold 
Since #471 is merged now, i think we can close this PR and any related issues/pull requests now?

Related: #462 #463 #464

unoebauer:
Yes, let's close it - doing it now

PR #471, opened 2016-02-01
Quickfix to solve Issue #455 and #464

yeganer:
Deleted atom_data._lines and atom_data._levels
This change ensures that the BasePlasma always gets prepared data.
A proper fix will take more time and will be done later.


unoebauer:
I strongly suspect, that the Travis testing process will fail, at lest for the full_spectrum test. The spectra won't be bit-equal to the recorded values. You should re-calculate these tests offline, ensure that the new results are sensible and update the data files.

wkerzendorf:
@yeganer looks very good, we need to think about updating the tests. maybe grab a coffee now at ESO?

yeganer:
@wkerzendorf @unoebauer: I fixed all tests. They are running fine locally. If you could please look over the changes and run the test locally so we can be sure everything is working fine and we can merge this and finally fix the bug for master.

There are still 2 Issues with the code. 
* `test_atomic.py:test_atomic_reprepare()` should be rewritten and the logic should be changed( see comment in code) to check for  `len(lines) == N` and not `if len(lines) > 0`.

* `test_simulation.py` seems like a  "small" version of a full tardis test to me. Either move the logic ( comparing nubar, t_rad, w, j_est) to `test_tardis_full.py` or load a fixed model with a known plasma state and run the simulation on that. The current implementation will fail on any change in the codebase, whether it is in plasma, atomic, montecarlo or somewhere else. In that case the bug would be catched by `test_tardis_full.py` anyway.

Since these have nothing to do with the bug, I'll open new Issues for them. I just wanted to document them.

tardis/plasma/properties/atomic.py
```diff
@@ -58,18 +59,19 @@ class Lines(BaseAtomicDataProperty):
...
     def _filter_atomic_property(self, lines, selected_atoms):
-        return lines[lines.atomic_number.isin(selected_atoms)]
+        # return lines[lines.atomic_number.isin(selected_atoms)]
+        return lines
 
     def _set_index(self, lines):
 # Filtering process re-arranges the index. This re-orders it.
 # It seems to be important that the lines stay indexed in the correct order
 # so that the tau_sobolevs values are in the right order for the montecarlo
 # code, or it returns the wrong answer.
```
unoebauer: I would really remove the comments - this may lead to confusions in the future if somebody looks at it and doesn't remember the entire #455 discussion

tardis/tests/test_plasma_simple.py
```diff
@@ -7,6 +7,7 @@
 import warnings
 from tardis.plasma.standard_plasmas import LegacyPlasmaArray
 from tardis.io.util import parse_abundance_dict_to_dataframe
+import numpy as np
```
unoebauer: Looks like an unnecessary import to me

yeganer: You are right. That line is a leftover from debugging. I'll remove it.

tardis/plasma/properties/atomic.py
yeganer: My reasoning behind not removing it was that someday in the future when all the filtering and atom data preparation gets implemented in the plasma part having the old functionality at hand might be useful.

But your comment reminded me that BasicAtomDataProperty definitely needs a change.

unoebauer: Hmm - but I think that the entire idea behind the enforced re-indexing (which should only achieve compatibility between unsorted _lines and sorted lines) is no longer applicable. Moreover, the re-indexing procedure left in the comments is wrong. Given all that, I'd strongly suggest removing the comments.

unoebauer:
Apart from my concern about the comments in the Plasma part, I am happy with this PR and am in favour of merging ASAP. Opinions, @aoifeboyle, @wkerzendorf, @orbitfold?

tardis/plasma/properties/atomic.py
```diff
@@ -58,18 +59,19 @@ class Lines(BaseAtomicDataProperty):
...
-        try:
-            reindexed = lines.reindex(lines.index)
-        except:
-            reindexed = lines.reindex(lines.index)
-        lines = reindexed.dropna(subset=['atomic_number'])
+        # try:
+        #     reindexed = lines.reindex(lines.index)
+        # except:
+        #     reindexed = lines.reindex(lines.index)
+        # lines = reindexed.dropna(subset=['atomic_number'])
```
yeganer: @unoebauer I'm fine with removing that part since you are right, it is wrong and obsolete. However i'd leave the line filtering in place and maybe add a commented part to `_set_index()` which mirrors the current set_index that is done in atomic.py

unoebauer: Fine with me.

yeganer:
@unoebauer Could you please run the original setup where this bug first appeared and compare the output with 45eca24, just to be 100% sure.

tardis/conftest.py
```diff
@@ -133,10 +133,10 @@ def pytest_report_header(config):
     if opts:
         s += "Using Astropy options: {0}.\n".format(" ".join(opts))
 
-    if six.PY3 and (config.getini('doctest_rst') or config.option.doctest_rst):
+    if not six.PY2 and (config.getini('doctest_rst') or config.option.doctest_rst):
```
wkerzendorf: what's happening here?

tardis/io/tests/data/tardis_configv1_verysimple.yml
```diff
@@ -4,8 +4,8 @@ supernova:
     luminosity_requested: 2.8e9 solLum
     time_explosion: 13 day
 
-atom_data: kurucz_atom_pure_simple.h5
-
+# atom_data: kurucz_atom_pure_simple.h5
+atom_data: kurucz_cd23_chianti_H_He.h5
```
wkerzendorf: Why is that changed?

tardis/plasma/properties/atomic.py
wkerzendorf: I think that is a good idea - leave the `_set_index` method in there even if it doesn't do anything.

tardis/plasma/properties/base.py
```diff
@@ -79,8 +84,9 @@ def _update_inputs(self):
         """
         calculate_call_signature = self.calculate.func_code.co_varnames[
                                    :self.calculate.func_code.co_argcount]
-        self.inputs = [item for item in calculate_call_signature if
-                      item != 'self']
+        self.inputs = [
+                item for item in calculate_call_signature
+                if item != 'self']
```
wkerzendorf: PEP8?

tardis/simulation/tests/test_simulation.py
```diff
@@ -7,8 +7,15 @@
 from tardis.simulation import Simulation
 
 
+''' I don't get the purpose of this test. It looks like it's trying to do unit
```
wkerzendorf: comment's such as these should be added to the PR and not in the code. Code should only contain facts and not personal comments or questions.

tardis/atomic.py
```diff
@@ -463,11 +464,14 @@ def prepare_atom_data(self, selected_atomic_numbers, line_interaction_type='scat
             maximum ion number to be included in the calculation
 
         """
+        if not self.prepared:
+            self.prepared = True
+        else:
+            raise "AtomData was already prepared"
```
wkerzendorf: One should not raise generic Exceptions. Either raise one of the built-in exceptions (not sure if `ValueError` is descriptive) or better yet make your own Exception by subclassing from Exception or one of its subclasses e.g.:

```python
class AtomDataNotPreparedError(Exception):
    pass
```

and then `raise that`

wkerzendorf:
@yeganer - I've added a few comments, but it is almost done.

tardis/conftest.py
yeganer: It's changing the logic to treat python2 as the exception. That way this part doesn't need to be rewritten in the future once there is python 3 and python 4 around.
The old code would treat python4 the same way as python2 which would result in an undesired behavior.

tardis/io/tests/data/tardis_configv1_verysimple.yml
yeganer: I changed that since i wasn't able to find the specified atom_data file and it seemed to not matter because this setting is overwritten when running the tests anyway.

It bight be a good idea to discuss what should happen.
For now i can revert that to the old setting

tardis/plasma/properties/base.py
yeganer: Yes.

tardis/conftest.py
orbitfold: I wouldn't worry about Python 4. Even if it does eventually happen, I doubt compatibility issues will be handled by the six module, which is purely for compatibility issues between 2 and 3.

tardis/atomic.py
yeganer: Added a custom exception.

tardis/conftest.py
yeganer: I just ran across that and thought, why not fix it. If I run across a potential bug, is there any reason to not fix it?

yeganer:
@wkerzendorf I addressed most of your concerns. The only remaining issues are:

- verysimpleconfig.yaml restore original behavior?
- PY2/PY3 restore original behavior?

If that's settled, we can merge.

tardis/conftest.py
orbitfold: I'd argue the original behavior is not a bug.

wkerzendorf: okay - I think we should leave it as the old one. The only reason is that when our astropy friends will help us, they are not confused. please change it back.

tardis/io/tests/data/tardis_configv1_verysimple.yml
```diff
@@ -5,7 +5,6 @@ supernova:
     time_explosion: 13 day
 
 atom_data: kurucz_atom_pure_simple.h5
```
wkerzendorf: this is fine - if as you say this is never really used we could put a dummy string in. but leave it for now.

wkerzendorf:
I think checkout `conftest.py` and then merge from my side.
@orbitfold - do you agree to merge?
@unoebauer - do you agree to merge?

orbitfold:
Fine by me.

unoebauer:
@wkerzendorf Just a second - let me do the last final check. I'd like to rerun the original #455 problem. If this passes let's merge.

unoebauer:
@wkerzendorf The test looks excellent:
virt_spec_cmp

Let's merge

PR #475, opened 2016-02-04
Reorganize import statements in atomic.py

Pratik151:
came across this TODO when I was looking at codebase. 

Should I remove <code>from pandas import DataFrame</code> as we are already using <code> import pandas as pd</code> and use `pd.DataFrame` instead of `DataFrame`


wkerzendorf:
@orbitfold @yeganer - sensible change?

orbitfold:
Aesthetically I think this is better.

Issue #473, opened 2016-02-02
test_atomic.py:test_atomic_reprepare() improve logic


yeganer:
Currently `test_atomic.py:test_atomic_reprepare()` only checks `if len(lines) > 0`.

To improve it's logic and the coverage two improvements come to my mind:
- Rewrite to check `if len(lines) == N`. To get `N` the best approach would be to open the atomic data file, extract the lines, filter them and count them. 
- In a second step one could parametrize the function to work for different `atomic_numbers`. This would ensure that everything is working as expected. ( see pytest.mark.parametrize )

unoebauer:
This sounds very reasonable to me. We should adopt this. Other thoughts, @wkerzendorf?

yeganer:
I think this issue would be a great exercise for GSoC students. That's why i didn't implement it right away.

unoebauer:
Good idea

wkerzendorf:
@yeganer thanks!

unoebauer:
Let's try to close this issue by adopting #476!

PR #476, opened 2016-02-06
added two tests that replace test_atomic_reprepare()

mishinma:
@yeganer @unoebauer @wkerzendorf 
Hi, this is my first pull request! I've implemented the idea from #473 . 
- Firstly, I removed the  `pytest.mark.skipif` marker and moved all the verification to the `atomic_data_fname` fixture in the tardis/conftest.py. This way we don't have to check it in each test that requests an atomic database. I also included two other fixtures: `lines_dataset` that returns  lines data from hdf5 file and `atom_data_from_dataset` that requests the `atomic_data_fname` fixture and returns an AtomData object.
- Then I replaced `test_atomic_reprepare` with two tests: `test_prepare_atom_data_set_lines` and `test_prepare_atom_data_set_lines_w_max_ion_number`. I decided to go with to tests because I didn't want to overload them with logic - they are unit tests after all. So the second one does basically the same thing as the first but also checks max_ion_number. As suggested in #473 the number of lines is calculated from the provided atomic data file and then compared with the value from `prepare_atom_data` method. 

Look forward to a review

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -145,18 +145,41 @@ def pytest_report_header(config):
 import os
 import tardis
 import yaml
+import h5py
 
 from tardis.io.config_reader import Configuration
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def atomic_data_fname():
+    """The fixture provides the atomic dataset filename if it was passed as an argument."""
     atomic_data_fname = pytest.config.getvalue("atomic-dataset")
     if atomic_data_fname is None:
+        # if the name wasn't provided that the test requesting it would be skipped
         pytest.skip('--atomic_database was not specified')
     else:
-        return os.path.expandvars(os.path.expanduser(atomic_data_fname))
+        # check that the atomic dataset exists
+        atomic_data_fname = os.path.expandvars(os.path.expanduser(atomic_data_fname))
+        if not os.path.exists(atomic_data_fname):
+            raise AttributeError("{0} atomic datafiles does not seem to exist".format(atomic_data_fname))
+        return atomic_data_fname
+
+
+@pytest.fixture(scope="session")
+def lines_dataset(atomic_data_fname, request):
+    """ The fixture returns the dataset containing lines data from the provided dataset"""
+    h5_file = h5py.File(atomic_data_fname, 'r')
+    def fin():
+        h5_file.close()
+    request.addfinalizer(fin)
+    return h5_file['lines_data']
 
 from tardis.atomic import AtomData
+
+@pytest.fixture
+def atom_data_from_dataset(atomic_data_fname):
+    """The fixture returns the AtomData instance that contains atomic data from the provided dataset"""
+    atom_data = AtomData.from_hdf5(atomic_data_fname)
+    return atom_data
 
 @pytest.fixture
 def kurucz_atomic_data(atomic_data_fname):
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -145,9 +145,10 @@ def pytest_report_header(config):
 import os
 import tardis
 import yaml
-import h5py
 
 from tardis.io.config_reader import Configuration
+from tardis.atomic import AtomData
+
 
 @pytest.fixture(scope="session")
 def atomic_data_fname():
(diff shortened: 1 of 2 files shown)


tardis/conftest.py
```diff
@@ -145,20 +145,43 @@ def pytest_report_header(config):
...
-        return os.path.expandvars(os.path.expanduser(atomic_data_fname))
+        # check that the atomic dataset exists
+        atomic_data_fname = os.path.expandvars(os.path.expanduser(atomic_data_fname))
+        if not os.path.exists(atomic_data_fname):
+            raise AttributeError("{0} atomic datafiles does not seem to exist".format(atomic_data_fname))
+        return atomic_data_fname
+
+
+@pytest.fixture(scope="session")
+def lines_dataset(atomic_data_fname, request):
```
yeganer: I think this fixture belongs to test_atomic.py

tardis/conftest.py
```diff
@@ -145,20 +145,43 @@ def pytest_report_header(config):
...
+    h5_file = h5py.File(atomic_data_fname, 'r')
+    def fin():
+        h5_file.close()
+    request.addfinalizer(fin)
+    return h5_file['lines_data']
 
 from tardis.atomic import AtomData
 
 @pytest.fixture
+def atom_data_from_dataset(atomic_data_fname):
```
yeganer: This fixture should be moved to test_atomic.py, too. Accessing atom_data without calling atom_data.prepare_atom_data is dangerous.

tardis/tests/test_atomic.py
```diff
@@ -38,22 +38,40 @@ def test_atomic_symbol():
...
-    atom_data.prepare_atom_data([14])
-    assert len(atom_data.lines) > 0
-    # Fix for new behavior of prepare_atom_data
-    # Consider doing only one prepare_atom_data and check
-    # len(atom_data.lines) == N where N is known
-    atom_data = atomic.AtomData.from_hdf5(atom_data_filename)
-    atom_data.prepare_atom_data([20])
-    assert len(atom_data.lines) > 0
+
+param_atom_num = [[14], [20], [14, 20]]
```
yeganer: I wonder if this is needed. Something like
```
@pytest.mark.parametrize("selected_atomic_numbers", [[14], [20], [14, 20]])
```
should to the trick, right? (Haven't tested that but the goal should be something like that.)

yeganer:
@mishinma Thank you for the contribution. I really like your documentation of the code.

I've added some comments to the code. But it looks already really good.

tardis/tests/test_atomic.py
mishinma: Sure, but I wanted to use ids to make test names more informative.  
In the latest version of pytest ids can be either a list of string names or a function that returns names, but the version used by Travis seems only to support list of strings. So without using a global variable it could be like this: 
```
@pytest.mark.parametrize("selected_atomic_numbers", [[14], [20], [14, 20]],
  pytest.config.getvalue('atomic-dataset')))
                       ids = ["atom_num: {}".format(_) for _ in [[14], [20], [14, 20]]])
```
Without ids at all pytest doesn't show atomic numbers, only the name of the list with index:
`tardis/tests/test_atomic.py::test_prepare_atom_data_set_lines[selected_atomic_numbers0] PASSED`\
Well, that was my reasoning, but I can remove ids altogether.

mishinma:
@yeganer I moved both fixtures to test_atomic.py. And
- used yield fixture mechanism to handle hdf5 in the lines_dataset fixture
- changed the name of the second one to atom_data_f_dataset_not_prepared

mishinma:
After giving it more thought I removed the `atom_data_f_dataset_not_prepared` fixture completely - it only adds ambiguity.

yeganer:
That looks really good.

@wkerzendorf @unoebauer Can you please have a look and merge if you have no further comments?

unoebauer:
Looks very good! I am however confused about our decrease in coverage, given that we have actually added and improved tests with this PR.

In any case, having looked over the source code changes, I'm fine with merging this one. @wkerzendorf, what do you think?

yeganer:
The decrease in coverage comes from parameterized tests not being
recognized.

Ulrich Noebauer <notifications@github.com> schrieb am Di., 9. Feb. 2016
10:33:

wkerzendorf:
@embray - did you setup coveralls for astropy? Does astropy also have a problem with parameterized tests not appearing in the coverage?

unoebauer:
@wkerzendorf - what to do with this one (extends to #473)?

unoebauer:
Closing this one - @wkerzendorf reopen if you think this is still relevant.

Issue #478, opened 2016-02-07
test_simulation.py fails with the latest version of numpy


mishinma:
@yeganer @unoebauer @wkerzendorf 
In PR #471 the following (a few of them actually) asset statement was added to test_simulation.py:
```
npt.assert_allclose(
            t_rad, simulation_compare_data['test1/t_rad'], atol=0.0)
```
It is calculated like this: absolute(`a` - `b`) <= (`atol` + `rtol` * absolute(`b`))
The problem here is that t_rad is a Quantity while simulation_compare_data['test1/t_rad'] is just a numpy array. And if somehow works with numpy v 1.9.0, with v 1.10 it fails. An error is raised in the `__array_prepare__` method that is overridden in Astropy:

```

try:
                if _can_have_arbitrary_unit(maybe_arbitrary_arg):
                    converters = [None, None]
                else:
                    raise UnitsError("Can only apply '{0}' function to "
                                     "dimensionless quantities when other "
                                     "argument is not a quantity (unless the "
                                     "latter is all zero/infinity/nan)"
>                                    .format(function.__name__))
E                                    UnitsError: Can only apply 'subtract' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)
```

mishinma:
Actually, it would be nice to compare Quantities with a special util function that would validate the units of its arguments. One version of such a function is implemented in Astropy's [module tests/helpers.py]( https://github.com/astropy/astropy/blob/master/astropy/tests/helper.py ): assert_quantity_allclose

yeganer:
I think the right approach should be to use astropy's assert_quantity_allclose. One would have to test if that solution is compatible with different versions of numpy and astropy.

mishinma:
But units still should be added to the tests. `assert_quantity_allclose still` raises an error in this case as well, although with more clear message:

```
(Pdb) assert_quantity_allclose(t_rad, simulation_compare_data['test1/t_rad'], atol=0.0)
*** UnitsError: Units for 'desired' () and 'actual' (K) are not convertible
(Pdb) npt.assert_allclose(t_rad, simulation_compare_data['test1/t_rad'], atol=0.0)
*** UnitsError: Can only apply 'subtract' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)
```

yeganer:
So we have two options. Strip the result of its unit or add the unit to the
compare data. In that case we should do the second option.

[...]

PR #479, opened 2016-02-07
added units to test_simulation.py

mishinma:
I just added some units to tests in test_simulation.py. This seems to resolve the issue #478

### DIFF
--- a/tardis/simulation/tests/test_simulation.py
+++ b/tardis/simulation/tests/test_simulation.py
@@ -5,7 +5,7 @@ import pytest
 from tardis.io import config_reader
 from tardis.model import Radial1DModel
 from tardis.simulation import Simulation
-
+from astropy import units as u
 
 @pytest.fixture
 def tardis_config(kurucz_atomic_data, tardis_config_verysimple):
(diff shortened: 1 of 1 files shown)


yeganer:
I'm fine with that change and i think we should use `assert_quantity_allclose` whenever possible.

@wkerzendorf @unoebauer Any comments? If you are fine i think we should merge.

mishinma:
Do we need to pass `atol=0.0` in every assert statement? Its default value is 0 already. 
And when providing atol value to `assert_quantity_allclose` one must specify units because of the implementation in Astropy:
```
if atol is None:
        # by default, we assume an absolute tolerance of 0
        atol = u.Quantity(0)
    else:
        atol = u.Quantity(atol, subok=True, copy=False)
        try:
            atol = atol.to(actual.unit)
        except u.UnitsError:
            raise u.UnitsError("Units for 'atol' ({0}) and 'actual' ({1}) "
                               "are not convertible"
                               .format(atol.unit, actual.unit))

```

unoebauer:
Looks very good - thanks for the effort @mishinma.

I am wondering, though, if (maybe for future reference) it would make more sense to generate a proper reference data set, containing quantities instead of bare numpy arrays. Thoughts on this, @wkerzendorf, @yeganer?

wkerzendorf:
@unoebauer - yes it would make sense to do this, however it's not clear how to store those. But for now we are merging this.

PR #482, opened 2016-02-10
added control skipping of tests with command line option

mishinma:
@wkerzendorf 
I've read this year description of the testing project and made this PR for the first [objective.](http://opensupernova.org/gsoc2016/doku.php?id=ideas_page#testing) 
There is an example of how to skip tests with command line option in the [py.test documentaion](http://pytest.org/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option).
In this PR I just followed the example since it is surely the right way to implement this. The execution of slow tests can be enabled with `--runslow` command line option.
I also considered creating a custom marker, like `pytest.mark.slow`, but it will require implementing two more pytest hooks in `conftest.py` (at least the way I see it). This only makes sense to me if there will be a lot of slow test

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -152,9 +152,6 @@ import yaml
 from tardis.atomic import AtomData
 from tardis.io.config_reader import Configuration
 
-def slow():
-    return pytest.mark.skipif(not pytest.config.getoption("--runslow"),
-            reason="need --runslow option to run")
 
 @pytest.fixture
 def atomic_data_fname():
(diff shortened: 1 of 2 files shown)


unoebauer:
@mishinma: Very nice - just tested your PR and it works perfectly.

Concerning the number of slow tests: the original idea was (@wkerzendorf - correct me if I'm wrong) to have actually quite a few of those, testing the various parameter combinations (in particular different combinations of the physics modules).

wkerzendorf:
@mishinma looks good.

PR #484, opened 2016-02-11
Minor improvement of vpacket logging scheme

unoebauer:
This PR implements minor changes into the vpacket logging scheme. Currently, the virtual_packet property arrays are set to None if the vpacket logging is inactive. This has caused some confusion in DALEK (multiplication of None with units, etc.). To this end, this behaviour is changed by this PR. Now the virtual packet property arrays are not None but empty numpy arrays in case the vpacket logging is off

- [x] change type of vpacket_properties from None to np.zeros(0)
- [x] update tests
- [x] check with tardis example


unoebauer:
@orbitfold, @wkerzendorf 
Thoughts on this?

yeganer:
The implementation looks fine and it seems to be reasonable change.

PR #485, opened 2016-02-12
Montecarlo progress

yeganer:
This is a quick implementation that prints and updates a status line
containing the how many packets are already finished, how many total
packets are there and the percentage of finished packets.

The format of the statusline can be changed by the gobal macro
STATUS_FORMAT.

The implmentation is exact when running without OMP and an approximation
when activating OMP. This is not garanteed to work bugfree in case the
sheduler for the for-loop changes.
The alternative would be to have a shared variable but that could lead
to a minor performance reduction.

### DIFF
--- a/tardis/montecarlo/src/cmontecarlo.c
+++ b/tardis/montecarlo/src/cmontecarlo.c
@@ -5,6 +5,8 @@
 #include "abbrev.h"
 #include "cmontecarlo.h"
 
+#define STATUS_FORMAT "\r\033[2K\t[%" PRId64 "%%] Packets(finished/total): %" PRId64 "/%" PRId64
+
 /** Look for a place to insert a value in an inversely sorted float array.
  *
  * @param x an inversely (largest to lowest) sorted float array
@@ -816,6 +818,7 @@ montecarlo_one_packet_loop (storage_model_t * storage, rpacket_t * packet,
 void
 montecarlo_main_loop(storage_model_t * storage, int64_t virtual_packet_flag, int nthreads, unsigned long seed)
 {
+  int64_t finished_packets = 0;
   storage->virt_packet_count = 0;
 #ifdef WITH_VPACKET_LOGGING
   storage->virt_packet_nus = (double *)safe_malloc(sizeof(double) * storage->no_of_packets);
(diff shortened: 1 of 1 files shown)


unoebauer:
I'm not sure if we want to have the indentation fix clog up the diff... Also, do we actually have clear C-style conventions, @wkerzendorf, @orbitfold?

wkerzendorf:
yes we do have conventions - @orbitfold ?

yeganer:
Sure, i can delete the indentation fix and make a new PR for it. There were mixed tabs and spaces so i just did `:retab and gg=G` in vim. I'll look into the conventions and adjust the indentation and remove it from this PR.

orbitfold:
We do. We have agreed to follow gnu conventions for C code.

orbitfold:
Also yes. For formatting/conventions/bookkeeping purposes please do a
separate OR.

unoebauer:
Particularly in light of GSoC 2016, but also in general, we should prominently mention and link these conventions.

yeganer:
@orbitfold @unoebauer @wkerzendorf I'm finished with the implementation. If you have anything to add to the feature, i'd be happy to implement it otherwise please have a quick look and then merge.

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -829,41 +832,56 @@ montecarlo_main_loop(storage_model_t * storage, int64_t virtual_packet_flag, int
 #ifdef WITHOPENMP
   omp_set_dynamic(0);
   omp_set_num_threads(nthreads);
-#pragma omp parallel
+
+#pragma omp parallel firstprivate(finished_packets)
```
wkerzendorf: why is this change in there? I don't think that has anything to do with the logging, right?

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -829,41 +832,56 @@ montecarlo_main_loop(storage_model_t * storage, int64_t virtual_packet_flag, int
...
+#ifdef WITHOPENMP
+        // WARNING: This only works with a static sheduler and gives an approximation of progress.
+        // The alternative would be to have a shared variable but that could potentially decrease performance when using many threads.
+        if (omp_get_thread_num() == 0 )
+          fprintf(stderr, STATUS_FORMAT,
+              finished_packets * omp_get_num_threads() * 100 / storage->no_of_packets,
+              finished_packets * omp_get_num_threads(),
+              storage->no_of_packets);
+#else
+        fprintf(stderr, STATUS_FORMAT, finished_packets*100/storage->no_of_packets, finished_packets, storage->no_of_packets);
```
wkerzendorf: that is not PEP8 compatible (which applies to python, I know). Does GNU C suggest not having spaces between `*`?

tardis/montecarlo/src/cmontecarlo.c
yeganer: i made finished_packets ( my variable to count ) firstprivate so it won't be uninitialized in the threads. I made it private so it won't impact performance.

tardis/montecarlo/src/cmontecarlo.c
yeganer: Oh, i overlooked that part. Maybe i should copy the thing from above. That would look cleaner, too.

unoebauer:
Tested it with tardis_example and it works both with the OMP on and off.

tardis/montecarlo/src/cmontecarlo.c
orbitfold: Well pep8 actually allows to omit spaces before and after operators it
looks better. But yes you should surround * with spaces unless it is a
pointer.

unoebauer:
@wkerzendorf, @orbitfold - happy with this? I'd merge

PR #486, opened 2016-02-12
Bugfix radiative rates type detailed

chvogl:
In the current version of TARDIS the detailed mode for the estimation of radiative rates does not work (see #487). This PR fixes the problem.

### DIFF
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -207,7 +207,7 @@ class Radial1DModel(object):
             logger.info('Calculating J_blues for radiate_rates_type=detailed')
 
             self.j_blues = pd.DataFrame(
-                self.j_blue_estimators.transpose() *
+                self.j_blue_estimators *
                 self.j_blues_norm_factor.value,
                 index=self.atom_data.lines.index,
                 columns=np.arange(len(self.t_rads)))
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -216,7 +216,7 @@ class Radial1DModel(object):
                 zero_j_blues = self.j_blues[i] == 0.0
                 self.j_blues[i][zero_j_blues] = (
                     w_epsilon * intensity_black_body(
-                        self.atom_data.lines.nu.values[zero_j_blues],
+                        self.atom_data.lines.nu[zero_j_blues].values,
                         self.t_rads.value[i]))
 
         else:
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -208,9 +208,9 @@ class Radial1DModel(object):
 
             self.j_blues = pd.DataFrame(
                 self.j_blue_estimators *
-                self.j_blues_norm_factor.value,
-                index=self.atom_data.lines.index,
-                columns=np.arange(len(self.t_rads)))
+                    self.j_blues_norm_factor.value,
+                    index=self.atom_data.lines.index,
+                    columns=np.arange(len(self.t_rads)))
 
             for i in xrange(self.tardis_config.structure.no_of_shells):
                 zero_j_blues = self.j_blues[i] == 0.0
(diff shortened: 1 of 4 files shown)


wkerzendorf:
I guess these things are only slow-testable. So merge now and do a test later - what do people think @tardis-sn/tardis-core

unoebauer:
If we want to properly test it, we need a slow test. However, one could think about simple, quick test, which just runs Tardis with few packets in the detailed mode and checks that it doesn't break and that j_blues are returned.

@chvogl, @wkerzendorf - thoughts on this?

unoebauer:
@wkerzendorf - with the last commit, we have introduced a simple test to check the j_blue_estimators. Naturally, the slow test should probe the detailed mode thoroughly.

Still, a simple test is in place and I'd merge it - objections, @wkerzendorf, @orbitfold?

wkerzendorf:
@unoebauer - signing off on this!

tardis/model.py
```diff
@@ -207,7 +207,7 @@ def calculate_j_blues(self, init_detailed_j_blues=False):
             logger.info('Calculating J_blues for radiate_rates_type=detailed')
 
             self.j_blues = pd.DataFrame(
-                self.j_blue_estimators.transpose() *
+                self.j_blue_estimators *
                 self.j_blues_norm_factor.value,
```
orbitfold: I don't like the way this is indented, the bottom line in the multiplication should be indented one level to the right.

unoebauer:
Happy with it @chvogl?

tardis/model.py
wkerzendorf: Is this a specific PEP8 violation?

orbitfold: Don't know but it should be. Now it looks like it is another argument at first glance and not a part of an expression.

wkerzendorf: You are the style person - can you suggest how to make it PEP8 conform?

yeganer: There are people that fancy the operator on the next line for that reason.
(Which is against PEP)

The pythonic way might be to save the calculation in a temporary variable
and pass that as the argument.

Wolfgang Kerzendorf <notifications@github.com> schrieb am Di., 16. Feb.
2016 17:11:

wkerzendorf: My personal opinion - whatever pep8 says is what we do. But you guys need to make decision and enforce it on future happenings.

orbitfold: I think simply indenting the second line one level to the right is sufficient.

orbitfold: PEP8 does not address everything. And in a lot of cases it is not entirely clear what is meant. For example it says to use blank lines inside method bodies "sparingly, to indicate logical sections". What does that mean? I think 80% of TARDIS's Python code is in violation of that.

unoebauer: Ok - it seems as if this innocent little line of code is sparking a fundamental discussion of TARDIS coding style. I'd suggest we address this properly either in a TEP or on the tardis documentation (after first agreeing on a style, naturally) and just get on with the fix at hand.

Pragmatic solution - just leave it as it is until we have decided on conventions for cases which are not unambiguously covered by PEP8 and merge the PR

yeganer: After getting informed more about coding styles I have to admit indenting the next line would look better and improve readability.

We should definitely write some code guidelines for the documentation and every time we encounter something not handled yet add it.

wkerzendorf: can you do a PR on @chvogl's PR and implement that then.

unoebauer: Guys - let's not get sidetracked here - we want to fix the detailed mode! Let's merge and not get bugged down in style questions (which we also have not yet been formalised!)

wkerzendorf:
Okay @yeganer @orbitfold - we will merge this PR tomorrow at 2pm. If you want to change the current style, make a PR to this PR by then.

PR #491, opened 2016-02-13
Include IDE specific auto-generated files in .gitignore. Fixes #490

### LINKED ISSUE #490: IDE specific auto-generated files need to be in gitignore.
In the recent times many IDEs have become popular for Python development. Among them are Jetbrain Community's [Pycharm](https://www.jetbrains.com/pycharm/) and [PyDev](http://www.pydev.org/), an IDE extended from Eclipse for Python development.

Upon opening a project within, these IDEs generate certain files and directories which are specific to them, and are not project specific.

These files and directories need not be pushed on Github with the main source. So they must be included in _.gitignore_ file.

karandesai-96:
As described in #490, the necessary files and directories for the following two IDEs have been included in .gitignore :
- **Pycharm IDE** - Jetbrains.
  - `.idea/` directory in top level project directory.
- **PyDev IDE** - Eclipse.
  - mainly `*.project` and `*.pydevproject` files, along with some other common eclipse generated files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,6 @@ htmlcov
 
 ### PyDev IDE - Eclipse
 .metadata
-bin/
 tmp/
 *.tmp
 *.bak
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,22 @@ __pycache__
 htmlcov
 .coverage
 
+# IDE Specific files
+### Pycharm IDE - Jetbrains
+.idea/*
+
+### PyDev IDE - Eclipse
+.metadata
+bin/
+tmp/
+*.tmp
+*.bak
+local.properties
+.settings/
+.loadpath
+*.project
+*.pydevproject
+
 # Sphinx
 _build
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,16 +5,19 @@
 *.so
 __pycache__
 
+
 # Ignore .c files by default to avoid including generated code. If you want to
 # add a non-generated .c extension, use `git add -f filename.c`.
 *.c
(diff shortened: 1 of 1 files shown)


karandesai-96:
The coverage check fails wit the message - "_Coverage decreased (-0.0%) to 54.25%_".

I can't understand this, decreased by 0.0% ? But it hasn't decreased at all..

I haven't added any source specific file in .gitignore so the coverages wil not decrease by a non zero value.. :smiley:

orbitfold:
Well then the message is correct. Furthermore a decrease of -0% is actually
an increase of 0% so an improvement.

karandesai-96:
Ok, so I have both the IDEs and I opened the project in both of them. The IDEs generated certain files and they weren't show as **untracked** under `git status`.

Is this fit to be merged ?

yeganer:
This looks good. Can you please add `.ycm_extra_conf.py` just for completeness? This file is used by the vim plugin YouCompleteMe.

I think @wkerzendorf should have a quick look, since he is the one using PyCharm.

karandesai-96:
Pycharm has just ne directory :  `.idea` and everything inside it,, so I just had to write `.idea/*` for that :smile: It was painless !

Sure I will push a commit right in a moment.

wkerzendorf:
This looks good - if there are no other additions: I will merge.

karandesai-96:
Nothing I can find to add now, these two IDEs are popular for Python, and I covered one for Vim, I didn't have idea about, thanks to @yeganer .. This can be merged :smiley:

PR #494, opened 2016-02-15
Fix #493

### LINKED ISSUE #493: Progress updates have undesired effect when redirecting to file
When running `tardis tardis_example.yml out 1>stdout.log 2>stderr.log` the progress bar creates this output:
```

[2K	[1%] Packets(finished/total): 400/40000
[2K	[2%] Packets(finished/total): 800/40000
[2K	[3%] Packets(finished/total): 1200/40000
```

This can be solved by introducing a switch to disable that functionality or by writing to a seperate tty device.

yeganer:
When redirecting the output to a file the escape sequences which are
used to produce the progress updates don't work. This resulted in long
lines in the logile. To avoid this, one can check if the outputstream is
a terminal with isatty().
This functionality was implemented and wrapped in a function defined in
the new header file "io.h". This header will be used in the future to
implement sending logging messages to the python logging framework.

Resolves: #493

### DIFF
--- a/tardis/montecarlo/src/cmontecarlo.c
+++ b/tardis/montecarlo/src/cmontecarlo.c
@@ -1,11 +1,13 @@
+#define _POSIX_C_SOURCE 1
+
 #include <inttypes.h>
 #ifdef WITHOPENMP
 #include <omp.h>
 #endif
 #include "abbrev.h"
 #include "cmontecarlo.h"
+#include "io.h"
 
-#define STATUS_FORMAT "\r\033[2K\t[%" PRId64 "%%] Packets(finished/total): %" PRId64 "/%" PRId64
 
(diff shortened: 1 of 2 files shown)


unoebauer:
@yeganer - very nice. I just checked it and both modes of operations, i.e. redirecting stderr to a log file and printing to the terminal work as expected. In the former, the status bar is suppressed, in the latter it is displayed.

I would merge this - any objections, @wkerzendorf, @orbitfold?

orbitfold:
I don't really mind, but I don't think we should add features unless they are absolutely necessary for work. Every new line of code is a liability not an asset.

wkerzendorf:
I think this seems to make it easy to work with the new progress bar. We should still look into the logging feature - that would allow us to log to a file and ditch any progress bars. -- merging

Issue #448, opened 2015-12-01
Missing relativistic CMF-LF frame transformation effects - Overview


unoebauer:
This issue is an meant to summarise issues #335, #57, #56 and #55 and to collect any other issues concerning omitted relativistic frame transformation effects.

unoebauer:
## Current status:

When transforming between the local co-moving frame (CMF) and the lab frame (LF) only the following effects are taken into account:

- frequency changes due to the Doppler effect 

## Omitted/Neglected effects:

- changes in the propagation direction due to angle aberration (c.f. #57)
- correct transformation of opacity (c.f. #55)
- correct transformation of photon path segments (**needs confirmation**)
- related to everything above: correct frame treatment in Monte Carlo estimators (c.f. #56)

## Tasks:

- decide which of these effects should be implemented
- the documentation should have some notes about which effects are incorporated and which are neglected

chvogl:
@unoebauer 
## Further problems
### Energy conservation in the comoving frame
For each interaction the energy of the packet should be conserved in the comoving frame.
However, at the moment energy conservation is enforced between the comoving energy at the position of the last interaction and the comoving energy at the interaction point.
#### Current approach for energy conservation
#### Step 1
First the comoving energy of the packet is calculated:

    double doppler_factor = move_packet (packet, storage, distance);
    double comov_energy = rpacket_get_energy (packet) * doppler_factor;
Since the return value of move_packet is the doppler factor at the position of the last interaction (or movement across a shell boundary), the value for the comoving energy is also for this point.
#### Step 2
In the next step this comoving energy is transformed back into the lab frame using the inverse doppler factor at the position of the interaction.

    rpacket_set_mu (packet, 2.0 * rk_double (mt_state) - 1.0);
    double inverse_doppler_factor = 1.0 / rpacket_doppler_factor (packet, storage);
    rpacket_set_energy (packet, comov_energy * inverse_doppler_factor);

Therefore, energy conservation is enforced between the comoving energy at the position of the last interaction and the comoving energy at the interaction point.
As a result the current approach retrieves the correct values only asymptotically for an increasing number of shells and/or high optical depths.

unoebauer:
@chvogl, @ssim 

After doing the math, I am fairly convinced that the current interaction scheme is not owed to consistency with the calculation of the Sobolev point:

[...]

PR #495, opened 2016-02-24
Fix energy conservation

unoebauer:
As found by @chvogl and reported in issue #448, an inconsistency in the energy conservation calculation exists. Instead of the Doppler factor at the interaction location, the initial Doppler factor, determined at the beginning of the trajectory segment is used. For typical applications, the consequences of this inconsistency should be small (as discussed in #448). Nevertheless, this PR aims at fixing this inconsistency.

- [x] fix  energy conservation, i.e. use Doppler factors at the interaction location
- [x] investigate consequences (check tardis_example)
- [x] decide whether move_packet should actually return the doppler factor
- [x] adjust and update tests


yeganer:
@unoebauer When I first looked at the C code I was wondering why move_packed returned the doppler_factor.

My suggestion: add `rpacket_get_doppler_factor(*rpacket) `returning the dopplerfactor of the current packet ( current r, mu). I don't think we need that factor saved to rpacket. then any time we need the doppler factor we can just call that function and save it in a local variable.
If there are too many calls to `rpacket_get_doppler_factor`, maybe we should rename it to `update_doppler_factor` and then save the doppler_factor as a property of rpacket

unoebauer:
Fix does not seem to induce significant changes for standard setups:

spec_cmp

unoebauer:
@yeganer, @ssim, @wkerzendorf, @chvogl 

Many tests had to be updated - for future references: we should write a script which recalculates all the data files, once critical changes have been made within the Monte Carlo routine. Of course, before using these files, they should be carefully compared with the current ones. I have started a very very crude version of such a script.

yeganer:
@unoebauer Last time we did this I advised against creating such a script because it could be used carelessly. But I guess with future slow tests coming we might need that.

unoebauer:
Well - true, there is a temptation to use it carelessly. But manually building all the data files is really painfull!

unoebauer:
Ok - Travis is finally happy. Should this be merged @wkerzendorf, @yeganer?

It would be good if you, @ssim and @chvogl, could also have a careful look at it.

tardis/montecarlo/src/test_cmontecarlo.c
```diff
@@ -261,17 +261,17 @@ test_rpacket_doppler_factor(void){
...
-	double DISTANCE = 1e13;
-        rpacket_t rp;
-        storage_model_t sm;
-        init_rpacket(&rp);
-        init_storage_model(&sm);
-	double res = move_packet(&rp, &sm, DISTANCE);
-        dealloc_storage_model(&sm);
-        return res;
-}
+//double
```
wkerzendorf: @unoebauer should we just delete these lines?

unoebauer: Well - I wouldn't for now. Eventually, we want to test move_packet, just in a different way. I'd keep these lines as a reminder.

yeganer:
I think this is ready to get merged.

Regarding `test_move_packet`:
I think it's fine to have it in the c code. We could even uncomment it because the test is skipped anyway. Deleting the lines is dangerous because when the tests get an overhaul one could easily forget implementing this one.

ssim:
I think that yes, I agree with this fix. I wonder how it came about that there was an error here but I think that yes, it is was wrong. It would have been correct if the Dopper factor were re-calculated at the end of "move packet" before being returned, but that was not being done. This seems like the right fix to me.

chvogl:
I agree that this can and should be merged.

PR #496, opened 2016-02-24
Added counting estimator of packet resonance per line and global coutner

vaditya04:
@unoebauer - Please have a look. I'm getting 1e7 resonances on the global counter. 

Critical changes in storage.h, cmontecarlo.c, and some initialization parameters in a few other files.

I've coded by the assumption that we want no. of resonances 'per line' rather than just the total. So there's an array in storage.h which stores these things, indexed by line_id.


unoebauer:
@vaditya04
First of all, thanks for the PR. I have a couple of remarks to it (all of which have also been posted on our Gitter channel):

- for future references, please don't include file permission changes in your PRs - they bloat the PR, making it hard to focus on the actual changes - @yeganer posted instructions on how to avoid this on Gitter
- your counter does not really count the number of packets which come into resonance with each line
- instead they count how many close lines (i.e. lines whose frequencies are very close together) are encountered by packets
- try to look at and understand the j_blue_estimator and its increments - a lot of the infrastructure necessary for your counters is already there


There are couple of this

Issue #438, opened 2015-11-23
Bug in Tardis atomic / standard data set?


ssim:
It looks to me like the set of atomic data we currently are providing has missing basic ionizating data for Z=4 (Be). This data is not actually important for current applications (abundances are low); but if running with full ARTIS model this leads to a crash since the (new?) plasma insists that the full data is available (and although negligibly small, the Be abundance is not = 0 in the models).

Data appears to be missing the top ion stage (expected).

@wkerzendorf can you easily regenerate our standard atomic dataset will all ions for Be and replace the existing one?

aoifeboyle:
Hi @ssim. Yeah, the missing Be data has been a problem for ages now. I think we agreed that rebuilding the atomic dataset was complicated. 

In my own version of Tardis I have a temporary solution that manually adds the Be IV ionisation data. It's not a long-term solution but it does work. I'll make a PR with it (not to be merged ever, just as a temporary solution).

unoebauer:
Just realised that this issue is quite severe because it breaks a key feature of Tardis, namely setting up a stratified abundance structure (using the abundances config parameters type: file, filetype: simple_ascii). Pushing up the priority.

unoebauer:
@ssim, @aoifeboyle, @wkerzendorf - maybe we should, for the time being, merge the temporary Be fix (PR #439) to restore the possibility of calculating stratified models. If desired, we could add a warning message, notifying the user that the Be treatment is incomplete in the current version once Be is present in his models and that this issue will be resolved in future releases. 

The other possibility is of course, as @ssim suggested, to re-generate the atomic data file now containing the missing data...(I have the feeling that this may turn out to be a bit tricky given the obscurity of the involved scripts).

unoebauer:
@ssim, @aoifeboyle, @wkerzendorf: I have to go back on my statement from before: the Be issue does not break the generic Tardis capability of handling stratified abundances, as long as all Be abundances are set to zero.

chvogl:
@unoebauer: As a temporary solution it is possible to manually add the missing Be data to the existing atomic data file:


    with h5py.File('kurucz_cd23_chianti_H_He.h5') as atom_data:
        new_ion_data = np.insert(atom_data['ionization_data'][:],9,(4,4,217.71865))
        del atom_data['ionization_data’]
        atom_data.create_dataset(u'ionization_data', data=new_ion_data)

unoebauer:
[...]

PR #502, opened 2016-03-03
Check and zero Be abundances - circumvents #438

unoebauer:
This PR provides a quickfix for #438. It does not solve the underlying problem but ensures that the Tardis calculation proceeds even if Be is present in the model.

If it is, its abundance is set to zero everywhere. This is done before the normalization step. The rationale behind this is that in any reasonable model, Be should only of minor importance. This quickfix should be adopted and kept until all the necessary atomic data has been added to the standard atomic data set (or if the capability is implemented in Tardis to not require always all ionisation stages of a given element).

- [x] add checker for Be abundance
- [x] set Be abundance to zero and log warning
- [ ] log warning if specified Be abundance was significant
- [x] test with tardis_example
- [ ] test with uniform and stratified models containing Be

### DIFF
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -628,7 +628,7 @@ class ConfigurationNameSpace(dict):
 
         return cls(ConfigurationValidator(config_definition,
                                        config_dict).get_config())
-    
+
     marker = object()
     def __init__(self, value=None):
         if value is None:
@@ -773,7 +773,7 @@ class Configuration(ConfigurationNameSpace):
             in the `data` directory that ships with TARDIS
 
         validate: ~bool
-            Turn validation on or off. 
+            Turn validation on or off.
 
 
         Returns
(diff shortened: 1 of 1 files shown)


wkerzendorf:
@unoebauer I think we should sanitize our inputs beforehand and not put a special treatment for `Be` into TARDIS.

unoebauer:
@wkerzendorf - look at the discussion to issue #438. This is a quickfix and it should only be there until we have a proper solution.

And I think until the standard atomic data set is incomplete but Tardis and the same time requires this data, we have to have a checker.

wkerzendorf:
I realize the discussion on #438 and I agree that  setting the abundances to 0 is a good quick fix. I understood it as changing the abundances file so it doesn't show Be. But I advise against putting this into the actual master code for TARDIS as specialized treatments can lead to unexpected results down the line. I think a quickfix that can be a sensible merge is exiting with an error to say that Be abundances are currently not supported.

unoebauer:
@wkerzendorf - I see your point. Opinions on that, @ssim, @chvogl, @aoifeboyle? I can adapt the PR...

yeganer:
@wkerzendorf @unoebauer If you want to exit with a warning when there is Be present, then I think we should have an option, to override this behaviour and set the abundances to 0.

unoebauer:
@wkerzendorf - I've updated the quickfix to a simple check and abort.

unoebauer:
hmm - Travis is unhappy because one of our config test files (artis_abundances) has minute amounts of Be in it. It could be solved by modifying the data file, i.e. setting Be to zero in artis_abundances.dat (ugly hack again...)

Any Suggestions, @wkerzendorf?

wkerzendorf:
@unoebauer I'll have a look.

unoebauer:
Closing this PR - see comments to #438.

PR #503, opened 2016-03-03
NF - running full test from comand line

Minhmo:
Added a pytest.mark.slow to run test from command line.

### DIFF
--- a/tardis/tests/test_tardis_full.py
+++ b/tardis/tests/test_tardis_full.py
@@ -17,14 +17,13 @@ from tardis.plasma.standard_plasmas import LegacyPlasmaArray
 
 def data_path(fname):
     return os.path.join(tardis.__path__[0], 'tests', 'data', fname)
-
+@pytest.mark.slow
 @pytest.mark.skipif(not pytest.config.getvalue("atomic-dataset"),
                     reason='--atomic_database was not specified')
 class TestSimpleRun():
     """
     Very simple run
     """
-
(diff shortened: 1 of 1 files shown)


wkerzendorf:
@Minhmo how will this be able to be switched on from the command line

Minhmo:
@wkerzendorf adding "-m slow" to the py.test command. The whole command to run the full (slow) test would look like "py.test -m slow -q --atomic-dataset=<path to dataset>".

wkerzendorf:
@mishinma sorry I should have clarified, it should normally not run these tests only with the commandline option. As far as I can tell - your code will always also run the slow tests. with the `-m slow` option only this test will be run. Have you tested it?

mishinma:
@wkerzendorf you probably meant @Minhmo

Minhmo:
@wkerzendorf full test can be excluded by adding -m "not slow". Yeah, I have just checked -m "not slow" and it automatically deselects full test.

wkerzendorf:
@mishinma sorry wrong name. @Minhmo I want slow tests only if we specifically select them.

orbitfold:
@Minhmo why did you close this? You don't have to close pull requests to respond to comments. If you want to make changes then pushing to your fork will update this pull request.

Minhmo:
@orbitfold I just lost control, my bad. However, I opened a new pull request. Won't do this in future.

Issue #441, opened 2015-11-27
Gui does not work if LaTeX is activated


unoebauer:
This is just a minor issue, but it seems as if the gui does not run properly if the LaTeX is used to render math fonts. This is potentially an issue when the user has customised matplotlibrc files, specific mplstyle files or hast set LaTeX processing by hand. 

There are two possible solutions:
- change all strings in the gui to be compatible with LaTeX and normal processing (requires some work)
- force deactivate LaTex by calling `ptl.rcParams["text.usetex"] = False` right at the start (easy bugfix)

unoebauer:
Some more diagnostics on this issue: 

Starting the gui without any custom rcParams settings (no matplotlibrc files, no styles), the gui works fine (since LaTex is deactivated by default)

Activating LaTeX with ```plt.rcParams["text.usetex"] = True``` just before calling the gui results in the follwing LaTeX error:

```
 LaTeX was not able to process the following string:
'$\\\\times\\\\mathdefault{10^{39}}\\\\mathdefault{}$'
Here is the full report generated by LaTeX: 

This is pdfTeX, Version 3.14159265-2.6-1.40.15 (TeX Live 2015/dev/Debian) (preloaded format=latex)
 restricted \write18 enabled.
entering extended mode
(./4f01cccb7421a182ba64f251faccce1f.tex
LaTeX2e <2014/05/01>
Babel <3.9l> and hyphenation patterns for 7 languages loaded.
(/usr/share/texlive/texmf-dist/tex/latex/base/article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(/usr/share/texlive/texmf-dist/tex/latex/base/size10.clo))
(/usr/share/texlive/texmf-dist/tex/latex/type1cm/type1cm.sty)
(/usr/share/texlive/texmf-dist/tex/latex/psnfss/helvet.sty
(/usr/share/texlive/texmf-dist/tex/latex/graphics/keyval.sty))
(/usr/share/texlive/texmf-dist/tex/latex/psnfss/courier.sty)
(/usr/share/texlive/texmf-dist/tex/latex/base/textcomp.sty
(/usr/share/texlive/texmf-dist/tex/latex/base/ts1enc.def))
(/usr/share/texlive/texmf-dist/tex/latex/geometry/geometry.sty
(/usr/share/texlive/texmf-dist/tex/generic/oberdiek/ifpdf.sty)
[...]

PR #511, opened 2016-03-07
Gui does not work if LaTeX

theaverageguy:
Fixes https://github.com/tardis-sn/tardis/issues/441


yeganer:
I'm not familiar with that part of the codebase but it is possible to switch LaTeX rendering on and of, right? In that case I think your fix should be wrapped so that it's only executed if LaTeX rendering is activated.

Issue #489, opened 2016-02-12
Misleading header file in C extension (cmontecarlo1.h)


yeganer:
For some reason there is a header file called `cmontecarlo1.h` declaring a single function ( which would produce compiler errors if the order of import in `cmontecarlo.c` was changed).

I think this function declaration should be moved to `cmontecarlo.h` and in that process I suggest to check if there are other potential errors.

I also suggest adding missing declarations from functions in `cmontecarlo.c` to it's header.

wkerzendorf:
sure - I would also like to know how to convert an issue to a PR. I think this issue is a good candidate.

yeganer:
I haven't found any new information but it seems this feature isn't available anymore.
See discussion here: https://github.com/github/hub/commit/4f70dd126f46dec14fc341c97c18efae417743c7

When i am done with some implementation i'll create a PR. That seems the best approach.

orbitfold:
I'd ask whoever added it why they did it. It's easy to assume people made a
mistake but sometimes there are reasons you may not be aware of

yeganer:
When i noticed it I tried to think of a reason and didn't find one. I moved the declaration and deleted `cmontecarlo1.h` and it compiled and ran without errors.

I'm happy to hear a reason and be proven wrong.

orbitfold:
You should see when it was added. I think it was added as part of an effort
to improve performance around this September. By someone from mpa

yeganer:
It was added by @mreinecke during the performance improvements.
But i don't see a reason for that function to have it's own header.
To me it looks like an artifact from some test or something like that.

wkerzendorf:
okay @mreinecke is there a reason for `cmontecarlo1.h`

orbitfold:
I'm curious too actually

mreineck:
The reason for this curious construct was (if I remember correctly) that the function line_search() is defined (not only declared) in cmontecarlo.h, but it does not have the "static" attribute. As a consequence, if cmontecarlo.h is included by two different C files, the function line_search() turns up in each of the .o files compiled from these, and the linker will most likely complain about a function being defined twice.
There are different ways to avoid this:
- declare line_search() to be "static"
- move the definition of line_search() to montecarlo.c, but leave the declaration in montecarlo.h
- play confusing games with the header files like I did.

I must admit that I don't remember why I went for this awkward solution. There probably was a reason, but maybe not a very good one.

orbitfold:
[...]

PR #512, opened 2016-03-07
Misleading header file

theaverageguy:
Fixes https://github.com/tardis-sn/tardis/issues/489


wkerzendorf:
@chvogl @yeganer @mreinecke can you guys look at this? I can't remember if that was a sensible fix.

yeganer:
@theaverageguy Thank you for the contribution.
@wkerzendorf The change is fine but there are two issues:

- The 1st commit won't compile since there will be two declarations of line_search
- After the 2nd commit there will be an import error because `cmontecarlo1.h` is deleted.

I suggest you remove the `include "cmontecarlo1.h"` from `cmontecarlo.h` and then combine all commits into one commit that covers the whole change with a litte explanation in the body of the commit message summarizing what and why it changed.

mreineck:
Also, you need to remove the
#include "cmontecarlo1.h"
from rpacket.h and add
#include "cmontecarlo1.h"
to rpacket.c.

mreineck:
Sorry, correction ....
of course you need to 
#include "cmontecarlo.h"
from rpacket.c.

yeganer:
@mreineck thanks, I totally forgot that rpacket.c is missing this `#include`.

In general there needs to be some cleanup of header files. Headers should only `#include ...` stuff the header needs. Everything else goes into the corresponding source file.

@wkerzendorf I'm not sure if it is a good idea to include these changes within this PR or if we want a new PR for header cleanup.

yeganer:
#531 Fixed the corresponding issue. Closing this

Issue #497, opened 2016-02-25
Optimization: cmontecarlo.c:compute_distance2boundary restructure logic


yeganer:
In cmontecarlo.c:compute_distance2boundary the 1st thing that is calculated is d_outer in [https://github.com/tardis-sn/tardis/blob/master/tardis/montecarlo/src/cmontecarlo.c#L147](L147). Since we are not always using this value, we should only calculate it when we need it.
Thus I propose the following logic:

```pseudocode
if mu>0 // travelling outward
   return d_outer
else //travelling inward
  if (cross inner boundary) // inner boundary is always shorter than outer
    return d_inner
  else // didn't hit inner boundary? calculate d_outer
    return d_outer
```

A profiled run we did some time ago showed 25% execution time for compute_distance2boundary. Reducing the overhead is a simple task and should improve performance a bit.

unoebauer:
I agree that this routine could be improved. The following is taken from my thesis:

This illustrates the propagation problem in spherical symmetry

![spherical](https://cloud.githubusercontent.com/assets/6351165/13320074/7efef5e4-dbc6-11e5-82e9-02da9ba2d8f5.png)

The distance to the next boundary, can then be calculated according to:

![distance](https://cloud.githubusercontent.com/assets/6351165/13320124/cabb433e-dbc6-11e5-81d4-c6e5420dd862.png)

PR #514, opened 2016-03-08
Performance

yeganer:
This is a small collection of performance boosts and improvement of code readability.

Included changes:

- remove confusion from line_scatter by setting the next line only when it is known
- change memory layout of line_list_tau_sobolevs and transition_probabilities so the montecarlo routine can interprete it as no_of_shells x no_of_lines
- Fix for #497: distance2boundary is more structured now and has no overhead anymore

- [x] `tardis_example.yml`
- [x] `tardis_w7.yml` ~20-25%
- [x] `abn_tom_test.yml` ~31%

### DIFF
--- a/scripts/tardis
+++ b/scripts/tardis
@@ -36,6 +36,8 @@ parser.add_argument('--profile', action='store_true', help=
 parser.add_argument('--profiler_log_file', default='profiler.log', help=
 'name of the profiler output file')
 
+parser.add_argument('--gdb', action='store_true', help='print pid and pause')
+
 args = parser.parse_args()
 
 packet_logging_fname = 'tardis_packets.log'
@@ -63,9 +65,11 @@ if args.packet_log_file:
 tardis_config = config_reader.Configuration.from_yaml(args.config_fname)
 radial1d_mdl = model.Radial1DModel(tardis_config)
 
+if args.gdb:
+    import os; print(os.getpid()); raw_input() # Workaround to attach gdb
 if args.profile:
     import cProfile
-    cProfile.runctx('simulation.run_radial1d(radial1d_mdl)', locals(),
+    cProfile.runctx('run_radial1d(radial1d_mdl)', locals(),
                     globals(), filename=args.profiler_log_file)
 else:
     run_radial1d(radial1d_mdl)
--- a/tardis/montecarlo/src/cmontecarlo.c
+++ b/tardis/montecarlo/src/cmontecarlo.c
@@ -137,67 +137,62 @@ void calculate_chi_bf(rpacket_t * packet, storage_model_t * storage)
   rpacket_set_chi_boundfree(packet, bf_helper * doppler_factor);
 }
 
-double
+void
 compute_distance2boundary (rpacket_t * packet, const storage_model_t * storage)
 {
   double r = rpacket_get_r (packet);
   double mu = rpacket_get_mu (packet);
   double r_outer = storage->r_outer[rpacket_get_current_shell_id (packet)];
   double r_inner = storage->r_inner[rpacket_get_current_shell_id (packet)];
-  double d_outer =
-    sqrt (r_outer * r_outer + ((mu * mu - 1.0) * r * r)) - (r * mu);
-  double d_inner;
+  double check, distance;
   if (rpacket_get_recently_crossed_boundary (packet) == 1)
     {
       rpacket_set_next_shell_id (packet, 1);
-      return d_outer;
+      distance = sqrt (r_outer * r_outer + ((mu * mu - 1.0) * r * r)) - (r * mu);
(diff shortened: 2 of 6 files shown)


yeganer:
@wkerzendorf @unoebauer I tested this setup with `tardis_example.yml` and `tardis_w7.yml`. In with both setups I got identical results compared to the current master (`diff ref.txt performance.txt` reported nothing at least) and the tests ran fine.

The speedup I measured was ~20-25%.
I think restructuring the way the distances are calculated could yield another 25% or even more but that change would be less straightforward.

Can you please point me to other setups I can run to ensure I didn't implement any nasty bugs?

unoebauer:
Hi @yeganer - having a quick at the changes, it looks to me as if the sequence of RNG calls is not altered. To double check that the optimization tweaks do not interfere with the implemented physics, I'd suggest that you do a bit comparison of the spectra before and after the optimization changes for

* tardis_example
* tardis_w7
* abn_tom_test

wkerzendorf:
@unoebauer - are they in the TEP?

unoebauer:
@wkerzendorf yes (apart from tardis_example)

yeganer:
@unoebauer @wkerzendorf @orbitfold I updated this PR and I'm not seeing any difference between spectra. `np.all(spec_ref == spec_pr) : True`
Can you have a look please and merge once you are fine with the changes.

scripts/tardis
```diff
@@ -36,6 +36,8 @@ parser.add_argument('--profile', action='store_true', help=
 parser.add_argument('--profiler_log_file', default='profiler.log', help=
 'name of the profiler output file')
 
+parser.add_argument('--gdb', action='store_true', help='print pid and pause')
```
wkerzendorf: do we still need this - or is this a debug statement from you?

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -137,67 +137,62 @@ void calculate_chi_bf(rpacket_t * packet, storage_model_t * storage)
   rpacket_set_chi_boundfree(packet, bf_helper * doppler_factor);
 }
 
-double
```
wkerzendorf: why is that `void`

scripts/tardis
yeganer: This is a debug statement I added specifically to enable debugging. If we don't want that capanility I can remove that commit but I think it's a handy option.

I wasn't able to start tardis with `gdb` to debug the C code, so I added it.

wkerzendorf: that's fine - just wanted to know.

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -137,67 +137,62 @@ void calculate_chi_bf(rpacket_t * packet, storage_model_t * storage)
...
+void
```
wkerzendorf: the `void` we should talk about.

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -296,20 +291,20 @@ compute_distance2continuum(rpacket_t * packet, storage_model_t * storage)
 int64_t
 macro_atom (const rpacket_t * packet, const storage_model_t * storage, rk_state *mt_state)
 {
-  int emit = 0, i = 0, probability_idx = -1;
-  int activate_level =
-    storage->line2macro_level_upper[rpacket_get_next_line_id (packet) - 1];
+  int emit = 0, i = 0, offset = -1;
+  uint64_t activate_level =
```
wkerzendorf: why `uint64_t` and not `int`

yeganer:
@unoebauer @wkerzendorf @chvogl @orbitfold I rerun my comparison with this commit and the previous one(with recently_crossed_boundary missing) and both gave me spectra which were bit equal  for all three configurations.

With that Information I think it is save to assume that I didn't change any logic. As for the Issue with packets potentially having a discriminant < 0 which would result in the packet missing the outer boundary. A quick calculation showed that `mu < sqrt( 2 e_m + e_m**2)` where `e_m` is the machine precision. This results in a probability of roughly 1e-9. In addition to the exact angle the packet needs a position of `r=r_outer + e_m`. We can safely assume this because the maximum error done during move_packet is the machine precision.

wkerzendorf:
@tardis-sn/tardis-core are we happy to merge this?

yeganer:
@wkerzendorf I'm happy to merge. My local branch has `recently_crossed_boundary` removed and all tests are passing. Should I update the PR again?

wkerzendorf:
@yeganer I'm happy to merge this! I need one more confirmation from @tardis-sn/tardis-core

orbitfold:
I'm fine as well

Issue #57, opened 2013-08-27
angle aberation


wkerzendorf:

unoebauer:
There is an entire series of issues dealing with omitted relativistic effects: #335, #57, #56, #55. Maybe we should combine them in one issue/milestone, @ssim, @wkerzendorf?

wkerzendorf:
@unoebauer sounds like a good idea - issue or milestone; your choice.

unoebauer:
Closing this one in favour of #448

PR #515, opened 2016-03-09
[config-system] Add a custom YAML Loader to parse astropy quantities

ftsamis:
The `yaml.Loader` class supports adding constructors and implicit resolvers to parse a YAML scalar as a Python object.

In this PR the following changes are introduced:

- Subclass the default `yaml.Loader`, add a constructor and an implicit resolver for `quantity` properties to the subclass and allow the user to choose which loader they will use when parsing the YAML configuration.
- Monkeypatch pyyaml's [bug #57](https://bitbucket.org/xi/pyyaml/issues/57/add_implicit_resolver-on-a-subclass-may) which affected this implementation.
- Change the `tardis.io.config_validator` to accept already constructed `astropy.units.Quantity` objects as valid quantity properties.
- Write a `pytest` test which checks if `yaml.Loader` and the new `config_reader.YAMLLoader` both produce the same `Configuration` object for a given input.

### DIFF
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import re
 import pprint
 
 from astropy import constants, units as u
@@ -728,15 +729,86 @@ class ConfigurationNameSpace(dict):
         return ConfigurationNameSpace(copy.deepcopy(dict(self)))
 
 
+class YAMLLoader(yaml.Loader):
+    """
+    A custom YAML loader containing all the constructors required
+    to properly parse the tardis configuration.
+    """
+    @classmethod
+    def add_implicit_resolver(cls, tag, regexp, first=None):
+        """
+        Parameters
+        ----------
+
+        tag:
+            The YAML tag to implicitly apply to any YAML scalar that matches `regexp`
+
+        regexp:
+            The regular expression to match YAML scalars against `tag`
+
+
+        Notes
+        -----
+
+        This classmethod is a monkey-patch for a copy() related bug
+        in the original class method which affects this yaml.Loader subclass.
+
+        This class method is to be removed when this bug gets fixed upstream.
+
+        https://bitbucket.org/xi/pyyaml/issues/57/add_implicit_resolver-on-a-subclass-may
+        """
+        if 'yaml_implicit_resolvers' not in cls.__dict__:
+            yaml_implicit_resolvers = {}
+            for k, v in cls.yaml_implicit_resolvers.items():
+                yaml_implicit_resolvers[k] = copy.copy(v)
+            cls.yaml_implicit_resolvers = yaml_implicit_resolvers
+        if first is None:
+            first = [None]
+        for ch in first:
+            cls.yaml_implicit_resolvers.setdefault(ch, []).append((tag, regexp))
+
+    def construct_quantity(self, node):
+        """
+        A constructor for converting quantity-like YAML strings to
+        `astropy.units.Quantity` objects.
+
+        Parameters
+        ----------
+
+        node:
+            The YAML node to be constructed
+
+        Returns
+        -------
+
+        `astropy.units.Quantity`
+
+        """
+        data = self.construct_scalar(node)
+        value_str, unit_str = data.split(None, 1)
+        value = float(value_str)
+        if unit_str.strip() == 'log_lsun':
+            value = 10 ** (value + np.log10(constants.L_sun.cgs.value))
+            unit_str = 'erg/s'
+        return value * u.Unit(unit_str)
+
+
+YAMLLoader.add_constructor('!quantity', YAMLLoader.construct_quantity)
+# This regex matches anything that is a number (scientific notation supported) followed by whitespace
+# and any other characters (which should be a unit).
+pattern = re.compile(r'^-?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?\s+.+$')
+YAMLLoader.add_implicit_resolver('!quantity', pattern)
+
+
 class Configuration(ConfigurationNameSpace):
     """
     Tardis configuration class
     """
 
     @classmethod
-    def from_yaml(cls, fname, test_parser=False):
+    def from_yaml(cls, fname, test_parser=False, loader=YAMLLoader):
         try:
-            yaml_dict = yaml.load(open(fname))
+            yaml_dict = yaml.load(open(fname), Loader=loader)
         except IOError as e:
             logger.critical('No config file named: %s', fname)
             raise e
--- a/tardis/io/config_validator.py
+++ b/tardis/io/config_validator.py
@@ -331,6 +331,13 @@ class PropertyTypeFloat(PropertyTypeInt):
 
 class PropertyTypeQuantity(PropertyType):
     def check_type(self, value):
+        if hasattr(value, 'unit') and hasattr(value, 'value'):
+            if self._default is not None:
+                try:
+                    self._default.to(value.unit)
+                except ValueError:
+                    return False
+            return True
         try:
             quantity_split = value.strip().split()
             quantity_value = quantity_split[0]
@@ -353,15 +360,21 @@ class PropertyTypeQuantity(PropertyType):
             return False
 
     def to_type(self, value):
-        quantity_split = value.strip().split()
-        quantity_value = quantity_split[0]
-        quantity_unit = ' '.join(quantity_split[1:])
-        if quantity_unit.strip() == 'log_lsun':
-            quantity_value = 10 ** (float(quantity_value) +
-                                    np.log10(constants.L_sun.cgs.value))
-            quantity_unit = 'erg/s'
+        if hasattr(value, 'unit') and hasattr(value, 'value'):
+            return value
+        elif hasattr(value, 'split'):
+            quantity_split = value.strip().split()
+            quantity_value = quantity_split[0]
+            quantity_unit = ' '.join(quantity_split[1:])
+            if quantity_unit.strip() == 'log_lsun':
+                quantity_value = 10 ** (float(quantity_value) +
+                                        np.log10(constants.L_sun.cgs.value))
+                quantity_unit = 'erg/s'
 
-        return float(quantity_value) * units.Unit(quantity_unit)
+            return float(quantity_value) * units.Unit(quantity_unit)
+        else:
+            raise ConfigError
+
 
 
 class PropertyTypeQuantityRange(PropertyTypeQuantity):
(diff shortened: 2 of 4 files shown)


wkerzendorf:
@ftsamis that looks interesting (in good interesting :wink: )

ftsamis:
@wkerzendorf 
Unfortunately, this implementation is affected by a [bug in _pyyaml_](https://bitbucket.org/xi/pyyaml/issues/57/add_implicit_resolver-on-a-subclass-may) which essentially messes up the parent class `yaml.loader.Loader` when you subclass it and use the subclass with implicit resolvers.

The fix for the _pyyaml_ package is trivial but the bug was originally reported 4 years ago. I reopened a report in their new tracker (linked above) just in case it gets noticed.

This bug will probably **not** affect _TARDIS_ when the only YAML file parsed will be the configuration file (when the current ad-hoc YAML schema will be replaced that is). But it's totally not a good idea to ignore it.

Lastly, we can easily monkey-patch the bug, by overriding the buggy class method and doing the proper deep copy of `yaml.loader.Loader.yaml_implicit_resolvers` ourselves in the modified loader. I would appreciate your opinion on if you find this a good idea.

wkerzendorf:
@ftsamis let's leave this problem for now. I think this was very thorough research and a good fix suggestion. Well done!

ftsamis:
This one saw some progress according to @orbitfold's suggestion to make the default and my custom yaml loader interchangeable and it needs some attention.

Changes have been made to the validator in order to support ``Quantity`` objects as well instead of only strings which would be converted to Quantities.

@wkerzendorf, @orbitfold: Please kindly review!

**Not ready to be merged**: I also aim to write a test that loads a configuration file with both the default yaml loader and the custom YAMLLoader and then, after validation and post-processing, compare if the final configuration objects' properties match.

tardis/io/config_validator.py
```diff
@@ -331,47 +331,64 @@ def to_type(self, value):
 
 class PropertyTypeQuantity(PropertyType):
     def check_type(self, value):
-        try:
-            quantity_split = value.strip().split()
-            quantity_value = quantity_split[0]
-            quantity_unit = ' '.join(quantity_split[1:])
-            if quantity_unit.strip() == 'log_lsun':
-                quantity_unit = 'erg/s'
+        if isinstance(value, units.Quantity):
```
wkerzendorf: @ftsamis typechecking is frowned upon in python: just use the ducktyping. let me know if you have further questions for this.

tardis/io/config_reader.py
```diff
@@ -728,15 +730,67 @@ def deepcopy(self):
...
 
 
+class YAMLLoader(yaml.Loader):
+    """
+    A custom YAML loader containing all the constructors required
+    to properly parse the tardis configuration.
+    """
+    @classmethod
+    def add_implicit_resolver(cls, tag, regexp, first):
+        """
```
wkerzendorf: try to use `numpydoc`-style to document your methods.

tardis/io/config_validator.py
```diff
@@ -331,47 +331,64 @@ def to_type(self, value):
...
+            return float(quantity_value) * units.Unit(quantity_unit)
+        else:
+            raise ConfigError
 
-        return float(quantity_value) * units.Unit(quantity_unit)
 
 
 class PropertyTypeQuantityRange(PropertyTypeQuantity):
     @staticmethod
     def _to_units(los):
```
wkerzendorf: document your function

ftsamis: Which one? I did not introduce any new function here.

tardis/io/config_reader.py
ftsamis: @wkerzendorf Will do that!

tardis/io/config_validator.py
wkerzendorf: in this case we want it to be `if hasattr(value, 'unit') and hasattr(value, 'value'):` If things will change and we do not use astropy quantity anymore it should be fine in the future.

tardis/io/config_validator.py
```diff
@@ -331,47 +331,64 @@ def to_type(self, value):
...
+            return True
+        elif isinstance(value, basestring):
             try:
+                quantity_split = value.strip().split()
+                quantity_value = quantity_split[0]
+                quantity_unit = ' '.join(quantity_split[1:])
+                if quantity_unit.strip() == 'log_lsun':
+                    quantity_unit = 'erg/s'
+                try:
+                    float(quantity_value)
```
wkerzendorf: this is a ducktyping thing already.

tardis/io/config_validator.py
wkerzendorf: I apologize - misread this.

tardis/io/config_validator.py
ftsamis: I thought you were speaking about introducing more try/except blocks. Using hasattr instead of isinstance actually sounds good, so I'll do that! Thanks for the suggestion!

ftsamis:
Since `config_validator` can accept either a `dict` of `{'start': .., 'stop': ..., 'num': ...}` or a `list` of `[<start>, <stop>, <num>]` as valid inputs for `quantity_range_sampled`, there was no need to convert input `dict`s for this type of properties to `list` objects in the `YAMLLoader`. Therefore, I reverted that commit.

tardis/io/config_validator.py
```diff
@@ -353,25 +360,33 @@ def check_type(self, value):
...
+            quantity_split = value.strip().split()
+            quantity_value = quantity_split[0]
+            quantity_unit = ' '.join(quantity_split[1:])
+            if quantity_unit.strip() == 'log_lsun':
+                quantity_value = 10 ** (float(quantity_value) +
+                                        np.log10(constants.L_sun.cgs.value))
+                quantity_unit = 'erg/s'
+
+            return float(quantity_value) * units.Unit(quantity_unit)
+        except:
```
wkerzendorf: never do blank excepts. The code can then not even be `Ctrl-Ced`. It should work with the standard datatypes that we expect. It's fine if it raises another exception because new data was put in.

wkerzendorf: it's fine because you reraise the error. But still try to avoid this. rather use a `hasattr(value, 'strip')` and then on an `else` raise the `ConfigurationError`

tardis/io/tests/test_config_reader.py
```diff
@@ -351,5 +352,37 @@ def test_ascii_reader_exponential_law():
...
+                    return False
+            return True
+        elif isinstance(item1, collections.Sequence) and not isinstance(item1, basestring):
+            for val1, val2 in zip(item1, item2):
+                if not check_equality(val1, val2):
+                    return False
+            return True
+        else:
+            try:
+                return np.allclose(item1, item2)
```
wkerzendorf: this should be `np.allclose(item1, item2, atol=0.0)`

yeganer:
@ftsamis The test looks good. I was just wondering if it would make sense to monkeypatch pandas Dataframe here.
If we did that we could simply use `numpy.testing.assert_allclose`.
Other than that I'd suggest writing a simple __eq__ method for `AtomData` which compares the uuid1 of both atomdata objects. That way we shouldn't need to delete atom_data from the config.

[This](http://stackoverflow.com/questions/14246983/compare-assert-equality-of-two-complex-data-structures-containing-numpy-arrays) looks like a simple function which could be monketpatched into `pd.DataFrame`.
But if that's too much hassle and your function works(tested with one config loader but different//equal config settings and returned the expected result) then you can leave it as it is.

ftsamis:
@yeganer On your first point, I think we cannot simply replace the recursive checking with checking the entire object with `numpy.testing.assert_allclose` since although it works for lists it fails for dicts (no matter what they contain):

```
>>> numpy.testing.assert_allclose({'a':1},{'a':1})
TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Regarding the `AtomData` equality checking, I just noticed that because I was checking interactively how everything behaves, I had forgotten to pass the `test_parser=True` argument in the `Configuration.from_yaml` function. When setting that to True the `atom_data` is actually set to `None` so there should be no need to either delete it from the dict, or monkeypatching its `__eq__` method.

ftsamis:
I think this is finally ready to merge. I squashed the commits, and rebased to upstream because there was a merge conflict.

tardis/io/tests/test_config_reader.py
```diff
@@ -367,5 +368,36 @@ def test_absolute_relative_config_paths(monkeypatch, cwd):
...
+def test_custom_yaml_loader():
+    filename = 'tardis_configv1_density_exponential_test.yml'
+    default_conf = config_reader.Configuration.from_yaml(data_path(filename), test_parser=True,
+                                                         loader=yaml.Loader)
+    custom_conf = config_reader.Configuration.from_yaml(data_path(filename), test_parser=True,
+                                                        loader=config_reader.YAMLLoader)
+    assert len(default_conf) == len(custom_conf)
+    assert set(default_conf.keys()) == set(custom_conf.keys())
+
+    def check_equality(item1, item2):
```
yeganer: I think defining this function in a global namespace would make more sense.

Additionally I think splitting this function into two logical parts would be useful. You would have one function that's responsible for traversing the dictionary and another function which does the comparing. That way you separate the two different parts of the function.
For the traversing part it should be save to use something similiar to [http://stackoverflow.com/questions/3860813/recursively-traverse-multidimensional-dictionary-dimension-unknown](url). You just need to modify that a bit to allow traversing of two dictionaries at once.

tardis/io/config_reader.py
```diff
@@ -806,9 +806,9 @@ class Configuration(ConfigurationNameSpace):
     """
 
     @classmethod
-    def from_yaml(cls, fname, test_parser=False):
+    def from_yaml(cls, fname, test_parser=False, loader=yaml.Loader):
         try:
-            yaml_dict = yaml.load(open(fname), Loader=YAMLLoader)
```
yeganer: Can you please rebase and introduce this commit as a fixup of a10d98d8fbf75272a602d29c8752f87c60e6dce4 ?
That prevents switching back and forth between yaml loaders.

tardis/io/tests/test_config_reader.py
```diff
@@ -13,6 +14,48 @@ def data_path(filename):
...
+        for k in base:
+            traverse_configs(base[k], other[k], func, *args)
+    elif isinstance(base, collections.Iterable) and not isinstance(base, basestring) and not hasattr(base, 'shape'):
+        for val1, val2 in zip(base, other):
+            traverse_configs(val1, val2, func, *args)
+    else:
+        func(base, other, *args)
+
+
+def check_equality(item1, item2):
```
yeganer: The implementation looks good but currently `traverse_configs(default_conf, custom_conf, check_equality)` would not work because you are not returning anything in `traverse_configs`. That's why I suggested using `assert_equality` as your function that gets passed to `traverse_configs`. Then you could implement `check_equality` as a simple `try: ...; except AssertionError: return False; else: return True;`. In it's current implementation I don't see `check_equality` working.

yeganer:
@ftsamis Sorry for the delay, I added 2 final comments. If those are fixed I'll merge.

yeganer:
@tardis-sn/tardis-core @wkerzendorf This PR looks good and I'd merge it. Any objections?

wkerzendorf:
looks good - merging

PR #516, opened 2016-03-10
parser for tomic dataset table

soura97:
this python file parses the atomic dataset table and stores what has been parsed into a data.txt file.


soura97:
The code can extract different data for an atom from the table and this code can be quickly added with the sqlalchemy part and put all data into a table. I would be very grateful if somebody reviews this code and tells me what should be changed in the parsing method. Thanks! :)

wkerzendorf:
please put the resulting table in a pandas dataframe

soura97:
@wkerzendorf I changed the code to give you an idea about how the table from the websigte is parsed and stored. Please review this code and give your kind feedback. Thanks! :)

PR #520, opened 2016-03-10
Add check for dangerously misplaced packets (Debugging PR not for merging)

unoebauer:
@yeganer, @wkerzendorf, @ssim: Following our discussion, I've added a check for what I think are the potentially dangerous packets. For these, the r_outer calculation should produce a NAN (due to negative values in the sqrt). 

I've tested it with tardis_example and did not find a single packet falling into this.


wkerzendorf:
just a debug PR - decided that it's irrelevant for production

PR #523, opened 2016-03-13
to_hdf function written to MonteCarloRunner class

Prabhnith:
as the first objective TEP-002

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -117,6 +117,25 @@ class MontecarloRunner(object):
         return ['scatter', 'downbranch', 'macroatom'].index(
             line_interaction_type)
 
+    def to_hdf(self, path_or_buf):
+        """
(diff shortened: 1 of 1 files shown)


yeganer:
Thank you for the contribution. There are still some issues with your implementation.

It seems like you are not writing to the HDF file you created.
Additionally I suggest using `os.path` for dealing with filenames, paths etc.
For checking whether you are dealing with filename or hdf file I suggest doing a `isinstance` or something similar instead of assuming so after creating `pd.HDFStore` fails.

Prabhnith:
need help in writing data to the file, rest changed according to previous commet.

yeganer:
I suggest you have a look at how pandas HDFStores work and figure out how to work with them. Once you've done that you should decide which members you want to store in the HDF5 file.
If you can't decide whether one particular variable is important and needs to be stored feel free to ask.
A general rule of thumb would be that is should be possible to restore an instance from a file.

PR #525, opened 2016-03-17
Counting Estimator: Preliminary Task

SooryaN:
@unoebauer, here are the line and cell based counters. I've written them into the `montecarlo.pyx` file too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 *.so
 __pycache__
 
-
+tardis/tardis_example
 # Ignore .c files by default to avoid including generated code. If you want to
 # add a non-generated .c extension, use `git add -f filename.c`.
 *.c
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -43,6 +43,9 @@ class MontecarloRunner(object):
         self.j_estimator = np.zeros(no_of_shells, dtype=np.float64)
         self.nu_bar_estimator = np.zeros(no_of_shells, dtype=np.float64)
         self.j_blue_estimator = np.zeros(tau_sobolev_shape)
+        #Counters
+        self.line_list_resonance_counter = np.zeros(tau_sobolev_shape[0], dtype=np.int64)
+        self.cell_list_resonance_counter = np.zeros(tau_sobolev_shape)
 
 
     def _initialize_geometry_arrays(self, structure):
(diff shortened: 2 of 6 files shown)


yeganer:
It's just a small suggestion, but can you please make the counters available on the python level?
That is after running tardis with `mdl = run_tardis('tardis_example.yml')` I can access them somehow. For example as `mdl.runner.line_counters` or something similiar?
Thanks in advance.

unoebauer:
@SooryaN - thanks for the update. @yeganer is right - the idea is to make the counter available on the python level. 

Additionally, something is clearly wrong with your PR. The Travis testing process completely fails.

yeganer:
@unoebauer That's because the travis fix isn't merged in this branch.

@SooryaN I'd suggest doing 
```
git fetch upstream
git rebase -i upstream master
git push origin mine -f
```
But be carefull with `git rebase -i` and consider doing a backup before. In the end you should have only commits with your code, no merges etc.

unoebauer:
Hi @SooryaN - looking at the C source code, I am fairly confident, that the cell-based counter does the right thing. What remains to be done is to transport the counter fully to the python level. With respect to this task, you are also very close. Introduce a new runner attribute and initialize it outside of the C/Cython parts. Again, take ``runner.j_blue_estimator`` as an example.

SooryaN:
@unoebauer, On running `mdl = run_tardis('tardis_example.yml')` you can access the counters as `mdl.line_list_resonance_counter` and `mdl.cell_list_resonance_counter`

@yeganer I followed the steps you had outlined with regards to git. Not sure why Travis still fails.

yeganer:
@SooryaN It's just a guess but it looks like during test_cmontecarlo.py something crashes. Most likely it's a segfault because you didn't initialize the arrays when running the tests.

Were the tests working locally? `./setup.py test`

unoebauer:
@yeganer is correct - the cell-based counter now only counts how many resonances occur in each shell. We want to know how many resonances occur with each line per cell. I seem to remember that at an earlier stage of the PR (yesterday, when I posted my previous comment), the counter looked as if it would fulfil these specifications.

SooryaN:
@unoebauer , I made a mistake earlier. Now, the counters should be fine. 
@yeganer, I believe both the counters are now passed to the c part without a hitch. Though I don't think it would have been an issue earlier as the `storage.cell_list_resonance_counter` would have still behaved the same considering both the variables are np.zeros to begin with.

unoebauer:
@SooryaN - nice job! Your counter does the right thing. I've just tested your PR on tardis_example and got the same number of resonances as my own counter.

unoebauer:
Concerning the Travis problems - I am confused. When starting the testing process offline, with ``./setup.py test``, it runs smoothly.

unoebauer:
@SooryaN - the counter is working, which is very good. However, any PR should eventually pass the testing process. So, please try to fix the remaining issue with Travis. 

Travis crashes due to a segmentation fault during the test_cmontecarlo process. As @yeganer suspected from the start, this is a consequence of not allocating the resonance counting arrays you introduced in the ``test_increment_j_blue_estimator``.

yeganer:
The lates version looks good. Although using `calloc` instead of `malloc` would be the preferred way to allocate memory that is initialized to zero.

PR #527, opened 2016-03-23
Unit tests for untested methods of tardis/atomic.py

karandesai-96:
This PR will be consisting a series of commits targetting to improve code coverage of [**tardis/atomic.py**](https://github.com/tardis-sn/tardis/blob/master/tardis/atomic.py).

This checklist maintains the methods imporved in **tardis/atomic.py**. The commits enclosed in brackets show where these tests were modified/ added.

**Methods:**
- [x] data_path (0645546)
- [x] read_basic_atom_data (6da1b9d)
- [x] read_ionization_data (6da1b9d)
- [x] read_levels_data (6da1b9d)
- [x] read_synpp_refs (24c3796)
- [x] read_lines_data (6da1b9d)
- [x] read_zeta_data (24c3796)
- [x] read_collision_data (fa7d778)
- [ ] read_ion_cx_data
- [x] read_macro_atom_data (013a625)

### DIFF
--- a/tardis/atomic.py
+++ b/tardis/atomic.py
@@ -19,15 +19,13 @@ class AtomDataNotPreparedError(Exception):
 
 logger = logging.getLogger(__name__)
 
-default_atom_h5_path = os.path.join(os.path.dirname(__file__), 'data', 'atom_data.h5')
 
 def data_path(fname):
-    data_dir = os.path.join(os.path.dirname(__file__), 'data')
-    return os.path.join(data_dir, fname)
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), 'data', fname)
 
-atomic_symbols_data = np.recfromtxt(data_path('atomic_symbols.dat'),
-                                    names=['atomic_number', 'symbol'])
 
+default_atom_h5_path = data_path('atom_data.h5')
+atomic_symbols_data = np.recfromtxt(data_path('atomic_symbols.dat'), names=['atomic_number', 'symbol'])
 symbol2atomic_number = OrderedDict(zip(atomic_symbols_data['symbol'], atomic_symbols_data['atomic_number']))
 atomic_number2symbol = OrderedDict(atomic_symbols_data)
 
--- a/tardis/atomic.py
+++ b/tardis/atomic.py
@@ -21,12 +21,16 @@ logger = logging.getLogger(__name__)
 
 
 def data_path(fname):
-    return os.path.join(os.path.dirname(os.path.realpath(__file__)), 'data', fname)
+    return os.path.join(
+            os.path.dirname(os.path.realpath(__file__)), 'data', fname)
 
 
 default_atom_h5_path = data_path('atom_data.h5')
-atomic_symbols_data = np.recfromtxt(data_path('atomic_symbols.dat'), names=['atomic_number', 'symbol'])
-symbol2atomic_number = OrderedDict(zip(atomic_symbols_data['symbol'], atomic_symbols_data['atomic_number']))
+atomic_symbols_data = np.recfromtxt(data_path('atomic_symbols.dat'),
+                                    names=['atomic_number', 'symbol'])
+
+symbol2atomic_number = OrderedDict(zip(atomic_symbols_data['symbol'],
+                                       atomic_symbols_data['atomic_number']))
 atomic_number2symbol = OrderedDict(atomic_symbols_data)
 
 
--- a/tardis/tests/test_atomic.py
+++ b/tardis/tests/test_atomic.py
@@ -72,6 +72,27 @@ def test_read_zeta_data():
         atomic.read_zeta_data(atomic.default_atom_h5_path)
 
 
+def test_read_collision_data():
+    data = atomic.read_collision_data(chianti_he_db_h5_path)
+    assert data[0]['atomic_number'][0] == 2
+    assert data[0]['ion_number'][0] == 0
+    assert data[0]['level_number_upper'][0] == 18
+    assert data[0]['level_number_lower'][0] == 2
+    assert data[0]['g_ratio'][0] == 1.0
+    testing.assert_almost_equal(data[0]['delta_e'][0], 35484.251143, decimal=4)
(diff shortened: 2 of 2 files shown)


wkerzendorf:
@karandesai-96 I think this is nice to add more coverage. But testing C-functions is much more important to us. Maybe @yeganer can show you what needs to be done there.

wkerzendorf:
@tardis-sn/tardis-core I think this adds more tests to our code. I need one other to sign off or reject this PR.

karandesai-96:
It came to my knowledge that this code is a new implemented feature to be included in further versions and under heavy development, so writing tests right now might not be good idea as they can break and need to be updated. I have not added tests for scientific code here, only the tests are those which ensure that reading from an hdf5 file is done properly.

tardis/tests/test_atomic.py
```diff
@@ -1,32 +1,122 @@
...
     assert data['symbol'][13] == "Si"
-    si_mass = data['mass'][13]
-    testing.assert_almost_equal(si_mass, 28.085, decimal=4)
-    pass
+    testing.assert_almost_equal(data['mass'][13], 28.085, decimal=4)
 
-def test_ionization_h5_readin():
+
+def test_read_ionization_data():
     data = atomic.read_ionization_data(atomic.default_atom_h5_path)
```
yeganer: I'd suggest using fixtures for the data paths.
That way they could easily be changed for all tests in case something changed.
This goes in line with the DRY principle.

karandesai-96: Yes, I will correct it.

karandesai-96:
If this is signed off, it would be really nice to have some sample data for `ion_cx_data`. I will add one or two rows in **chianti_He_db.h5.hdf5** and assert accordingly - just for sake of completeness. This will complete the PR with tests of all hdf5 interactions in this file.

wkerzendorf:
@karandesai-96 let me be clear about this: currently there is no need to test this experimental part of tardis as it is still being developed.

tardis/tests/test_atomic.py
```diff
@@ -4,8 +4,16 @@
 from tardis import atomic
 
 
-chianti_he_db_h5_path = os.path.join(
-    os.path.dirname(os.path.realpath(__file__)), 'data', 'chianti_he_db.h5')
+@pytest.fixture(scope="module")
+def default_atom_h5_path():
```
yeganer: I think It's fine calling `atomic.data_path` since you are testing that function, too.

karandesai-96:
@wkerzendorf So I would close this PR and keep it mind about this. Over the time this code will be developed and I think we have **v1.5** tags floating around here. Once this code gets stable enough, I will reopen it, rebase it and then work on it. It is the best to close it as of now. Sounds like a plan ?

wkerzendorf:
@karandesai-96 - just don't test that part. As far as I can tell there is a lot of good in this. Is there a reason not to merge?

tardis/atomic.py
```diff
@@ -19,15 +19,13 @@ class AtomDataNotPreparedError(Exception):
 
 logger = logging.getLogger(__name__)
 
-default_atom_h5_path = os.path.join(os.path.dirname(__file__), 'data', 'atom_data.h5')
 
 def data_path(fname):
-    data_dir = os.path.join(os.path.dirname(__file__), 'data')
-    return os.path.join(data_dir, fname)
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), 'data', fname)
```
wkerzendorf: is that more than 140 characters?

yeganer:
I think this should be merged.
Just because one method is missing doesn't mean we shouldn't merge. There
are always methods not worth testing ( to simple or to complicated ).

Wolfgang Kerzendorf <notifications@github.com> schrieb am Fr., 1. Apr. 2016
um 15:04 Uhr:

tardis/atomic.py
yeganer: If you are talking about the line length PEP, the value is 80. I'm not sure whether this is too long but line 29 looks like it definitely is to long

tardis/tests/test_atomic.py
karandesai-96: - [x] Done

tardis/atomic.py
wkerzendorf: sorry 80 yes.

karandesai-96: Is the rule followed as 80 chars or 120 ? Many orgs compromise this PEP8 rule as 120 characters. I'll follow whichever used and push a final cosmetic concluding commit.

yeganer: As far as I know we stick with 80. Although we won't do PEP fixes on their own. If I stumble upon code that doesn't follow PEP I'll edit it on the fly.

karandesai-96: I'll take care about it. But why merge if we know of existent PEP8 errors ? I'll include the fixes to cleanup this PR, if we are going with "_edit on the fly_" way, I'd better squash the PEP8 fixes into previous single commit.

yeganer: I'm trying to look out for PEP8 errors but since I'm not doing a checkout of every PR I have a look at I won't see them (I have a pep checker for my editor only).

karandesai-96: I have a built-in PEP8 checker in Pycharm. I had compromised line wrap to 120 characters, but I switched it to 80 characters just now. I removed PEP8 errors from the part of code in this PR. Rest will be "_edited on the fly_" :wink:

karandesai-96:
I cannot find more additions here now. If I missed anything, I'll include it as asked.

wkerzendorf:
Maybe just one more comment on PEP8: Every PEP8 fix will show a changed line in the files changed part. So for us to review PRs that come from a very junior person contributing makes it harder to look at the actual part of the code. So we encourage PEP8 changes of existing code for the more senior members. Obviously we want everyone (senior & junior) contributors to stick to PEP8 in general.

karandesai-96:
Acknowledged. :+1: I will add pep8 complaint code from now, **without making PEP8 fixes to existing code**.

PR #529, opened 2016-03-23
Fix comparison in AtomData.from_hdf5() of atomic.py

karandesai-96:
In **tardis/atomic.py**:

In `AtomData.from_hdf5` method, there's this block:
```python
if 'ion_cx_data' in h5_datasets and 'ion_cx_data' in h5_datasets:
    ion_cx_data = read_ion_cx_data(fname)
else:
    ion_cx_data = None
```

Whereas in `read_ion_cx_data` method, there's something else:
```python
ion_cx_th_data = h5_file['ionization_cx_threshold']
ion_cx_sp_data = h5_file['ionization_cx_support']
```

It should be:
```python
if 'ionization_cx_threshold' in h5_datasets and 'ionization_cx_support' in h5_datasets:
    ion_cx_data = read_ion_cx_data(fname)
else:
    ion_cx_data = None
```
Then it would denote that call the method **only if both columns exist** in dataset.

I discovered that this part of code was contributed last by @mklauser . Please review it and confirm whether I have fixed it correctly.

### DIFF
--- a/tardis/atomic.py
+++ b/tardis/atomic.py
@@ -332,7 +332,7 @@ class AtomData(object):
         else:
             synpp_refs = None
 
-        if 'ion_cx_data' in h5_datasets and 'ion_cx_data' in h5_datasets:
+        if 'ionization_cx_threshold' in h5_datasets and 'ionization_cx_support' in h5_datasets:
             ion_cx_data = read_ion_cx_data(fname)
         else:
             ion_cx_data = None


karandesai-96:
Further in the `__init__` method of `AtomData` class, there is this code block:
```python
if ion_cx_data is not None:
    # ... code ...
    self.ion_cx_th_data = DataFrame(np.array(ion_cx_data[0]))
    self.ion_cx_th_data.set_index(['atomic_number', 'ion_number', 'level_id'], inplace=True)

    self.ion_cx_sp_data = DataFrame(np.array(ion_cx_data[1]))
    self.ion_cx_sp_data.set_index(['atomic_number', 'ion_number', 'level_id'])
else:
    self.has_ion_cx_data = False
```
This means that `ion_cx_data` would be a tuple of `ionization_cx_threshold` and `ionization_cx_support` columns, which is separated out later. So this matter is taken care of.

wkerzendorf:
@karandesai-96 both of these methods are related to a new feature and I we should not deal with fixing them now. I will close this for now - if there is further discussion feel free to reopen

karandesai-96:
Yes, suitable.

PR #531, opened 2016-03-30
Organize headers in C-Extension

yeganer:
Header files should only contain includes that are necessary for the
compiler to understand the header. Includes needed by the implementation
should be included by the corresponding source.

With this philosophy in mind the includes in all c files were
reorganized.
As part of this reorganization the header `cmontecarlo1.h` was removed
and it's declaration moved to `cmontecarlo.h`

Resolves: #489


wkerzendorf:
@chvogl your comments would be interesting. I'm fine to merge this as this is only of cosmetic nature.

yeganer:
@chvogl @wkerzendorf Since these changes affect mostly imports I'd be happy to help resolve any merge conflicts arising from this change.

chvogl:
@wkerzendorf No objections on my account. Fire away!

Issue #517, opened 2016-03-10
distance2boundary: sometimes returns the wrong distance


yeganer:
While looking at the code I came across a weird behaviour in distance2boundary.
I thought the recently_crossed_boundary flag was necessary because the old implementation couldn't handle packets close to the boundary. I was wrong. This flag was implemented to correct numerical errors in the case that moving a packet resulted in a radius outside it's shell boundary. This happens quite a lot actually, around 10% of the invocations to distance2boundary have either `r<r_inner` or `r>r_outer`.Most of those cases are `r<r_inner` and this is checked for with
```
if (recently_crossed_boundary ==1):
    return d_outer
```
But in the other case where `r>r_outer` nothing is checked and the distances are calculated the normal way. For d_outer the formula used is
`d = sqrt(r_outer ** 2 + (mu**2 - 1) r**2 ) - mu * r`.
For r>r_outer this gives the distance to the nearest crossing instead of the farthest.
I think this is not a huge issue but since I haven't found anything contradicting me this can result in a packet crossing the inner boundary multiple times in a row and being absorbed in that process or that the r and shell_id of the packet are messed up and are not synchronised anymore.

In my opinion there are two ways to fix this:
1. Implement a check similiar to the one for `r<r_inner` and return the correct distance. This will leave these numerical errors in the code.
2. Remove the numerical errors by explicitly setting `r=r_inner|r_outer` when crossing the shell boundary. I'm not sure of the consequences of this approach. @unoebauer might have some experience with this particular problem.

yeganer:
Sorry for the confusion. The calculated distance seems to always be correct.

So the only thing that's strange now is why removing `recently_crossed_boundary` changes the result.
I'll do some more tests and come back later.

wkerzendorf:
@yeganer there is no need to apologize. It's very important to question the code! @tardis-sn/tardis-core to make sure that everyone has a look.

wkerzendorf:
@ssim did we discuss setting `r=r_inner` at some stage?

unoebauer:
@yeganer; what about the possibility we talked about yesterday (see PR #520). I think this is technically still possible in the current master

unoebauer:
@yeganer - is this still relevant? Which your modifications, I thought this issue is resolved. Or am I wrong?

yeganer:
[...]

PR #533, opened 2016-03-31
Remove rpacket.recently_crossed_boundary flag

yeganer:
Using the new distance2boundary calculation, we don't need recently_crossed_boundary flag, so it is
removed from all occurences.
This flag covered only outward propagating packets which have mu>0
anyway and result in the same calculation. This commit removes that
overhead and improves readability of the code.

This is in line with the discussion in Issue #517 .
If we are not 100% sure that there is no negative side effect, I'd suggest waiting for the improved C tests. Adding simple counters showed that `recently_crossed_boundary == 1` implies `mu > 0` for all testes setups (w7, abn_tom, example).


ssim:
This looks good to me - happy to sign off is someone else is too!

PR #536, opened 2016-04-01
Fix Issue 534

yeganer:
This is a quickfix approximating each bin bin the bin_center instead of
one bin edge.

The tests comparing the spectrum are currently failing. I'm thinking about doing a proper fix later.

This would include logic to use the packet_nu and packet_energy if available to create `luminosity_density_lambda` and, if that's not available, transfrom `luminosity_density_nu` into `luminosity_density_lambda`.

### DIFF
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -372,7 +372,7 @@ class TARDISSpectrum(object):
 
     @property
     def frequency(self):
-        return self._frequency[:-1]
+        return (self._frequency[1:] + self._frequency[:-1]) / 2
 
 
     @property


wkerzendorf:
@yeganer - I think we should not get stuck in this. This is a minor known issue, but requires quite a bit to fix it. Just leave any of the binning in dalek for now.

wkerzendorf:
can we close this for now?

yeganer:
Sure, we can close this and #559 . Should the issue be closed, too?

PR #539, opened 2016-04-15
Adding OSX to our travis build

wkerzendorf:
This restructures our TRAVIS runs slightly and mainly adds OSX build to test it working on macs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+os:
+  - linux
 language: python
 
 python:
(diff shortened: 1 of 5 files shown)


wkerzendorf:
@tardis-sn/tardis-core can someone sign off on this so I can merge.

karandesai-96:
```yml
env:
  global:
    COMPILER=gcc
    PANDAS_VERSION=0.16
```

Back when I made libgfortran fix, I had upgraded pandas to 0.16.2
I wonder if this is a suitable time to include this change here.

yeganer:
These changes look reasonable but since I'm not familiar with travis-ci I'd recommend somebody else looking at these, too.
The github diff for your first commit looks broken for me. It shows stupid stuff but viewing the file in your repository it looks okay.

wkerzendorf:
@yeganer - not sure anybody else knows about the innards of travis :(

karandesai-96:
Since normal conda env and rtd's conda env has everything same except, four dependencies regarding coverage, more in normal env, we can share a single yml config file for both, if four more installed dependencies related to code coverage aren't a problem.

karandesai-96:
This suggestion directs towards the _DRY_ principle. If both requirement files are kept as one, we can do the following steps:
1. Delete **conda-requirements.txt**
2. Update documentation about creating an environment using:
        ```conda env create --name tardis -f environment.yml```
3. **`literalinclude`** with **`linenos`** from **environment.yml** file to reflect it in readthedocs.

For further change in dependencies, one has to then make only 1 update instead of 3 ;)

wkerzendorf:
@karandesai-96 I think this would be useful for a future feature to get rid of conda requirements. Can you put it as an issue. I think for now this works and has a substantial improvement over the current situation.

karandesai-96:
@wkerzendorf I agree. Opening an issue and handling it in near future is a good idea, I will do it right away.

PR #540, opened 2016-04-16
Mirroring C enums in Python for CMontecarlo tests.

karandesai-96:
There are certain parameters, such as `RPacket.status` and `StorageModel.cont_status` which are C enum datatypes. Enums were considered as `ctypes.c_uint` or `ctypes.c_int` in Python, and implemented in that manner.

This PR adds a new file, **enum.py** which sets up a mirror for C enums in python. Enums are not supported directly in `ctypes` so a Meta Class and corresponding Class were written.

Following were mirrored for Python tests and used:
- [X] TardisError
- [X] RPacketStatus
- [X] ContinuumProcessesStatus

This targets improvement in readability of tests.

### DIFF
--- a/tardis/montecarlo/enum.py
+++ b/tardis/montecarlo/enum.py
@@ -11,10 +11,10 @@ class EnumerationType(type(c_int)):
 
             dictionary["_members_"] = _members_
         else:
-            _members_ = dict["_members_"]
+            _members_ = dictionary["_members_"]
 
         dictionary["_reverse_map_"] = {value: key for key, value in _members_.items()}
-        cls = type(c_int).__new__(metacls, name, bases, dict)
+        cls = type(c_int).__new__(metacls, name, bases, dictionary)
 
         for key, value in cls._members_.items():
             globals()[key] = value
@@ -39,3 +39,8 @@ class CEnumeration(c_int):
         return "<%s.%s: %d>" % (self.__class__.__name__,
                                 self._reverse_map_.get(value, '(unknown)'),
                                 value)
+
+
+class ContinuumProcessesStatus(CEnumeration):
+    OFF = 0
+    ON = 1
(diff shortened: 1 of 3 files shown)


wkerzendorf:
That looks like an interesting addition and makes the code more readable - well done @karandesai-96

tardis/montecarlo/enum.py
```diff
@@ -0,0 +1,58 @@
+from ctypes import c_int
+
+
+class EnumerationType(type(c_int)):
```
wkerzendorf: Normally, when we copy code from the internet we like to attribute it with the original source. In science this is an important part to understand the source of information. Can you add one?

tardis/montecarlo/enum.py
```diff
@@ -0,0 +1,58 @@
...
+        if isinstance(other, int):
+            return self.value == other
+
+        return type(self) == type(other) and self.value == other.value
+
+    def __repr__(self):
+        value = self.value
+        return "<%s.%s: %d>" % (self.__class__.__name__,
+                                self._reverse_map_.get(value, '(unknown)'),
+                                value)
```
yeganer: You can directly write self.value here

karandesai-96:
Thanks @wkerzendorf. I will work upon the suggestions received here and wrap up here with a couple of docstrings ;)

tardis/montecarlo/enum.py
karandesai-96: - [X] Added.

tardis/montecarlo/enum.py
karandesai-96: - [X] Done.

wkerzendorf:
I'm signing off on this piece of code. It looks good (as far as I understand). @yeganer - you can merge it tomorrow if you are happy as well.

yeganer:
Looks good to me, I'll merge.

karandesai-96:
This modification is causing occasional segfaults, so it needs to be modified.

PR #541, opened 2016-04-18
Move progress bar update after calculation

yeganer:
This ensures that the progress bar is most accurate.


wkerzendorf:
sure - signing off.

PR #542, opened 2016-04-18
Updates to He NLTE treatment.

aoifeboyle:
In the process of ensuring that the results for my paper were reproducible with the new plasma, I found some discrepancies and made some changes. I think there are parts of this that could be better coded but at least it works. I will try to improve/condense it in a while. This now gives the correct results. 

Aoife

### DIFF
--- a/tardis/plasma/properties/nlte.py
+++ b/tardis/plasma/properties/nlte.py
@@ -6,7 +6,7 @@ import pandas as pd
 
 from tardis.plasma.properties.base import (PreviousIterationProperty,
                                            ProcessingPlasmaProperty)
-from tardis.plasma.properties import PhiSahaNebular, PhiSahaLTE
+from tardis.plasma.properties.ion_population import PhiSahaNebular
 
 __all__ = ['PreviousElectronDensities', 'PreviousBetaSobolev',
            'HeliumNLTE', 'HeliumNumericalNLTE']
@@ -45,7 +45,8 @@ class PreviousBetaSobolev(PreviousIterationProperty):
 class HeliumNLTE(ProcessingPlasmaProperty):
     outputs = ('helium_population',)
 
-    def calculate(self, level_boltzmann_factor, electron_densities,
+    @staticmethod
+    def calculate(level_boltzmann_factor, electron_densities,
         ionization_data, beta_rad, g, g_electron, w, t_rad, t_electrons,
         delta, zeta_data, number_density, partition_function):
         """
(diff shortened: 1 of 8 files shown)


wkerzendorf:
@aoifeboyle is this the new code that is used to make the paper. Why doesn't it pass?

aoifeboyle:
Hi @wkerzendorf. Sorry, busy with other stuff. Will add this to the to-do list but might need to wait til the weekend. I changed the `ion_number_density` inputs so that the helium ion populations could be calculated correctly in that property so now I need to reorganise `property_collections`. There may be a cleaner way to do this so I need to think about it for a while.

aoifeboyle:
@wkerzendorf How is this as far as removing the duplicate code?

wkerzendorf:
it's still showing duplicates. but we can look at that later.

aoifeboyle:
@wkerzendorf Does that mean we can merge?

wkerzendorf:
@aoifeboyle We should get rid of them - if we sit together this should not take much longer than an hour. 13.30 today?

tardis/plasma/properties/nlte.py
```diff
@@ -68,38 +66,38 @@ def calculate(self, level_boltzmann_factor, electron_densities,
...
     @staticmethod
     def calculate_helium_three(t_rad, w, zeta_data, t_electrons, delta,
-        g_electron, beta_rad, ionization_data, electron_densities, g):
+        g_electron, beta_rad, ionization_data, g):
         """
         Calculates the He III level population values.
         """
         zeta = PhiSahaNebular.get_zeta_values(zeta_data, 2, t_rad)[1]
-        he_three_population = (2 / electron_densities) * \
+        he_three_population = 2 * \
```
yeganer: You can wrap the whole equation in `()` then you don't need `\` at the end of line which should be avoided.

tardis/plasma/properties/nlte.py
```diff
@@ -45,20 +45,18 @@ def set_initial_value(self, kwargs):
 class HeliumNLTE(ProcessingPlasmaProperty):
     outputs = ('helium_population',)
 
-    def calculate(self, level_boltzmann_factor, electron_densities,
+    @staticmethod
```
yeganer: why are all these methods staticmethods?

aoifeboyle:
@wkerzendorf Sorry, only saw this yesterday evening. I'm focusing on the draft right now so I think it will have to wait til next week.

tardis/plasma/properties/nlte.py
aoifeboyle: Remnant from the original method I tried to get this to work before I thought of a different way. I can change it.

tardis/plasma/properties/nlte.py
aoifeboyle: Ok, sure. What's the difference?

yeganer: To me `\` at the end of a line looks ugly and PEP recommends to use `( )` instead.
The added benefit of the 2nd approach is that most editors will indent the following lines differently which looks nicer, too.

Issue #535, opened 2016-04-01
Treat rel. paths inside config relative to the config file


yeganer:
I think this would be a useful addition if one is storing all files to describe a model in one directory.
With this feature one would be able to run a config without being in the config directory.

PR #543, opened 2016-04-18
Fix issue #535: Treat rel. paths inside config relative to the config file

lukeshingles:
This is a quickfix for issue #535. File paths in the config file are now treated as relative to the location of the config file.

### DIFF
--- a/tardis/base.py
+++ b/tardis/base.py
@@ -23,11 +23,14 @@ def run_tardis(config, atom_data=None):
 
     from tardis.io import config_reader
     from tardis import model, simulation, atomic
+    import os
 
     try:
         config_dict = yaml.load(open(config))
+        config_dirname = os.path.dirname(config)
     except TypeError:
         config_dict = config
+        config_dirname = ''
 
     if atom_data is not None:
         try:
@@ -36,7 +39,7 @@ def run_tardis(config, atom_data=None):
             atom_data = atom_data
 
     tardis_config = config_reader.Configuration.from_config_dict(
-        config_dict, atom_data=atom_data)
+        config_dict, atom_data=atom_data, config_dirname=config_dirname)
     radial1d_mdl = model.Radial1DModel(tardis_config)
 
     simulation.run_radial1d(radial1d_mdl)
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -745,11 +745,15 @@ class Configuration(ConfigurationNameSpace):
         if tardis_config_version != 'v1.0':
             raise ConfigurationError('Currently only tardis_config_version v1.0 supported')
 
-        return cls.from_config_dict(yaml_dict, test_parser=test_parser)
+        config_dirname = os.path.dirname(fname)
+        
+        return cls.from_config_dict(yaml_dict, test_parser=test_parser,
+                                    config_dirname=config_dirname)
 
     @classmethod
     def from_config_dict(cls, config_dict, atom_data=None, test_parser=False,
-                         config_definition_file=None, validate=True):
+                         config_definition_file=None, validate=True,
+                         config_dirname=''):
         """
         Validating and subsequently parsing a config file.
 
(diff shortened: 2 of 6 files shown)


wkerzendorf:
This is an interesting proposal. Not clear if we want to have it that way as I think most software does it relative to the CWD for these user defined config files. But what does @tardis-sn/tardis-core think?

yeganer:
I think this change will help to keep things more organized. If I'm defining paths inside a file I expect them to be treated relative to that file, if I don't want that behavior I'd use absolute paths.

The current implementation requires me to constantly link the atom_data database to my current working directory.

lukeshingles:
If nothing else, at least this was a good example problem for me to work on. Finally passes the tests :).

lukeshingles:
Could include an extra setting for each input file with options to set the base path to either the config directory or CWD. Alternatively, special keywords like $CWD or $CONFIGDIR could be replaced with their values during input. Both of these can preserve the default behaviour of giving paths relative to the CWD.

wkerzendorf:
@lukeshingles I think it's fine. If we need this later - we can add it. I think we just need to review it from the code and then merge.

yeganer:
I think it's fine as it is. In the next month the config system might see significant changes so I wouldn't add these features now.

As for the implementation. I think it's fine. For a rewrite of the config system I'd convert everything to absolute paths but that's not necessary.

wkerzendorf:
@tardis-sn/tardis-core are we merging this now?

yeganer:
@wkerzendorf I'm happy to merge. If @lukeshingles wants to add a small test for that functionality it would be even better but since this is not a major change I wouldn't make it a requirement for merging.

lukeshingles:
I'm happy to do it, although it would take me a couple of working days to have it done. What kind of test did you have in mind? Moving the working directory up a level and rewriting the config dir to match?

Also, any idea why Scrutinizer failed this commit?

yeganer:
@lukeshingles I was thinking of the following:

- Use some config file already in the repository, get it's absolute path

- Call the config validator with the path to that file in different ways. (absolute path, local path while in that directory, local path while directory above/below)
- Test if reading in the abundance file works (for example, any other 'external' file would work, too)

ftsamis:
Since I plan to write/update tests for the changes I make in the configuration system and I have no extensive experience in writing tests, this would be a good chance for me to write one for the changes @lukeshingles did. May I?

yeganer:
@ftsamis Sure, feel free to add them and make a PR to @lukeshingles 's branch.
I think it's a really good idea actually because you can make yourself familiar with `py.test` and use that knowledge during your own development. But don't spend too much time on the tests as they will most likely change because of the rewrite you are doing.

ftsamis:
Please check my PR on @lukeshingles' branch which adds tests for this PR. This is also my first attempt to write pytest tests, so bear with me if something is not as it should be.

 https://github.com/lukeshingles/tardis/pull/1

wkerzendorf:
@lukeshingles @ftsamis @tardis-sn/tardis-core is that done?

lukeshingles:
Looks good to me.

wkerzendorf:
@lukeshingles @ftsamis well done!

yeganer:
@tardis-sn/tardis-core Sorry I'm a bit late to the party.
I wouln't have merged this version. There is an empty commit (which could have been deleted easily) and I was about to suggest using [`monkeypatch`](https://pytest.org/latest/monkeypatch.html) to simulate the directory changes.

Pytest provides this fixture automatically to do exactly what we want.

ftsamis:
You are right @yeganer, monkeypatch.chdir is exactly what I should have used in this one. Opened a new PR with this change: https://github.com/tardis-sn/tardis/pull/555 , please review it.

Regarding the empty commit, I'm not sure anything can be done now on my side, except not pushing empty commits again on PR branch. I was exploring Travis, but I should have done that on a separate test branch (which I did, after this empty commit) from the beginning.

yeganer:
@ftsamis There is no problem with pushing anything to a PR branch, but please make sure to clean up the commit history of your PR before finishing it (have a look at `git rebase -i` and `git push -f`).

PR #530, opened 2016-03-27
Refactor cmontecarlo tests for exercising parametrization.

karandesai-96:
Current design procedure of tests:
- CMontecarlo methods are compiled into **montecarlo.so** (shared object)
- Corresponding tests are compiled into **test_montecarlo.so**.
- For testing these using pytest framework, **test_montecarlo.so** is wrapped and C tests are exposed to **test_cmontecarlo.py** using Python `ctypes` library.

This PR refactors the current tests such that:
- They directly wrap **montecarlo.so** and expose the C methods and struct parameters to Python instead of C tests.
- Pytest's parametrization can be used to test the methods extensively.
- Parameters of each Python test method are the ones which are used by the corresponding C method.
- RPacket and StorageModel structs are defined inside the test every time, using those parameters.

PR #544, opened 2016-04-19
Continue cmontecarlo tests refactor. Extend #530.

karandesai-96:
This PR extends the work of #530. The tests which wrapped the C tests from **test_cmontecarlo.so** were redesigned to directly wrap C methods from **cmontecarlo.so**. A couple of _TODOs_ commented in the file during the work of #530 have been resolved here.
The checklist states the tests which were added / refactored. The commit where the test was added is mentioned within parenthesis:

_(No commit in parenthesis means the work on that method was done in #530.)_

- [X] `line_search` ( 5441c00 )
- [X] `reverse_binary_search` ( 50cf9f0 )
- [X] `rpacket_doppler_factor`
- [X] `compute_distance2boundary`
- [X] `compute_distance2line`
- [X] `compute_distance2continuum`
- [X] `move_packet` ( e7a7b4e )
- [X] `increment_j_blue_estimator`
- [X] `move_packet_across_shell_boundary` ( fd9d4c1 )
- [X] `montecarlo_thomson_scatter` ( a776364 )
- [X] `montecarlo_line_scatter` 

#### Not yet relevant:
These tests will have to be rewritten in future as the underlying code is experimental and at early development stage. Few of these tests have a `@pytest.mark.skipif` decorator too.
- [X] `montecarlo_continuum_event_handler` ( 870e5e4 )
- [X] `montecarlo_free_free_scatter` ( 4f4a1e8 )
- [X] `montecarlo_bound_free_scatter` ( 6e7be8c )
- [X] `bf_cross_section` ( 1f099af )
- [X] `calculate_chi_bf` ( 0de7cba )

----
After writing the above mentioned tests, following are undertaken as a routine to wrap up this task:
- [X] Every function in C has a `test_<func>` counterpart in python ( 02d1f13, acec56c )
- [X] Not yet implemented tests are marked `skip` with a descriptive reason ( ^ same )
- [X] Tests are grouped by 'importance', block comments separate the groups ( ^ same )
- [X] Removal of other `test_*.c` files ( 2b8465c ) and the building of `test_cmontecarlo.so` ( 70dcfac )

----
These methods are yet not tested. Refer #556 for details:
- `montecarlo_one_packet`
- `montecarlo_one_packet_loop`
- `montecarlo_main_loop`
- `montecarlo_event_handler`
- `macro_atom`


tardis/montecarlo/tests/test_cmontecarlo.py
```diff
@@ -202,6 +202,25 @@ def test_bf_cross_section(packet_params, expected, packet, model):
...
 
+# TODO: fix underlying method and update expected values in testcases.
+# For loop is not being executed in original method, and hence bf_helper
+# always remains zero. Reason for for loop not executed:
+#         "current_continuum_id = no_of_continuum edges"
+@pytest.mark.parametrize(
+    ['packet_params', 'expected'],
+    [({'nu': 0.1, 'mu': 0.3, 'r': 7.5e14}, 0.0),
+     ({'nu': 0.2, 'mu': -.3, 'r': 7.7e14}, 0.0)]
+)
```
karandesai-96: @wkerzendorf @yeganer There might be a problem in `calculate_chi_bf`s implementation, or I wonder `chi_bf` is zero for these test cases. I found a _FIXME_ note in the underlying method, and I suspect that it is indeed not thread-safe.

yeganer: skip `calculate_chi_bf`

karandesai-96:
After #547 was merged, all the methods stopped throwing a segfault, except `move_packet_across_shell_boundary`. This is still troublesome.

yeganer:
I don't think it's `move_packet_accross_shell_boundary`.

Is the list in the first post of this PR complete? If not, can you please add ALL functions which are in cmontecarlo.c, including those you already wrote tests for in the previous PR.

This gives us a nice overview of all functions and we can categorize them. For the moment I think the tests are mostly done because most of the functions left aren't relevant at the moment.

yeganer:
#### Important
- [ ] `line_search`
- [ ] `reverse_binary_search`
- [X] `rpacket_doppler_factor`
- [X] `compute_distance2boundary`
- [X] `compute_distance2line`
- [X] `compute_distance2continuum`
- [ ] `macro_atom`
- [X] `move_packet`
- [X] `increment_j_blue_estimator`
- [ ] `move_packet_across_shell_boundary`
- [ ] `montecarlo_thomson_scatter`
- [X] `montecarlo_line_scatter` 
- [X] `move_packet` ( e7a7b4e )
- [X] `move_packet_across_shell_boundary` ( b5ec6bf )

### Difficult to test
- [ ] `montecarlo_one_packet`
- [ ] `montecarlo_one_packet_loop`
- [ ] `montecarlo_main_loop`
- [ ] `montecarlo_event_handler`

### Not yet relevant
- [ ] `montecarlo_continuum_event_handler`
- [X] `montecarlo_free_free_scatter` ( 4f4a1e8 )
- [X] `montecarlo_bound_free_scatter` ( 6e7be8c )
- [X] `bf_cross_section` ( 1f099af )
- [X] `calculate_chi_bf` ( 0de7cba )

tardis/montecarlo/struct.py
```diff
@@ -99,15 +99,6 @@ class StorageModel(Structure):
         ('virt_array_size', c_int64)
     ]
 
-
-class RKState(Structure):
-    _fields_ = [
-        ('key', POINTER(c_ulong)),
```
karandesai-96: This meant 4 bytes for POINTER in stack, while the whole array in heap.

tardis/montecarlo/struct.py
```diff
@@ -121,3 +112,15 @@ class RKState(Structure):
...
 CONTINUUM_OFF = 0
 CONTINUUM_ON = 1
+
+# `rk_state` specific macros.
+RK_STATE_LEN = 624
+
+
+class RKState(Structure):
+    _fields_ = [
+        ('key', c_ulong * RK_STATE_LEN),
```
karandesai-96: This means all 624 blocks sequentially in stack

wkerzendorf:
@karandesai-96 there are several checkboxes not checked. You need to tell us if this PR is finished and what you want to do with the rest of the textboxes.

karandesai-96:
@wkerzendorf These tests were classified in three different categories by @yeganer a few days back and I set my priorities accordingly. So here it goes:

1. **Not yet relevant**
    - These are kept out of scope of this PR, whichever are left. Though only test for `montecarlo_continuum_event_handler` remains in this section. What is  better thing to do- should I write it in this PR and add a `@pytest.mark.skipif(True...)` or tackle all of these when time arrives ? 

2. **Difficult to test**
    - I have been involved with these tests since quite a while now, and wish to switch over to some other part for a change. These four tests might consume time and delay this PR. So I am planning to open a new PR after doing one or two small tasks related to integration test. It would contain work for these tests.

3. **Important**
    - I have worked on this PR assuming the tests falling under this category are the topmost priority. All the tests are written, and they pass, except test for `macro_atom`.
    - `macro_atom` uses certain parameters of `StorageModel` for calculations, which are not set to default values in the fixture. For example: `line2macro_level_upper` is supposed to be an array - but I am unaware of its length, what decides its length ? What is the general range of numerical values which this parameter takes up ? There are a couple of them more.

I would need your opinion about the further work to do here. Should I list out the parameters which I need as a test case for `macro_atom` test so that we can form a suitable test case ? Rest all work in this PR except the ones mentioned above are complete from my side unless I do modifications based on the reviews ;)

karandesai-96:
Moreover, before merging we will make sure to remove **test_cmontecarlo.c** and **test_montecarlo.so** completely. These tests dirctly wrap around C methods and no longer rely on tests written in C.

wkerzendorf:
@karandesai-96 finish this PR in the sense that you document

1. what it tests
2. list what you skip
3. why you skip it
4. open an issue where you list what needs to be done

As it stands you are asking us to merge an unfinished PR. When you are done with something like this - the PR description should be cleaned up. @unoebauer's PRs are a very good example of how to do this. I should be able to go into a PR and figure out exactly what it does - what remains to be done, etc.

yeganer:
### Summary

- [ ] Every function in C has a `test_<func>` counterpart in python 

- [ ] Not yet implemented tests are marked `skip` with a descriptive reason

- [ ] Tests are grouped by 'importance', block comments separate the groups 

- [ ] Removal of all `test_*.c` files and the building of `test_cmontecarlo.so`

- [ ] New Issue about missing tests for `rpacket`

### Explanation

@karandesai-96 I had a look at the PR and I think it's mostly done. There is still some room for improvements.
For example I think it's a good idea to have a `test_<method_name>` for **every** function of the C-extension. For the "difficult to test" and "not important" categories they would be empty functions marked with `pytest.mark.skip` and a reason why this test should be skipped.

@wkerzendorf you can correct me if I'm wrong but in my opinion we should skip all functions listed as "not yet relevant", too. These functions are related to the continuum interactions which are developed by @chvogl. Running them for the current master doesn't seem necessary to me.

Finally I'd suggest you group the functions inside `test_cmontecarlo.py` by importance. This means all tests which are currently run should be at the top. Below them would be a block comment about the difficult to test methods and below that a block comment stating the next group isn't relevant because the implementation isn't final.

This PR aims to completely replace the tests for the C-extension. As such we should remove ALL files associated with testing the C part. I know there is `test_rpacket.c` which will be untested but there is only one relevant function (`rpacket_init`) and not testing that is safe for now. The removal of all files associated with the c-tests should be reflected by a new Issue about the newly untested methods.
You should remove the building of `test_cmontecarlo.so`, as it's not needed anymore.

karandesai-96:
Hi @yeganer I have done most of the work suggested. I have a doubt:

> Removal of all test_*.c files and the building of test_cmontecarlo.so

**test_rpacket.py** wraps around this shared object. Should I remove that as well ? Or should I apply `skip` marker to all methods inside that file ?

yeganer:
Remove test_rpacket.py
Last time I looked the file did nothing but testing the getters and setters. Since they are inlined we can't test them with ctypes. But they shouldn't break. If they break, then we'll know it because nothing works anymore.

karandesai-96:
- [X] Every function in C has a `test_<func>` counterpart in python
- [X] Not yet implemented tests are marked `skip` with a descriptive reason
- [X] Tests are grouped by 'importance', block comments separate the groups
- [X] Removal of all `test_*.c` files and the building of `test_cmontecarlo.so`
- [ ] New Issue about missing tests for `rpacket`

@yeganerI have completed the points mentioned above.

If the methods are `static inline` they aren't meant to be tested by python. Do you think opening an issue about that is a good idea ? I am anyways opening an issue for the requirement of **difficult to test** category.

karandesai-96:
@yeganer Letting out a gentle reminder. This PR is ready from my side.

wkerzendorf:
@tardis-sn/tardis-core what is going on here? From my side this looks good.

wkerzendorf:
@yeganer @unoebauer I'm signing off on this. Can one of you check and merge?

unoebauer:
Reviewing the PR now...

unoebauer:
Looks good - merging.

PR #545, opened 2016-04-21
added mailmap to compile the commits for each person

wkerzendorf:
@tardis-sn/tardis-core  can you guys have a quicklook and merge this - it will just allow us to compile commits from names and attribute them to a single person.


yeganer:
I don't know how this works exactly but Aoife seems to be in there 2 times with exactly the same entry. Is this relevant?

Issue #239, opened 2015-03-06
Update Configuration Validator


mavidser:
The [Configuration file validator](https://github.com/tardis-sn/tardis/blob/master/docs/configuration/config_validator.rst) seems a bit outdated.

> For every specified container item, the code expects the declaration of all sub items. The keywords for this are “_“ + “name of the container item”. 

But the current config file has elements like:

```
property_type: container-declaration
containers: ['branch85_w7','exponential','power_law', 'uniform']
_uniform: ['value']
_branch85_w7: []
+branch85_w7: ['w7_time_0', 'w7_rho_0', 'w7_v_0']m”
_power_law: ['time_0', 'rho_0', 'v_0', 'exponent']
_exponential: ['time_0', 'rho_0', 'v_0']
```

I'm not sure what does  “+“ + “name of the container item” specifies as in `+branch85_w7`.

It looks like it specifies non-mandatory items, but that seems redundant due to a mandatory property in every item, so I may be wrong.

wkerzendorf:
@mavidser I have to look into it. It might outdated, we're considering switching to a third party YAML schema validator but have not found a good one (maybe another GSoC project)

mavidser:
Okay. 

Meanwhile, I think I should consider `_container` as mandatory and `+container` as optional, for the pull requests for the web-app.

wkerzendorf:
I think that's probably right anyways. Sorry, I just have to deal with some
other things currently. good luck and let me know how it goes.

unoebauer:
I am closing this one, since it should have been addressed by @ftsamis work. If not, please re-open, @ftsamis.

PR #554, opened 2016-05-04
Update config_validator.rst

ftsamis:
This one was really badly formatted, outdated and incomplete.

I mostly made these changes so I could use this as a sensible reference when redesigning the configuration system but as a side-effect I also learned the basics of RST syntax which will save me some time later when writing the documentation for the new configuration system.

If merged, this should also close https://github.com/tardis-sn/tardis/issues/239

### DIFF
--- a/docs/configuration/config_validator.rst
+++ b/docs/configuration/config_validator.rst
@@ -2,75 +2,118 @@
 Configuration Validator
 ***********************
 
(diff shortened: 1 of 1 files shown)


ftsamis:
If you want to compare the complete result of my changes to the old page:
With my changes: https://github.com/ftsamis/tardis/blob/patch-1/docs/configuration/config_validator.rst
The old page: https://github.com/tardis-sn/tardis/blob/2804869ec183652ed7a9bc8cd545241f7a1efddd/docs/configuration/config_validator.rst

wkerzendorf:
@unoebauer this is probably useful to insert into the new scheme.

PR #555, opened 2016-05-04
Use pytest's monkeypatch.chdir to change the working directory in the config_reader tests

ftsamis:
Following @yeganer's suggestion at https://github.com/tardis-sn/tardis/pull/543#issuecomment-216568575 and replacing the custom function in ``test_config_reader``, for temporarily changing the working directory in a test, with pytest's ``monkeypatch.chdir``.

### DIFF
--- a/tardis/io/tests/test_config_reader.py
+++ b/tardis/io/tests/test_config_reader.py
@@ -358,16 +358,14 @@ def test_ascii_reader_exponential_law():
         assert_almost_equal(structure['mean_densities'][i].value,mdens)
         assert structure['mean_densities'][i].unit ==  u.Unit(expected_unit)
 
-@pytest.mark.parametrize("path", [os.path.curdir, os.path.pardir, os.path.abspath(data_path(''))])
-def test_absolute_relative_config_paths(monkeypatch, path):
+@pytest.mark.parametrize("cwd", [os.path.curdir, os.path.pardir])
+def test_absolute_relative_config_paths(monkeypatch, cwd):
     filename = 'tardis_configv1_ascii_density_abund.yml'
-    if os.path.isabs(path):
-        filepath = os.path.join(path, filename)
-    else:
-        monkeypatch.chdir(data_path(path))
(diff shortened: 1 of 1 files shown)


yeganer:
I think you could go even so far as to have one test function and parametrize it. I think you wouldn't need a class in that case.

wkerzendorf:
@yeganer @ftsamis are we done here?

ftsamis:
@yeganer Checked parametrize, and I think it would probably make sense to only use it for 2 of the 3 functions, since the one with the absolute path has slightly different behaviour (no `monkeypatch`).

If you don't insist on using parametrize, I'd say we can merge this @wkerzendorf.

wkerzendorf:
@yeganer @ftsamis looks good - signing off.

yeganer:
Actually I'd use parametrize because the functionality is the same in all 3 functions.
```
@pytest.parametrize(['.', '..', os.path.abs(os.cwd())])
```
something like this.
That would make it very obvious what's happening.
@wkerzendorf If you don't share my opinion, you can merge. The rest looks good.

wkerzendorf:
@ftsamis I agree with @yeganer. It's a quick change and will reduce the code. Will merge once this is in.

wkerzendorf:
@ftsamis can you fix this?

ftsamis:
@wkerzendorf I'm now working on it.
@yeganer The first two functions need to change working directory, and to have a relative path. The third doesn't need to change working directory but needs an absolute path. That's the slightly different behaviour I meant.

Anyhow, I'll use parametrize.

wkerzendorf:
@yeganer what do you think?

yeganer:
I think, I'm a bit nitpicking here but you could have the cwd as parameter for your function and then read the config two times, once with a relative path to the destination and once with an absolute path. That should cover everything and you don't have to do if/else in the test.

ftsamis:
That sounds indeed better. Done.

yeganer:
Thanks, that looks great.
@tardis-sn/tardis-core Ready to merge from my side.

PR #557, opened 2016-05-05
Enable skipping slow tests.

karandesai-96:
Certain tests can be skipped in normal run of tests and would be executed only on the provision of `--run-slow` parameter. This means:
* `./setup.py test` will skip all slow marked tests inherently.
* `./setup.py test --slow` will include all the tests while running.

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -32,6 +32,10 @@ def pytest_addoption(parser):
     parser.addoption("--atomic-dataset", dest='atomic-dataset', default=None,
                      help="filename for atomic dataset")
 
+    parser.addoption("--with-slow", action="store_true",
+                     help="include running slow tests during run")
+
+
 def pytest_report_header(config):
 
     stdoutencoding = getattr(sys.stdout, 'encoding') or 'ascii'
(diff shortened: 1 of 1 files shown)


karandesai-96:
@wkerzendorf Just as we decided, a separate PR which _just_ adds the skip marker in **conftest.py**. I will be basing my next PR to add the integration test for stratified W7 setup on this branch itself. I will be doing the rebase to **master** once this is merged.

tardis/conftest.py
```diff
@@ -32,6 +32,10 @@ def pytest_addoption(parser):
     parser.addoption("--atomic-dataset", dest='atomic-dataset', default=None,
                      help="filename for atomic dataset")
 
+    parser.addoption("--with-slow", action="store_true",
```
wkerzendorf: just change this to `--slow`. otherwise it looks good.

karandesai-96: - [X] Modified

wkerzendorf:
@yeganer - I'm fine to merge this one.

unoebauer:
Easy enough - looks good - merging

PR #508, opened 2016-03-06
GSoC '16 Testing Project, prelim task: skip slow tests (not to be merged)

karandesai-96:
* This PR is for discussions related to my preliminary task for **Tardis Testing** project ([TEP001](https://github.com/tardis-sn/tep/blob/master/TEP001_extensive_test_suite.rst)).

* The task was to mark the full Tardis test as **slow** and skip it automatically during normal run of tests, for example at Travis CI, and include slow tests only while running on a **dedicated server**.

* The full Tardis test is marked as slow and will only run when `--run-slow` parameter is provided while running the tests. In a nutshell:
    * `./setup.py test` will skip full Tardis test inherently, no special parameters to provide.
    * `./setup.py test --run-slow` will include Tardis slow test while running.

* A brief description is provided as a note in the _diffs_ below, please refer to it.

PR #558, opened 2016-05-09
Integration Test class for Stratified W7 setup.

karandesai-96:
This PR focuses on addition and proper functioning of an integration test class for Stratified W7 setup. This test is classified as a **slow test** due to its time taken for a run on single core.

Important components added:
- [X] **Configuration file**: `tardis/tests/tests_slow/w7/config_w7.yml`
- [X] **Abundances profile**: `tardis/tests/tests_slow/w7/abundances_w7.dat`
- [X] **Densities profile**: `tardis/tests/tests_slow/w7/densities_w7.dat`
- [X] **Integration test class**: `tardis/tests/tests_slow/test_w7.py`

Other relevant documentation is added in Integration test itself.


tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,155 @@
...
+        * last_line_interaction_out_id   |     * luminosity_inner
+        * last_line_interaction_in_id    |     * montecarlo_luminosity
+        * j_estimators                   |     * montecarlo_virtual_luminousity
+        * j_blue_estimators              |     * time_of_simulation
+        * last_line_interaction_shell_id |     * montecarlo_nu
+        * nubar_estimators               |     * last_line_interaction_angstrom
+        * ws                             |     * j_blues_norm_factor
+
+    Also assumed `kurucz_cd23_chianti_H_He.h5` file exists in `tmp` directory.
+    """
```
karandesai-96: Docstring for explanation related to work in progress.

wkerzendorf:
you need to rebase this commit on the other one.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,155 @@
...
+from tardis.io.config_reader import Configuration
+
+
+def data_path(fname):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), "w7", fname)
+
+
+@pytest.mark.skipif(not pytest.config.getoption("--with-slow"),
+                    reason="slow tests can only be run using --with-slow")
+class TestW7(object):
```
wkerzendorf: This can likely be subclassed.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,155 @@
...
+from tardis.simulation.base import Simulation
+from tardis.model import Radial1DModel
+from tardis.io.config_reader import Configuration
+
+
+def data_path(fname):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), "w7", fname)
+
+
+@pytest.mark.skipif(not pytest.config.getoption("--with-slow"),
```
wkerzendorf: wrong decorator

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # We now do a run with prepared config and get radial1d model.
+        self.obtained_radial1d_model = Radial1DModel(tardis_config)
+        simulation = Simulation(tardis_config)
+        simulation.legacy_run_simulation(self.obtained_w7_radial1d_model)
+
+        # The baseline data against which assertions are to be made is ingested
+        # from already available compressed binaries (.npz). These will return
+        # dictionaries of numpy.ndarrays for performing assertions.
+        self.expected_ndarrays = np.load(data_path("baseline_ndarrays.npz"))
```
wkerzendorf: they should come from outside of the tardis directory and one should specify the baseline configuration directory with a command line option.

karandesai-96:
@wkerzendorf @yeganer @unoebauer I am experiencing an **IOError** everytime now, and this test doesn't work. The problem is that **tests_slow** directoy and its contents are not included the the directory tree created after the build. This problem was not faced earlier while I implemented it pre-GSoC in #508. Infact, the test was run properly and I have reported it in one of the comments. Is there any recent change in build pattern which I am unaware of ?

I had experimentally let this slow test run on Travis of my fork, whether to make sure that my loca copy itself is not faulty. [Here is the build job about that.](https://travis-ci.org/karandesai-96/tardis/jobs/128748508) I am facing the same error locally.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,163 @@
...
+        * last_interaction_type          |     * t_rads
+        * last_line_interaction_out_id   |     * luminosity_inner
+        * last_line_interaction_in_id    |     * montecarlo_luminosity
+        * j_estimators                   |     * montecarlo_virtual_luminousity
+        * j_blue_estimators              |     * time_of_simulation
+        * last_line_interaction_shell_id |     * montecarlo_nu
+        * nubar_estimators               |     * last_line_interaction_angstrom
+        * ws                             |     * j_blues_norm_factor
+
+    Also assumed `kurucz_cd23_chianti_H_He.h5` file exists in `tmp` directory.
```
wkerzendorf: why?

karandesai-96: It needs an atomic dataset. Once the test gets running, I will update this. Atomic dataset will be provided by user through command line (before this PR gets merged). It is actually too long to type on command line :sweat_smile:

wkerzendorf: you should use the existing infrastructure to pass the atomic datafile

wkerzendorf:
your errors are related to not including these files in a `setup_package.py` - have a look throughout TARDIS for this file and it shows you how you need to include your files.

tardis/tests/tests_slow/test_w7.py
karandesai-96: Sure, I will do it same as **test_tardis_full.py**

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,163 @@
...
+
+def data_path(fname):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), "w7", fname)
+
+
+@pytest.mark.skipif(not pytest.config.getoption("--slow"),
+                    reason="slow tests can only be run using --slow")
+@pytest.mark.skipif(not pytest.config.getvalue("baseline-data"),
+                    reason="--baseline-data was not specified")
+class TestW7:
```
wkerzendorf: inherit from object to get new-style classes.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        self.config_file = data_path("config_w7.yml")
+        self.abundances = data_path("abundancies_w7.dat")
+        self.densities = data_path("densities_w7.dat")
+
+        # First we check whether atom data file exists at desired path.
+        self.atom_data_filename = os.path.expanduser(os.path.expandvars(
+                                    pytest.config.getvalue('atomic-dataset')))
+        assert os.path.exists(self.atom_data_filename), \
+            "%s atom data file does not exist" % self.atom_data_filename
```
wkerzendorf: use new style formatting. In this case `"{0} atom data file does not exist".format(self.atom_data_filename)`

wkerzendorf:
@yeganer I'm signing off on this.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,165 @@
...
+        This method does initial setup of creating configuration and performing
+        a single run of integration test.
+        """
+        self.config_file = data_path("config_w7.yml")
+        self.abundances = data_path("abundancies_w7.dat")
+        self.densities = data_path("densities_w7.dat")
+
+        # First we check whether atom data file exists at desired path.
+        self.atom_data_filename = os.path.expanduser(os.path.expandvars(
+                                    pytest.config.getvalue('atomic-dataset')))
```
unoebauer: We may want to add a check that the atomic database is the same as used for the baseline data. Otherwise, this test most likely won't pass. Not sure if we have to include this check in the current PR, but something to keep in mind, though. Thoughts, @wkerzendorf, @karandesai-96, @yeganer?

unoebauer:
Apart from my inline comment concerning atomic data integrity, I am very happy with the PR. As soon as we've come to a decision concerning this point, we can merge.

tardis/tests/tests_slow/test_w7.py
karandesai-96: This didn't even come to my mind ! Using a different atomic database would generate different results. So far there is **kurucz_cd23_chianti_H_He.h5** everywhere. If this check is added when need arrives, then the purpose of adding that additional check would be more clear and could be specified in PR description more effectively. Waiting for others' opinions :smile:

wkerzendorf: can you just check the uuid of the atomic data set - that would be good.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,175 @@
...
+        assert os.path.exists(self.atom_data_filename), \
+            "{0} atom data file does not exist".format(self.atom_data_filename)
+
+        # We now check that the baseline data for slow tests was obtained using
+        # the same atomic dataset as the one used in current test run.
+        # TODO: This is a quick workaround, generalize it.
+
+        kurucz_data_file_uuid = "5ca3035ca8b311e3bb684437e69d75d7"
```
unoebauer: Not too happy about this quick workaround. You can directly access the UUID and the MD5 checksum of the atomic data file via ``model.atom_data.uuid1`` and  ``model.atom_data.md5``. I suggest you add these to your baseline data and then check them against the one of test.

karandesai-96: Oh, so after the run, I obtain the `radial1d_model` and that would enable comparison, right ? But wouldn't it be a bit time expensive ? Waiting for twenty minutes or more and finally seeing that the test failed because atom data files weren't same ? Is it fine ?

karandesai-96: > I suggest you add these to your baseline data and then check them against the one of the test run.

Add them in the sense, like a json or yml file, right ?
```yml
atom_data:
    name: kurucz_cd23_chianti_H_He.h5
    uuid1: <uuid1>
    md5: <md5>
```
I think this would conflict with `--atomic-dataset` argument. In that case we can give the yml file a priority and **NOT** skip this test if `--atomic-dataset` is explicitly not provided. Does it sound like a plan ?
@wkerzendorf @yeganer Your thoughts ?

wkerzendorf: @unoebauer I suggested this quick workaround. I agree that in the long run we want to have this done properly, but my suggestion is to build parts of the framework and then refine. Would that work for you?

unoebauer: @wkerzendorf, @karandesai-96 - sure thing. Sorry for holding up the process. Just wanted to flag this again.

unoebauer:
Happy to merge if nobody objects...

tardis/tests/tests_slow/test_w7.py
karandesai-96: Okay if we not do apply this change, though please do let me know about this idea of putting in a yml file like this. If it is good enough, I  would keep it in my mind and apply it immediately when required :wink:

yeganer: You don't have to wait for the run to be completed to access AtomData.
Afaik Atomdata is already an object inside tardis_config(the parsed, validated one) so doing something like `tardis_config.atom_data.uuid` should work.

karandesai-96:
```
config_yaml = yaml.load(open(self.config_file))
config_yaml['atom_data'] = self.atom_data_filename
```
We open the config file and did an overwrite with the new path of atom data filename (the one which we provided from command line). It doesn't seem like an object. Did I miss anything ?

tardis/tests/tests_slow/test_w7.py
```diff
@@ -0,0 +1,175 @@
...
+        # densities and abundances profile files as desired. We form dictionary
+        # from the config file and override those parameters by putting file
+        # paths of these three files at proper places.
+        config_yaml = yaml.load(open(self.config_file))
+        config_yaml['atom_data'] = self.atom_data_filename
+        config_yaml['model']['abundances']['filename'] = self.abundances
+        config_yaml['model']['structure']['filename'] = self.densities
+
+        # The config hence obtained will be having appropriate file paths.
+        tardis_config = Configuration.from_config_dict(config_yaml)
```
yeganer: That's where the config get's validated and where atom_data becomes an object.

wkerzendorf:
@yeganer if you do not object - I'm merging now.

yeganer:
@wkerzendorf I'm objecting fornow

tardis/tests/tests_slow/test_w7.py
karandesai-96: Oh sorry, it was a misunderstanding, @yeganer I did as you suggested and it works fine. :)

tardis/tests/tests_slow/test_w7.py
```diff
@@ -78,6 +69,12 @@ def setup(self):
         # The config hence obtained will be having appropriate file paths.
         tardis_config = Configuration.from_config_dict(config_yaml)
 
+        # We now check that the baseline data for slow tests was obtained using
+        # the same atomic dataset as the one used in current test run.
+        with h5py.File(self.atom_data_filename, 'r') as slow_test_atom_data_file:
+            slow_test_data_file_uuid1 = slow_test_atom_data_file.attrs['uuid1']
+        assert slow_test_data_file_uuid1 == tardis_config.atom_data.uuid1
```
yeganer: this is translates to  `assert a == a`
compare tardis_config with your hardcoded uuid

tardis/tests/tests_slow/test_w7.py
karandesai-96: This strategy sounds better. I am trying it.

tardis/tests/tests_slow/test_w7.py
wkerzendorf: and - that is sensible. if someone uses a different atomic dataset with the same name it will fail.

yeganer:
Looks good now.

Issue #534, opened 2016-03-31
Inconsistency in representation of "spectrum.frequency"?


yeganer:
The config reader initializes the spectrum and and creates `N` bins. The result is an array with `N+1` bin edges. Later, when the spectrum gets constructed these `N+1` edges are cut to `N` by removing the last value. Shouldn't it convert the edges to bin centers?

Later the cropped version is used to trasform `luminosity_density_nu` into `luminosity_density_lambda`. Because we are already dealing with bins instead of single packets we are doing a small error compared to calculating the `luminosity_density_lambda` directly from the packets. But we are additionally using the left bin edge for the transformation instead of the bin center.

During my tests I found a relative difference of 6% between `luminosity_density_lambda` and my manually calculated `luminosity_density`.

yeganer:
I made a plot showing the relative error we are doing compared to the spectrum of `tardis_example`.
In this plot the error is magnified by 1e42. This results in a maximum error of `0.0039` which is far smaller than the 6% I got earlier. I don't know what went wrong there.
![issue_534_2](https://cloud.githubusercontent.com/assets/6164110/14201797/7903071c-f7f4-11e5-82a3-c1d76d22807c.png)

unoebauer:
@yeganer - just saw this issue. The plot above looks very severe! I would be surprised, however, that we missed such a big effect. Update?

yeganer:
@unoebauer The effect is just a linear systematic error.
It's not that severe. In my opinion this should be fixed (use bin edges for every calculation and construct the bin_centers for plotting only when needed)
A fix wouldn't be difficult (one line of code) but it would change the output spectrum, so some tests would need new reference data etc.

unoebauer:
@yeganer, @wkerzendorf, @ssim: then I propose to wait with this until the slow testing infrastructure of @karandesai-96 is up and running and then fix it.

unoebauer:
Closing due to inactivity and lack of relevance.

PR #559, opened 2016-05-11
Fix Issue #534

unoebauer:
This small PR addresses issue #534. TARDISSpectrum frequency now returns frequency bin midpoints instead of left edges.

Travis will fail, since the real spectrum is now slightly different (on the order of 1e-4 - 1e-2). We should wait until @karandesai-96's project has progressed to a point where the process of exchanging baseline data is facilitated.

### DIFF
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -372,8 +372,7 @@ class TARDISSpectrum(object):
 
     @property
     def frequency(self):
-        return self._frequency[:-1]
-
+        return 0.5 * (self._frequency[1:] + self._frequency[:-1])
 
     @property
     def flux_nu(self):


unoebauer:
The following plot compares the emergent spectra (in terms of ``luminosity_density_lambda``) between the current master, 356f2dcfe4bffa52c5d43d57fe01c19a267d27a0, and this PR, cbd24016739ccbdc51e1d24c4fa608350e896102. Also, the absolute value of the relative difference between the two spectra is shown:

spectral_bins_fix

yeganer:
You really know, how to make these plots look nice :)

I agree, we should wait for @karandesai-96 's project to have progressed farther.

wkerzendorf:
we will merge once testing framework is operational.

wkerzendorf:
@unoebauer can we close this? I think what we have is not ideal but it works.

unoebauer:
Sure thing.

PR #561, opened 2016-05-13
TestW7: Assertions for spectrum quantities, fixture for handling baseline data.

karandesai-96:
Radial1D model's spectrum contains these quantities which weren't tested before this PR:
* `luminosity_density_nu`
* `delta_frequency`
* `wavelength`
* `luminosity_density_lambda`

-----
Work included in this PR can be followed by this checklist:
#### Assertions for spectrum quantities
- [X] Existing `test_spectrum` method had an assertion for `luminosity_inner`. This assertion is shifted to a separate method named `test_luminosity_inner`.
- [X] Add assertions for the above mentioned members in `test_spectrum` method.
- [X] Relevant docstring for spectrum assertions and `spectrum.npz`.

#### Fixture for baseline data
- [X] Fixture named `baseline` has been created for collecting all baseline data in the form of dictionary.
- [X] Unpacked numpy arrays of astropy quantities are associated with their CGS units in this fixture.
- [X] Use `astropy.tests.helper`'s `assert_quantities_allclose` wherever required.
- [X] Relocate docstrings about baseline data to from `TestW7` to this fixture.

#### Miscellaneous
- [X] Modify pytest skip marker.
- [X] Convert `data_path` to a fixture.
- [X] Use fixtures for atomic data and slow test data from top level **conftest.py**.
- [X] Remove `--slow` option as `--slow-test-data` is sufficient and necessary.

By this, `TestW7` and its methods only contain statements for assertions all preprocessing of baseline data has been shifted to `baseline` fixture.


karandesai-96:
This is a test run of `TestW7` class, which reduced packets (2e3).
image
Everything is working fine for 2e3 packets and I am safely assuming that it would be more or less same for actual W7 setup (2e5 packets).

unoebauer:
@karandesai-96: Looks good! Just one remark - eventually, we will do the same thing also for the virtual packet spectrum. However, as @wkerzendorf stressed yesterday, adding all the quantities to be eventually checked from the beginning is not a top priority. Rather it is more important to get the testing infrastructure up and running.

I'm signing off on this PR - if @yeganer and/or @wkerzendorf are happy as well, we can merge

unoebauer:
@karandesai-96 one more thing: It would be good if you could include the plotting capability (showing the comparison between the current and reference spectra and their relative deviations).

karandesai-96:
@unoebauer I have it ready on a separate branch. Since this PR contains work specific to "Assertions", I decided to keep it isolated from another _to-be-done_ PR containing lines of `matplotlib` involved code. I have minimal working code as well as description ready. I would rebase my plotting functionality branch to **master** after this one gets merged and would open a PR immediately.

yeganer:
@karandesai-96 In the end we want to have one file with reference data per setup. I'd suggest to use `pd.HDFStore` for that purpose. Since you are in the process of expanding the current framework I wanted to mention this early so you don't spent too much time with the current numpy data saving.

As for the PR. I think it would be nice to include the capability to plot data in this PR as this feature seems almost done, too.

karandesai-96:
@yeganer
* Point noted about `pd.HDFStore` When we finally decide of adding **all** the assertions (that would be in one separate PR I guess), so at that Time I would reference this conversation there and add a commit which replaces all `np.load` and `np.save` with _HDF5 I/O_ methods.

* Adding the plotting functionality, sure I will `cherry-pick` the commits from that branch to this one.

yeganer:
The implementation of the baseline fixture looks good and handling the units there is a good approach, too.

However I suggest using `units.Quantity(value, unit)` to initialize the quantities as it looks more organized to me and it's very easy to understand what's happening there.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -39,11 +37,66 @@ class TestW7(object):
...
+            os.path.expandvars(request.config.getvalue('slow-test-data'))),
+            "w7")
+
+    ndarrays = dict(np.load(os.path.join(slow_test_data_dir, "ndarrays.npz")))
+    quantities = dict(np.load(os.path.join(slow_test_data_dir, "quantities.npz")))
+    spectrum = dict(np.load(os.path.join(slow_test_data_dir, "spectrum.npz")))
+
+    # Associate CGS units to ndarrays of baseline quantities.
+    quantities['j_blues_norm_factor'] = \
+        quantities['j_blues_norm_factor'] * u.Unit('1 / (cm2 s)')
```
yeganer: As I wrote in the other comment, use `units.Quantity(quantities['j_blues_norm_factor'], '1 / (cm2 s)')`.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -39,11 +37,66 @@ class TestW7(object):
...
+
+    # Associate CGS units to ndarrays of baseline quantities (spectrum).
+    spectrum['luminosity_density_nu'] = spectrum['luminosity_density_nu'] * u.Unit('erg')
+    spectrum['delta_frequency'] = spectrum['delta_frequency'] * u.Unit('Hz')
+    spectrum['wavelength'] = spectrum['wavelength'] * u.Unit('Angstrom')
+
+    spectrum['luminosity_density_lambda'] = \
+            spectrum['luminosity_density_lambda'] * u.Unit('erg / (Angstrom s)')
+
+    return dict(ndarrays.items() + quantities.items() + spectrum.items())
```
yeganer: I'm not sure this a good way to combine dictionaries. I think using `dict.update` would be the safer approach.

karandesai-96: Yes, I was confused between those two.
- [X] Done.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -55,32 +55,35 @@ def baseline(request):
 
     # Associate CGS units to ndarrays of baseline quantities.
     quantities['j_blues_norm_factor'] = \
```
yeganer: It's just a cosmetic thing but don't use `\` for line breaks. If it's a huge formula wrap it in braces, otherwise split after the function opening brace.
IE:
```
    quantities['last_line_interaction_angstrom'] = u.Quantity(
        quantities['last_line_interaction_angstrom'], 'Angstrom')
```

tardis/tests/tests_slow/test_w7.py
```diff
@@ -39,18 +37,75 @@ class TestW7(object):
...
+    spectrum['luminosity_density_nu'] = (
+        u.Quantity(spectrum['luminosity_density_nu'], 'erg'))
+
+    spectrum['delta_frequency'] = u.Quantity(spectrum['delta_frequency'], 'Hz')
+    spectrum['wavelength'] = u.Quantity(spectrum['wavelength'], 'Angstrom')
+
+    spectrum['luminosity_density_lambda'] = (
+        u.Quantity(spectrum['luminosity_density_lambda'], 'erg / (Angstrom s)'))
+
+    ndarrays.update(quantities, **spectrum)
```
yeganer: This looks unintuitive. Just update the dictionary two times.
Or what would be even better is you create an empty dictionary and add the elements manually.
```
data = dict()
data['t_rads'] = u.Quantity(quantities['t_rads'], 'K')
```

tardis/tests/tests_slow/test_w7.py
```diff
@@ -39,18 +37,75 @@ class TestW7(object):
...
+    return ndarrays
+
+
+@pytest.mark.skipif(not pytest.config.getoption("--slow"),
+                    reason="slow tests can only be run using --slow")
+@pytest.mark.skipif(not pytest.config.getvalue("slow-test-data"),
+                    reason="--slow-test-data was not specified")
+@pytest.mark.skipif(not pytest.config.getvalue("atomic-dataset"),
+                    reason="--atomic-dataset was not specified")
```
yeganer: I suggest you create a mark that combines these.
In the end it should look like this:
```
@runslow
class TestW7(object):
```

yeganer:
This PR starts to look really good. Keep on the good work!

tardis/tests/tests_slow/test_w7.py
karandesai-96: How about the method I suggested back in chat room ? Writing all kwargs ?

yeganer: I think doing it this way is the easiest to understand and most robust solution

tardis/tests/tests_slow/test_w7.py
```diff
@@ -12,21 +12,19 @@
 from tardis.io.config_reader import Configuration
 
 
-def data_path(fname):
-    return os.path.join(os.path.dirname(os.path.realpath(__file__)), "w7", fname)
+@pytest.fixture(scope="module")
```
wkerzendorf: that can go into the `conftest.py`

tardis/tests/tests_slow/test_w7.py
```diff
@@ -12,21 +12,19 @@
...
 
 
-@pytest.mark.skipif(not pytest.config.getoption("--slow"),
-                    reason="slow tests can only be run using --slow")
-@pytest.mark.skipif(not pytest.config.getvalue("slow-test-data"),
-                    reason="--slow-test-data was not specified")
-@pytest.mark.skipif(not pytest.config.getvalue("atomic-dataset"),
-                    reason="--atomic-dataset was not specified")
-class TestW7(object):
+@pytest.fixture(scope="module")
```
wkerzendorf: that as well into conftest

tardis/tests/tests_slow/test_w7.py
```diff
@@ -1,81 +1,15 @@
 import os
```
yeganer: move this file into tests_slow/w7/

karandesai-96: I think this will eventually become a common class for running multiple setups, and putting it in w7 dir might not suite it. When more setups join in, we can surely change the filename and class name to something more generalized. Thoughts ?

tardis/tests/tests_slow/test_w7.py
```diff
@@ -67,8 +35,12 @@ def setup(self):
         config_yaml['model']['abundances']['filename'] = self.abundances
         config_yaml['model']['structure']['filename'] = self.densities
 
+        # First we check whether atom data file exists at desired path.
+        assert os.path.exists(atomic_data_fname), (
```
yeganer: You are using the fixture which already does this check. I'd remove it.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -84,93 +56,86 @@ def setup(self):
...
-                luminosity_inner,
+    def test_luminosity_inner(self):
+        assert_quantity_allclose(
+                self.baseline['luminosity_inner'],
                 self.obtained_radial1d_model.luminosity_inner)
 
-    def test_montecarlo_properties(self):
-        montecarlo_luminosity = self.expected_quantities['montecarlo_luminosity']
-        montecarlo_luminosity = montecarlo_luminosity * u.Unit('erg / s')
+    def test_spectrum(self):
```
yeganer: We should test the virtual spectrum, too.

tardis/tests/tests_slow/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+        return os.path.expandvars(os.path.expanduser(slow_tests_datadir))
+
+
+@pytest.fixture(scope="session")
+def data_path():
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), "w7")
+
+
+@pytest.fixture(scope="session")
+def baseline(request, slow_tests_datadir):
```
yeganer: I don't like the naming of this fixture but if I'm the only one, don't bother changing it.
Something like `reference_data` would be far more descriptive in my opinion.

Same goes for `self.baseline`. Either call it `self.data` or `self.reference` or any combination of the two.

tardis/tests/tests_slow/test_w7.py
unoebauer: Definitely - should it be part of this PR already?

wkerzendorf: nope - I think we build up the structure first and then think about specific plots and tests.

wkerzendorf:
@yeganer I think we are ready?

unoebauer:
Looking very good - happy to merge.

yeganer:
@wkerzendorf Yes, I think so, too.

Issue #560, opened 2016-05-12
Segfault during tardis cmontecarlo run


yeganer:
This is what appears to cause the issue.

```
[26] last_line = 1822360143409613.250000
[26] next_line = 1822353496842714.500000
[26] ERROR: Comoving nu less than nu_line!
[26] comov_nu = -58783631258883757404258453611703772493428311659382084935097109061384145004566790513369500832759808.000000
[26] nu_line = 1822355712359626.750000
[26] (comov_nu - nu_line) / nu_line = -32256946797049520942182563511351622606622620138438708712185411031008847627110318080.000000
[26] r = 999999999999999967336168804116691273849533185806555472917961779471295845921727862608739868455469056.000000
[26] mu = -0.522198
[26] nu = -3489787318773531.500000
[26] doppler_factor = 16844473857376206614733575835140392265551777909465454558809109268882602250184163328.000000
[26] cur_zone_id = 1
```

I'll update later as I gather more information.
### SETUP:

```
dalek         1ac519cbb9f91941fbc4d51165e1942d9afc1e95
tardis        3370d8d67a9abf4102d9f9e1fb7fbcff8bcede18
```

`run_failed.py` is a simple script that reproduces the problem. You have to run with OMP_NUM_THREADS=8. The way I use it, I'm exporting that environment variable before running the file (nthreads = 0 in config allows for dynamically changing the number of threads)
[packet.zip](https://github.com/tardis-sn/tardis/files/261436/packet.zip)

unoebauer:
@yeganer A bit of context would be good - when does this segfault occur, which setup is used, is there anything particular about the run?

unoebauer:
@yeganer: Before going anywhere further, you should check that ``v_inner`` < ``v_outer``. @chvogl cautioned that this is in principle possible...

yeganer:
I can verify that but I think it's very unlikely that this is causing the
problem. I'm setting v_inner in my config an the rest should be calculated.
As the crash happens during the spectral run and v_inner isn't changed
during the iterations, I think we have an other issue here.

Ulrich Noebauer <notifications@github.com> schrieb am Do., 12. Mai 2016
12:05:

yeganer:
It looks to me like the packet in question is a virtual packet which was moved to miss distance.

unoebauer:
@yeganer - still missing the basic background:
* problem setup
* yaml file

yeganer:
[...]

PR #562, opened 2016-05-13
Fix #552 #560

### LINKED ISSUE #552: distance2boundary: Result is not checked for NaN
During the calculation of the distance to boundary it is assumed that the packet's radius is between r_inner and r_outer. Because of rounding errors this is not always true and there exists a small range of parameters (radius and mu) for which the discriminant gets negative (the packed doesn't hit the boundary). In that case the distance is NaN.

Possible ways to fix this:
- When crossing the boundary explicitly set the radius of the packet
- If dboundary is NaN, set it to a constant maximum (2 \* r_outer for example)
- Return some TARDIS_ERROR and discard packet

yeganer:
This aims to fix Issue #552 and #560.

The code from move_packet is mirrored in move_packet_across_boundary.  But instead of calculating the new position it's known value is used.
The current implementation relies on rpacket.d_boundary being set only for the calculation of the estimators.


tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -234,6 +248,7 @@ compute_distance2line (rpacket_t * packet, const storage_model_t * storage)
           fprintf (stderr, "nu = %f\n", nu);
           fprintf (stderr, "doppler_factor = %f\n", doppler_factor);
           fprintf (stderr, "cur_zone_id = %" PRIi64 "\n", cur_zone_id);
+          rpacket_set_d_line (packet, NAN);
```
unoebauer: Is there a reason why we set d_line to NAN and not any or all of the other distances?

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -489,7 +504,37 @@ void
...
+          comov_energy * distance;
+#ifdef WITHOPENMP
+#pragma omp atomic
+#endif
+      storage->nubars[rpacket_get_current_shell_id (packet)] +=
+          comov_energy * distance * comov_nu;
+    }
+  // done moving packet
+
+  // reset tau event
```
unoebauer: Why is this necessary, why not keep move_packet here? **DRY**

tardis/montecarlo/src/cmontecarlo.c
yeganer: My reasoning was, that since the calculation of `distance2line` fails we set the corresponding distance to NaN. Setting `d_boundary` in a function that calculates `d_line` seems counter intuitive to me.

The proper 'solution' would be to check the error of the distance calculations and terminate if something goes wrong.

unoebauer:
Quite a few changes for a simple fix, but fine. However, I have a few remarks (see inline comments). In particular, I am not happy about copying the contents of move_packet into the cross_boundary function (Don't repeat yourself).

I'd suggest one of the following:
* leave move_packet in there and just set distance explicitly afterwards; this will introduce a slight inconsistency between the distance to boundary used in the estimator increments and the position of the packet; however, **if** the deviations between the shell interface location and the packet location are on the level of machine precision, these discrepancies should be of the same order
* modularize the move packet routine - separate actual packet propagation from estimator increments, distance recalculation, etc

I'd favour the latter approach...

unoebauer:
Following up on the thoughts from above, I wonder whether it may be prudent to introduce a sanity check here. In particular ensuring that the relative difference between the packet location determined traditionally (move packet) and the interface location is not larger than, say 1e-10. This way, we would make sure that we only do this explicit boundary fix for cases in which machine precision introduces discrepancies and not in cases where some other error introduces differences. Thoughts @ssim, @wkerzendorf, @chvogl, @yeganer?

tardis/montecarlo/src/cmontecarlo.c
yeganer: I asked myself the same question and came to the conclusion that `move_packet` and `move_packet_across_boundary` are two different functions which happen to have some bits in common.

- `move_packet` just moves the packet a known distance but 
- `move_packet_across_boundary` moves the packet TO a known point.
To me that's two different things. In theory the right thing to do would be to calculate `distance2boundary` again but we know that `d_boundary` didn't change between the last calculation and calling `move_packet_across_boundary` so I left it out.
Moving the packet and then setting the radius explicitly would work of course but that would repeat the calculation while using existing code. So both variants have their trade-offs.

The best solution in my opinion would be to have a function `increment_estimators` which is called by both methods. This would keep the (different) logic separated but combine the common one.

yeganer:
@unoebauer 
I think a sanity check isn't necessary because we use the exact same formula to calculate the `distance2boundary` and use the result to calculate the new_r. Mathematically we can prove that `new_r == r_boundary`. Because of machine precision this is not the case. But if we assume similiar orders of magnitude for `r` and `r_boundary` we are only doing errors on the level of machine precision.

unoebauer:
@yeganer  - I am aware of that. However, my concern was more that due some possible future changes we may create situations in which we introduce differences between the calculated r_packet and the expected one. This discrepancy would be absorbed by our quickfix (which was not designed to compensate this type of error). However, if we all think that the unit tests which are checking all c Monte Carlo functions would catch such a situation, we can also not explicitly include an additional sanity check.

tardis/montecarlo/tests/test_cmontecarlo.py
```diff
@@ -65,6 +66,8 @@
 cmontecarlo_filepath = os.path.join(path[0], 'montecarlo', 'montecarlo.so')
 cmontecarlo_methods = CDLL(cmontecarlo_filepath)
 
+INVERSE_C = 3.33564095198152e-11
```
wkerzendorf: I think we should get those from astropy at some stage.

yeganer: Sure but when I did that the tests failed so I hardcoded the value from the C part

unoebauer: At least add a comment detailing the current need to hardwire this.

yeganer:
@tardis-sn/tardis-core This PR is ready to be merged from my side.

wkerzendorf:
We are not merging this - right? For the moment, we will only use this as a branch to do dalek at the moment?

yeganer:
@wkerzendorf I'm not in a position to decide whether to merge or not. Although the underlying bug is affecting tardis and could happen during any run I'm completely fine if we postpone merging and I'm running dalek on this branch instead of master.

wkerzendorf:
merge when integration tests.

unoebauer:
Since the integration tests are up and running, we should merge this. @yeganer can you resolve the conflicts?

PR #565, opened 2016-05-17
Change the way AtomData quantities are initialized

yeganer:
astropy<1.1.0 is uses a different value for u for converting quantities
than the one specified in constants.u.
In astropy>=1.1 this is changed so that constants.u is always used. By
making this change we are independent of future changes.

Additionally the other quantities are changed to use units.Quantity for
initialization.


yeganer:
@tardis-sn/tardis-core This PR is ready to be merged from my side.

unoebauer:
My first thought when looking at this was - why should we care about these minor discrepancies. We could just acknowledge these minor inconsistencies in astropy, update the test baseline data when we move to astopy>=1.1 and be done with it.

However, given the small changes in the code base which are necessary to fix it properly, I am fine with merging this PR.

wkerzendorf:
For now I'm against merging this PRs. I think we should wait for the detailed integration tests. I don't believe this is currently necessary to change this as we can keep the astropy version fixed below 1.1.

unoebauer:
I agree with @wkerzendorf - let's wait for now

PR #566, opened 2016-05-18
Initial laydown of comparison plots for TestW7.

karandesai-96:
This PR implements a new design pattern - having a `plot_foo` method in `TestW7` class, for each `test_foo` method. As a first prototype, `plot_spectrum` has been added.

-----
Checklist holds a track of work done so far and work _TODO_:
- [X] Method `plot_spectrum` to generate a plot for deviation in  `spectrum.luminosity_density_lambda` against `spectrum.wavelength` is added within.
- [X] Show whether that test passed or failed within the plot itself (a badge on top right of plot).
- [X] Enclose assertions in `try-except` and make suitable calls.
- [X] Form a directory structure for saving the plots.

#### Future enhancements
* Find a way to put in description about build status corresponding to the plots. (**[pytest-html](https://www.github.com/davehunt/pytest-html)**)
    * Commit hash of build.
    * Starting time, time taken.
    * Number of passes, failures, skips etc.

### DIFF
--- a/tardis/tests/tests_slow/test_w7.py
+++ b/tardis/tests/tests_slow/test_w7.py
@@ -18,7 +18,8 @@ class TestW7(object):
 
     @classmethod
     @pytest.fixture(scope="class", autouse=True)
-    def setup(self, reference, data_path, atomic_data_fname):
+    def setup(self, request, reference, data_path, atomic_data_fname,
+              slow_tests_datadir, base_plot_dir):
         """
         This method does initial setup of creating configuration and performing
         a single run of integration test.
@@ -55,6 +56,12 @@ class TestW7(object):
 
         # Get the reference data through the fixture.
         self.reference = reference
+
+        # Form a base directory to save plots for `W7` setup.
+        # TODO: Rough prototyping, parametrize this as more setups are added.
+        self.name = "w7"
+        self.base_plot_dir = os.path.join(base_plot_dir, self.name)
+        os.makedirs(self.base_plot_dir)
 
     def test_j_estimators(self):
         assert_allclose(
@@ -152,7 +159,7 @@ class TestW7(object):
 
         # Figure is saved in `tmp` directory right now, till a suitable way of
         # saving them is decided.
-        plt.savefig(os.path.join(self.plot_savedir, "spectrum.png"))
+        plt.savefig(os.path.join(self.base_plot_dir, "spectrum.png"))
 
     def test_montecarlo_properties(self):
         assert_quantity_allclose(
--- a/tardis/tests/tests_slow/test_w7.py
+++ b/tardis/tests/tests_slow/test_w7.py
@@ -1,6 +1,7 @@
 import os
 import yaml
 import pytest
+import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
 from astropy.tests.helper import assert_quantity_allclose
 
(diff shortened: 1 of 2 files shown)


tardis/tests/tests_slow/test_w7.py
```diff
@@ -102,22 +104,56 @@ def test_luminosity_inner(self):
...
+
+    def plot_spectrum(self, passed):
+        plt.suptitle("Deviation in spectrum_quantities", fontweight="bold")
+
+        # `ldl_` prefixed variables associated with `luminosity_density_lambda`.
+        # Axes of subplot are extracted, if we wish to make multiple plots
+        # for different spectrum quantities all in one figure.
+        ldl_ax = plt.subplot(111)
+        ldl_ax.set_title("Deviation in luminosity_density_lambda")
+        ldl_ax.set_xlabel("N-th packet")
```
yeganer: You are plotting the luminosity density which is used together with the wavelength to form the spectrum.
labeling the x axis N-th packet is wrong.
Plot the data as (x,y) where x is the wavelength and y the luminosity density at that x and call the x-label Wavelength.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -48,90 +51,128 @@ def setup(self, reference, data_path, atomic_data_fname):
...
+        plt.suptitle("Deviation in spectrum_quantities", fontweight="bold")
+
+        # `ldl_` prefixed variables associated with `luminosity_density_lambda`.
+        # Axes of subplot are extracted, if we wish to make multiple plots
+        # for different spectrum quantities all in one figure.
+        ldl_ax = plt.subplot(111)
+        ldl_ax.set_title("Deviation in luminosity_density_lambda")
+        ldl_ax.set_xlabel("Wavelength")
+        ldl_ax.set_ylabel("Relative error (1 - obtained / baseline)")
+        deviation = 1 - (
```
yeganer: @wkerzendorf We want the absolute value of relative differences, right?

tardis/tests/tests_slow/test_w7.py
```diff
@@ -48,90 +51,128 @@ def setup(self, reference, data_path, atomic_data_fname):
...
+            self.result.spectrum.luminosity_density_lambda.value /
+            self.reference['luminosity_density_lambda'].value)
+
+        if passed:
+            ldl_ax.text(0.8, 0.8, 'passed', transform=ldl_ax.transAxes,
+                        bbox={'facecolor': 'green', 'alpha': 0.5, 'pad': 10})
+            ldl_ax.plot(self.reference['wavelength'], deviation, "g+")
```
yeganer: what's the difference between 'g+' and 'rx'? can you please use a keyword argument so it's clear what option this argument refers to?

tardis/tests/tests_slow/test_w7.py
unoebauer: @yeganer - looks good to me as it is. It calculates the relative error. I think this is the most sensible thing to do.

tardis/tests/tests_slow/test_w7.py
unoebauer: @yeganer: these are matplotlib style shortcuts consisting of a color identifier (g for green, r for red) and a style key word (+ for plus markers and x for x markers). I am ok with using them, but you have more control over the plotting style when fully using the key words (for example, color="green", marker="+", ls="")

tardis/tests/tests_slow/test_w7.py
```diff
@@ -48,90 +51,128 @@ def setup(self, reference, data_path, atomic_data_fname):
...
+            raise e
+
+    def plot_spectrum(self, passed):
+        plt.suptitle("Deviation in spectrum_quantities", fontweight="bold")
+
+        # `ldl_` prefixed variables associated with `luminosity_density_lambda`.
+        # Axes of subplot are extracted, if we wish to make multiple plots
+        # for different spectrum quantities all in one figure.
+        ldl_ax = plt.subplot(111)
+        ldl_ax.set_title("Deviation in luminosity_density_lambda")
```
unoebauer: Personally, I think that we, in addition to showing the relative error, should also show a comparison between the two spectra. This is very helpful in case the test fails. Then seeing difference in the absolute spectrum may help pinpointing the cause - thoughts, @yeganer, @wkerzendorf?

wkerzendorf: I agree with that - but I think we want to get the entire framework built before moving on. Here I just plot something.

tardis/tests/tests_slow/test_w7.py
wkerzendorf: @yeganer I would just move on for now.

wkerzendorf:
@yeganer @unoebauer is this mergeable - with the notice that we will change the plots later? Are there any serious issues left?

unoebauer:
@wkerzendorf, @karandesai-96 - fine with me. The PR looks good and some aesthetic details or additional plotting capabilities can be added later. Merge at will!

yeganer:
@wkerzendorf @unoebauer Yes I think we can merge this (I'd autosquash it because of all the back and forth during the  PR)

PR #567, opened 2016-05-20
Accept keyword arguments for Configuration.from_yaml

yeganer:
This is a small change which allows all kwargs passed to `Configuration.from_yaml()` to be redirected to `Configuration.from_dict()`.

### DIFF
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -805,21 +805,24 @@ class Configuration(ConfigurationNameSpace):
     """
 
     @classmethod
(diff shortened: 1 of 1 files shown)


unoebauer:
Without having checked the PR in detail: given @ftsamis work on the config system, do we really need this one?

yeganer:
@unoebauer Have a look, it's only 4 lines with 'logic'. I think of this more as a bugfix.

ftsamis:
This is indeed a small bugfix, and it will probably stay even after my work with the config system is done.

PR #569, opened 2016-05-23
Added a conda environment for tardis

wkerzendorf:
There have already been some conda environments for TARDIS. I have added one into the root directory.

This should replace the conda requirements file. We see a whole bunch of Libraries currently undergoing changes that have direct influence on TARDIS working. I suggest the team uses a fixed environment for development with only a few people trying out new releases of software and curating the environment for the rest of the team


yeganer:
I suggest to stick to numpy 1.9 until we have fixed all inconsistencies with astropy 1.1 which allows us to use numpy 1.11.
The reason I suggest moving to numpy 1.11 is the increased support for astropy quantities. As a result one would be able to rely on library code in several cases where we currently use self-written functions.

wkerzendorf:
done!

yeganer:
You haven't updated the PR yet.

wkerzendorf:
@tardis-sn/tardis-core  I want to merge this ASAP. @ftsamis that is interesting for you.

tardis_env27.yml
```diff
@@ -1,15 +1,14 @@
...
-- numpy=1.10.0
+- numpy=1.9
 - scipy=0.17.0
 - pandas=0.16.2
 - pytables=3.2.2
 - h5py=2.5
 - matplotlib=1.4.3
 - astropy=1.0.5
 - PyYAML=3.11
-- jsonschema=2.5.1
```
ftsamis: huh? :P Did I miss something?

.travis.yml
```diff
@@ -71,7 +71,7 @@ before_install:
     - if [[ $TEST_MODE == 'spectrum' ]]; then mv kurucz_cd23_chianti_H_He.h5 $HOME/; fi
 
 install:
-   - conda env create -f tardis_env27.yml
+   - conda env create -f tardis_env27.yml -n tardis-ci
    - source activate tardis-ci
```
ftsamis: Your best solution currently is to change both lines to the following:
```bash
conda env create -f tardis_env27.yml
source activate tardis
```

ftsamis:
There is a bug in newer conda versions where if there is an environment name specified in the environment file, the `-n` option doesn't override it. The [easy solution](https://github.com/tardis-sn/tardis/pull/569/files/b2d95b40ad08d0624b50cd370420c4e4664d096d..9a0307a0289a04a803c61032811b0fb504ddbd6c#r68331168) now is to just keep the environment named "tardis" instead of "tardis-ci".

PR #570, opened 2016-05-24
Enable generation of html report of integration tests, and upload to dokuwiki.

karandesai-96:
This PR introduces a set of pytest functions which would be complaint with [pytest-html](https://www.github.com/davehunt/pytest-html). These fixtures introduce a layer of customization over default html report provided by plugin.

* As mentioned in README of **pytest-html**, to generate a report, an extra argument has to be included as:
```
--html=path/to/file.html
```

* This command line option no longer has to be specified, due to invocation of `pytest_configure` method, before test execution. It assigns a _tempfile_  path to `config.option.filepath` for writing html report.

* Another method `pytest_unconfigure`, serves as a global teardown. It uploads the test report to dokuwiki.

* The command line argument `--slow-tests-data` has been renamed to `--integration-tests` and it accepts a path to YAML configuration for running slow tests. All other related fixtures have been modified. For now the schema of configuration file is:
```yml
reference: "/absolute/path/to/reference/dir"

dokuwiki:
  username: "username"
  password: "password"
```

This configuration file is parsed as a dict in a fixture created in **conftest.py**.

The checklist here maintains the status of work on this PR:
- [x] Show commit hash of the build on which current test run is being executed.
- [x] Make pytest-html write the html report to a tempfile, and `--html` may no longer be required.
- [x] Upload report to dokuwiki and leave no trace of tempfiles in filesystem.
- [x] Accept configuration for integration test run as YAML instad of individual fixtures.

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -1,4 +1,5 @@
 import os
+import tempfile
 import tardis
 from astropy.tests.pytest_plugins import *
 
@@ -47,8 +48,9 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
+    html_file = tempfile.NamedTemporaryFile(delete=False)
     # Html test report will be generated at this filepath by pytest-html plugin
-    config.option.htmlpath = "/tmp/{0}.html".format(tardis.__githash__)
+    config.option.htmlpath = html_file.name
 
 
 def pytest_unconfigure(config):
@@ -78,7 +80,7 @@ def pytest_unconfigure(config):
 
     # Remove the local report file. Keeping the report saved on local filesystem
     # is not desired, hence deleted.
-    os.remove(config.option.htmlpath)
+    os.unlink(config.option.htmlpath)
     print "Deleted html report previously existing at {0}".format(
         config.option.htmlpath
     )
@@ -224,7 +226,6 @@ def test_data_path():
 
 @pytest.fixture
 def included_he_atomic_data(test_data_path):
-    import os, tardis
     atomic_db_fname = os.path.join(test_data_path, 'chianti_he_db.h5')
     return AtomData.from_hdf5(atomic_db_fname)
 
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -1,9 +1,11 @@
+import os
 import tardis
 from astropy.tests.pytest_plugins import *
 
 # -------------------------------------------------------------------------
 # Initialization
 # -------------------------------------------------------------------------
+
 
 def pytest_addoption(parser):
     parser.addoption("--remote-data", action="store_true",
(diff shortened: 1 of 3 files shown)


karandesai-96:
The resources referred were:
https://pytest.org/dev/plugins.html
https://pytest.org/dev/_modules/_pytest/hookspec.html
https://www.github.com/davehunt/pytest-html

Test execution looks like this:
image

Here is the report for fafa9e5, which is specified in the screenshot:
http://opensupernova.org/~karandesai96/integration2/doku.php?id=reports:fafa9e5

tardis/conftest.py
```diff
@@ -35,6 +42,54 @@ def pytest_addoption(parser):
...
+def pytest_configure(config):
+    html_file = tempfile.NamedTemporaryFile(delete=False)
+    # Html test report will be generated at this filepath by pytest-html plugin
+    config.option.htmlpath = html_file.name
+
+
+def pytest_unconfigure(config):
+    # Html report created by pytest-html plugin is read here, uploaded to
+    # dokuwiki and finally deleted.
+    try:
```
wkerzendorf: you should do a try and except statement at the beginning with then just having dokuwiki_available boolean variable.

tardis/conftest.py
```diff
@@ -35,6 +42,54 @@ def pytest_addoption(parser):
     parser.addoption("--slow-test-data", dest="slow-test-data",
```
wkerzendorf: I thought we changed this to integration test (at least @yeganer and I discussed this).

tardis/conftest.py
```diff
@@ -35,6 +42,54 @@ def pytest_addoption(parser):
...
                      help="path to directory having baseline data for slow tests")
 
+    # Accept dokuwiki credentials through command line.
+    parser.addoption("--username", dest='username', help="Dokuwiki username")
```
wkerzendorf: I think you should make a yaml file that contains all of that information.

tardis/conftest.py
```diff
@@ -35,6 +42,54 @@ def pytest_addoption(parser):
...
+        # Upload report on dokuwiki. Temporary link due to prototyping purposes.
+        doku_conn.pages.append("reports:{0}".format(githash[:7]), report_content)
+        print "Uploaded report of {0} on Dokuwiki".format(githash[:7])
+    except:
+        print "Dokuwiki python bindings not found!"
+
+    # Remove the local report file. Keeping the report saved on local filesystem
+    # is not desired, hence deleted.
+    os.unlink(config.option.htmlpath)
+    print "Deleted html report previously existing at {0}".format(
```
wkerzendorf: I think we can completely remove this or put this in the debug log.

tardis/conftest.py
karandesai-96: I'll do it that way then.

tardis/conftest.py
karandesai-96: I would remove `print` statements for now.

tardis/conftest.py
wkerzendorf: the try-except block should only contain things that fail. Move this block up to the import statements and only use if here.

tardis/conftest.py
```diff
@@ -32,8 +50,48 @@ def pytest_addoption(parser):
...
+
+            report_content = (
+                "Test executed on commit "
+                "[[https://www.github.com/tardis-sn/tardis/commit/{0}|{0}]]\n\n"
+                "{1}".format(githash, report_content)
+            )
+            # Upload report on dokuwiki. Temporary link due to prototyping purposes.
+            doku_conn.pages.append("reports:{0}".format(githash[:7]), report_content)
+        except gaierror, dokuwiki.DokuWikiError:
+            pass
```
wkerzendorf: again - stick the things that generate the error in the try except block and nothing else.

tardis/conftest.py
```diff
@@ -1,5 +1,23 @@
...
+from tardis.atomic import AtomData
+from tardis.io.config_reader import Configuration
 from astropy.tests.pytest_plugins import *
 
+# For specifying error while exception handling
+from socket import gaierror
+
+try:
+    import dokuwiki
+    dokuwiki_available = True
```
wkerzendorf: same goes here - the things that raise the error need to be in the try, but nothing else.

tardis/conftest.py
```diff
@@ -32,8 +51,50 @@ def pytest_addoption(parser):
...
+    if dokuwiki_available:
+        githash = tardis.__githash__
+        report_content = open(config.option.htmlpath, 'rb').read()
+        report_content = report_content.replace("<!DOCTYPE html>", "")
+
+        report_content = (
+            "Test executed on commit "
+            "[[https://www.github.com/tardis-sn/tardis/commit/{0}|{0}]]\n\n"
+            "{1}".format(githash, report_content)
+        )
```
karandesai-96: @wkerzendorf These four lines are good to be here or in `else` block ? I spoke to myself: _try to make a connection, except these two errors, else upload the report we have_. Hence, I kept the report ready beforehand (compromising the time of execution of these lines if connection isn't at all made.)

wkerzendorf: @karandesai-96 it's not about execution time it's about making the code easy to understand. What really happens when the if is True, what happens when the if is False. I think its fine the way it is for now.

tardis/conftest.py
```diff
@@ -73,62 +134,60 @@ def pytest_report_header(config):
```
karandesai-96: @wkerzendorf Do we require a `try-except-else` for numpy as well ?

PR #576, opened 2016-05-31
[config-system] Validator rewrite: Use jsonschema

ftsamis:
**Note: Tests for the validator are not yet implemented**

With this PR I rewrite the `io.config_validator` which was a custom-built validator written specifically for TARDIS. With the changes I made, the validator now uses jsonschema, a well-known and broadly-used validation module which validates schemas according to the json-schema specification.

For this purpose the configuration definition needed to be transcoded in json schema syntax, which was done at #549.

All currently used types are supported by this validator, as well as setting default values.


tardis/io/config_reader.py
```diff
@@ -1031,19 +1031,19 @@ def from_config_dict(cls, config_dict, atom_data=None, test_parser=False,
...
         virtual_spectrum_section = montecarlo_section['virtual_spectrum_range']
         montecarlo_section['virtual_spectrum_range'] = {}
         montecarlo_section['virtual_spectrum_range']['start'] = \
-            virtual_spectrum_section[0]
+            virtual_spectrum_section['start']
         montecarlo_section['virtual_spectrum_range']['end'] = \
-            virtual_spectrum_section[1]
+            virtual_spectrum_section['stop']
         montecarlo_section['virtual_spectrum_range']['samples'] = \
-            virtual_spectrum_section[2]
```
yeganer: What's the reason we keep this?
I propose to delete lines 1031-1046 and rename `montecarlo.virtual_spectrum_range.samples` to `montecarlo.virtual_spectrum_range.num` wherever it is accessed

tardis/io/config_reader.py
```diff
@@ -1018,7 +1018,7 @@ def from_config_dict(cls, config_dict, atom_data=None, test_parser=False,
 
 
 
-        if montecarlo_section['convergence_strategy'] is None:
+        if 'convergence_stragegy' not in montecarlo_section:
             logger.warning('No convergence criteria selected - '
                            'just damping by 0.5 for w, t_rad and t_inner')
             montecarlo_section['convergence_strategy'] = (
```
yeganer: Since we decided to use default values for now, I propose to move these default values to either the schema or the Simulation class

tardis/io/config_reader.py
```diff
@@ -486,7 +487,9 @@ def parse_spectrum_list2dict(spectrum_list):
     """
     Parse the spectrum list [start, stop, num] to a list
     """
-
+    if isinstance(spectrum_list, dict):
```
wkerzendorf: how about we change to 'if 'start' in spectrum_list and 'stop' in spectrum list and 'num' in spectrum_list`: :wink:

tardis/io/config_reader.py
ftsamis: This is part of the next and final big milestone/PR (configuration system restructure). My purpose in this PR was to only do changes related and absolutely necessary to the validator.

wkerzendorf: @yeganer that's dangerous as older configuration files need to work. I would flag it but leave it for now.

tardis/io/config_reader.py
wkerzendorf: good idea.

tardis/io/config_reader.py
ftsamis: I was expecting a comment on this line.
I even thought about adding a "Dear Wolfgang, this is as temporary as it can get, please act like you didn't notice the missing ducks" comment. :stuck_out_tongue: 

I'll do the change...

tardis/io/config_reader.py
yeganer: I think we should have a function/module that provides the backwards compatibility.
```
if config.version < latest_config_version:
    config_dict = interpret_legacy_config(config_dict)
```

tardis/io/config_reader.py
```diff
@@ -487,7 +487,8 @@ def parse_spectrum_list2dict(spectrum_list):
     """
     Parse the spectrum list [start, stop, num] to a list
     """
-    if isinstance(spectrum_list, dict):
+    if 'start' in spectrum_list and 'stop' in spectrum_list \
+            and 'num' in spectrum_list:
```
yeganer: This looks complicated, why not `try/except`?

tardis/io/config_reader.py
```diff
@@ -1007,6 +1007,10 @@ def from_config_dict(cls, config_dict, atom_data=None, test_parser=False,
         ##### Monte Carlo Section
 
         montecarlo_section = validated_config_dict['montecarlo']
+        montecarlo_section['no_of_packets'] = \
+            int(montecarlo_section['no_of_packets'])
+        montecarlo_section['last_no_of_packets'] = \
+            int(montecarlo_section['last_no_of_packets'])
```
yeganer: That's a sensible hotfix but I suggest to keep in mind to move this conversion to the Simulation object in the process of refactoring the config system?

wkerzendorf: agreed - make an issue.

yeganer: I think there will be several small issues like this which can be 'fixed' during the refactor. Should I make one Issue for all of them?

wkerzendorf: yep it can be one big issue - that summarizes the small ones. but @ftsamis should be able to edit it as he will be leading that.

tardis/io/config_reader.py
```diff
@@ -27,7 +27,6 @@
 
 default_config_definition_file = os.path.join(data_dir,
```
yeganer: Do we still need this line?

tardis/io/config_reader.py
ftsamis: Done, nice and clean in the schema!

tardis/io/config_reader.py
ftsamis: No, removed all the references to `tardis_config_definition.yml` and the file itself.

PR #578, opened 2016-06-02
Shift slow test related pytest hooks in lower level conftest.py

karandesai-96:
There were two primary issues with infrastructure of integration tests developed till now:
* With the merge of #570 , the html report was uploaded on dokuwiki. Html report was generated using pytest-html plugin, which requires `--html` command line option to specify path to save the report locally. But with the help of tempfile module, and also pytest hooks: `pytest_configure` and `pytest_unconfigure` declared in top level conftest.py, the report was saved in a temp file, uploaded to dokuwiki and finally deleted. 
* Astropy test helpers were imported as `*` in top level conftest.py and the configure-unconfigure hooks I declared, actually had overwritten the former ones. This loosened the protection of `@remote_data` marker provided by astropy test helpers.

    * **Fix**: All these hooks are moved in lower level conftest.py . It also helps to keep slow tests related hooks and fixtures separate.

* The plots were saved in `/tmp/plots/w7/tardis.__githash__`. I used `os.rmdir` to cleanup the temp directory. But `os.rmdir` cannot remove non empty directories. Hence running multiple tests successively on same commit required manual deletion of that directory, else the tests would raise `OSError`.

    * **Fix**: This directory path was provided using `mkdtemp()` method of tempfile module. During `pytest_unconfigure`, recursive deletion was performed using `shutil` module.

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -10,23 +10,11 @@ from astropy.tests.pytest_plugins import (
         )
 
 import yaml
-import tempfile
 import tardis
 import pytest
 from tardis.atomic import AtomData
 from tardis.io.config_reader import Configuration
 from tardis.io.util import yaml_load_config_file
-
-# For specifying error while exception handling
-from socket import gaierror
-
-try:
-    import dokuwiki
-except ImportError:
-    dokuwiki_available = False
-else:
-    dokuwiki_available = True
-
 
 ###
 # Astropy
@@ -86,45 +74,6 @@ def pytest_addoption(parser):
                      help="path to configuration file for integration tests")
 
 
-def pytest_configure(config):
-    _pytest_configure(config)
-    html_file = tempfile.NamedTemporaryFile(delete=False)
-    # Html test report will be generated at this filepath by pytest-html plugin
-    config.option.htmlpath = html_file.name
-
-
-def pytest_unconfigure(config):
-    _pytest_unconfigure()
-    # Html report created by pytest-html plugin is read here, uploaded to
-    # dokuwiki and finally deleted.
-    if dokuwiki_available:
-        githash = tardis.__githash__
-        report_content = open(config.option.htmlpath, 'rb').read()
-        report_content = report_content.replace("<!DOCTYPE html>", "")
-
-        report_content = (
-            "Test executed on commit "
-            "[[https://www.github.com/tardis-sn/tardis/commit/{0}|{0}]]\n\n"
-            "{1}".format(githash, report_content)
-        )
-
-        try:
-            doku_conn = dokuwiki.DokuWiki(
-                url=integration_tests_config["dokuwiki"]["url"],
-                user=integration_tests_config["dokuwiki"]["username"],
-                password=integration_tests_config["dokuwiki"]["password"])
-        except gaierror, dokuwiki.DokuWikiError:
-            print "Dokuwiki connection not established, report upload failed!"
-        else:
-            # Upload report on dokuwiki. Temporary link due to prototyping purposes.
-            doku_conn.pages.append("reports:{0}".format(githash[:7]), report_content)
-            print "Uploaded report on Dokuwiki."
-
-    # Remove the local report file. Keeping the report saved on local filesystem
-    # is not desired, hence deleted.
-    os.unlink(config.option.htmlpath)
-    print "Deleted temporary file containing html report."
-
 # -------------------------------------------------------------------------
 # project specific fixtures
 # -------------------------------------------------------------------------
@@ -165,15 +114,3 @@ def included_he_atomic_data(test_data_path):
 def tardis_config_verysimple():
     return yaml_load_config_file(
         'tardis/io/tests/data/tardis_configv1_verysimple.yml')
-
-
-@pytest.fixture(scope="session")
-def integration_tests_config(request):
-    integration_tests_configpath = request.config.getvalue("integration-tests")
-    if integration_tests_configpath is None:
-        pytest.skip('--integration-tests was not specified')
-    else:
-        integration_tests_configpath = os.path.expandvars(
-            os.path.expanduser(integration_tests_configpath)
-        )
-        return yaml.load(open(integration_tests_configpath))
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -4,12 +4,9 @@
 
 from astropy.tests.pytest_plugins import *
 from astropy.tests.pytest_plugins import (
-        pytest_addoption as _pytest_add_option,
-        pytest_configure as _pytest_configure,
-        pytest_unconfigure as _pytest_unconfigure
-        )
+        pytest_addoption as _pytest_add_option
+    )
 
-import yaml
 import tardis
 import pytest
 from tardis.atomic import AtomData
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -5,6 +5,7 @@ import yaml
 import numpy as np
 import pytest
 from astropy import units as u
+from astropy.tests.helper import remote_data
 import tardis
 
 # For specifying error while exception handling
@@ -28,6 +29,7 @@ def pytest_configure(config):
     config.option.htmlpath = html_file.name
 
 
+@remote_data
 def pytest_unconfigure(config):
     # Html report created by pytest-html plugin is read here, uploaded to
     # dokuwiki and finally deleted.
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -20,6 +20,14 @@ else:
 
 
 def pytest_configure(config):
+    integration_tests_configpath = config.getvalue("integration-tests")
+    if integration_tests_configpath is not None:
+        integration_tests_configpath = os.path.expandvars(
+            os.path.expanduser(integration_tests_configpath)
+        )
+        config.option.integration_tests_config = yaml.load(
+            open(integration_tests_configpath))
+
     # A common tempdir for storing plots / PDFs and other slow test related data
     # generated during execution.
     tempdir_session = tempfile.mkdtemp()
@@ -47,13 +55,12 @@ def pytest_unconfigure(config):
         # These steps are already performed by `integration_tests_config` but
         # all the fixtures are teared down and no longer usable, when this
         # method is being called by pytest, hence they are called as functions.
-        integration_tests_configdict = integration_tests_config()
-
+        integration_tests_config = config.option.integration_tests_config
         try:
             doku_conn = dokuwiki.DokuWiki(
-                url=integration_tests_configdict["dokuwiki"]["url"],
-                user=integration_tests_configdict["dokuwiki"]["username"],
-                password=integration_tests_configdict["dokuwiki"]["password"])
+                url=integration_tests_config["dokuwiki"]["url"],
+                user=integration_tests_config["dokuwiki"]["username"],
+                password=integration_tests_config["dokuwiki"]["password"])
         except gaierror, dokuwiki.DokuWikiError:
             print "Dokuwiki connection not established, report upload failed!"
         else:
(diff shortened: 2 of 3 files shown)


tardis/tests/tests_slow/conftest.py
```diff
@@ -87,3 +88,20 @@ def reference(request, reference_datadir):
...
+def pytest_runtest_makereport(item, call):
+    pytest_html = item.config.pluginmanager.getplugin('html')
+    outcome = yield
+    report = outcome.get_result()
+    extra = getattr(report, 'extra', [])
+    if report.when == 'call':
+        test_method_name = report.nodeid.split('::')[-1]
+        if test_method_name == "test_spectrum":
+            # Just making it work for this commit. Will be changes subsequently
+            with open(os.path.join("/tmp/plots", tardis.__githash__[0:7],
```
wkerzendorf: don't use the png file as a workaround - directly encode to base64 binary thingie

karandesai-96:
There are certain changes with the storage mechanism of plots (and PDFs within few commits). The `base_plot_dir` fixture used to return absolute path of directory to store plots. Now that we've decided to uploaded everything online and not saving it locally, the `base_plot_dir` fixture is removed, and instead we have a common, session wide `tempdir` made available during `pytest_configure` as `config.option.tempdir`. Right now, the spectrum plot is saved in `tempdir/w7/spectrum.png` and accessed by the hookwrapper of `pytest-html` plugin accordingly.

#### Why passing through `report` object isn't possible ?
By referring `pytest`'s documentation, I discovered that `TestReport` class objects (reports) are responsible for making the report look like it is. The hookwrapper `yield`s once and a report object is obtained by getting the result out of generator. We need to append an image in html_report, which is done by setting it is `report.extra` parameter. We have no other way to transsfer from test_method to hookwrapper, and even if we have - we don't need to append `base64` string in `report.extra`, but a formatted `` tag. All in all, it isn't feasible - even if I do it the crooked way and make it work, the code might not be understandable in one glance.

#### Note:
Pytest has its own `tmpdir` fixture but is is function scoped, there is a `tmpdir_factory` fixture introduced in version 2.8, our conda requirements lists `pytest==2.2.4`. (Anyhow it looks like it isn't stable, doesn't work for me in `pytest==2.9.1` locally.)

I think that usage of `tempfile` built-in module is the most stable and simple approach to do here. Besides, I can imagine that interactions with a session wide temporary directory to store plots/ PDFs before they are uploaded (instead of function scoped scoped or direct `StringIO` object interactions), will be easily extendable when we parametrize the `data_path` fixture for multiple tests.

@wkerzendorf @unoebauer @yeganer I'll be happy to answer questions and accept suggestions and amendments :grinning:

tardis/conftest.py
```diff
@@ -56,6 +56,11 @@ def pytest_addoption(parser):
...
 
 def pytest_configure(config):
+    # A common tempdir for storing plots / PDFs and other slow test related data
+    # generated during execution.
+    tempdir_session = tempfile.mkdtemp()
+    config.option.tempdir = tempdir_session
+
     html_file = tempfile.NamedTemporaryFile(delete=False)
     # Html test report will be generated at this filepath by pytest-html plugin
     config.option.htmlpath = html_file.name
```
karandesai-96: A `tempdir` is added in `pytest_configure`. Would it be better to declare temporary html file inside `tempdir` now ?

tardis/conftest.py
```diff
@@ -75,22 +80,37 @@ def pytest_unconfigure(config):
...
+                # Upload report on dokuwiki. Temporary link due to prototyping purposes.
+                doku_conn.pages.set("reports:{0}".format(githash[:7]), report_content)
+                print "Uploaded report on Dokuwiki."
 
     # Remove the local report file. Keeping the report saved on local filesystem
     # is not desired, hence deleted.
     os.unlink(config.option.htmlpath)
     print "Deleted temporary file containing html report."
+    # Remove tempdir by recursive deletion
+    os.system("rm -r {0}".format(config.option.tempdir))
```
karandesai-96: `tempdir` won't be empty so `os.removedirs`, `os.rmdir`, `os.unlink` won't work. Hence I used `os.system` for recursive deletion.

Although there is a built-in module named `shutils` which provides high level file interactions, and recursive deletion of non-empty directory is handled as `shutil.rmtree("/path/to/dir")`. I was hesitant to use it and it is required just for one purpose, and we have an alternative (`os.system`), which quite clearly shows what the code is doing here. What should I do here - let it be or change it ?

tardis/conftest.py
```diff
@@ -56,6 +56,11 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
+    # A common tempdir for storing plots / PDFs and other slow test related data
+    # generated during execution.
+    tempdir_session = tempfile.mkdtemp()
```
yeganer: I don't think you need the `tempdir_session` variable. just assign it directly.

tardis/conftest.py
yeganer: So pytest-html requires `config.option.htmlpath` to be set to generate a report?
In any case, I'd store the html file handle in the config, too. Then you don't have to worry about deleting the temp file because it should go when `config` goes out of scope.

tardis/conftest.py
yeganer: I say don't use `os.system` if it can be avoided.

Other than that, why do we need to remove the directory? Doesn't `tempfile.mkdtemp()` delete the directory when it get's out of scope?

tardis/tests/tests_slow/conftest.py
```diff
@@ -87,3 +75,20 @@ def reference(request, reference_datadir):
         luminosity_density_lambda=
             u.Quantity(spectrum['luminosity_density_lambda'], 'erg / (Angstrom s)'))
     return ndarrays
+
+
+@pytest.mark.hookwrapper
+def pytest_runtest_makereport(item, call):
```
yeganer: This is the 'magic' line which generates the report, right?

tardis/conftest.py
```diff
@@ -75,22 +80,37 @@ def pytest_unconfigure(config):
...
+        # method is being called by pytest.
+        integration_tests_configpath = config.getvalue("integration-tests")
+        if integration_tests_configpath is not None:
+            integration_tests_configpath = os.path.expandvars(
+                os.path.expanduser(integration_tests_configpath)
+            )
+            integration_tests_config = yaml.load(open(
+                    integration_tests_configpath))
+
+            try:
```
yeganer: **DRY**
If you really need to create a connection at two places, import it the second time. Even if it's a fixture, you can call the function like a normal function.

tardis/conftest.py
```diff
@@ -56,6 +56,11 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
```
yeganer: Does py.test require `pytest_configure` to be in the root conftest? If that's not the case, we should everything related to slow_tests to the respective directory.

tardis/conftest.py
karandesai-96: Yes, `pytest-html` plugin wants a filepath (unicode or string) in `config.option.htmlpath`.

tardis/conftest.py
karandesai-96: > I say don't use `os.system` if it can be avoided.

Then I would use `shutil.rmtree` method. In one way I can think that Ican avoid using it as it is used just once, but it is a built-in module, so.. okay !

@yeganer No it doesn't, files remove themselves, but not directories. I quote this from `tempfile` [documentation](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp)

> Creates a temporary directory in the most secure manner possible. There are no race conditions in the directory’s creation. The directory is readable, writable, and searchable only by the creating user ID.

> The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

tardis/tests/tests_slow/conftest.py
karandesai-96: @yeganer Not exactly, it just appends custom details of our interests (like base64 of image). This is a `hookwrapper` which wraps pytest's own `pytest_runtest_makereport`, and runs before and after the creation (rather, appending details of each method as the tests run).

* `outcome = yield` will yield exactly once (details per test) and return a generator. report object is extracted by `outcome.get_result()`. The name of test method which was tested in a particular iteration is accessed by `report.nodeid` (this is how I am adding an image only to `test_spectrum`).

* Everything before `outcome = yield` is executed first, then `pytest`'s native method executes, which is being generated here by `yield` keyword, and everything after it will be executed afterwards. This happens for every `test_method`.

* This hookwrapper currently works only for `test_spectrum` as `report.nodeid` related if condition evaluates to `True` just once.

* After all the test methods are executed, `pytest-html` does its work by using report object, and essentially the `base64` of image which we had appended as `report.extra` and generating html of the report.

If I was unclear at some place, I'll clarify it further :grinning:

tardis/conftest.py
karandesai-96: > Even if it's a fixture, you can call the function like a normal function.

Oh right, this never came to my mind ! I will do it then.

tardis/conftest.py
karandesai-96: @yeganer No it doesn't, though it is true that `pytest_configure` in higher level directories will be called first, followed by any `pytest_configure`s in child directories. I don't think there are any conflicts and it is safe to move it in. Should I move integration tests config and dokuwiki connection fixture as well ? (I will be rewriting history in this case to avoid mess :sweat_smile: )

yeganer: Sure move it.

You can't rewrite history since part of the work is already merged.

tardis/conftest.py
yeganer: Okay, well I'd say just store the tmpfile in `config.option.html`

tardis/conftest.py
yeganer: Then use `shutil.rmtree`.

tardis/tests/tests_slow/conftest.py
```diff
@@ -1,9 +1,85 @@
...
+    # A common tempdir for storing plots / PDFs and other slow test related data
+    # generated during execution.
+    tempdir_session = tempfile.mkdtemp()
+    config.option.tempdir = tempdir_session
+    html_file = tempfile.NamedTemporaryFile(delete=False)
+    # Html test report will be generated at this filepath by pytest-html plugin
+    config.option.htmlpath = html_file.name
+
+
+@remote_data
```
yeganer: I'm not sure that works and I think that's not how this is supposed to to work.

yeganer:
I have some comments on `pytest_(un)configure`.
I think you should only apply the logic if `--integration-tests` is provided. If that's the case, it might be a good idea to populate the config during configure and then teardown in unconfigure. You can use the setup from `configure` in your fixture, too.
I'd preferr `request.config` instead of `pytest.config`.

tardis/tests/tests_slow/conftest.py
karandesai-96: It is working, [here](http://opensupernova.org/~karandesai96/integration/doku.php?id=reports:e0132d5) is the report which I generated this morning using this code.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -11,6 +11,8 @@
 from tardis.io.config_reader import Configuration
 
 
+@pytest.mark.skipif(not pytest.config.getvalue("integration-tests"),
+                    reason="integration tests are not included in this run")
```
karandesai-96: Should I keep this marker here or in `integration_tests_config` fixture ? I cannot keep it while unpacking it in `pytest_configure`, hence removed it from there.

tardis/tests/tests_slow/conftest.py
yeganer: and if you remove `@remote_data` it stops working?
and you have to add `--remote-data` in the commandline or it won't work?

tardis/tests/tests_slow/conftest.py
```diff
@@ -1,9 +1,85 @@
...
+def pytest_unconfigure(config):
+    # Html report created by pytest-html plugin is read here, uploaded to
+    # dokuwiki and finally deleted.
+    if dokuwiki_available:
```
yeganer: You added a check to `pytest_configure` to check if we are running the integration tests or not. Please do the same for `pytest_unconfigure`

tardis/tests/tests_slow/conftest.py
karandesai-96: I have to do both the things simultaneously to make it work.

yeganer:
http://opensupernova.org/~karandesai96/integration/doku.php?id=reports:52115b2 Is the report produced with the latest commit. I'm fine with merging.

PR #580, opened 2016-06-03
Store plasma to hdf

ftsamis:
Heavily based on #435, this allows storing the plasma to hdf. Additionally, `BasePlasma.to_hdf` now has a `property_type` argument, which filters the properties to be stored.

So for example:
```python
from tardis.io import config_reader
from tardis.model import Radial1DModel
from tardis.plasma.properties.base import Input

cfg = 'tardis_configv1_density_exponential_test.yml'

tardis_config = config_reader.Configuration.from_yaml(cfg)
model = Radial1DModel(tardis_config)

model.plasma_array.to_hdf('/tmp/plasma_inputs', property_type=Input)
```
would only store the input plasma properties to `/tmp/plasma_inputs`. For all parameters the `property_type` must be `BasePlasmaProperty` which is the default value and can be omitted.


ftsamis:
I now changed the way properties can be filtered. 

So, instead of specifying only one `BasePlasmaProperty` such as `Input`, and storing its and its subclasses instances, now a collection of properties (`property_collections.PlasmaPropertyCollection`) can be passed as a parameter and the `to_hdf` will only store parameters that are of a type included in the collection, and any of their subclasses.

Existing collections such as `property_collections.basic_inputs` can be used, as well as combinations of collections (expressed as a new collection), e.g.: `model.plasma_array.to_hdf('/tmp/plasma_inputs', collection=basic_inputs + basic_properties)`

wkerzendorf:
@ftsamis the filter is fine but goes well beyond what is needed. Next we should focus our efforts on storing model quantities

tardis/plasma/base.py
```diff
@@ -258,10 +260,35 @@ def remove_hidden_properties(self, print_graph):
                 print_graph.remove_node(str(item.name))
         return print_graph
 
-class StandardPlasma(BasePlasma):
+    def to_hdf(self, path_or_buf, path='', close_hdf=True, filter=None):
```
wkerzendorf: yes, I messed that up - `path` is clearly needed!

wkerzendorf: take out the filter. I think it's some complicated beast that we can just do with a list.

ftsamis: A complicated beast is an overstatement for these two simple classes, but the purpose there was to have some consistency in the different `to_hdf` signatures. This way, if we want to store part of plasma, and part of runner, and part of xyz, we only provide one Filter object to the first `simulation.to_hdf call`, instead of doing `simulation.to_hdf(plasma_properties, runner_properties, atom_properties, xyz_future_module_properties)`

wkerzendorf: I withdraw my "complicated beast" 😉 and simply say. Let's take it out for now and finish all the other `to_hdf` methods to see what we need and what we don't need.

yeganer: To add my 2 cents. I think we should have different methods. And I think we need to have only 2 distinctions.
1. case: save all the information required to 'reconstruct' a model. Let's say I run the simulation and save it at the end, I'd be able to continue with the saved information. Store the only uuid of atomdata
2. case: save everything, this includes all intermediately calculated values and the input from the atom database

I don't think we need more filtering. If we need, what would be the usecase?

wkerzendorf: I agree with @yeganer.

ftsamis: Well my only insight to this was [TEP002#Implementation](https://github.com/tardis-sn/tep/blob/master/TEP002_model_hdf5.rst#implementation), and specifically this part:

> This will also allow the higher level .to_hdf-methods to select specific information to store.
This might be grouped in hardcoded sets like "minimal" or "detailed". 

So, my thought was to have something similar to:
`minimal = FilterCollection(plasma=PlasmaFilter(...), runner=RunnerFilter(...), atom=AtomDataFilter(...))`

If you say that the only information that needs filtering are plasma properties, then, fine I'll remove the filter module. I can totally not provide you with a use case for extra filtering. At this point I don't even know which properties should be stored from all the objects:

>    Radial1Dmodel
    Plasma
    MontecarloRunner
    Simulation
    atomic data

I would even propose updating TEP002 with these information.

ftsamis:
Closing this in favor of #586, without the filters.

PR #581, opened 2016-06-04
Update several files to match package_template

yeganer:
With this change we will always use the latest version of
astropy.tests.pytest_plugins because we don't redefine their functions.

Some cleanup was done, too.

Currently pytest_configure and pytest_unconfigure are still defined in
tardis/conftest.py but these will be removed by a future commit by
@karandesai-96.

This fixes issue with `config_dir` and astropy> 1.1


wkerzendorf:
@yeganer are we done here? If so - you can merge.

yeganer:
@wkerzendorf I'll merge once travis finished for the rebase.

PR #583, opened 2016-06-07
Subclass pytest-html's HTMLReport to DokuReport class for report generation and upload.

karandesai-96:
This PR introduces a helper class named `DokuReport` which serves as a plugin for pytest. This class is subclassed from `pytest-html.plugin.HTMLReport` class. It handles logging of test execution and upload it to Dokuwiki. It is registered as a plugin by `config.pluginmanager` in `pytest_configure` hook and unregistered in `pytest_unconfigure`. Prior to this, `pytest-html` plugin was used for report generation - to force `pytest` use `DokuReport` instead of `HTMLReport`, `pytest-html` has o be unregistered as a plugin during `pytest_configure` execution.

Checklist maintains the work done so far / work _TO-DO_ in `DokuReport`:
- [X] Include summary and execution duration, generation time and commit hash.
- [X] Include test name, duration, result.
- [X] Embed plots, and error logs if any.
- [X] Handle dokuwiki connection and interaction from within the class.
- [X] Handle tempfile writing from within the class.
- [X] Make sure that this plugin is not activated when integration tests are not selected.

### DIFF
--- a/tardis/tests/tests_slow/report.py
+++ b/tardis/tests/tests_slow/report.py
@@ -63,7 +63,7 @@ class DokuReport(HTMLReport):
         except (TypeError, gaierror, dokuwiki.DokuWikiError):
             self.doku_conn = None
 
-    def _generate_report(self):
+    def _generate_report(self, session):
         """
         The method writes HTML report to a temporary logfile.
         """
@@ -120,6 +120,14 @@ class DokuReport(HTMLReport):
                 generated.strftime('%d-%b-%Y'),
                 generated.strftime('%H:%M:%S'))))
 
+        if session.config._environment:
+            environment = set(session.config._environment)
+            body.append(html.h2('Environment'))
+            body.append(html.table(
+                [html.tr(html.td(e[0]), html.td(e[1])) for e in sorted(
+                    environment, key=lambda e: e[0]) if e[1]],
+                id='environment'))
+
         body.extend(summary)
         body.extend(results)
 
@@ -152,7 +160,7 @@ class DokuReport(HTMLReport):
         This hook function is called by pytest when whole test run is completed.
         It calls the two helper methods `_generate_report` and `_save_report`.
         """
-        report_content = self._generate_report()
+        report_content = self._generate_report(session)
         self._save_report(report_content)
 
     def pytest_terminal_summary(self, terminalreporter):
--- a/tardis/tests/tests_slow/report.py
+++ b/tardis/tests/tests_slow/report.py
@@ -1,6 +1,7 @@
 import datetime
 import pkg_resources
 import os
+import tempfile
 import time
 
 # For specifying error while exception handling
@@ -18,7 +19,8 @@ except ImportError:
 
 class DokuReport(HTMLReport):
 
-    def __init__(self, logfile, dokuwiki_details):
+    def __init__(self, dokuwiki_details):
+        logfile = tempfile.TemporaryFile()
         super(DokuReport, self).__init__(logfile.name)
         self.logfile = logfile
 
@@ -109,7 +111,7 @@ class DokuReport(HTMLReport):
 
         if self.doku_conn is not None:
             self.doku_conn.pages.set("reports:{0}".format(
-                    tardis.__githash__[:7]), self.logfile.read())
+                tardis.__githash__[:7]), self.logfile.read())
             print "Uploaded report on Dokuwiki."
 
         self.logfile.close()
--- a/tardis/tests/tests_slow/report.py
+++ b/tardis/tests/tests_slow/report.py
@@ -1,3 +1,29 @@
+"""
+A helper class which works as a plugin to generate the test report and upload it
+to the group server's dokuwiki. It inheirts from the class `HTMLReport` of
+the `pytest-html` plugin. The test report contains the following details:
+
+* The git commit hash on which test run was executed.
+* The time of generation of test report.
+* Number of passes, fails, errors, skips etc.
+* Tabular representation of each method - name, result, duration.
+* Embedded image of plot(s) and error log below a particular method (if any).
+
+As a subclass, this class serves as a plugin and hence, `pytest-html` has to be
+unregistered during the test run for tis plugin to function.
+
+When the integration tests are selected for a particular test run, this class
+is registered as a plugin in `pytest_configure` and subsequently unregistered in
+`pytest_unconfigure`. As a plugin, it implements several "hook" functions
+specified in pytest's official documentation.
+
+
+References
+==========
+1. "Writing Plugins" ( https://pytest.org/latest/writing_plugins.html )
+2. "Hookspec Source" ( https://pytest.org/latest/_modules/_pytest/hookspec.html )
+3. "pytest-html" ( https://www.github.com/davehunt/pytest-html )
+"""
 import datetime
 import pkg_resources
 import os
@@ -20,7 +46,16 @@ except ImportError:
 class DokuReport(HTMLReport):
 
     def __init__(self, dokuwiki_details):
+        """
+        Initialization of a DokuReport object and registration as a plugin
+        occurs in `pytest_configure`, where a dict containing url, username and
+        password of dokuwiki is passed through `dokuwiki_details`.
+        """
+        # A tempfile to log report before it is uploaded to dokuwiki.
         logfile = tempfile.TemporaryFile()
+
+        # Base class accepts a file path to save the report, but we forcefully
+        # set `self.logfile` to be a file-object.
         super(DokuReport, self).__init__(logfile.name)
         self.logfile = logfile
 
@@ -39,6 +74,9 @@ class DokuReport(HTMLReport):
         self.is_report_uploaded = False
 
     def _generate_report(self):
+        """
+        The method writes HTML report to a temporary logfile.
+        """
         suite_stop_time = time.time()
         suite_time_delta = suite_stop_time - self.suite_start_time
         numtests = self.passed + self.failed + self.xpassed + self.xfailed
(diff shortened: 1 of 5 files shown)


tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def pytest_sessionstart(self, session):
+        """
+        This hook function is called by pytest after all the methods and
+        collected and before the test execution has been started. Here, the
+        starting time is being recorded.
+        """
+        super(DokuReport, self).pytest_sessionstart(session)
+
+    def pytest_runtest_logreport(self, report):
```
wkerzendorf: why are these functions here?

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,251 @@
...
+                self.doku_conn = None
+
+    def add_plot(self, path):
+        """
+        This method is used by any single test method to add the file path of
+        image of corresponding comparison plot.
+        """
+        self.image_path = path
+
+    def pytest_sessionstart(self, session):
```
wkerzendorf: same goes for this one.

tardis/tests/tests_slow/report.py
karandesai-96: These are pytest hooks, this class is registered as a plugin so pytest automatically calls these hook functions. I'll update this PR by their removal if removing them also works (it should, according to inheritance).

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # `generated` stores the date and time of generation of report.
+        # It is set in `pytest_sessionfinish` using datetime.now()
+        self.generated = None
+
+        # File path of image of comparison plots for a single test.
+        self.image_path = None
+
+        # Make a dokuwiki connection if dokuwiki import was successful.
+        if dokuwiki_available:
```
yeganer: you don't need this.
In the try/except for importing just define `dokuwiki = None` on import error
then you can catch a TypeError? in the next line.

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,233 @@
...
+                self.doku_conn = None
+
+    def add_plot(self, path):
+        """
+        This method is used by any single test method to add the file path of
+        image of corresponding comparison plot.
+        """
+        self.image_path = path
+
+    def _appendrow(self, result, report):
```
yeganer: I thought we settled for using the original `_appendrow` for now or did I miss something?

karandesai-96: I was waiting for the decision of using any one of `extras.html` and `extras.pdf`. We are clear with removing it now.

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,233 @@
...
+            self.doku_conn.medias.add(image_dokulink, self.image_path)
+            self.plot_images.append("{0}{1}|300x225{2} ".format(
+                "{{", image_dokulink, "}}")
+            )
+        else:
+            self.plot_images.append("")
+        # Reset image_path variable for next test method.
+        self.image_path = None
+
+    def _generate_report(self):
```
yeganer: I imagined `_generate_report` to return a string. This string would then be passed to `_save_report`.

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,233 @@
...
+@remote_data
+class DokuReport(HTMLReport):
+
+    def __init__(self, dokuwiki_details):
+        """
+        Initialization of a DokuReport object and registration as a plugin
+        occurs in `pytest_configure`, where a dict containing url, username and
+        password of dokuwiki is passed through `dokuwiki_details`.
+        """
+        logfile = tempfile.TemporaryFile()
```
yeganer: we don't need a logfile

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,233 @@
...
+        self._generate_report()
+        self._save_report()
+
+    def pytest_terminal_summary(self, terminalreporter):
+        """
+        This hook is called by pytest after session ends, and it adds an extra
+        summary at the end. Here, the success / failure of upload of report
+        to dokuwiki is logged.
+        """
+        if self.doku_conn is not None:
```
yeganer: DokuWiki must have a query if a page exists.
use that with try/except

tardis/tests/tests_slow/report.py
karandesai-96: Okay we'll do it this way.

tardis/tests/tests_slow/report.py
karandesai-96: Baseclass needs it. It is not an optional argument.

yeganer:
```
super(DokuReport, self).__init__('')
del self.logfile
```

tardis/tests/tests_slow/report.py
yeganer: For now you can copy/paste the code from pytest-html and note somewhere that it's copy/paste
obviously you need to split it into generate and 'save'

setup.cfg
```diff
@@ -10,6 +10,7 @@ show-response = 1
 [pytest]
 minversion = 2.2
 norecursedirs = build docs/_build
+addopts = -p no:html
```
wkerzendorf: Do we really need this?

tardis/tests/tests_slow/report.py
```diff
@@ -32,10 +32,12 @@ def __init__(self, dokuwiki_details):
                     password=dokuwiki_details["password"])
             except gaierror, dokuwiki.DokuWikiError:
                 self.doku_conn = None
-                print "Dokuwiki connection could not be established!"
         else:
```
yeganer: remove the if/else and add an except for the case dokuwiki=none

tardis/tests/tests_slow/report.py
```diff
@@ -32,10 +32,12 @@ def __init__(self, dokuwiki_details):
...
             self.doku_conn = None
 
+        # This variable will be set True when report gets uploaded on Dokuwiki.
+        self.is_report_uploaded = False
```
yeganer: I wouldn't use a variable but check in sessionfinish that the page is indeed accessible.

tardis/tests/tests_slow/report.py
karandesai-96: - [X] Done.

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,174 @@
...
+from socket import gaierror
+
+from py.xml import html, raw
+from pytest_html.plugin import HTMLReport
+import tardis
+
+try:
+    import dokuwiki
+except ImportError:
+    dokuwiki = None
```
wkerzendorf: why not `dokuwiki_available = False`

yeganer: There is no reason to have the `dokuwiki_available` variable. Since we are
ductyping anyway we can catch the case `dokuwiki == None` in the same block.

Wolfgang Kerzendorf <notifications@github.com> schrieb am Do., 9. Juni 2016
um 13:50 Uhr:

wkerzendorf: I think that makes for convoluted code, where it is not clear why it is excepting. But I withdraw my comment for now.

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,187 @@
...
+            html.p('Report generated on {0} at {1}'.format(
+                generated.strftime('%d-%b-%Y'),
+                generated.strftime('%H:%M:%S'))))
+
+        body.extend(summary)
+        body.extend(results)
+
+        doc = html.html(head, body)
+
+        self.logfile.write(
```
yeganer: As I wrote earlier, don't us the logfile, just save it to a string and return the string

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,187 @@
...
+        self.logfile.write(unicode_doc)
+
+    def _save_report(self):
+        """
+        The method uploads the report and closes the temporary file. Temporary
+        file is made using `tempfile` built-in module, it gets deleted upon
+        closing.
+        """
+        # Go back to the beginning to read everything
+        self.logfile.seek(0)
```
yeganer: add `contents` to the arguments and just add the string to the dokuwiki page

tardis/tests/tests_slow/report.py
```diff
@@ -0,0 +1,187 @@
...
+        occurs in `pytest_configure`, where a dict containing url, username and
+        password of dokuwiki is passed through `dokuwiki_details`.
+        """
+        # A tempfile to log report before it is uploaded to dokuwiki.
+        logfile = tempfile.TemporaryFile()
+
+        # Base class accepts a file path to save the report, but we forcefully
+        # set `self.logfile` to be a file-object.
+        super(DokuReport, self).__init__(logfile.name)
+        self.logfile = logfile
```
yeganer: As I wrote earlier, just `del self.logfile`

tardis/tests/tests_slow/report.py
```diff
@@ -166,7 +161,13 @@ def pytest_terminal_summary(self, terminalreporter):
...
-        if self.is_report_uploaded:
+        try:
+            uploaded_report = self.doku_conn.pages.get(
+                "reports:{0}".format(tardis.__githash__[0:7]))
+        except (gaierror, TypeError):
+            uploaded_report = ""
+
+        if len(uploaded_report) > 0:
             terminalreporter.write_sep(
                 "-", "Successfully uploaded report to Dokuwiki")
```
yeganer: It would be great if you could add the URL, too.

wkerzendorf:
@yeganer is this mergeable? I'm happy to.

PR #584, opened 2016-06-09
[WiP] Add a to_hdf method to the MontecarloRunner.

ftsamis:
For now this is a very simple `to_hdf` method which only stores `output_nu` and `output_energy`.

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -255,7 +255,8 @@ class MontecarloRunner(object):
         except TypeError:
             hdf_store = path_or_buf
 
-        properties = ['output_nu', 'output_energy']
+        properties = ['output_nu', 'output_energy', 'nu_bar_estimator',
+                      'j_estimator']
 
         for prop in properties:
             hdf_store[os.path.join(path, 'runner', prop)] = \
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -1,3 +1,4 @@
+import os
 import logging
 
 from astropy import units as u, constants as const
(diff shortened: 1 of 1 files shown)


tardis/montecarlo/base.py
```diff
@@ -220,3 +222,44 @@ def calculate_f_nu(self, frequency):
 
     def calculate_f_lambda(self, wavelength):
         pass
+
+    def property_to_hdf(self, property):
```
wkerzendorf: I would make this static and just use the value

wkerzendorf: nevermind

tardis/montecarlo/base.py
```diff
@@ -220,3 +222,44 @@ def calculate_f_nu(self, frequency):
...
+        -------
+            : None
+
+        """
+        try:
+            hdf_store = pd.HDFStore(path_or_buf)
+        except TypeError:
+            hdf_store = path_or_buf
+
+        properties = ['output_nu', 'output_energy']
```
wkerzendorf: okay the properties that we want to store are `output_nu`, `output_energy`, `nu_bar_estimator`, `j_estimator`

ftsamis: Added `nu_bar_estimator`, `j_estimator`

tardis/montecarlo/base.py
ftsamis: This needs some kind of change anyway. It does the same thing as `BasePlasmaProperty._get_output_to_hdf`. I'll probably introduce a similar method to `util` and use them everywhere needed.

wkerzendorf: don't let it live here. It's fine.

ftsamis:
Closing this in favour of #585

PR #585, opened 2016-06-09
[TEP002] Add a to_hdf method to the MontecarloRunner

ftsamis:
This is a very simple `to_hdf` method which only stores: `output_nu`, `output_energy`, `nu_bar_estimator`, `j_estimator` to an HDF file.

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -255,7 +255,8 @@ class MontecarloRunner(object):
         except TypeError:
             hdf_store = path_or_buf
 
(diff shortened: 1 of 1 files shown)


tardis/montecarlo/base.py
```diff
@@ -220,3 +222,45 @@ def calculate_f_nu(self, frequency):
 
     def calculate_f_lambda(self, wavelength):
         pass
+
+    def property_to_hdf(self, property):
```
wkerzendorf: let's add a documentation here. But other than the PR should be fine.

PR #586, opened 2016-06-09
[TEP002] Add to_hdf methods to the plasma

ftsamis:
Heavily based on #435. Allow storing the plasma properties to an HDF structure.

`property_collections.PlasmaPropertyCollection`s can be used to only store parameters that are of a type included in the collection, and any of their subclasses.

### DIFF
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -6,6 +6,7 @@ import os
 import numpy as np
 import pandas as pd
 from astropy import constants, units as u
+from astropy.utils.decorators import deprecated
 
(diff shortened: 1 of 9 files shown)


tardis/model.py
```diff
@@ -130,6 +131,15 @@ def __init__(self, tardis_config):
 
 
     @property
+    @deprecated('v1.5', obj_type='attribute')
```
wkerzendorf: Add the description here @ftsamis

PR #587, opened 2016-06-09
Move the three spectra attributes from Radial1DModel to MontecarloRunner

ftsamis:
This is probably a big and a risky change, but after @wkerzendorf's request, I attempted to move `spectrum`, `spectrum_virtual`, `spectrum_reabsorbed` from being `Radial1DModel`'s attributes to `MontecarloRunner`. 

I also changed every reference across the project and its tests to any of these three spectra objects to correctly point to the new location.

All the required code rellocations/removals to achieve the above are:
```
Moved spectrum, spectrum_virtual, spectrum_reabsorbed (Model -> Runner)
    Moved legacy_update_spectrum (Simulation -> Runner)
        Moved montecarlo_virtual_luminosity (Model -> Runner)
    Moved save_spectra (Model -> Runner)
    Completely removed Model.to_hdf5 (will reimplement as part of TEP002)
```
Also, now `MontecarloRunner` takes one extra argument, `distance`, in order to be able to create the spectra objects. I have deliberately made `distance` a required attribute and haven't added a `None` default value, so any call with the old signature will fail instead of silently run with wrong spectra.

In the process I found out that `spectrum_reabsorbed` is actually not used anywhere, there isn't even a reference of it in a test. Should it be removed?


wkerzendorf:
Make sure that model has legacy attributes to spectra with deprecation warnings.

yeganer:
May I suggest moving these properties to `Simulation` instead of `MonteCarloRunner`? As I understood, the runner *should* be only the interface between simulation and C extension.

wkerzendorf:
@yeganer - I disagree. I think that the output of the `montecarlorunner` is the data that produces the spectrum. The simulation has no idea about spectrum. It simply keeps track of iterations and convergence. The way this is done here is sensible in my opinion.

wkerzendorf:
@ftsamis I think this is again a good point where you see that this is a scientific code. Whatever is produced in the code is not necessarily used in the code. The output of this code is analysed using various means including plotting and statistics. so whatever we put out currently needs to stay there (including `spectrum reabsorbed`)

ftsamis:
I'll add the deprecated spectra properties to the model now. I wanted to make sure that Travis passes without them as well.

ftsamis:
I was also wondering if the `TARDISSpectrum` should be moved from model to a separate module.

wkerzendorf:
@ftsamis yes, but it in `spectrum.py` in montecarlo

wkerzendorf:
can you get rid of the distance property. There is nothing wrong running without a distance.

ftsamis:
Should I default it to None on the runner, or not support it at all? In the second case, do we also want it removed from the `TARDISSpectrum` class? Because it is used to calculate `flux_nu`.

wkerzendorf:
default to None.

wkerzendorf:
Then add a to_hdf method saving the luminosity_density.

ftsamis:
@wkerzendorf I think we should merge this and have the `to_hdf` on the to_hdf branch.

tardis/model.py
```diff
@@ -117,17 +118,23 @@ def __init__(self, tardis_config):
...
-        self.spectrum_virtual = TARDISSpectrum(
-            tardis_config.spectrum.frequency, distance)
-        self.spectrum_reabsorbed = TARDISSpectrum(
-            tardis_config.spectrum.frequency, distance)
-
         self.calculate_j_blues(init_detailed_j_blues=True)
         self.update_plasmas(initialize_nlte=True)
 
+    @property
+    @deprecated('v1.5', obj_type='attribute')
```
wkerzendorf: can we add a message saying ' this will disappear the spectrum is now at model.runner.xxx'

wkerzendorf: same goes for the plasma.

tardis/model.py
```diff
@@ -117,17 +118,23 @@ def __init__(self, tardis_config):
...
+    @deprecated('v1.5', 'spectrum will be removed from model. Use model.runner.spectrum instead.')
+    def spectrum(self):
+        return self.runner.spectrum
+
+    @property
+    @deprecated('v1.5', 'spectrum_virtual will be removed from model. Use model.runner.spectrum_virtual instead.')
+    def spectrum_virtual(self):
+        return self.runner.spectrum_virtual
+
+    @property
```
wkerzendorf: @ftsamis can you attach a note to plasma as well.

ftsamis: In the relevant PR (#586), yes, sure.

2016-06-13 18:03 GMT+03:00 Wolfgang Kerzendorf <notifications@github.com>:

PR #589, opened 2016-06-14
[TEP002] Add a to_hdf method to the Radial1DModel

ftsamis:
Added a `model.to_hdf` method which in turn calls `plasma.to_hdf` and also stores `t_inner`, `ws`, `t_rads`, `v_inner`, `v_outer`.

Additionally, `atom_data_uuid` is also stored in a Series under the HDF structure path `model/metadata`.

### DIFF
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -90,6 +90,9 @@ class Radial1DModel(object):
 
         self.t_inner = tardis_config.plasma.t_inner
 
+        self.v_inner = tardis_config.structure.v_inner
+        self.v_outer = tardis_config.structure.v_outer
+        self.v_middle = 0.5 * (self.v_inner + self.v_outer)
 
         self.ws = self.calculate_geometric_w(
             tardis_config.structure.r_middle,
@@ -97,7 +100,7 @@ class Radial1DModel(object):
 
         if tardis_config.plasma.t_rads is None:
             self.t_rads = self._init_t_rad(
-                self.t_inner, tardis_config.structure.v_inner[0], self.v_middle)
+                self.t_inner, self.v_inner[0], self.v_middle)
         else:
             self.t_rads = tardis_config.plasma.t_rads
 
@@ -115,8 +118,8 @@ class Radial1DModel(object):
             link_t_rad_t_electron=0.9,
             helium_treatment=tardis_config.plasma.helium_treatment,
             heating_rate_data_file=heating_rate_data_file,
-            v_inner=tardis_config.structure.v_inner,
-            v_outer=tardis_config.structure.v_outer)
+            v_inner=self.v_inner,
+            v_outer=self.v_outer)
 
         distance = tardis_config.supernova.get('distance', None)
         self.spectrum = TARDISSpectrum(
@@ -164,11 +167,6 @@ class Radial1DModel(object):
     @property
     def t_inner(self):
         return self._t_inner
-
-    @property
-    def v_middle(self):
-        structure = self.tardis_config.structure
-        return 0.5 * (structure.v_inner + structure.v_outer)
 
     @t_inner.setter
     def t_inner(self, value):
--- a/tardis/model.py
+++ b/tardis/model.py
@@ -250,116 +250,62 @@ class Radial1DModel(object):
         self.spectrum.to_ascii(fname)
         self.spectrum_virtual.to_ascii('virtual_' + fname)
 
-
-    def to_hdf5(self, buffer_or_fname, path='', close_h5=True):
+    def to_hdf(self, path_or_buf, path='', plasma_properties=None):
         """
-            This allows the model to be written to an HDF5 file for later analysis. Currently, the saved properties
-            are specified hard coded in include_from_model_in_hdf5. This is a dict where the key corresponds to the
-            name of the property and the value describes the type. If the value is None the property can be dumped
-            to hdf via its attribute to_hdf or by converting it to a pd.DataFrame. For more complex properties
-            which can not simply be dumped to an hdf file the dict can contain a function which is called with
-            the parameters key, path, and  hdf_store. This function then should dump the data to the given
-            hdf_store object. To dump  properties of sub-properties of  the model, you can use a dict as value.
-            This dict is then treated in the same way as described above.
+        Store the model and the runner to an HDF structure.
+        Properties that have 1D arrays as their values are stored
+        as columns of a pandas.DataFrame under `path`, called arrays.
+        Properties that have a scalar value are stored as labeled
+        elements in a pandas.Series under `path`, called scalars.
+
+        All Quantities will be stored as their SI values and without
+        their unit.
 
         Parameters
         ----------
+        path_or_buf
+            Path or buffer to the HDF store
+        path : str
+            Path inside the HDF store to store the plasma
+        plasma_properties
+            `None` or a `PlasmaPropertyCollection` which will
+            be passed as the collection argument to the
+            plasma.to_hdf method.
 
-        buffer_or_fname: buffer or ~str
-            buffer or filename for HDF5 file (see pandas.HDFStore for description)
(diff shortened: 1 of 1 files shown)


tardis/model.py
```diff
@@ -250,116 +250,48 @@ def save_spectra(self, fname):
...
-        hdf_store.flush()
-        if close_h5:
+        try:
+            hdf_store = pd.HDFStore(path_or_buf)
+        except TypeError:
+            hdf_store = path_or_buf
+
+        self.plasma.to_hdf(hdf_store, os.path.join(path, 'model'),
+                           False, plasma_properties)
+        try:
```
wkerzendorf: remove this - runner will be stored as part of simulation

tardis/model.py
```diff
@@ -250,116 +250,48 @@ def save_spectra(self, fname):
...
+            self.runner.to_hdf(hdf_store, path, False)
+        except AttributeError:
+            logger.warn("runner cannot be written to the HDF structure as "
+                        "there is no runner attached to the model.")
+        model_information = pd.Series({'atom_data_uuid': self.atom_data.uuid1})
+        hdf_store[os.path.join(path,
+                               'model', 'information')] = model_information
+
```
wkerzendorf: store t_inner, ws, trads, vinner, vouter

wkerzendorf: remove ws, trads from plasma.to_hdf

tardis/model.py
```diff
@@ -250,116 +250,48 @@ def save_spectra(self, fname):
...
-
-
-        hdf_store.flush()
-        if close_h5:
+        try:
+            hdf_store = pd.HDFStore(path_or_buf)
+        except TypeError:
+            hdf_store = path_or_buf
+
+        self.plasma.to_hdf(hdf_store, os.path.join(path, 'model'),
```
yeganer: I guess this should read `join(path, 'plasma')` right?

tardis/model.py
```diff
@@ -250,116 +250,48 @@ def save_spectra(self, fname):
...
+                           False, plasma_properties)
+        try:
+            self.runner.to_hdf(hdf_store, path, False)
+        except AttributeError:
+            logger.warn("runner cannot be written to the HDF structure as "
+                        "there is no runner attached to the model.")
+        model_information = pd.Series({'atom_data_uuid': self.atom_data.uuid1})
+        hdf_store[os.path.join(path,
+                               'model', 'information')] = model_information
```
yeganer: I'd prefer `model_information.to_hdf(...)`

tardis/model.py
```diff
@@ -250,116 +250,48 @@ def save_spectra(self, fname):
...
-                            include_from_model_in_hdf5[key][subkey](subkey, os.path.join(path, key), hdf_store)
-                        else:
-                            logger.critical('Can not save %s', str(os.path.join(path, key, subkey)))
-                except:
-                    logger.critical('An error occurred while dumping %s to HDF.', str(os.path.join(path, key)))
-
-
-        hdf_store.flush()
-        if close_h5:
+        try:
```
yeganer: I don't think we need this try/except block as we are working completely with pandas. That's why I think we should let pandas decide whether `path_or_buf` is a path or buf.

tardis/model.py
ftsamis: Nice point. Will change it.

tardis/model.py
ftsamis: Not really. I made it that each `to_hdf` method has to name itself, and so the path argument for each `to_hdf` should be only the parent path. Therefore in `plasma.to_hdf` will in its turn do: `prop.to_hdf(hdf_store, os.path.join(path, 'plasma'))`

tardis/model.py
ftsamis: This is another nice point. Though, one thing that works better in the current way, is that the HDF file (if it is a file and not an HDFStore) isn't constantly opened and closed like it would be if we did pass the path to pandas, but I guess we can take this, since we are not talking about a really huge amount of IO operations.

tardis/model.py
```diff
@@ -250,116 +250,44 @@ def save_spectra(self, fname):
...
-            This dict is then treated in the same way as described above.
+        Store the model and the runner to an HDF structure.
 
         Parameters
         ----------
+        path_or_buf
+            Path or buffer to the HDF store
+        path : str
+            Path inside the HDF store to store the plasma
+        close_hdf : bool
```
yeganer: Do we need `close_hdf`? I think we don't need it.

tardis/model.py
yeganer: Oh yeah, sorry forgot about that.

tardis/model.py
ftsamis: No, if we go with passing the `path_or_buf` straight to pandas, we don't. I removed it in another fixup for model. It will need to be removed from the rest `to_hdf` methods too.

tardis/model.py
```diff
@@ -268,6 +268,16 @@ def to_hdf(self, path_or_buf, path='', plasma_properties=None):
...
 
         """
+        properties = ['t_inner', 'ws', 't_rads', 'v_inner', 'v_outer']
+        for p in properties:
+            output_value = getattr(self, p)
+            if hasattr(output_value, 'si'):
+                output_value = output_value.si.value
+            if np.isscalar(output_value):
+                output_value = [output_value]
+            data = pd.DataFrame(output_value)
```
ftsamis: Yeah, I hadn't noticed that they are all the same length. Is that always the case? Because otherwise this won't work.

yeganer: Are they exactly the same length? I don't know but it's not important. If
one element is longer than the rest the others get filled with np.nan. One
would have to take care of that when reading the valus, that's all


Fotis Tsamis <notifications@github.com> schrieb am Mo., 20. Juni 2016 um
17:18 Uhr:

ftsamis: As far as I checked it does not do that automatically.

```python
>>>pd.DataFrame({'a': [1,2], 'b' : [1,2,3]})
ValueError: arrays must all be same length
```

while:
```python
>>> pd.DataFrame({'a': [1,2], 'b' : [1,2]})
   a  b
0  1  1
1  2  2
```

So one would have to take care of manually adding np.nan for the missing values of shorter columns as well.

yeganer: You want to do it this way:
```python
df = pd.DataFrame()
df['a'] = [1,2,3]
df['b'] = [1,2]
```

Fotis Tsamis <notifications@github.com> schrieb am Mo., 20. Juni 2016 um
18:56 Uhr:

ftsamis: That was unexpected to me. Alright, thanks for pointing this out.

wkerzendorf:
@tardis-sn/tardis-core can someone else look at this - from my point of view it looks mergeable.

tardis/model.py
ftsamis: Are you sure this works? Because the exact snippet you gave throws a `ValueError: Length of values does not match length of index` for me.

wkerzendorf: I think there is a high likelihood that this doesn't work. But all of the values in model have the same length.

ftsamis: I was actually trying to do the same with the plasma properties, so we could have a uniform way of storing things.

ftsamis:
Note to self: After this is merged, I should rebase #601 on to_hdf and change the `plasma.to_hdf` call which is made in the `model.to_hdf` to match the new signature.

wkerzendorf:
@tardis-sn/tardis-core looks good to me - should we merge?

ftsamis:
I think we should merge this, since either way it's going to face some minor changes because of some additions to the way data are stored to HDF in #601. Then, when both this and #601 are merged we can discuss if we are happy to merge the entire `to_hdf` branch to `master`. For now, this also blocks #601 from having its couple last commits.

PR #590, opened 2016-06-15
Link images of comparison plots in DokuReport using hookwrapper.

karandesai-96:
This PR introduces a common object for temporary holding comparison plots and uploading them on DokuWiki. For this purpose, a fixture named `plot_object` has been created in **tests_slow/conftest.py**. It returns an object of `PlotUploader` class.

Inclusions in this PR:
- [X] `PlotUploader` class in **tests_slow/plot_helpers.py**
- [X] `plot_object` fixture in **tests_slow/conftest.py** which returns object of `PlotUploader`class.
- [X] `plot_spectrum` of ``TestW7` returns a figure which is passed directly to `PlotUploader.upload()`
- [X] `PlotUploader.upload()` handles upload to dokuwiki.

### DIFF
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -48,23 +48,48 @@ def pytest_runtest_makereport(item, call):
     outcome = yield
     report = outcome.get_result()
 
-    dokuwiki_url = item.config.dokureport.dokuwiki_url
-    # TODO: remove hardcoded name w7 after fixture parametrization
-    plots = glob.glob(os.path.join(item.config.option.tempdir, "w7", "*.png"))
-    extra = getattr(report, "extra", [])
-
     if report.when == "call":
-        for plot in plots:
-            item.config.dokureport.doku_conn.medias.add("plots:{0}_{1}".format(
-                tardis.__githash__[0:7], plot.split("/")[-1]), plot
-            )
-            extra.append(extras.url(
-                "{0}lib/exe/fetch.php?media=plots:{1}_{2}".format(
-                    dokuwiki_url, tardis.__githash__[0:7], plot.split("/")[-1]
+        plot_obj = getattr(item, "plot_obj", None)
+        if plot_obj is not None:
+            report.extra = plot_obj.get_extras()
+
+
+@pytest.fixture(scope="function")
+def plot_object(request):
+    class PlotUploader(object):
+        def __init__(self):
+            self._plots = list()
+            self.plot_links = list()
+
+        def add(self):
+            # TODO: make this accept pyplot.figure later.
+            self._plots = glob.glob(os.path.join(
+                request.config.option.tempdir, "w7", "*.png"))
+
+        def upload(self):
+            dokuwiki_url = request.config.dokureport.dokuwiki_url
+            for plot in self._plots:
+                request.config.dokureport.doku_conn.medias.add("plots:{0}_{1}".format(
+                    tardis.__githash__[0:7], plot.split("/")[-1]), plot
                 )
-            ))
-            os.unlink(plot)
-        report.extra = extra
+                self.plot_links.append(extras.url(
+                    "{0}lib/exe/fetch.php?media=plots:{1}_{2}".format(
+                        dokuwiki_url, tardis.__githash__[0:7], plot.split("/")[-1]
+                    )
+                ))
+                os.unlink(plot)
+
+        def get_extras(self):
+            return self.plot_links
+    plot_obj = PlotUploader()
+    setattr(request.node, "plot_obj", plot_obj)
+
+    def fin():
+        # Adding plots in this manner is temporary.
+        plot_obj.add()
+        plot_obj.upload()
+    request.addfinalizer(fin)
+    return plot_obj
 
 
 @pytest.fixture(scope="session")
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -1,12 +1,15 @@
 import os
+import glob
 import shutil
 import tempfile
 import yaml
 import numpy as np
 import pytest
+from pytest_html import extras
 from astropy import units as u
 
 from tardis.tests.tests_slow.report import DokuReport
+import tardis
 
 
 def pytest_configure(config):
@@ -37,6 +40,31 @@ def pytest_unconfigure(config):
         config.pluginmanager.unregister(config.dokureport)
     # Remove tempdir by recursive deletion
     shutil.rmtree(config.option.tempdir)
+
+
+@pytest.mark.hookwrapper
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    report = outcome.get_result()
+
+    dokuwiki_url = item.config.dokureport.dokuwiki_url
+    # TODO: remove hardcoded name w7 after fixture parametrization
+    plots = glob.glob(os.path.join(item.config.option.tempdir, "w7", "*.png"))
+    extra = getattr(report, "extra", [])
+
+    if report.when == "call":
+        for plot in plots:
+            item.config.dokureport.doku_conn.medias.add("plots:{0}_{1}".format(
+                tardis.__githash__[0:7], plot.split("/")[-1]), plot
+            )
+            extra.append(extras.url(
+                "{0}lib/exe/fetch.php?media=plots:{1}_{2}".format(
+                    dokuwiki_url, tardis.__githash__[0:7], plot.split("/")[-1]
+                )
+            ))
+            os.unlink(plot)
+        report.extra = extra
 
 
 @pytest.fixture(scope="session")
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -45,11 +45,10 @@ def pytest_runtest_makereport(item, call):
     # execute all other hooks to obtain the report object
     outcome = yield
     report = outcome.get_result()
-
     if report.when == "call":
         if "plot_object" in item.fixturenames:
             plot_obj = item.funcargs["plot_object"]
-            plot_obj.upload()
+            plot_obj.upload(report)
             report.extra = plot_obj.get_extras()
 
 
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -61,23 +61,31 @@ def plot_object(request):
             self._plots = list()
             self.plot_links = list()
 
-        def add(self):
-            # TODO: make this accept pyplot.figure later.
-            self._plots = glob.glob(os.path.join(
-                request.config.option.tempdir, "w7", "*.png"))
+        def add(self, plot, name):
+            """
+            Accept a `matplotlib.pyplot.figure` and add it to self._plots.
+            """
+            self._plots.append((plot, name))
 
         def upload(self):
+            """
+            Upload the content in self._plots to dokuwiki.
+            """
             dokuwiki_url = request.config.dokureport.dokuwiki_url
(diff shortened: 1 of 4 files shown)


tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
     shutil.rmtree(config.option.tempdir)
 
 
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
```
karandesai-96: Nothing needs to be done before a test is executed.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
     shutil.rmtree(config.option.tempdir)
 
 
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
```
karandesai-96: I inserted `print(report.when)` here, and I got:
```
setup
call
.teardown
```
(. because of passing test)

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
...
+    outcome = yield
+    report = outcome.get_result()
+
+    dokuwiki_url = item.config.dokureport.dokuwiki_url
+    # TODO: remove hardcoded name w7 after fixture parametrization
+    plots = glob.glob(os.path.join(item.config.option.tempdir, "w7", "*.png"))
```
karandesai-96: The test method has saved its plots in this directory, so all of them are picked up.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
...
+    extra = getattr(report, "extra", [])
+
+    if report.when == "call":
+        for plot in plots:
+            item.config.dokureport.doku_conn.medias.add("plots:{0}_{1}".format(
+                tardis.__githash__[0:7], plot.split("/")[-1]), plot
+            )
```
karandesai-96: Image uploaded.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
...
+            extra.append(extras.url(
+                "{0}lib/exe/fetch.php?media=plots:{1}_{2}".format(
+                    dokuwiki_url, tardis.__githash__[0:7], plot.split("/")[-1]
+                )
+            ))
```
karandesai-96: URL added to html report.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
...
+            os.unlink(plot)
```
karandesai-96: Image deleted locally, so that it is not collected for next test.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,31 @@ def pytest_unconfigure(config):
...
+        report.extra = extra
```
karandesai-96: Finally `report.extra = extra` is required.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,53 @@ def pytest_unconfigure(config):
...
+                    )
+                ))
+                os.unlink(plot)
+
+    plot_obj = PlotUploader()
+    setattr(request.node, "plot_object", plot_object)
+
+    def fin():
+        # Adding plots in this manner is temporary.
+        plot_obj.add()
```
yeganer: The way I imagine it, `plot_obj.add()` is called by the test function.

karandesai-96: Yes it is temporary. It will be changed when `pyplot.figures` are directly passed on.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +42,53 @@ def pytest_unconfigure(config):
     shutil.rmtree(config.option.tempdir)
 
 
+@pytest.mark.hookwrapper
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    report = outcome.get_result()
+    plot_object = getattr(item, "plot_object", None)
```
yeganer: This hookwrapper get's executed for every test.
You should check whether the test has an associated `plot_obj` or not. Only if there is a `plot_obj` a plot should be uploaded and the report may be modified.

I suggest to put all the logic inside the PlotUploader and the to something like `report.extra = plot_obj.get_extra()`.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +40,25 @@ def pytest_unconfigure(config):
...
 
 
+@pytest.mark.hookwrapper
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    report = outcome.get_result()
+
+    if report.when == "call":
+        plot_obj = getattr(item, "plot_obj", None)
```
yeganer: I think we can use `item.funcargs['plot_object']`
Then we don't have to setattr in the fixture

tardis/tests/tests_slow/plot_helpers.py
```diff
@@ -0,0 +1,32 @@
+import tempfile
+from pytest_html import extras
+import tardis
+
+
+class PlotUploader(object):
```
yeganer: I'd add an `add` method for adding plots and call `upload()` in `pytest_runtest_makereport`. Then we can give the report object to `upload()` and `upload` adds stuff like the `passed/failed` text depending on the report

tardis/tests/tests_slow/plot_helpers.py
```diff
@@ -17,12 +17,21 @@ def add(self, plot, name):
...
         Upload the content in self._plots to dokuwiki.
         """
         for plot, name in self._plots:
             plot_file = tempfile.NamedTemporaryFile(suffix=".png")
+            axes = plot.axes[0]
+
+            if report.passed:
+                axes.text(0.8, 0.8, 'passed', transform=axes.transAxes,
+                            bbox={'facecolor': 'green', 'alpha': 0.5, 'pad': 10})
+            elif report.failed:
```
yeganer: I think we should just use `else:`
Otherwise it we might get nothing.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -112,31 +112,25 @@ def test_luminosity_inner(self):
...
+
+        assert_quantity_allclose(
+            self.reference['wavelength'],
+            self.result.runner.spectrum.wavelength)
+
+        assert_quantity_allclose(
+            self.reference['luminosity_density_lambda'],
+            self.result.runner.spectrum.luminosity_density_lambda)
+
+        plot_object.add(self.plot_spectrum(), "spectrum")
```
yeganer: We should plot first, then assert everything.

tardis/tests/tests_slow/test_w7.py
```diff
@@ -178,34 +164,21 @@ def test_montecarlo_properties(self):
...
-            plot = self.plot_t_rads(has_passed=True)
-
-        plot_object.add(plot, "t_rads")
-
-    def plot_t_rads(self, has_passed):
+        assert_quantity_allclose(
+            self.reference['t_rads'],
+            self.result.t_rads)
+
+        plot_object.add(self.plot_t_rads(), "t_rads")
```
yeganer: same here:
plot first then assert
otherwise we won't get a plot if there is an assert.

tardis/tests/tests_slow/plot_helpers.py
karandesai-96: Other case is `report.skipped`. I think we don't need a plot here.

tardis/tests/tests_slow/plot_helpers.py
karandesai-96: - [X] Done.

tardis/tests/tests_slow/conftest.py
```diff
@@ -39,6 +40,25 @@ def pytest_unconfigure(config):
...
+        if "plot_object" in item.fixturenames:
+            plot_obj = item.funcargs["plot_object"]
+            plot_obj.upload(report)
+            report.extra = plot_obj.get_extras()
+
+
+@pytest.fixture(scope="function")
+def plot_object(request):
+    plot_obj = PlotUploader(request)
+    setattr(request.node, "plot_obj", plot_obj)
```
yeganer: Wo don't need this line anymore.

tardis/tests/tests_slow/plot_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+                            bbox={'facecolor': 'red', 'alpha': 0.5, 'pad': 10})
+
+            plot.savefig(plot_file.name)
+
+            self.request.config.dokureport.doku_conn.medias.add(
+                "plots:{0}_{1}.png".format(tardis.__githash__[0:7], name),
+                plot_file.name
+            )
+
+            thumbnail_html = """
```
yeganer: Can you define this HTML snippet as a constant at the top?

tardis/tests/tests_slow/report.py
```diff
@@ -144,6 +144,11 @@ def _generate_report(self, session):
             )
         )
         report_content += doc.unicode(indent=2)
+
+        # Quick hack for preventing log to be placed in narrow left out space
```
yeganer: As this line only affects the result part of the whole log I'd move it so it only works on `result` just after `result = ...`

yeganer:
I made some last comments. I think we are done  and ready to merge once they are implemented.

@wkerzendorf Can you please have a quick look and merge once the comments are worked on?

tardis/tests/tests_slow/conftest.py
karandesai-96: @yeganer I'm further simplifying to:
```python
@pytest.fixture(scope="function")
def plot_object(request):
    return PlotUploader(request)
```
... in the next commit

tardis/tests/tests_slow/plot_helpers.py
karandesai-96: - [X] Done.

tardis/tests/tests_slow/plot_helpers.py
```diff
@@ -4,6 +4,15 @@
 import tardis
 
 
+thumbnail_html = """
+<div class="image" style="float: left">
+    <a href="#">
+        <img src= "{0}lib/exe/fetch.php?media=plots:{1}_{2}.png" />
```
yeganer: As a last exercise, please name the arguments, so it's obvious which is which

yeganer:
@wkerzendorf I think we are done here. Please merge if you are happy, too.

tardis/tests/tests_slow/plot_helpers.py
```diff
@@ -0,0 +1,61 @@
...
+class PlotUploader(object):
+    def __init__(self, request):
+        self.request = request
+        self._plots = list()
+        self.plot_html = list()
+        self.dokuwiki_url = self.request.config.dokureport.dokuwiki_url
+
+    def add(self, plot, name):
+        """
+        Accept a `matplotlib.pyplot.figure` and add it to self._plots.
```
wkerzendorf: can you please do `numpydoc` style documentation. for all functions that have documentation.

wkerzendorf:
@karandesai-96 looks good. Merging when it passes

PR #591, opened 2016-06-20
Fix path of CSS and JS files used in html report.

karandesai-96:
In **tardis/tests/tests_slow/report.py**:
```python
 style_css = pkg_resources.resource_string(
            __name__, os.path.join('resources', 'style.css'))
```
and
```python
 main_js = pkg_resources.resource_string(
            __name__, os.path.join('resources', 'main.js'))
```

...read from the `pytest-html`'s files. 

But here, `__name__` = 'tardis/tests/tests_slow/report', which is incorrect.  
So it is changed to `pytest_html.__name__` for avoiding `OSError`.

### DIFF
--- a/tardis/tests/tests_slow/report.py
+++ b/tardis/tests/tests_slow/report.py
@@ -33,6 +33,7 @@ import time
 from socket import gaierror
 
 from py.xml import html, raw
+from pytest_html import __name__ as pytest_html_path
 from pytest_html.plugin import HTMLReport
 import tardis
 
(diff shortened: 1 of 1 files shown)


tardis/tests/tests_slow/report.py
```diff
@@ -33,6 +33,7 @@
 from socket import gaierror
 
 from py.xml import html, raw
+import pytest_html
 from pytest_html.plugin import HTMLReport
```
yeganer: I'd add `import __name__ as pytest_path`

yeganer:
@tardis-sn/tardis-core I'm happy to merge. This is a small bugfix.

PR #595, opened 2016-06-21
Line absorption rate estimator

Tobychev:
Summary
========
A start on the adding of calculating spectra using exact source function integration: an  line absorption rate estimator. 

Background
=========
In Lucy 1999, A&A 345, a method of calculating emerging spectra based on the formal integral of the spectra is suggested, and part of that method requires adding an estimator of the line absorption rate that can be simplified to: 

Edotlu

where epsilon is the packet energy, tau is the optical depth of each transition, and the sum is over all packets that pass trough the _l_ -> _u_ transition in each cell. Only the sum explicitly needs to be inside cmontecarlo.c, the other factors can be done in python

To do
====
- [x] Add the sum estimator to cmonecarlo
- [x] Fix up the plumbing so the sum estimator can be reached from python
- [ ] Make unit tests for the sum estimator
- [x] Normalize the sums
- [x] Multiply with the extinction factor that depends on the tau for each transition


tardis/montecarlo/src/cmontecarlo.h
```diff
@@ -13,6 +13,10 @@
 #define LOG_VPACKETS 0
 #endif
 
+#define RESONANCE 0
+#define DOWNBRANCH 1
+#define MACROATOM 2
```
unoebauer: I'd rather work with enums and not use preprocessor statements.

tardis/montecarlo/src/cmontecarlo.h
```diff
@@ -13,6 +13,10 @@
 #define LOG_VPACKETS 0
 #endif
 
+#define RESONANCE 0
```
wkerzendorf: could this be an enum?

unoebauer:
@Tobychev - looks already very good. Maybe let's change the title of the PR because I think this one should be only about the line absorption rate (or at most about the level absorption rate). Also, please make a list of milestones which you want to achieve with this PR

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -373,6 +373,19 @@ increment_j_blue_estimator (const rpacket_t * packet, storage_model_t * storage,
...
 
+void
+increment_Edotlu_estimator (const rpacket_t * packet, storage_model_t * storage, int64_t line_idx)
+{
+  if (storage->line_lists_Edotlu != NULL)
+    {
+#ifdef WITHOPENMP
+#pragma omp atomic
+#endif
+    storage->line_lists_Edotlu[line_idx] += rpacket_get_energy (packet);
```
unoebauer: This is not yet the final estimator, right?

Tobychev: It's all that we said is needed in the montecarlo code. There is a factor $$1 - e^{\tau_{lu}}$$ and some normalization needed

unoebauer: no - the factor 1 - exp(-tau_s) is not global. Instead it is different for each line interaction. As the subscribed lu suggests, it is a quantity which depends on the line transition. In fact, this is the Sobolev line optical depth. It will be different in each shell and for each line.

unoebauer: **To clarify**: the factor (1 - exp(-tau_s)) describes the interaction probability, i.e. the probability of a packet which comes into resonance with the line to interact with it.

Tobychev: Yes it depends on the transition, I just meant to say that the tau for each line is added to the estimator(s, the code above is also for each transition and each shell) once all monte carlo steps have finished.

unoebauer: @Tobychev - hmm, I guess you could do that. But I think it should be easier to add the factor already during the increment. This way you avoid looping over all entries at the end again.

unoebauer:
@Tobychev - I think one milestone should be the addition of a unittest for your estimator increment routine

tardis/model.py
```diff
@@ -241,3 +242,11 @@ def update_plasmas(self, initialize_nlte=False):
...
             self.transition_probabilities = (
                 self.plasma_array.transition_probabilities)
+  
+    def postprocess(self):
+        self.Edotlu_norm_factor = (1 / 
+            (self.time_of_simulation * self.tardis_config.structure.volumes))
+        exptau = 1 - np.exp(- 
+                            self.runner.line_lists_tau_sobolevs.reshape(-1,
+                                self.tardis_config.structure["velocity"]["num"]) ) 
+        self.Edotlu = self.Edotlu_norm_factor*exptau*self.Edotlu_estimators
```
unoebauer: I wouldn't do this post-processing here. As I've said, the 1- exp(-tau) factor should already be included in the estimator increments in the C part.

unoebauer:
@Tobychev - I wouldn't do any "estimator post-processing" in the python part. We want to have the line absorption rate estimators ready and available on the C or at least on the cython level. This way, we can do the intensive number crunching (calculating the level absorption rates and ultimately performing the integral on the C or cython level)

unoebauer:
@wkerzendorf, @yeganer - it is unclear to us why the unit tests for the estimator doesn't work. Apparently ``line_lists_Edotlu`` is not properly allocated. Any ideas?

I'd like to merge this PR as soon as possible, even if the unit tests are broken (we can mark them skip, right?)

tardis/montecarlo/tests/test_cmontecarlo.py
```diff
@@ -119,6 +119,9 @@ def model():
         line_lists_j_blues=(c_double * 2)(*([1.e-10] * 2)),
         line_lists_j_blues_nd=0,
 
+        line_lists_Edotlu=(c_double * 3)(1e-10,1e-10,1.0), # Init to an explicit array
```
karandesai-96: @Tobychev Please replace this initialization with
```
(c_couble * 3)(*[1.e-10, 1.e-10, 1.0])
```
I think this might be causing the problem. @yeganer what do you think about it ?

tardis/montecarlo/struct.py
```diff
@@ -61,6 +61,7 @@ class StorageModel(Structure):
         ('line_lists_tau_sobolevs_nd', c_int64),
         ('line_lists_j_blues', POINTER(c_double)),
         ('line_lists_j_blues_nd', c_int64),
+        ('line_lists_Edotlu', POINTER(c_double)),
```
yeganer: I think here's your bug.
The order matters for the definition of the storage_model struct fields. You defined them as
```
   double *line_lists_j_blues;
  double *line_lists_Edotlu;
   int64_t line_lists_j_blues_nd;
```

tardis/montecarlo/tests/test_cmontecarlo.py
```diff
@@ -119,6 +119,9 @@ def model():
         line_lists_j_blues=(c_double * 2)(*([1.e-10] * 2)),
         line_lists_j_blues_nd=0,
 
+        line_lists_Edotlu=(c_double * 3)(1e-10,1e-10,1.0), # Init to an explicit array 
+        line_lists_Edotlu_nd=0,
```
yeganer: this doesn't exist and shouldn't, as it's the same as tau_sob_nd

tardis/montecarlo/tests/test_cmontecarlo.py
```diff
@@ -426,7 +429,23 @@ def test_montecarlo_line_scatter(packet_params, expected_params, packet, model,
...
+----------------
+These test check very simple pices of code still work.
+"""
+
+@pytest.mark.parametrize(
+    ['packet_params', 'line_idx', 'expected'],
+    [({'energy':0.0}, 0, 0),
+     ({'energy':1.0}, 1, 1),
+     ({'energy':0.5}, 2, 1.5)]
+)
```
yeganer: I know the parametrization in the code is currently done mostly this way but I think we should do new tests differently.
Have a look at https://github.com/yeganer/tardis/blob/4511369c245b6a6fb9a33a6891089e20efac56a3/tardis/montecarlo/tests/test_cmontecarlo.py#L282 to see how I wrote the tests.

yeganer:
@unoebauer I added some comments including where I mention a possible cause for the bug.
Concerning the merge, I have to disagree with you. I'm not familiar with the planned implementation but I'm there are some thing with the current implementation I don't like, which could be improved. As we have the policy `don't touch a running system unless it's needed` I'd fix them right away.

tardis/montecarlo/base.py
```diff
@@ -106,7 +107,7 @@ def legacy_update_spectrum(self, no_of_virtual_packets):
             self.spectrum_virtual.update_luminosity(
                 self.montecarlo_virtual_luminosity)
 
-    def run(self, model, no_of_packets, no_of_virtual_packets=0, nthreads=1):
+    def run(self, model, no_of_packets, no_of_virtual_packets=0, nthreads=1,last_run=False):
```
yeganer: I think we could get away without introducing this new argument.
What we should do (if possible), is to move the postprocess to `MontecarloRunner.calculate_Edotlu(self, model)`.
This method could then be called anytime. It would make sense for Simulation to call this function after It's done with it's last iteration.

unoebauer: Hmm - @Tobychev had this idea initially. I was against this approach, since we need the post-processed estimators on the C or cython level. That's why I favoured the current approach.

tardis/montecarlo/tests/test_cmontecarlo.py
Tobychev: I think the line you link to is quite a bit harder to read... is it very much faster to declare in tuples, or it is the immutability you advocate for?

unoebauer:
Looks good to me now - I'd like to merge. @tardis-sn/tardis-core, objections?

tardis/montecarlo/tests/test_cmontecarlo.py
```diff
@@ -119,8 +119,7 @@ def model():
         line_lists_j_blues=(c_double * 2)(*([1.e-10] * 2)),
         line_lists_j_blues_nd=0,
 
-        line_lists_Edotlu=(c_double * 3)(1e-10,1e-10,1.0), # Init to an explicit array 
-        line_lists_Edotlu_nd=0,
+        line_lists_Edotlu=(c_double * 3)(*[0.0,0.0,1.0]), # Init to an explicit array
```
yeganer: Why are we initializing this estimator to something nonzero?

tardis/montecarlo/base.py
wkerzendorf: Postprocessing the results can still be done on the C/Cython level even if we did this using `calculate_Edotlu` later. I think having a postprocessing class is a sensible step (this also goes for spectral creation out of the packets).

wkerzendorf: But we can definitely do this in another PR.

tardis/montecarlo/tests/test_cmontecarlo.py
wkerzendorf: not quite sure what you meant.

Tobychev: To see that in properly increments in that case too, might as well test that right?

Issue #593, opened 2016-06-20
Downbranch crashing using tardis_example


Tobychev:
Setting `line_interaction_type: downbranch` in tardis_example.yml while also introducing the changes necessary to get around #577 leads to the example crashing with an unbound local exception:
```
tardis.simulation.base - INFO - t_inner 10458.724 K -- next t_inner 10795.859 K
tardis.model - INFO - Calculating J_blues for radiative_rates_type=dilute-blackbody
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tobychev/ownCloud/socis/tardis/tardis/base.py", line 39, in run_tardis
    simulation.run_radial1d(radial1d_mdl)
  File "/home/tobychev/ownCloud/socis/tardis/tardis/simulation/base.py", line 347, in run_radial1d
    simulation.legacy_run_simulation(radial1d_model)
  File "/home/tobychev/ownCloud/socis/tardis/tardis/simulation/base.py", line 261, in legacy_run_simulation
    convergence_section.global_convergence_parameters.
UnboundLocalError: local variable 'convergence_section' referenced before assignment
```

yeganer:
As nobody is currently using the convergence_strategy I'm going to voice the (probably very unpopular) opinion to delete everything related to the current convergence and reimplement it (from scratch) when the changes to the config, model and runner are done.

unoebauer:
Yes - as pointed out in #577, the convergence criteria have been utterly neglected in the recent past. I am however not sure if a small subset of the convergence scheme is used by @orbitfold. Maybe @wkerzendorf can comment on that. In principle, I'd be okay with just deactivating all of this since in practice we are typically using a fixed number of iterations.

yeganer:
@unoebauer The way the convergence criteria is currently used by @orbitfold and me is by setting the threshold to 0.001 which translates to disabling the convergence and running 20 iterations fixed

unoebauer:
Quickfix for now:

- change ``convergence_criteria`` to ``convergence_strategy``
- change ``hold`` to ``hold_iterations``

This should do the trick!

Tobychev:
That helps one around #577, if you also want the `line_interaction_type` to be downbranch, it crashes with the reported error (after doing 20 steps).

Tobychev:
Following @yeganer's suggestion to set the threshold to 0.001 does avoid this crash for me.

yeganer:
[...]

PR #598, opened 2016-06-22
Fix convergence handling in run_legacy_simulation - fix issue #593

unoebauer:
This PR fixes the convergence handling in simulation/base.py, but only in the legacy_simulation function. The changes (should) ensure that all accessed quantities exist. However, this PR does not change anything in the (dubious) convergence handling.

### DIFF
--- a/tardis/simulation/base.py
+++ b/tardis/simulation/base.py
@@ -213,19 +213,22 @@ class Simulation(object):
     def legacy_run_simulation(self, model):
         start_time = time.time()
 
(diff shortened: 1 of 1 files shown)


wkerzendorf:
I'm happy to merge this. @yeganer I wouldn't say we need a test now, since we might go down your road and redo the convergence methods.

yeganer:
@wkerzendorf That makes sense.
@tardis-sn/tardis-core I'll merge in 20 minutes if nobody is objecting.

PR #604, opened 2016-06-24
Carsus/atomic

mishinma:
This PR moves `atomic.py` to the `io` package and replaces the old `from_hdf5` method with a new method `from_hdf`. The `from_hdf` method is used for instantiating `AtomData` from HDFStores created with carsus. 

The following DataFrame names are used in Carsus HDFStores: `atom_data`, `ionization_data`, `levels`, `lines`, `macro_atom_data`, `macro_atom_references`, `zeta_data`, `collision_data`, `collision_data_temperatures`, `synpp_refs`, `ion_cx_th_data`, `ion_cx_sp_data`. The DataFrames are described in `AtomData`'s docstring.
Some of the DataFrames are related to each other, e.g. `collision_data` and `collision_data_temperatures`. Related dataframes can only be supplied together; otherwise, the error is raised. All related groups are defined in the `related_groups` attribute of `AtomData` class.

Lastly, the `prepare_atom_data` method was refactored. The main objective of the refactoring was to ensure  #that views of the DataFrames were not set as the attributes of `AtomData`, only copies. This is important because when views were set as the attributes,  the simulation **was not performing correctly**.  Apparently,  the bug is related to the C part of the code (maybe when passing numpy arrays using `.values`; `.values` returns a view, not a copy). Another objective of the refactoring was to improve readability of the code  and its conformity with PEP8 and pandas guidelines. It is likely that in future the `prepare_atom_data` method will be removed from `AtomData` and moved to the `plasma` module.


unoebauer:
Closing this one - reopen if you think otherwise, @wkerzendorf, @yeganer

PR #609, opened 2016-06-28
From hdfstore

mishinma:
In this PR I have:
- Ditched all the "read" methods.
- Replaced the `from_hdf5` classmethod of `AtomData` with `from_hdfstore`. I used  the try-except blocks to check if the dataframe is present in the hdfstore. A question: which dataframes are mandatory for the initialization?
- Refactored the `__init__` method of `AtomData`. A question: do we really need the "has_some_dataframe" attribures on the instances of `AtomData`? 
- Written a better docstring for `AtomData.


tardis/io/atomic.py
```diff
@@ -310,51 +128,85 @@ def from_hdf5(cls, fname=None):
...
-                        ion_cx_data=ion_cx_data)
-
-        with h5py.File(fname, 'r') as h5_file:
-            atom_data.uuid1 = h5_file.attrs['uuid1']
-            atom_data.md5 = h5_file.attrs['md5']
-            atom_data.version = h5_file.attrs.get('database_version', None)
-
-            if atom_data.version is not None:
-                atom_data.data_sources = pickle.loads(h5_file.attrs['data_sources'])
+        with pd.HDFStore(fname) as store:
```
yeganer: I have a **very bold** suggestion:
```
kwargs={}
for k in store.keys():
    kwargs[k] = store[k]

cls(**kwargs)
```

wkerzendorf: not really - then we should expose the HDFStore directly.

mishinma: Some of the arguments - like `collision_data` - are tuples.

yeganer: Yes, I know. But it would be possible, to go around that. For the `collision data`, one could simply add the unit straight away. It makes far more sense for `__init__` to accept a dataframe instead of a tuple.

mishinma: I agree, but I'm afraid to change the __init__ parameters at the moment (and for collision data it takes tuples). Anyway, the current goal is to make a successful simulation with carsus's dataframes. Once that is accomplished I will make more substantial changes to the `AtomData` class.

PR #612, opened 2016-06-29
Add facility to generate and accept HDF files as reference data for Integration Tests.

karandesai-96:
This PR introduces facility to generate reference data for Integration tests. Sometimes, due to breaking changes, the integration tests may not pass, but as the changes are suitable, the reference data has to be updated.

#### From the "outside":
* A parameter named `generate_reference` is to be added in YAML file holding configuration for integration tests. It specifies the path to directory where the reference data will be generated.
* This parameter is normally ignored, but becomes active when `--generate-reference` is specified as command line argument.
* At the end of test run, the model for each setup (for example _W7, AT_) will be stored to an _HDF5_ file and saved in the directory path mentioned, as per githash.

#### From the "inside":
* When reference data is to be generated, only _TARDIS_ run with setup is needed, and assertions are to be skipped. Moreover the dokureport uploading mechanism must be turned off.
* A key named `gen_ref_dirpath` is added to dict return by `data_path` fixture, which provides the path for storing models, to `TestIntegration` class.

#### Checklist for the PR:
- [X] Facilitate generation of reference data and storing as HDF files as described above.
- [X] Make `reference` fixture to ingest directly from HDF files instead of `.npz` compressed binaries.
- [X] Make suitable changes in assertions of `TestIntegration` class.

[Latest Report](http://opensupernova.org/~karandesai96/integration/doku.php?id=reports:363256d)

### DIFF
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -30,8 +30,8 @@ def pytest_configure(config):
 
 def pytest_unconfigure(config):
     # Unregister only if it was registered in pytest_configure
-    integration_tests_configpath = config.getvalue("integration-tests")
-    if integration_tests_configpath and not config.getoption("--generate-reference"):
+    if (config.getvalue("integration-tests") and not
+            config.getoption("--generate-reference")):
         config.pluginmanager.unregister(config.dokureport)
 
 
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -69,25 +69,20 @@ def plot_object(request):
 def data_path(request):
     integration_tests_config = request.config.option.integration_tests_config
     setup_name = os.path.basename(request.param)
-    return {
+
+    path = {
         'config_dirpath': request.param,
         'reference_dirpath': os.path.join(os.path.expandvars(
             os.path.expanduser(integration_tests_config["reference"])), setup_name
         ),
+        'gen_ref_dirpath': os.path.join(os.path.expandvars(os.path.expanduser(
+            integration_tests_config["generate_reference"])), tardis_githash[:7]
+        ),
         'setup_name': setup_name
     }
-
-
-@pytest.fixture(scope="session")
-def gen_ref_dirpath(request):
-    if request.config.getoption("--generate-reference"):
-        integration_tests_config = request.config.option.integration_tests_config
-        path = os.path.join(os.path.expandvars(os.path.expanduser(
-            integration_tests_config["generate_reference"])), tardis_githash[:7]
-        )
-        os.makedirs(os.path.join(path))
-    else:
-        path = None
+    if (request.config.getoption("--generate-reference") and not
+            os.path.exists(path['gen_ref_dirpath'])):
+        os.makedirs(path['gen_ref_dirpath'])
     return path
 
 
@@ -117,6 +112,11 @@ def reference(request, data_path):
         * luminosity_density_nu          | * wavelength
         * delta_frequency                | * luminosity_density_lambda
     """
+    # Reference data need not be loaded and provided if current test run itself
+    # generates new reference data.
+    if request.config.getoption("--generate-reference"):
+        return
+
     reference_dirpath = data_path['reference_dirpath']
 
     # TODO: make this fixture ingest data from an HDF5 file.
--- a/tardis/tests/tests_slow/conftest.py
+++ b/tardis/tests/tests_slow/conftest.py
@@ -1,9 +1,8 @@
 import glob
 import os
 import yaml
-import numpy as np
+import pandas as pd
 import pytest
-from astropy import units as u
 
 from tardis import __githash__ as tardis_githash
 from tardis.tests.tests_slow.report import DokuReport
@@ -69,17 +68,17 @@ def plot_object(request):
 ])
 def data_path(request):
     integration_tests_config = request.config.option.integration_tests_config
-    setup_name = os.path.basename(request.param)
+    hdf_filename = "{0}.h5".format(os.path.basename(request.param))
 
     path = {
         'config_dirpath': request.param,
-        'reference_dirpath': os.path.join(os.path.expandvars(
-            os.path.expanduser(integration_tests_config["reference"])), setup_name
+        'reference_filepath': os.path.join(os.path.expandvars(
+            os.path.expanduser(integration_tests_config["reference"])), hdf_filename
         ),
         'gen_ref_dirpath': os.path.join(os.path.expandvars(os.path.expanduser(
             integration_tests_config["generate_reference"])), tardis_githash[:7]
         ),
-        'setup_name': setup_name
+        'setup_name': hdf_filename[:-3]
     }
     if (request.config.getoption("--generate-reference") and not
             os.path.exists(path['gen_ref_dirpath'])):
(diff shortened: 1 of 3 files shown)


tardis/tests/tests_slow/conftest.py
```diff
@@ -19,6 +22,12 @@ def pytest_configure(config):
         config.option.integration_tests_config = yaml.load(
             open(integration_tests_configpath))
 
+        if "generate_reference" not in config.option.integration_tests_config:
+            config.option.generate_reference['generate_reference'] = None
```
yeganer: What's happening here?

tardis/tests/tests_slow/conftest.py
```diff
@@ -11,6 +11,9 @@
 
 
 def pytest_configure(config):
+    # This boolean is set True when integration tests as run, provided current
+    # integration test run is not for generating reference data.
```
karandesai-96: **_TODO_**: I need to rephrase it. I typed it wrong.

tardis/tests/tests_slow/conftest.py
```diff
@@ -33,13 +42,23 @@ def pytest_configure(config):
 
 
 def pytest_unconfigure(config):
-    integration_tests_configpath = config.getvalue("integration-tests")
-    if integration_tests_configpath is not None:
+    # Unregister only if it was registered in pytest_configure
+    if config.option.dokureport_active:
         config.pluginmanager.unregister(config.dokureport)
```
yeganer: Is it necessary to unregister the plugin?

karandesai-96: Usually third party plugins have this `pytest_unconfigure`, where there are unregistered, once their job is done. As they do not assume anything outside in the main test run. But in our case, I think we can remove this part.

tardis/tests/tests_slow/conftest.py
karandesai-96: If `generate_reference` is not at all mentioned, it throws `KeyError`. Just to avoid `KeyError` in the fixture down below. I can do this in fixture itself, but I had to make this comparison for setting `dokureport_active` anyhow.

yeganer: I was refering to `config.option.generate_reference` which looks like a typo

karandesai-96: Oh that, I fixed it.

tardis/tests/tests_slow/conftest.py
```diff
@@ -5,34 +5,53 @@
...
 def pytest_unconfigure(config):
+    # Unregister only if it was registered in pytest_configure
     integration_tests_configpath = config.getvalue("integration-tests")
-    if integration_tests_configpath is not None:
+    if integration_tests_configpath and not config.getoption("--generate-reference"):
         config.pluginmanager.unregister(config.dokureport)
 
 
+def pytest_terminal_summary(terminalreporter):
+    if terminalreporter.config.getoption("--generate-reference") and (
```
yeganer: The parenthesis should wrap the whole condition, not only the last part.

karandesai-96: Then I would require a `\` linebreak, which we decided to avoid earlier

yeganer: No.
```
if (foo and
    bar):
    print('foobar')
```
is valid python

yeganer:
Other than that this looks really good.

tardis/tests/tests_slow/conftest.py
karandesai-96: Oh okay, I think I should have a close look at these very micro lints :grinning:

wkerzendorf:
@yeganer good to merge?

yeganer:
Not yet. I'd like to reproduce this locally first.

tardis/tests/tests_slow/conftest.py
```diff
@@ -66,8 +78,21 @@ def data_path(request):
     }
 
 
+@pytest.fixture(scope="session")
+def gen_ref_dirpath(request):
+    if request.config.getoption("--generate-reference"):
+        integration_tests_config = request.config.option.integration_tests_config
+        path = os.path.join(os.path.expandvars(os.path.expanduser(
+            integration_tests_config["generate_reference"])), tardis_githash
```
karandesai-96: @wkerzendorf @yeganer How about naming the directory as shortened githash ( `tardis_githash[:7]` ) ?

yeganer: That was my original idea.
We might want fixtures for `githash` and `gitshort`.

yeganer:
I think this will still fail if I don't have some *.npy reference files. Can you please fix that?

tardis/tests/tests_slow/test_integration.py
```diff
@@ -64,67 +64,67 @@ def setup(self, request, reference, data_path, atomic_data_fname):
...
 
     def test_j_blue_estimators(self):
         assert_allclose(
-                self.reference['j_blue_estimators'],
+                self.reference['/simulation/model/plasma/j_blues'],
                 self.result.j_blue_estimators)
 
         assert_quantity_allclose(
-                self.reference['j_blues_norm_factor'],
+                self.reference['j_blues_norm_factor'],   # FIXME
```
karandesai-96: No guess.

tardis/tests/tests_slow/test_integration.py
```diff
@@ -64,67 +64,67 @@ def setup(self, request, reference, data_path, atomic_data_fname):
...
 
     def test_last_line_interactions(self):
         assert_allclose(
-                self.reference['last_line_interaction_in_id'],
+                self.reference['last_line_interaction_in_id'],   # FIXME
                 self.result.last_line_interaction_in_id)
 
         assert_allclose(
-                self.reference['last_line_interaction_out_id'],
+                self.reference['last_line_interaction_out_id'],   # FIXME
```
karandesai-96: `/simulation/model/plasma/lines_upper_level_index` ?

tardis/tests/tests_slow/test_integration.py
```diff
@@ -64,67 +64,67 @@ def setup(self, request, reference, data_path, atomic_data_fname):
...
 
         assert_quantity_allclose(
-                self.reference['j_blues_norm_factor'],
+                self.reference['j_blues_norm_factor'],   # FIXME
                 self.result.j_blues_norm_factor)
 
     def test_last_line_interactions(self):
         assert_allclose(
-                self.reference['last_line_interaction_in_id'],
+                self.reference['last_line_interaction_in_id'],   # FIXME
```
karandesai-96: `/simulation/model/plasma/lines_lower_level_index` ?

tardis/tests/tests_slow/test_integration.py
```diff
@@ -64,67 +64,67 @@ def setup(self, request, reference, data_path, atomic_data_fname):
...
                 self.result.last_line_interaction_in_id)
 
         assert_allclose(
-                self.reference['last_line_interaction_out_id'],
+                self.reference['last_line_interaction_out_id'],   # FIXME
                 self.result.last_line_interaction_out_id)
 
         assert_allclose(
-                self.reference['last_line_interaction_shell_id'],
+                self.reference['last_line_interaction_shell_id'],   # FIXME
```
karandesai-96: `/simulation/model/plasma/lines` ?

tardis/tests/tests_slow/test_integration.py
```diff
@@ -44,17 +44,31 @@ def setup(self, reference, data_path, atomic_data_fname):
...
         self.result = Radial1DModel(tardis_config)
-        simulation = Simulation(tardis_config)
-        simulation.legacy_run_simulation(self.result)
+
+        # If current test run is just for collecting reference data, store the
+        # output model to HDF file, save it at specified path. Skip all tests.
+        # Else simply perform the run and move further for performing
+        # assertions.
+        if request.config.getoption("--generate-reference"):
+            run_radial1d(self.result, os.path.join(
```
yeganer: We should pass the path as kwarg so it's obvious what the path is

yeganer:
Please upload a report with reference files generated by this PR.

tardis/tests/tests_slow/conftest.py
```diff
@@ -1,38 +1,50 @@
...
+                config.dokureport = DokuReport(
+                    config.option.integration_tests_config['dokuwiki'])
+                config.pluginmanager.register(config.dokureport)
 
 
 def pytest_unconfigure(config):
+    # Unregister only if it was registered in pytest_configure
     integration_tests_configpath = config.getvalue("integration-tests")
-    if integration_tests_configpath is not None:
+    if integration_tests_configpath and not config.getoption("--generate-reference"):
```
yeganer: We don't need integration_tests_configpath as variable here, use `config.getvalue` directly

karandesai-96: I used it for maintaining 80 characters limit :sweat_smile: Though I'll add a linebreak anyhow

PR #613, opened 2016-06-29
Fix bug in montecarlo.pyx

yeganer:
The number of shells was extracted from tardis_config which fails if the
structure is given by a file. Now the no_of_shells is extracted from the
shape of the j_estimator.

Trailing spaces were removed.


yeganer:
@tardis-sn/tardis-core Please merge ASAP as the underlying bug breaks all setups with a structure file.

PR #614, opened 2016-06-29
Omit the integration_tests when running coveralls

yeganer:


yeganer:
@wkerzendorf @tardis-sn/tardis-core Looks like this is working. Please merge.

PR #615, opened 2016-06-29
Formal source function integral

Tobychev:
Implementing the source integral method of Lucy 99b.
# To do
- [x] Add way to calculate source function
- [x] Add way to integrate source function
- [ ] Add continuum source of opacity
- [ ] Make sure integrator gives sensible results
- [ ] Consider  if 448 could be done without making lots of NaN's
- [ ] Consider adding a p at exactly R_min


tardis/simulation/tests/test_simulation.py
```diff
@@ -61,3 +73,15 @@ def test_packet_output(simulation_one_loop, simulation_compare_data):
     assert_quantity_allclose(simulation_one_loop.runner.output_energy,
                         simulation_compare_data['test1/output_energy'] * u.Unit('erg'),
                         atol=0.0 * u.Unit('erg'))
+
+
+def test_make_source_function(simulation_one_loop,raw_model):
+    # For now just test that the method can be called without exceptions
+    succes = False
```
yeganer: You don't need these exception guards. You can simply run the function. If it throws an error, the test fails. If you expect an error, you can use `with pytest.raise(TypeError):` which will cause the test to fail if that exception was NOT raised.

Tobychev:
There is a bit of the integral in this one now...

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+        dtau = 1 # Just to remember it 
+
+        L_nu  = np.zeros(nus.shape)
+
+        import pdb
+        for nu_idx,nu in enumerate(nus.value):
+            I_outer = np.zeros(ps_outer.shape)
+            for p_idx,p in enumerate(ps_outer):
+                z_cross_p = z_ct_outer[z_ct_outer[:,p_idx] > 0,p_idx]
+                z_cross_p = np.hstack((-z_cross_p,z_cross_p[::-1][1:],0)) # Zero ensures empty ks in last step below
```
yeganer: You are add the `0` so `nu_end` will be 0 for the last iteration, right?

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+
+        dtau = 1 # Just to remember it 
+
+        L_nu  = np.zeros(nus.shape)
+
+        import pdb
+        for nu_idx,nu in enumerate(nus.value):
+            I_outer = np.zeros(ps_outer.shape)
+            for p_idx,p in enumerate(ps_outer):
+                z_cross_p = z_ct_outer[z_ct_outer[:,p_idx] > 0,p_idx]
```
yeganer: You can 'precompute' `z_cross_p` as a 2D array outside of the loop and filter for `>0` there (might not be easy if different rows have a different number of columns.

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+        transitions = model.atom_data.macro_atom_data[model.atom_data.macro_atom_data.transition_type == -1].copy()
+        transitions_index = transitions.set_index(['atomic_number', 'ion_number', 'source_level_number']).index.copy()
+        tmp = model.plasma.transition_probabilities[(model.atom_data.macro_atom_data.transition_type == -1).values]
+        q_ul = tmp.set_index(transitions_index)
+        return (model.atom_data.lines.wavelength[transitions.transition_line_id].values.reshape(-1,1) * 
+                 (q_ul * e_dot_u) * model.tardis_config.supernova.time_explosion / (4*np.pi) )
+
+    def integrate(self,model):
+        num_shell, = self.runner.volume.shape
+        ps         = np.linspace(0.999, 0, num = 3 * num_shell)
```
yeganer: Any reason this is backwards. It seems the order doesn't make a difference.

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+        J_blues    = model.j_blues
+        J_rlues    = model.j_blues.values * np.exp( -model.plasma.tau_sobolevs.values) + self.runner.att_S_ul
+
+        r_shells = np.zeros((num_shell+1,1))
+        # Note the reorder from outer to inner
+        r_shells[1:,0],r_shells[0,0] = self.runner.r_inner_cgs[::-1] / R_max, 1.0
+        z_crossings = np.sqrt(r_shells**2 - ps**2)
+
+        z_ct = z_crossings/ct
+        z_ct[np.isnan(z_crossings)] = 0
```
yeganer: I can't think of a situation (where everything went smoothly) where `z_crossings` turns out to be nan as `r_shells` goes from 1 to `R_min_rel` where `ps` goes from 0.999 to 0. If `z_crossings` is nan, something went wrong, right?

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+                    shell = shell_idx[idx]
+                    # Note the direction of the comparisons
+                    ks, = np.where( (line_nu < nu_start) & (line_nu >= nu_end) )                    
+
+                    if len(ks) < 2:
+                        continue
+
+                    I_outer[p_idx] = dtau * ( 
+                                        ( J_rlues.iloc[ks[0],shell] + J_blues.iloc[ks[1],shell] ) / 2 )
+
```
mkromer: @Tobychev Is this supposed to be an implementation of eqs. 27 and 28 in Lucy99b? As I told you yesterday on Skype my python skills are limited, but I think there might be a problem here. Also: the idea was to start the implementation of the formal integral without continuum processes. For this case eqs. 27 and 28 are not required, just proceed recursively through the line list, i.e. I_{k+1}^b=I_k^r

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+                    nu_end   = nu / (1 + z_cross_p[idx+1])
+                    shell = shell_idx[idx]
+                    # Note the direction of the comparisons
+                    ks, = np.where( (line_nu < nu_start) & (line_nu >= nu_end) )                    
+
+                    if len(ks) < 2:
+                        continue
+
+                    I_inner[p_idx] = intensity_black_body(nu,temps[shell])
+
```
mkromer: @Tobychev Does this initialize the intensity to the Planck function at the photosphere (i.e. photospheric temperature). I am surprised that you use a shell index. But again I am not familiar enough with the syntax or the data structures.

yeganer: As far as I know, the blackbody uses `t_inner`.

tardis/simulation/base.py
Tobychev: Lines 503 and 504 are indeed 27 & 28, but they are just included as a reminder.  They are run once per impact parameter, at the start, so when I_k^r = 0. The dtau is completely fake however, so maybe I should just remove it and see if that helps. (I doubt it, earlier I had a bug in the code and while attempting to fix it I verified that these lines didn't produce very significant contributions, they are of order 1 at most)

tardis/simulation/base.py
Tobychev: Yes, I use a shell temperature as a stopgap because I couldn't find an "official" photosphere temperature in the final model. That said, it is the inner most shell and it differs from what the simulation states is the photosphere temperature with some hundred Kelvin at most, which for this model is at the percent level.

tardis/simulation/base.py
Tobychev: The short answer is no, due to some dank broadcasting magic line 448 results in an upper triangular matrix, where the lower half is all NaN:s. 

I think what happens is that when r_shells gets extended in the broadcast it gets padded with zeros, but I'm actually not sure what is going on.

 It is a good point that now I lose the ability to detect code elsewhere giving rise to NaN before this line.

yeganer: I see, r_shells is a 2D array. Do you need it to be a 2D array?
Because in 447 you are only setting one dimension.

Please enlighten me on what `z_crossings` should look like? What dimension do you expect?

tardis/simulation/base.py
Tobychev: I can't tell what you refer to, please specify further.

yeganer: the ordering of all your arrays seems to be from outside to the inside. Any reason for that?
iirc if you did for example `np.linspace(0,1,3*num_shell)` there is an option to exclude the last boundary, i think.
In any case you'd wouldn't have to use `[:,:,-1]` for accessing the radius for example.

tardis/simulation/base.py
Tobychev: Kinda, z enters as 1 + z, so actually what happens is that nu_end < nu_start. Still this makes for a empty selection which is the important part.

tardis/simulation/base.py
Tobychev: The outside in order leads to a situation where complexity increases as you step further, which I think is a simpler way to organize the work. For example, if you start from the outside, rays at the first impact parameter (the ps) will only intersect the outer shell, while at smaller impact parameters many more shells could be intersected and need to be included.

yeganer: Let's say I just did the calculation for shell 2, can my result be used in any way for shell 1?
If that's the case:
Iterative approach favors out->in while recursive approach favors in->out (for readability and consistency with the rest of the codebase among other things)

tardis/simulation/base.py
```diff
@@ -401,6 +406,137 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+                    nu_start = nu / (1 + z_cross) 
+                    nu_end   = nu / (1 + z_cross_p[idx+1])
+                    shell = shell_idx[idx]
+                    # Note the direction of the comparisons
+                    ks, = np.where( (line_nu < nu_start) & (line_nu >= nu_end) )                    
+
+                    if len(ks) < 2:
+                        continue
+
+                    I_inner[p_idx] = intensity_black_body(nu,temps[0])
```
yeganer: I think there might be an error here as you loop over `z_cross_p` while only modifying `I_Inner[p_index]`

tardis/simulation/base.py
yeganer: In any case you should comment them out if they aren't doing what you want them to do. I see several problems here.
`ks` could be anything and you are explicitly accessing `ks[0]` and `ks[1]` without knowing if there's more.
the way I understand it, `ks` is a mask of all the lines that have a matching nu. so accessing only one element looks suspicious to me

tardis/simulation/base.py
yeganer: Regarding the possible `NaNs` in `z_ct`:
`np.nan_to_num(array)` converts all nans to 0, -inf to -1e308 and inf to 1e308.

tardis/simulation/base.py
```diff
@@ -401,6 +406,145 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+                    #dtau * ( 
+                    #( J_rlues.iloc[ks[0],shell] + J_blues.iloc[ks[1],shell] ) / 2 )
+
+                    for k in ks:
+                        I_inner[p_idx] = I_inner[p_idx] * np.exp(-taus.iloc[k,shell]) + att_S_ul.iloc[k,shell]
+
+            if ( nu_idx % 100 ) == 0:
+                print "{:3.0f} %".format( 100*float(nu_idx)/len(nus))
+            ps = np.hstack((ps_outer,ps_inner))*R_max
+            I_nu = np.hstack((I_outer,I_inner))*ps
```
mkromer: I would prefer a different variable name here, since this is not I_nu but I_nu*p

Tobychev: Sure

tardis/simulation/base.py
```diff
@@ -401,6 +406,147 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+
+        upper_level_index = model.atom_data.lines.set_index(['atomic_number', 'ion_number', 'level_number_upper']).index.copy()
+        e_dot_lu = pd.DataFrame(self.runner.Edotlu, index=upper_level_index)
+        e_dot_u = e_dot_lu.groupby(level=[0, 1, 2]).sum()
+        e_dot_u.index.names = ['atomic_number', 'ion_number', 'source_level_number'] # To make the q_ul e_dot_u product work, could be cleaner
+        transitions = model.atom_data.macro_atom_data[model.atom_data.macro_atom_data.transition_type == -1].copy()
+        transitions_index = transitions.set_index(['atomic_number', 'ion_number', 'source_level_number']).index.copy()
+        tmp = model.plasma.transition_probabilities[(model.atom_data.macro_atom_data.transition_type == -1).values]
+        q_ul = tmp.set_index(transitions_index)
+        return (model.atom_data.lines.wavelength_cm[transitions.transition_line_id].values.reshape(-1,1) *
```
yeganer: You might want to split this line across multiple lines to make it more readable.

tardis/simulation/base.py
```diff
@@ -401,6 +406,147 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+        Formally it calculates the expresion ( 1 - exp(-tau_ul) ) S_ul but this product is what we need later,
+        so there is no need to factor out the source function explicitly.
+
+        Parameters
+        ----------
+        model : tardis.model.Radial1DModel
+
+        Returns
+        -------
+        DataFrame containing ( 1 - exp(-tau_ul) ) S_ul
```
yeganer: This is wrong, there are ~3 variables returned

tardis/simulation/base.py
```diff
@@ -401,6 +406,150 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+        """
+
+        upper_level_index = model.atom_data.lines.set_index(['atomic_number', 'ion_number', 'level_number_upper']).index.copy()
+        e_dot_lu          = pd.DataFrame(self.runner.Edotlu, index=upper_level_index)
```
yeganer: Please don't aligning in python.
Rather group the assignments and use empty lines to seperate them.

Tobychev: Is that a pep8 requirement?

tardis/simulation/base.py
```diff
@@ -401,6 +406,146 @@ def to_hdf(self, model, path_or_buf, path='', plasma_properties=None):
...
+        e_dot_u.index.names = ['atomic_number', 'ion_number', 'source_level_number'] # To make the q_ul e_dot_u product work, could be cleaner
+        transitions       = model.atom_data.macro_atom_data[model.atom_data.macro_atom_data.transition_type == -1].copy()
+        transitions_index = transitions.set_index(['atomic_number', 'ion_number', 'source_level_number']).index.copy()
+        tmp  = model.plasma.transition_probabilities[(model.atom_data.macro_atom_data.transition_type == -1).values]
+        q_ul = tmp.set_index(transitions_index)
+        t    = model.tardis_config.supernova.time_explosion.value
+        wave = model.atom_data.lines.wavelength_cm[transitions.transition_line_id].values.reshape(-1,1)
+        reorder  = wave[:,0].argsort()
+        att_S_ul =  ( wave * (q_ul * e_dot_u) * t  / (4*np.pi) ).iloc[reorder,:]
+
```
unoebauer: Not sure that everything goes as planned here. Maybe use a more simplistic approach for testing purposes:
```python
        line_ids = model.atom_data.lines.index
        upper_levels = model.atom_data.lines.level_number_upper.values
        lower_levels = model.atom_data.lines.level_number_lower.values
        t = model.tardis_config.supernova.time_explosion

        Edot_u = {}
        j_ul = {}
        S_ul = {}

        for i in range(1, int(max(upper_levels)+1), 1):
            Edot_u[i] = model.runner.Edotlu[upper_levels == i].sum()

        for i in xrange(len(line_ids)):
            line_id = line_ids[i]
            j_ul[line_id] = 1. / (4. * np.pi) * model.plasma.transition_probabilities.ix[line_id].values * Edot_u[iup[i]]
            S_ul[line_id] = (j_ul[line_id] * t * u.AA * model.atom_data.lines.ix[line_id].wavelength).to("1/cm^2")
```

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,161 @@
...
+        }
+        printf("\n\n");
+    }
+}
+
+
+void integrate_source_functions(double* L_nu, double* line_nu, double* taus, double* att_S_ul, double* I_BB, double* nus, 
+              double* ps, double* Rs, double R_ph, int64_t* lens)
+{
+    double I_nu[lens[SHELLEN]];
```
yeganer: I'd declare as
`double  *I_nu = calloc(lens[PLEN], sizeof(double))`

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,161 @@
...
+    double inv_t = 1.8432e+42;
+    int no_of_shell_cr;
+    indexpair_t nu_lims;
+    printf("Test: bajs!");
+    for (int nu_idx = 0; nu_idx < lens[NULEN]; ++nu_idx)
+    {
+        memset(I_nu,0.0,sizeof(I_nu));
```
yeganer: `memset(I_nu, 0.0, lens[PLEN] * sizeof(I_nu))`

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,161 @@
...
+        for (int p_idx = 0; p_idx < lens[PLEN]; ++p_idx)
+        {
+            if (ps[p_idx] < R_ph) 
+            {
+                I_nu[nu_idx] = I_BB[nu_idx];
```
yeganer: `I_nu[p_idx]`?

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,161 @@
...
+            }           
+            no_of_shell_cr = get_num_shell_cr(ps[p_idx],Rs,lens[SHELLEN]);
+            for (int cr_idx = get_cr_start(no_of_shell_cr, ps[p_idx], R_ph); cr_idx < 2*no_of_shell_cr; ++cr_idx)
+            {
+                nu_lims = find_nu_limits_for_crossing_and_p(nus[nu_idx], ps[p_idx], cr_idx, no_of_shell_cr, Rs, inv_t);
+                for (int k_idx = nu_lims.start; k_idx < nu_lims.end; ++k_idx)
+                {
+                    I_nu[nu_idx] = I_nu[nu_idx] * exp(-taus[k_idx * lens[SHELLEN] + get_sh_idx(cr_idx,no_of_shell_cr)])
```
yeganer: `I_nu[p_idx]`?

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,161 @@
...
+            {
+                nu_lims = find_nu_limits_for_crossing_and_p(nus[nu_idx], ps[p_idx], cr_idx, no_of_shell_cr, Rs, inv_t);
+                for (int k_idx = nu_lims.start; k_idx < nu_lims.end; ++k_idx)
+                {
+                    I_nu[nu_idx] = I_nu[nu_idx] * exp(-taus[k_idx * lens[SHELLEN] + get_sh_idx(cr_idx,no_of_shell_cr)]) 
+                                    + att_S_ul[k_idx * lens[SHELLEN] + get_sh_idx(cr_idx,no_of_shell_cr)];
+                }
+            }
+        }
+        L_nu[nu_idx] = integrate_intensity(I_nu, ps, lens);
```
yeganer: `lens[PLEN]`

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,161 @@
...
+                I_nu[nu_idx] = I_BB[nu_idx];
+            }           
+            no_of_shell_cr = get_num_shell_cr(ps[p_idx],Rs,lens[SHELLEN]);
+            for (int cr_idx = get_cr_start(no_of_shell_cr, ps[p_idx], R_ph); cr_idx < 2*no_of_shell_cr; ++cr_idx)
+            {
+                nu_lims = find_nu_limits_for_crossing_and_p(nus[nu_idx], ps[p_idx], cr_idx, no_of_shell_cr, Rs, inv_t);
+                for (int k_idx = nu_lims.start; k_idx < nu_lims.end; ++k_idx)
+                {
+                    I_nu[nu_idx] = I_nu[nu_idx] * exp(-taus[k_idx * lens[SHELLEN] + get_sh_idx(cr_idx,no_of_shell_cr)]) 
+                                    + att_S_ul[k_idx * lens[SHELLEN] + get_sh_idx(cr_idx,no_of_shell_cr)];
```
yeganer: `k * no_of_shells + current_shell_id` is a suboptimal way of indexing the data. If you use `storage_model_t` the data will be in the `k + current_shell_id * no_of_shells` format. Please keep that in mind and switch to `storage_model_t` because that makes EVERYTHING easier (coding on your part, bug fixing for all of us, maintenance for us)

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,223 @@
...
+        z_cr = sqrt( Rs[cr_idx]*Rs[cr_idx] - p*p );
+        nu_blu = nu * (1 + z_cr*inv_ct);
+        nu_red = nu * (1 - z_cr*inv_ct);
+    }
+
+    pair = nu_idx_from_nu_pair(nu_blu, nu_red, line_nu, len);
+ 
+    return pair;
+}
+indexpair_t nu_idx_from_nu_pair(double nu_blu, double nu_red, const double* line_nu, int len)
```
yeganer: have a look at `line_search` from cmontecarlo.c/.h
This does the same but uses an optimized algorithm.

tardis/montecarlo/src/rpacket.c
```diff
@@ -15,9 +15,11 @@ rpacket_init (rpacket_t * packet, storage_model_t * storage, int packet_index,
   double current_nu = storage->packet_nus[packet_index];
   double current_energy = storage->packet_energies[packet_index];
   double current_mu = storage->packet_mus[packet_index];
-  double comov_current_nu = current_nu;
+//  double comov_current_nu = current_nu;
   int current_shell_id = 0;
   double current_r = storage->r_inner[0];
+  double comov_current_nu = current_nu * (1 - (current_mu * current_r * storage->inverse_time_explosion * INVERSE_C)); 
+/*
```
unoebauer: Why did you change this? If this is really a bug, it is disjunct from the formal integral and should be fixed in separate patch (i.e. PR)

tardis/montecarlo/src/integrator.c
unoebauer: Good point!

tardis/montecarlo/src/rpacket.c
Tobychev: What do you mean? Its exactly like you told me to change it. This is part of handling the CMF - LF initialization of the BB field at the photosphere, which we had to to to match the MC. Discussion [here](https://gitter.im/tardis-sn/tardis?at=579b83b0f1da4f376e186512) (hopefully)

unoebauer: Oh - I remember - sorry, my bad!

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,259 @@
...
+        blu_R = get_r(cr_idx,no_of_cr_shells,Rs);
+        red_R = get_r(cr_idx+1,no_of_cr_shells,Rs);
+        z_blu = sqrt( blu_R*blu_R - p*p );
+        z_red = sqrt( red_R*red_R - p*p );
+        nu_blu = nu * (1 - get_cr_sign(cr_idx,no_of_cr_shells)*z_blu*inv_ct);
+        nu_red = nu * (1 - get_cr_sign(cr_idx+1,no_of_cr_shells)*z_red*inv_ct);
+    }
+    else 
+    {
+        if (p > Rs[1]) {
```
yeganer: This line is confusing. Is this some testing for the single shell case?

Tobychev: yes, its the single shell branch of the test on line 26.

tardis/montecarlo/src/integrator.c
```diff
@@ -0,0 +1,259 @@
...
+    }
+    result += I_nu[len-1]*h*ps[len-1]/2.0;
+    return result*8*M_PI*M_PI;
+}
+void integrate_source_functions(double* L_nu, const double* line_nu, const double* taus, const double* att_S_ul, const double* nus,
+        const double* ps, const double* Rs, double T, double R_ph, double inv_ct, const int64_t* lens)
+{
+    double* I_nu  = calloc(lens[PLEN], sizeof(double));
+    int no_of_cr_shells, sh_idx, cr_start, cr_end;
+    int kludge = 0;
```
Tobychev: I forgot, but it also shouldn't matter since it serves to fix a off-by-one problem with the indexing once our p-rays cross the photosphere. Since we never leave the photosphere after entering it during correct execution, resetting it probably isn't an issue.

But yeah, a kludge to get the shell indexing working correctly in the photosphere. The python version doesn't count the photosphere as a shell that is crossed, but it is natural to do so in the c version since it is one loop: however, because of this mismatch I had to put in this fix to get the C version to index the shells correctly since I based much on the python version.

yeganer: You leave the photosphere in the next nu iteration

Tobychev: True, fixed that now.

PR #619, opened 2016-06-30
Fixed incorrect name in error message

Tobychev:
Some tests are skipped if `--atomic-dataset` is not given to `test`, but the diagnostic message suggests `--atomic_database` is missing. This PR changes the the message to give the correct argument name.

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -80,7 +80,7 @@ def pytest_addoption(parser):
 def atomic_data_fname():
     atomic_data_fname = pytest.config.getvalue("atomic-dataset")
     if atomic_data_fname is None:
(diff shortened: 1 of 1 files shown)


tardis/conftest.py
```diff
@@ -80,7 +80,7 @@ def pytest_addoption(parser):
 def atomic_data_fname():
     atomic_data_fname = pytest.config.getvalue("atomic-dataset")
     if atomic_data_fname is None:
-        pytest.skip('--atomic_database was not specified')
+        pytest.skip('--atomic-datset was not specified')
```
yeganer: Please correct the typo

yeganer:
@tardis-sn/tardis-core This is a small bugfix and can be merged at any time.

PR #625, opened 2016-07-18
Automatically download and cache atomic data files.

karandesai-96:
This PR makes changes with mechanism of making use of atomic dataset. Initially an atomic dataset had to be kept separately and provided through command line, but with the merge of this PR, it will be automatically downloaded, cached and used. The command line argument `--atomic-dataset` has been replaced by a flag `--remote-data`.

- [X] A session scoped fixture named `_atom_data` is made in top level **conftest.py**, which will download and cache atom data (currently only `kurucz_cd23_chianti_H_He.h5.zip`).
- [X] A fixture named `atom_data` is made alongwith it, to return a copy of this `AtomData` object returned by `_atom_data` fixture, to avoid re prepare.
- [X] Old fixtures in **conftest.py** have been removed and other tests have been modified to use this structure henceforth.
- [X] Suitable changes have been made in **.travis.yml** as well. Manually `wget` and `unzip` commands of atomic dataset file have been removed.

[Latest Report](http://opensupernova.org/~karandesai96/integration/doku.php?id=reports:c0cf29d)

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -114,3 +114,11 @@ def included_he_atomic_data(test_data_path):
 def tardis_config_verysimple():
     return yaml_load_config_file(
         'tardis/io/tests/data/tardis_configv1_verysimple.yml')
+
+
+@pytest.fixture(scope="session")
+def atom_data_url():
+    return {
+        'kurucz_cd23_chianti_H_He.h5':
+            'http://www.mpa-garching.mpg.de/~michi/tardis/data/kurucz_cd23_chianti_H_He.zip'
+    }
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -98,7 +98,6 @@ def tardis_config_verysimple():
         'tardis/io/tests/data/tardis_configv1_verysimple.yml')
 
 
-@remote_data
 @pytest.fixture(scope="session")
 def _atom_data(request):
     atom_data_name = 'kurucz_cd23_chianti_H_He.h5'
@@ -124,7 +123,6 @@ def _atom_data(request):
     return atom_data
 
 
-@remote_data
 @pytest.fixture(scope="class")
 def atom_data(_atom_data):
     if _atom_data.md5 != '21095dd25faa1683f4c90c911a00c3f8':
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -15,7 +15,6 @@ from astropy.utils.data import download_file
 
 import tardis
 from tardis.atomic import AtomData
-from tardis.io.util import yaml_load_config_file
 
 ###
 # Astropy
@@ -94,8 +93,7 @@ def included_he_atomic_data(test_data_path):
 
 @pytest.fixture(scope="session")
 def tardis_config_verysimple():
-    return yaml_load_config_file(
-        'tardis/io/tests/data/tardis_configv1_verysimple.yml')
+    return 'tardis/io/tests/data/tardis_configv1_verysimple.yml'
 
 
 @pytest.fixture(scope="session")
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -83,26 +83,6 @@ def pytest_addoption(parser):
 # -------------------------------------------------------------------------
 
 
-@pytest.fixture(scope="session")
-def atomic_data_fname():
-    atomic_data_fname = pytest.config.getvalue("atomic-dataset")
-    if atomic_data_fname is None:
-        pytest.skip('--atomic-dataset was not specified')
-    else:
-        return os.path.expandvars(os.path.expanduser(atomic_data_fname))
-
-
-@pytest.fixture
-def kurucz_atomic_data(atomic_data_fname):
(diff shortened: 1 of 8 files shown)


yeganer:
Please do the following:

- Create a fixture `_atom_data` inside tardis/conftest.py returning the cached atom_data
- Create a fixture `atom_data` returning `copy.copy(_atom_data)`
- Rewrite ALL tests so they use the top level atom_data fixture

wkerzendorf:
@karandesai-96 rebase this branch

karandesai-96:
@wkerzendorf Rebased.
@yeganer Directory name was changed `tests_slow` -> `integration_tests` hence overagerc didn't work. Rectified that as well.

tardis/tests/integration_tests/test_integration.py
```diff
@@ -28,6 +30,7 @@ def setup(self, request, reference, data_path, atomic_data_fname):
         self.name = data_path['setup_name']
 
         self.config_file = os.path.join(data_path['config_dirpath'], "config.yml")
+        self.config_dict = yaml.load(self.config_file)
```
yeganer: That's bad, don't do this. We modify the `YAMLLoader`.
Use `from tardis import load_yaml` instead.

yeganer: Why did you change it anyway? To me it seems like `Configuration.from_yaml(...)` does everything we want.

We might even go so far as to use `tardis.run_tardis(config)`. Any reason to not use this?

tardis/conftest.py
```diff
@@ -110,7 +92,40 @@ def included_he_atomic_data(test_data_path):
...
+    request.addfinalizer(fin)
+    return atom_data
+
+
+@pytest.fixture(scope="class")
+def atom_data(_atom_data):
+    if _atom_data.md5 != '21095dd25faa1683f4c90c911a00c3f8':
+        pytest.skip('Need default Kurucz atomic dataset '
+                    '(md5="21095dd25faa1683f4c90c911a00c3f8"')
+    return copy.deepcopy(_atom_data)
```
yeganer: As `copy.deepcopy` doesn't seem to work here we might as well get rid of `atom_data` and rename `_atom_data` to `atom_data`

tardis/simulation/tests/test_simulation.py
```diff
@@ -2,16 +2,18 @@
...
+from astropy.tests.helper import assert_quantity_allclose, remote_data
+
 
 @pytest.fixture
-def tardis_config(kurucz_atomic_data, tardis_config_verysimple):
-    return config_reader.Configuration.from_config_dict(
-        tardis_config_verysimple, atom_data=kurucz_atomic_data)
+def tardis_config(atom_data, tardis_config_verysimple):
+    return Configuration.from_config_dict(
+        tardis_config_verysimple, atom_data=atom_data
```
yeganer: I think we need a copy here

tardis/simulation/tests/test_simulation.py
```diff
@@ -37,6 +39,7 @@ def simulation_compare_data(simulation_compare_data_fname):
     return h5py.File(simulation_compare_data_fname, mode='r')
 
 
+@remote_data
```
yeganer: I think `@remote_data` for the `atom_data` fixture should be enough, or is there a problem?

tardis/tests/integration_tests/test_integration.py
```diff
@@ -1,15 +1,17 @@
...
-from astropy.tests.helper import assert_quantity_allclose
+from astropy.tests.helper import assert_quantity_allclose, remote_data
 
-from tardis.atomic import AtomData
 from tardis.simulation.base import run_radial1d
 from tardis.model import Radial1DModel
 from tardis.io.config_reader import Configuration
 
 
+@remote_data
```
yeganer: again I think we need `@remote_data` only for the top level fixture

tardis/tests/test_atomic.py
```diff
@@ -1,6 +1,7 @@
 import os
 import pytest
 from numpy import testing
+from astropy.tests.helper import remote_data
```
yeganer: Same as other comments about `@remote_data`

tardis/tests/integration_tests/test_integration.py
karandesai-96: `run_tardis` was proposed way back earlier but I exactly cant remember why it was turned down. For the first comment, it was to simply check the name of atom data but we dont need it yet so I am removing it.

tardis/simulation/tests/test_simulation.py
karandesai-96: There is a problem.

tardis/tests/integration_tests/test_integration.py
karandesai-96: Yes, problem here as well.

tardis/tests/integration_tests/test_integration.py
karandesai-96: Oh this diff is outdated one.

PR #629, opened 2016-07-21
From hdfstore rebased

mishinma:


yeganer:
I think, there is something wrong here as this doesn't look like it's based on current master.
Did you `git fetch upstream` before running `git rebase -i upstream/master`

PR #630, opened 2016-07-21
Create an overview page on Dokuwiki.

karandesai-96:
This PR introduces maintenance of an overview page on our Integration Tests Dokuwiki. It is selected as the [landing page](http://opensupernova.org/~karandesai96/integration/doku.php?id=start) of the wiki.
Each time a report it uploaded n Wiki, its information in appended on this page.
Information includes commit hash, commit message description, start time in UTC, time elapsed, and status of test run (pass / fail / error).

### DIFF
--- a/tardis/tests/integration_tests/report.py
+++ b/tardis/tests/integration_tests/report.py
@@ -25,8 +25,9 @@ References
 3. "pytest-html" ( https://www.github.com/davehunt/pytest-html )
 """
 import datetime
-import pkg_resources
+import json
 import os
+import pkg_resources
 import time
 
 # For specifying error while exception handling
@@ -39,8 +40,10 @@ import tardis
 
 try:
     import dokuwiki
+    import requests
 except ImportError:
     dokuwiki = None
+    requests = None
 
 
 class DokuReport(HTMLReport):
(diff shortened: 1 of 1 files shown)


karandesai-96:
@wkerzendorf @yeganer This one is ready from my side, it is a simple enhancement.

wkerzendorf:
@yeganer any other questions on this one?

yeganer:
@wkerzendorf Looks good to me.

tardis/tests/integration_tests/report.py
```diff
@@ -67,14 +70,16 @@ def __init__(self, dokuwiki_details):
         else:
             self.dokuwiki_url = dokuwiki_details["url"]
 
+    def pytest_sessionstart(self, session):
```
yeganer: I *think* you don't need this as you should be able to convert suite_start_time to utc

karandesai-96: @yeganer `suite_start_time` is time zone unaware hence I cannot directly convert it to UTC. I won't have any idea how much offset is to be added / subtracted. Hardcoding +0200 according to CEST may not be a good idea either, I am in +0530 while debugging.

yeganer: Have you confirmed that the line has any effect?
Because I assume `super(DokuReport, self).pytest_sessionstart(session)` does the same thing.
This will essentially overwrite the previously defined `self.suite_start_datetime`

karandesai-96: I found a method which could do the trick, I updated it in my subsequent commit.

PR #632, opened 2016-07-22
Restructure Radial1DModel

ftsamis:
The purpose of this PR is to rewrite the `Radial1DModel` class to make it more dynamic and add a proper `__init__` method to it, which accepts all the required arguments for the model. Also, attributes and methods that do not belong to the model should be removed from it with this PR.

The main idea is to have some base attributes to the model, and then provide dynamic arguments (i.e. python properties) which are (re)calculated every time they are accessed from the base attributes. This way, if a base attribute is changed, everything that depends on it will reflect the changes.

Additionally, `__init__` should now take the actual base attributes that model needs instead of a config object, while we will still have a `from_config` classmethod which will create a model from a config object.


ftsamis:
I have added an option to provide `time_0` for uniform densities as well. Additionally, now `time_0` for all specific density types is not a required property, and if it's not provided, time_0 will take the value of `time_explosion`.

ftsamis:
Closing in favor of #652.

PR #635, opened 2016-07-25
Replace LegacyPlasmaArray with a function

ftsamis:
With this PR `LegacyPlasmaArray` is replaced with `assemble_plasma` which takes a `Configuration` object and a `Radial1DModel` (in the form it has in #632) and returns a `BasePlasma` instance.

### DIFF
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -957,39 +957,6 @@ class Configuration(ConfigurationNameSpace):
 
 
 
-
-
-        ##### NLTE subsection of Plasma start
-        nlte_validated_config_dict = {}
-        nlte_species = []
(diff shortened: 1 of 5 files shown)


wkerzendorf:
@ftsamis should we merge this one?

ftsamis:
@wkerzendorf Not yet, I'm doing some tests first.

ftsamis:
**Note: It's not possible to initialize property classes with the current implementation. That is a problem for `LevelBoltzmannFactorNLTE` which needs to be instantiated.**

Details: https://gitter.im/tardis-sn/tardis?at=57a9f1f2f7f11a7936af6afc

ftsamis:
Closing in favor of #652.

PR #636, opened 2016-07-26
Restructure the MontecarloRunner

ftsamis:
This PR introduces the necessary changes to the `MontecarloRunner` in order to make it work with the restructured `Radial1DModel`.

### DIFF
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -460,11 +460,6 @@ def parse_supernova_section(supernova_dict):
 
     config_dict['time_explosion'] = parse_quantity(supernova_dict['time_explosion']).to('s')
 
-    if 'distance' in supernova_dict:
-        config_dict['distance'] = parse_quantity(supernova_dict['distance'])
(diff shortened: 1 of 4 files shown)


wkerzendorf:
looking good.

ftsamis:
@wkerzendorf It doesn't. :( I realized that I need to change `montecarlo.pyx` as well because it uses the now gone `model.tardis_config` a lot.

Working on it now...

yeganer:
I think of `MonteCarloRunner` as the bridge between python and C. So all information that needs to be passed to the C should be available in the runner.
I don''t know if that makes sense but that's how I see it, do you agree/disagree?

ftsamis:
@yeganer That's right, it already gets the model and the plasma, but `montecarlo.pyx` was accessing `model.tardis_config` which is no longer available. So the point now is changing those access attempts to the correct thing.

Should be relatively easy.

ftsamis:
Closing in favor of #652.

PR #637, opened 2016-07-28
From hdf:

mishinma:
This PR:
- Ditches all `read_something` methods.
- Replaces the `from_hdf5` classmethod of `AtomData` with `from_hdf`. I used  the try-except blocks to check if the dataframe is present in an hdfstore. 
- Refactores the `__init__` method of `AtomData`: 
          1. All `has_some_data` attributes are removed ( check if the dataframe `is not None` instead)
          2. Basically the dataframes are just set as attributes in `__init__()`
- Rewrites the `prepare_atom_data`. The main purpose is to avoid creating views (because of the issues with the C code). 
- Writes a better docstring for `AtomData.


tardis/io/atomic.py
```diff
@@ -38,270 +37,106 @@ def tests_data_path(fname):
...
-
 class AtomData(object):
     """
     Class for storing atomic data
 
-    AtomData
-    ---------
-
     Parameters
     ----------
```
mishinma: @wkerzendorf @yeganer  I used the legacy names for the Dataframes but I suggest changing `atom_data` -> `atom_masses`, `ionization_data` -> `ionization_energies` . Not sure about other dataframes.

tardis/io/atomic.py
```diff
@@ -38,270 +37,106 @@ def tests_data_path(fname):
...
-    lines : ~astropy.table.Table
-        containing the lines data: wavelength, z, ion, levels_number_lower,
-        levels_number_upper, f_lu, f_ul
-
-    macro_atom_data : tuple of ~astropy.table.Table
-        default ~None, a tuple of the macro-atom data and macro-atom references
-
-    zeta_data : ~dict of interpolation objects
-        default ~None
+    atom_data: pandas.DataFrame
```
mishinma: @wkerzendorf @yeganer I used the legacy names but I suggest changing `atom_data` -> `atomic_masses`, `ionization_data` -> `ionization_energies` (and I used these names in carsus). Not sure about other dataframes.

tardis/io/atomic.py
```diff
@@ -310,149 +145,69 @@ def from_hdf5(cls, fname=None):
...
-
-        self.levels = DataFrame(levels_data.__array__())
-        self.levels.energy = units.Quantity(self.levels.energy.values, 'eV').cgs
-
-        self.lines = DataFrame(lines_data.__array__())
-        self.lines['nu'] = units.Quantity(self.lines['wavelength'], 'angstrom').to('Hz', units.spectral())
-        self.lines['wavelength_cm'] = units.Quantity(self.lines['wavelength'], 'angstrom').cgs
-
-
+        self.atom_data = atom_data
```
mishinma: @wkerzendorf @yeganer Currently, I am just setting all dataframes as instance attributes in `__init__()`. Should I do some checks in the method as well? E.g. check that if `collision_data` was provided than `collision_data_temperatures` must be provided as well.

tardis/io/atomic.py
```diff
@@ -310,149 +145,69 @@ def from_hdf5(cls, fname=None):
...
-            if atom_data.version is not None:
-                atom_data.data_sources = pickle.loads(h5_file.attrs['data_sources'])
-
-            logger.info('Read Atom Data with UUID=%s and MD5=%s', atom_data.uuid1, atom_data.md5)
 
         return atom_data
 
-    def __init__(self, atom_data, ionization_data, levels_data, lines_data, macro_atom_data=None, zeta_data=None,
-                 collision_data=None, synpp_refs=None, ion_cx_data=None):
+    def __init__(self, atom_data, ionization_data, levels=None, lines=None,
```
mishinma: Should any of the frames be mandatory to instantiate `AtomData`? Currently, `atom_data` and `ionization_data` are positional arguments so they are mandatory, but that is not sensible (at least `levels` and `lines` also should be positional args then)

tardis/io/atomic.py
wkerzendorf: a check for this would be good.

tardis/io/atomic.py
wkerzendorf: probably, not clear how we want this `__init__` method to look like in the future.

tardis/io/atomic.py
mishinma: @wkerzendorf I have the following check in mind: there is a list of related attributes:
```
    related_attrs = [("macro_atom_data", "macro_atom_references"),
                     ("collision_data", "collision_data_temperatures")]
```
It is checked for each tuple from the list if all attributes from the tuple are `None` or if all of them are not `None`. If any or the attributes is `None` while there are attributes that are not `None`, then `AtomDataMissingError` is raised with a message informing the user about  the related attributes that weren't provided. Does it make sense?

wkerzendorf: sounds good.

tardis/io/atomic.py
```diff
@@ -310,147 +153,83 @@ def from_hdf5(cls, fname=None):
...
+            check_list = [name for name in group if getattr(self, name) is None]
 
-
-        self.ion_cx_data = ion_cx_data
-
+            if len(check_list) != 0 and len(check_list) != len(group):
+                raise AtomDataMissingError(
+                    "The following dataframes from the related group [{0}] "
+                    "were not given: {1}".format(", ".join(group), ", ".join(check_list))
+                )
```
mishinma: @wkerzendorf so this method does the check. An example of the error message:
```
AtomDataMissingError: The following dataframes from the related group [collision_data, collision_data_temperatures] were not given: collision_data_temperatures
```

tardis/io/atomic.py
```diff
@@ -482,113 +261,110 @@ def prepare_atom_data(self, selected_atomic_numbers, line_interaction_type='scat
...
+                ])
 
-                self.macro_atom_data.insert(len(
-                    self.macro_atom_data.columns), 'destination_level_idx',
-                    pd.Series(self.macro_atom_references['references_idx'].ix[
-                    tmp_macro_destination_level_idx].values.astype(
-                        np.int64), index=self.macro_atom_data.index))
+                self.macro_atom_data.loc[:, 'destination_level_idx'] = self.macro_atom_references.ix[
+                    tmp_macro_destination_level_idx, "references_idx"
+                ].astype(np.int64).values  # why it is named `destination_level_idx` ?! It is reference index
```
mishinma: @wkerzendorf @yeganer `destination_level_idx` is ambiguous here because, in fact,  it is  the reference index.

mishinma:
I have visualized all attributes of a prepared `AtomData` instance (not all of them are DataFrames in fact, but this is pretty illustrative)   https://drive.google.com/file/d/0BxXCddocl9m3dlVmNDdwVGtJY28/view?usp=sharing

tardis/io/atomic.py
wkerzendorf: Looks good

tardis/io/atomic.py
wkerzendorf: Hmm - not clear. Can you make sure to keep track of this question even if we merge

mishinma: Sure

mishinma:
The conversion to CGS units has been returned to `__init__()`

PR #638, opened 2016-08-01
Fix hold iterations

ftsamis:
Fix the following issues:

- There were cases where the hold_iterations mode was triggered by the convergence_strategy criteria, and the simulation would get in an infinite loop.
- Improve counting the simulation iterations in the log.
- Also count the last iteration in the final summary message.

### DIFF
--- a/tardis/io/schemas/montecarlo.yml
+++ b/tardis/io/schemas/montecarlo.yml
@@ -164,6 +164,7 @@ definitions:
         hold_iterations:
           type: number
           multipleOf: 1.0
+          minimum: 1
           default: 3
           description: the number of iterations that the convergence criteria need
             to be fulfilled before TARDIS accepts the simulation as converged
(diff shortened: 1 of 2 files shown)


ftsamis:
@wkerzendorf If the `hold_iterations` in the configuration are set to 0, and an iteration converges, should the simulation be over?

wkerzendorf:
@ftsamis I would say hold-iterations >=1 as a constraint

ftsamis:
@wkerzendorf Done

yeganer:
I don't know if there are plans for further changes to the Simulation logic but while we're at it, I suggest to make all the variables dealing with convergence and iterations local.
Or, if `Simulation.iterations_executed` are really needed, we should initialize them in init and not there.

unoebauer:
Should be merged before v2.0

ftsamis:
#652 has a better version of the fix, along with all the other changes I introduced during the summer. Let's close this and focus on #652.

PR #642, opened 2016-08-06
Use pytest-html==1.10.0, modify DokuReport class accordingly.

karandesai-96:
This PR reduces a chunk of code from `DokuReport` class as per the new version of `pytest-html`.

[Latest Report](http://opensupernova.org/~karandesai96/integration/doku.php?id=reports:8e56f40)

### DIFF
--- a/conda-requirements
+++ b/conda-requirements
@@ -11,4 +11,4 @@ numexpr==2.4.4
 Cython==0.21
 networkx==1.10
 pytest==2.6.3
-pytest-html==1.9.0
+pytest-html==1.10.0
--- a/conda-requirements
+++ b/conda-requirements
@@ -11,4 +11,4 @@ numexpr==2.4.4
 Cython==0.21
 networkx==1.10
 pytest==2.6.3
-pytest-html==1.8.1
+pytest-html==1.9.0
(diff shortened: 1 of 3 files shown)


conda-requirements
```diff
@@ -11,4 +11,3 @@ numexpr==2.4.4
 Cython==0.21
 networkx==1.10
 pytest==2.6.3
-pytest-html==1.8.1
```
yeganer: You are doing a change here without providing a 'backup'. This is generally bad.

It is correct, that we consider `conda-requirements` obsolete but as long as they are included in our repository we should keep them updated, too.

tardis/tests/integration_tests/report.py
```diff
@@ -26,22 +26,20 @@
...
-from pytest_html.plugin import HTMLReport
 import tardis
 
 try:
+    from pytest_html import __name__ as pytest_html_path
+    from pytest_html.plugin import HTMLReport
     import dokuwiki
     import requests
 except ImportError:
+    pytest_html = None
```
karandesai-96: @yeganer This was the only location where pytest-html caused an ImportError even with integration tests turned off. With that enclosed in try-except, it won't raise any Exception. Not even NoneType error just because it won't be used anywhere.

conda-requirements
yeganer: for now please add the correct version to conda-requirements. We can discuss removing this file entirely in a new issue/PR.

karandesai-96: :white_check_mark: Done

karandesai-96:
This PR is still WIP

PR #643, opened 2016-08-11
Add artis setup configuration for TestIntegration.

karandesai-96:
This PR includes setups from the first paper, with line interaction types macroatom, downbranch and scatter. Report on OSN for less number of packets is [here](http://opensupernova.org/~karandesai96/integration/doku.php?id=reports:d9fb8a3).


tardis/tests/integration_tests/conftest.py
```diff
@@ -73,12 +73,15 @@ def data_path(request):
...
         ),
         'gen_ref_dirpath': os.path.join(os.path.expandvars(os.path.expanduser(
-            integration_tests_config["generate_reference"])), tardis_githash[:7]
+            integration_tests_config['generate_reference'])), tardis_githash[:7]
         ),
-        'setup_name': hdf_filename[:-3]
+        'setup_name': hdf_filename[:-3],
+        # Temporary hack for providing atom data per individual setup.
+        # This url has all the atom data files hosted, for downloading.
+        'atom_data_url': integration_tests_config['atom_data_url']
```
karandesai-96: @mishinma atom_data_url

yeganer:
Can you please add some label (i.e. 'CHANGED') to the configuration files at *every* location where you made changes to the original config file?

yeganer:
Actually I think you should add the files **exactly** as @wkerzendorf gave them to you. These won't work of course but having the changes as separate commits will help us keep track of them.

PR #645, opened 2016-08-15
Restructure the Simulation

ftsamis:
This PR restructures the `Simulation` class in order to make it work with the restructured `Radial1DModel`, `MontecarloRunner`, and the new `assemble_plasma` function.

Additionally, it fixes the `hold_iterations` logic (see #638 for details), and better organizes the code, removing unneeded class properties as well as legacy and repeated code.

PR #623, opened 2016-07-13
Cleanup of the config_reader module.

ftsamis:
This PR is to progressively solve issue #579.
- [ ] Remove unnecessary conversions (spectrum_list2dict)
- [ ] Interpret values where they are used (`no_of_packes`, ...)
- [ ] Clean up Configuration.from_config_dict (move unit conversions to validator maybe?)
- [x] Remove unused functions(parse_abundance_file_section, parse_supernova_section, parse_density_file_section, parse_model_file_section, etc.)

Issue #644, opened 2016-08-13
virtual_spectrum_range.num is never used


ftsamis:
When cleaning up the `config_reader` I've noticed that the configuration property `montecarlo.virtual_spectrum_range.num` is never used in the code.

Shall we remove it from the schemas or is this a bug?

yeganer:
You are right, there some ambiguity.

I had a quick look at the code and noticed several things.

- To determine if virtual packets should be created the code compares with `spectrum_virt_start_nu` which comes from `montecarlo.virtual_spectrum_range`
- When 'saving' the virtual packet, the code uses `spectrum_start_nu` and `spectrum_delta_nu` to determine which bin the packet corresponds to(`spectrum.[start|stop|num]`)
- The `spectrum_delta_nu` is calculated as the difference of the first two elements of the real spectrum frequency(constant bin width assumed but that's okay as we don't support anything else)

So there is some 'duplicate information' passed to the montecarlo as well as some hidden 'mechanics'.
The current state of the code allows to control the virtual spectrum in two ways. Either one could simply crop the spectrum with `spectrum.[start|end]` or one uses `montecarlo.virtual_spectrum_range` to control virtual spectrum creation with the result of only having data between 6000-7000 angstrom in a 1000-20000 angstrom spectrum.

The proper solution to this issue would be to 'streamline' the montecarlo and get rid of all duplicate information in `storage_model_t`(namely `spectrum_virt_*`). I don't see a reason to have both `spectrum_virt_start_nu` and `spectrum_start_nu`.
The result would be to have only one method to control the range of the spectrum.

TLDR:
There are currently two ways to control the virtual spectrum. Let's remove one of them.

unoebauer:
Closing - reopen if necessary!

PR #652, opened 2016-08-18
[TEP006] Aggregate TARDIS restructure and configuration system cleanup

ftsamis:
This is an aggregate of:
- Restructure Radial1DModel - #632 
- Replace `LegacyPlasmaArray` with a function - #635 
- Restructure the `MontecarloRunner` - #636
- Restructure the `Simulation` - #645 
- Cleanup of the `config_reader` module - #623 
- Fix hold iterations - #638


tardis/plasma/base.py
```diff
@@ -14,11 +14,11 @@
 class BasePlasma(object):
 
     outputs_dict = {}
-    def __init__(self, plasma_properties, **kwargs):
+    def __init__(self, plasma_properties, property_args=None, **kwargs):
```
wkerzendorf: @ftsamis that looks awesome! I'm wondering if we should call this `property_kwargs`? But if you think `args` is better let's go with this.

ftsamis: You are right, `property_kwargs` is a better name. Done

wkerzendorf:
@tardis-sn/tardis-core I would like to have as many collaboration members as possible look over this. This is a **major** change, but seemingly has passed all the tests for now.

tardis/montecarlo/montecarlo.pyx
```diff
@@ -298,8 +295,8 @@ def montecarlo_radial1d(model, runner, int_type_t virtual_packet_flag=0,
 
 def postprocess(model, runner):
     Edotlu_norm_factor = (1 /
-        (model.time_of_simulation * model.tardis_config.structure.volumes))
+        (runner.time_of_simulation * model.volume))
     exptau = 1 - np.exp(-
                         runner.line_lists_tau_sobolevs.reshape(-1,
                             runner.j_estimator.shape[0]) )
-    model.Edotlu = Edotlu_norm_factor*exptau*runner.Edotlu_estimator
+    runner.Edotlu = Edotlu_norm_factor*exptau*runner.Edotlu_estimator
```
ftsamis: Out of scope for this PR. Introduced in #595.

tardis/model/density.py
```diff
@@ -0,0 +1,129 @@
...
+
+from tardis.util import quantity_linspace
+
+
+class HomologousDensity(object):
+    def __init__(self, density_0, time_0):
+        self.density_0 = density_0
+        self.time_0 = time_0
+
+    def after_time(self, time_explosion):
```
ftsamis: Propose an alternative.

wkerzendorf: `calculate_density_evolution`? @chvogl If no other alternative in the next hour just use that and we are done!!

chvogl: I think something like __calculate_density_at_time_of_explosion__ would be more comprehensible.

tardis/montecarlo/base.py
```diff
@@ -110,29 +120,30 @@ def legacy_update_spectrum(self, no_of_virtual_packets):
             self.spectrum_virtual.update_luminosity(
                 self.montecarlo_virtual_luminosity)
 
-    def run(self, model, no_of_packets, no_of_virtual_packets=0, nthreads=1,last_run=False):
+    def run(self, model, plasma, no_of_packets, no_of_virtual_packets=0, nthreads=1,last_run=False):
```
ftsamis: Fixed.

tardis/model/density.py
ftsamis: I really don't like long names, but done.

tardis/model/density.py
```diff
@@ -0,0 +1,129 @@
...
+        adjusted_velocity = velocity.insert(0, 0)
+        v_middle = (adjusted_velocity[1:] * 0.5 +
+                    adjusted_velocity[:-1] * 0.5)
+        no_of_shells = len(adjusted_velocity) - 1
+        time_explosion = config.supernova.time_explosion.cgs
+
+        if d_conf.type == 'branch85_w7':
+            density_0 = calculate_power_law_density(v_middle, d_conf.w7_v_0,
+                                                    d_conf.w7_rho_0, -7)
+            return cls(density_0, d_conf.w7_time_0)
```
ftsamis: I agree. Fixed.

tardis/montecarlo/base.py
```diff
@@ -302,11 +320,40 @@ def to_hdf(self, path_or_buf, path=''):
...
         self.spectrum_virtual.to_hdf(path_or_buf, runner_path,
                                      'spectrum_virtual')
         self.spectrum_reabsorbed.to_hdf(path_or_buf, runner_path,
                                         'spectrum_reabsorbed')
+
+    @classmethod
+    def from_config(cls, config):
+        if config.plasma.disable_electron_scattering:
+            logger.warn('Disabling electron scattering - this is not physical')
+            sigma_thomson = 1e-200 / (u.cm ** 2)
```
ftsamis: The logic was actually copied as-is from current master: https://github.com/tardis-sn/tardis/blob/60d1d1dcdcb92e608f3b4b1e5fb2be429398314c/tardis/io/config_reader.py#L948-L948. 

Since I don't really know what sigma_thomson or cross sections are in this context, I'd like a confirmation on if we should change that. @wkerzendorf?

tardis/montecarlo/base.py
```diff
@@ -302,11 +320,40 @@ def to_hdf(self, path_or_buf, path=''):
...
+        else:
+            logger.debug("Electron scattering switched on")
+            sigma_thomson = 6.652486e-25 / (u.cm ** 2)
```
ftsamis: Same answer as above. @wkerzendorf. On your second comment, if there is a constant for that it seems perfectly reasonable to use it as far as I am concerned.

tardis/montecarlo/base.py
```diff
@@ -28,9 +29,18 @@ class MontecarloRunner(object):
...
 
-    def __init__(self, seed, spectrum_frequency, distance=None):
+    def __init__(self, seed, spectrum_frequency, virtual_spectrum_range,
+                 sigma_thomson, enable_reflective_inner_boundary,
+                 inner_boundary_albedo, line_interaction_type, distance=None):
+
+        self.seed = seed
         self.packet_source = packet_source.BlackBodySimpleSource(seed)
         self.spectrum_frequency = spectrum_frequency
+        self.virtual_spectrum_range = virtual_spectrum_range # TODO: linspace handling
```
ftsamis: This is closely related to #644. I removed the TODO comment for now. Fixed.

tardis/montecarlo/base.py
```diff
@@ -302,11 +320,40 @@ def to_hdf(self, path_or_buf, path=''):
...
+            sigma_thomson = 6.652486e-25 / (u.cm ** 2)
+
+        spectrum_frequency = quantity_linspace(
+            config.spectrum.stop.to('Hz', u.spectral()),
+            config.spectrum.start.to('Hz', u.spectral()),
+            num=config.spectrum.num + 1)
+
+        return cls(seed=config.montecarlo.seed,
+                   spectrum_frequency=spectrum_frequency,
+                   # TODO: Linspace handling for virtual_spectrum_range
```
ftsamis: Removed the TODO comment. Fixed.

tardis/montecarlo/base.py
ftsamis: After some digging in the code, it seems that the unit of sigma_thomson is completely ignored afterwards, since cmontecarlo uses just its value. So, neither the wrong unit, neither correcting it would affect the computations in any way, which means it is probably safe to fix this by directly using, as @chvogl proposed, the `astropy.constants.sigma_T`.

tardis/base.py
```diff
@@ -19,23 +19,22 @@ def run_tardis(config, atom_data=None):
...
-            config, atom_data=atom_data)
+        tardis_config = Configuration.from_config_dict(config)
 
-    radial1d_mdl = model.Radial1DModel(tardis_config)
-    simulation.run_radial1d(radial1d_mdl)
+    simulation = Simulation.from_config(tardis_config, atom_data=atom_data)
+    simulation.run()
 
-    return radial1d_mdl
+    return simulation
```
unoebauer: This will break existing scripts which expect that radial1d_mdl is returned by run_tardis. 

I am tempted to swallow this and will update e.g. tardisanalysis.

wkerzendorf:
@ftsamis undo the last change. this breaks the tests - we shouldn't do this in this PR.

ftsamis:
@wkerzendorf That's because I forgot to convert sigma_T to cgs. I agree though that this could be best done in a different PR, so I reverted the change.

wkerzendorf:
@ftsamis congratulations - I'm merging this now.

PR #653, opened 2016-08-19
Facility to provide per-setup atom data as well as save report locally.

karandesai-96:
This PR focuses on adding flexibility to current integration testing infrastructure.

### Facility for Atom Data files;
* Till now only one atom data file ( generally `kurucz_cd23_chianti_H_He.h5` ) could be provided using the `--atomic-dataset` flag. Instead of this, the integration tests can now use the atom data file specified in the YAML config of a particular setup used for integration tests (in the sub directories of `tardis/tests/integration_tests` ).

* The user has to put all the atom data files specified in config files of various setups in one directory and provide the directory path in integration tests YAML config file. Alternatively, these files can be served at a URL and that URL can also be provided.

* The block schema in integration tests YAML config file looks like this:
```yml
atom_data:
    fetch: "local" # has two choices: "local" or "remote"
    url: "http://urlservingfiles.com"
    dirpath: "/path/to/directory"
```
* If `fetch` is set to local, then files will be taken from `dirpath` and `url` field will be ignored. Even writing it in isn't necessary in those cases. It will be the opposite case if `fetch: "remote"`.
* Example of URL serving atom data files: http://opensupernova.org/~karandesai96/atom_data

### Local saving of report:
* The _HTML_ report of integration tests was uploaded on a dokuwiki instance before now. This PR lets the user generate a report locally and view it in the browser.

* The block schema in integration tests YAML config file looks like this:
```yml
report:
    save_mode: "local" # has two choices: "local" or "remote"
    dokuwiki:
        url: "http://dokuwikiurl.com"
        username: "your_username"
        password: "your_password"
    dirpath: "/path/to/directory"
```

* If `save_mode` is set to local, then files will be taken from `dirpath` and `dokuwiki` section will be ignored. Even writing it in isn't necessary in those cases. It will be the opposite case if `save_mode: "remote"`.

### NOTE:
* Do add astropy's `--remote-data` flag as argument.

### DIFF
--- a/tardis/tests/integration_tests/test_integration.py
+++ b/tardis/tests/integration_tests/test_integration.py
@@ -1,8 +1,12 @@
 import os
+import urlparse
+import yaml
 import pytest
 import matplotlib.pyplot as plt
 from numpy.testing import assert_allclose
+
 from astropy.tests.helper import assert_quantity_allclose
+from astropy.utils.data import download_file
 
 from tardis.atomic import AtomData
 from tardis.simulation.base import run_radial1d
@@ -19,7 +23,7 @@ class TestIntegration(object):
 
     @classmethod
     @pytest.fixture(scope="class", autouse=True)
-    def setup(self, request, reference, data_path, atomic_data_fname):
+    def setup(self, request, reference, data_path):
         """
         This method does initial setup of creating configuration and performing
         a single run of integration test.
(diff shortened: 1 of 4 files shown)


tardis/tests/integration_tests/test_integration.py
```diff
@@ -29,14 +32,31 @@ def setup(self, request, reference, data_path, atomic_data_fname):
...
+
+        # Get the path to HDF file:
+        if 'atom_data_url' in data_path:
+            # If the atom data is to be ingested from url:
+            atom_data_filepath = download_file("{url}/{name}".format(
+                url=data_path['atom_data_url'], name=atom_data_name), cache=True
+            )
+        else:
+            # If the atom data is to be ingested from local file:
+            atom_data_filepath = "{dirpath}/{name}".format(
```
yeganer: Please use `os.path.join` instead of string formatting for the filepath.

tardis/tests/integration_tests/test_integration.py
```diff
@@ -29,14 +32,31 @@ def setup(self, request, reference, data_path, atomic_data_fname):
...
+        # A quick hack to use atom data per setup. Atom data is ingested from
+        # local HDF or downloaded and cached from a url, depending on data_path
+        # keys.
+        atom_data_name = yaml.load(open(self.config_file))['atom_data']
+
+        # Get the path to HDF file:
+        if 'atom_data_url' in data_path:
+            # If the atom data is to be ingested from url:
+            atom_data_filepath = download_file("{url}/{name}".format(
+                url=data_path['atom_data_url'], name=atom_data_name), cache=True
```
yeganer: For the url I'd suggest `urllib.parse.urljoin`

yeganer:
Thanks for this PR, it should solve the problem.

The way I imagined the fix would have been adding this functionality directly to tardis by adding an optional parameter to the config yml (`url: True` for example). If this is set the configuration would do the work.

But for now having this functionality exclusive to the integration tests is fine, too.

tardis/tests/integration_tests/test_integration.py
karandesai-96: Looks cleaner - I'll change it.

tardis/tests/integration_tests/test_integration.py
karandesai-96: Oops, mimicked from above block :sweat_smile:

Issue #654, opened 2016-08-22
Virtual Packet Cutoff during Montecarlo


yeganer:
In the montecarlo simulation there's a cutoff for virtual packets which boils down to:

```
if tau>10:
    tau == 100
```

This removes the contribution of weakly contributing packets almost completely (`exp(-100) ~ 1e-44`).
But because this isn't exactly 0 the packet will still register and add to the bin. If the packet is the only packet in that bin the resulting luminosity will be very small and numerical problems can surface when comparing different bins.

The solution I propose is to either set the `weight` or the `energy` of the packet to `0`. This removes any contribution and even allows small performance improvements by skipping the spectrum update(which needs to be atomic)

unoebauer:
Do we still need this? I agree conceptually with you that setting the weight or energy of the packet to 0 is the cleaner approach to discard virtual packets. However, I am wondering about cases where fluctuations in these parts of the spectrum ever become relevant.

yeganer:
The current implementation creates three possible 'states' for a bin. It's either zero meaning no packet at all, ~1e-100 meaning only discarded packets registered for the bin or it has an actual value.
The first two cases should be the same and currently one needs to have a cutoff to separate these.

There are two cases where this issue was relevant in the past.
1.  It makes comparing of the non-zero parts of spectra more difficult
2.  Fred had to differentiate between the two types of packets when doing his statistical analysis for the virtual packets. If I remember correctly ~80% of the packets were discarded packets. One could skip logging these packets reducing memory consumption on the way.

As the fix is easy, quick and has no bad side effects that I know of the important question that needs to be answered should be "Why shouldn't we merge this?" instead of "Why should we merge this?"

unoebauer:
Fine - I'd be in favour of merging.

unoebauer:
Will be dealt with once the testing framework is fully operational

unoebauer:
@wkerzendorf, @yeganer - thoughts on this? Should we merge this. Will require an update of reference data.

chvogl:
Solved with merge of #911 .

PR #655, opened 2016-08-22
Remove neglegible contributions to virtual spectrum

yeganer:
This small change results in bins with actual neglegible data being zero
instead of having very small contributions.

Fixes: #654


yeganer:
@tardis-sn/tardis-core What's the status on this PR? Merge/Discard/Improve?

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -816,7 +816,9 @@ montecarlo_one_packet_loop (storage_model_t * storage, rpacket_t * packet,
                                                                 distance, mt_state);
       if (virtual_packet > 0 && rpacket_get_tau_event (packet) > 10.0)
         {
-          rpacket_set_tau_event (packet, 100.0);
+          // Ideally we would like to set the weight 
+          // instead of the energy but this isn't possible yet
+          rpacket_set_energy (packet, 0.0);
           rpacket_set_status (packet, TARDIS_PACKET_STATUS_EMITTED);
```
yeganer: @unoebauer I'm not sure if this is 'correct'. From my understanding the reabsorbed status would make more sense here, right?

yeganer:
Let's move the discussion from #654 here.

Do we want to save virtual packets with 0 energy?

tardis/montecarlo/src/cmontecarlo.c
unoebauer: Not sure - for virtual packets "re-absorption" (i.e. intersection with the photosphere) does not make sense since these packets are always spawned in a way that they don't intersect the photosphere. For this reason, I am not sure if virtual packets with this flag would be handled properly.

yeganer: I checked. The packet status is only relevant for real packets and discarded for all virtual packets.

wkerzendorf:
@unoebauer @yeganer - what do we do with this thing. What kind of improvements does it offer to our science operations?

yeganer:
@wkerzendorf If we purely look at the algorithm and the implementation, the current behavior has to be considered a bug.
We made the decision to discard virtual packets at a threshold. However in the implementation we do not discard them completely but just reduce their contribution by ~50 orders of magnitude. They don't affect the generation of the spectrum by much (only in bins that would otherwise be 0). However, if virtual packet logging is used, they could screw the results.

As I consider the current behavior a bug (albeit a small one) and I'm in favor of fixing small and easy bugs, I would merge this and I would extend the PR to not save packets with an energy of 0 (to keep vlogging relevant).

unoebauer:
Pragmatically, I'd say it doesn't make a difference for the normal use of Tardis.

unoebauer:
Actually, the proper thing to do would be to introduce a "Russian Rouletting" scheme. I think @chvogl has done something like this for his Type IIP version of Tardis.

unoebauer:
Let's close this for now. If these low-weight packets become an issue again, let's try to address this with biasing techniques.

Issue #546, opened 2016-04-21
conda-requirements should be listed once in a common place.


karandesai-96:
After the merge of #539 which adds OSX builds to Travis, the conda-requirements were listed in a separate **tardis_ci/tardis_ci_env27.yml** file and a new conda environment named **tardis-ci** is created in Travis builds through this file. A similar yml config file **tardis_ci/tardis_rtd_env27.yml** has conda-requirements listed for read the docs. The differences between dependencies in both files are:
```
- coverage=3.7.1
- requests=2.9.1
- docopt=0.6.2
- pytest-cov=2.2.1
```
These four dependencies are included in **tardis_ci/tardis_ci_env27.yml** while not in another one. Having them there won't cause any problem. This issue requires certain work to do for a solution:

1. Delete **conda-requirements**.
2. Merge the mentioned config files into one and use it everywhere required.
3. Update the documentation to instruct users for creating a conda environment as:
```
conda env create --name tardis --file tardis_ci/tardis_ci_env27.yml
```
* [Optional] Files may be renamed such that they look suitable in documentation description.

The consequences would be helpful, when the dependencies to be updated will have to be updated only in one place instead of three.

wkerzendorf:
there are more differences in these which pertain to the documentation.

karandesai-96:
@wkerzendorf I did not understand what are "these" referred to here. Do you mean the amount of changes to be done in documentation after making changes with the configurations ?

wkerzendorf:
@karandesai-96 no you are right - nevermind.

mangalutsav:
I would like to know the status of this issue, can a beginner like me take it up ?

karandesai-96:
Hello @mangalutsav 
Yes, you can try your hand at this issue, it is an easy fix. Check the three files highlighted as **bold** in the description. You would see repeated content. The idea is to keep the content at one place and use it everywhere, for easy maintenance.

mangalutsav:
In **tardis_ci/tardis_rtd_env27.yml** and **conda-requirements** pytest==2.6.3 is used whereas in **tardis_ci/tardis_ci_env27.yml** pytest==2.9.1 is used will this be a prolem

karandesai-96:
I don't think pytest should create a problem. Still go ahead, do it and let us know what happened. **tardis_ci/tardis_ci_env27.yml** would be the most suitable to hold the conda requirements.

yeganer:
[...]

PR #659, opened 2016-09-19
Update installation docs

ftsamis:
Update the installation instructions to only refer to `tardis_env27.yml`, instead of `conda-requirements`, and also update the environment creation command.

If this is merged #546 should be solved as well.


wkerzendorf:
@ftsamis thanks!!

PR #660, opened 2016-10-09
Substitute reference data for unit tests

mishinma:
This PR substitutes the reference data used in the unit tests with the data created using the Carsus [hdfstore](https://github.com/tardis-sn/carsus-db/tree/master/test_hdfstores). This [document](https://docs.google.com/document/d/11NNajf1EaIlHE_i3Rag91ZvmQ-zmab4DD67nJxuNcNk/edit) shows plots for the most of failing unit tests. It can be seen from the plots that values are reasonably close; therefore, the substitution is reasonable. 

The only remaining failing unit tests are in `test_property_radiative.py`. I am not sure why they are failing, so I've left them.

The PR also explicitly  casts `lines_idx` to `int`.


unoebauer:
Closing - @wkerzendorf reopen if necessary.

PR #669, opened 2016-11-13
Py3 compatibility

yeganer:
As a small spare time project I wanted to see how much work it would be to add python 3 support to tardis.
With the help of the `futurize` tool this task was surprisingly easy.

Just to be clear, I don't expect a merge soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,11 @@
 os:
   - linux
+
 language: python
 
 python:
     - 2.7
-
-virtualenv:
-  system_site_packages: true
+    - 3.5
 
 cache:
   apt: true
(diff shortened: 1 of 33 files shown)


ftsamis:
I am wondering if there is any point to target both python2 and python3 instead of only python3.

wkerzendorf:
@ftsamis this was just a test - there is no plans to make tardis python 3 compatible in the next year or so.

ftsamis:
@wkerzendorf I understand, but I believe my question is still relevant. Unless the answer is that there is no answer yet.

wkerzendorf:
The forseeable future will not see any python 3 support @ftsamis

PR #670, opened 2016-11-24
Integrate continuum interactions into cmontecarlo

chvogl:
This PR imports the core routines of the continuum interaction procedure developed by @chvogl during his master's project. This PR constitutes only the first step, thus the immediate goal is not yet to have a fully fledged Tardis-continuum version but to import the necessary changes step by step in a series of PRs. The goal of the current one can be summarized as follows:

- [x] include core C routines of continuum interaction scheme
- [x] add necessary C unit tests for new additions and update and complete existing C unit tests
- [x] ensure that normal Tardis operation is not affected by changes


tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -94,6 +94,51 @@ line_search (const double *nu, double nu_insert, int64_t number_of_lines,
   return ret_val;
 }
 
+tardis_error_t
```
wkerzendorf: see if this is needed.

unoebauer:
Would be nice to have a brief description and list of milestone, so people see where this PR is heading to.

tardis/montecarlo/montecarlo.pyx
```diff
@@ -74,7 +79,8 @@ cdef extern from "src/cmontecarlo.h":
         double inverse_sigma_thomson
         double inner_boundary_albedo
         int_type_t reflective_inner_boundary
-        double *chi_bf_tmp_partial
+        photo_xsect_1level ** photo_xsect
```
yeganer: I'm not familiar with the exact usecase but I think double pointers can easily and should be avoided here.

chvogl: Using double pointers can be easily avoided by using an array of structures instead of an array of   
pointers to structures. Do you think this is preferable? Since the structures are small (two pointers and an integer) this should be ok. Alternatively it is possible to store the cross section data (frequencies and cross sections) consecutively in two arrays. In this case a reference array is needed to know where the data for a certain level starts and ends.

yeganer: If possible, you should store the data as a structure of arrays.

unoebauer:
I'd like this PR to get merged **before** the Christmas break! Can you identify the missing parts for doing so, @chvogl?

PR #671, opened 2016-11-25
Maintenance: Updated setup.py to remove cython files to force rebuild

orbitfold:


wkerzendorf:
@unoebauer @orbitfold I think we can do the `rm` ourselves with out the hack. I have no idea what other consequences this has. I'd be in favour of closing this PR.

PR #682, opened 2016-12-20
Import montecarlo.base in the __init__ module of the package.

ftsamis:
Similar to all other sub-packages, instead of requiring `from montecarlo.base import MontecarloRunner` allow simply doing `from montecarlo import MontecarloRunner`


wkerzendorf:
@tardis-sn/tardis-core any objections? can someone else green-light this as well?

PR #688, opened 2017-01-10
Use sigma_T from astropy constants instead of a hardcoded value

ftsamis:
Astropy now has a constant defined for sigma_thomson. This PR replaces the hardcoded value with it.

Props to @chvogl for discovering this.

For tardis_example outputs before and after this patch see: https://gist.github.com/ftsamis/eccbcc1b2a29053a412a2d74e5a11e7f


ftsamis:
@yeganer Not sure what you are referring to. Do you mean the test_cmontecarlo module? If so, done.

yeganer:
No, I mean `test_data.h5` and the numpy reference data files for `test_tardis_full.py`

wkerzendorf:
@ftsamis @chvogl I like this change. Can we get this into a mergeable state?

chvogl:
@wkerzendorf As @yeganer pointed out someone would have to update the reference data files (`test_data.h5` and the numpy reference data files for `test_tardis_full.py`) before we can merge this.

unoebauer:
I'd still like to merge this. Let's resolve the conflicts and update the reference data.

unoebauer:
I am a bit puzzled why some of the test differ by so much since the change in the value of the Thomson cross section is really minuscule:

Before we used the hard-coded value of 
```
6.652486e-25 
```
Now, we rely on astropy.constants, which includes a value for the Thomson cross section of
```
6.652458734e-25
```
in version 1.3.3.

Any thoughts on this, @wkerzendorf, @chvogl?

unoebauer:
Closing. Work continues in PR #863

PR #691, opened 2017-01-13
change to model refrence from Simulation model refrence

deep110:
Since simulation model contains base model object which contains _t_rad_ and _w_ (dilution_factor). I have taken _ws_ and _w_ same because I can't find any reference for _ws_. 

Also, after fixing this issue, gui is still crashing with

` interface.show(mdl)
  File "~/tardis/tardis/gui/interface.py", line 43, in show
    win.show_model(model)
  File "~/tardis/tardis/gui/widgets.py", line 1167, in showl
    self.mdv.fill_output_label()
  File "~/tardis/tardis/gui/widgets.py", line 473, in fill_l
    .format(self.model.iterations_max_requested, 
AttributeError: 'Simulation' object has no attribute 'iterations_max_requested'
`

is _iterations_ of simulation object and _iterations_max_requested_ same, if yes then should I also fix that.

### DIFF
--- a/tardis/gui/widgets.py
+++ b/tardis/gui/widgets.py
@@ -566,8 +566,8 @@ class ModelViewer(QtGui.QWidget):
         """Reset the model set in the GUI."""
         self.model = model
         self.tablemodel.arraydata = []
-        self.tablemodel.add_data(model.t_rad.value.tolist())
-        self.tablemodel.add_data(model.ws.tolist())
+        self.tablemodel.add_data(model.model.t_rad.value.tolist())
+        self.tablemodel.add_data(model.model.w.tolist())
 
     def change_spectrum_to_spec_virtual_flux_angstrom(self):
         """Change the spectrum data to the virtual spectrum."""


ftsamis:
The purpose is to have the GUI work again, so I'd say yes you should also fix any other bugs/errors.

Apart from that, simulation should be the toplevel object that is passed in the GUI, instead of model, so the naming should be changed. (model.model makes no sense)

deep110:
@ftsamis sure I will update the pull request with the changes

pfreddy:
@deep110 
Hi! I made the recommended changes in widgets.py (t_rad and w), which led to the error message copied by you. After changing _iterations_max_requested -> itarations_ , a new problem appeared:

/home/barna/anaconda3/envs/tardis/lib/python2.7/site-packages/tardis/gui/widgets.py in fill_output_label(self)
    476                         self.model.converged else
    477                         '<font color="red"><b>False</b></font>',
--> 478                         self.model.time_of_simulation.value,
    479                         self.model.t_inner.value,
    480                         self.model.current_no_of_packets,

AttributeError: 'Simulation' object has no attribute 'time_of_simulation'

How many variables have to be replaced?
Thank you in advance!

yeganer:
@pfreddy I had a quick look at the GUI and I noticed that quite a lot has to be changed.
`time_of_simulation` is now part of the `runner`. Several other properties are now part of the `model` so I guess in total I'll take 1-2 hours to redirect everything.
I'm currently quite busy but if it's urgent, I can try to fix most of it with some regex magic and make a PR.

Issue #687, opened 2017-01-10
Configuration version 2


ftsamis:
This is to collect and discuss all the changes that could be implemented in the configuration file syntax and would require a `tardis_config_version` bump due to backwards incompatibility with the current configuration schemata.

Feel free to edit adding ideas and/or comment on the existing ones.

**Proposed:**
1. _YAML implicit resolver for ranges_ (related to but not the same as #679)
    Instead of the long
    ```yaml
    black_body_sampling:
      start: 1 angstrom
      stop: 1000000 angstrom
      num: 1.e+6
    ```
     in the config file one would write something like `black_body_sampling: 1..1000000:1.e+6 angstrom` which would be interpreted as `start..stop[:num] [unit]` and is much shorter. The only drawback is that you cannot specify e.g. start in meters and stop in angstrom which is probably an edgy case anyway.

chinmay0301:
we can keep a default format to dodge the edgy case. `start..stop[:num] [unit_start] [unit_stop]`. If the units of the starting and ending wavelengths are same, only `unit_start` is specified, one can use that for both the start and stop wavelengths. Else, if both units are specified, we use them accordingly.

yeganer:
Spectrum should be a quantity linspace after the configuration validation

Do we want to assume units or always explicitly require them?

unoebauer:
I'll close this one due to inactivity. Reopen if necessary.

PR #693, opened 2017-01-15
 YAML implicit resolver of ranges: #687

chinmay0301:
This PR is in reference to the issue #687 
Identified the following set of features from the test example which we can resolve implicitly: 
1) `spectrum` 
2) `supernova` 

Allowing input in the YAML file in the format : `black_body_sampling: 1 1000000 1.e+6 angstrom angstrom` 
corresponding to the labels `start`, `stop`, `num`, and the units of start and stop correspondingly. 

Ensured that even in the inputs are given in the dictionary format as they are currently being given, the code still works. Can implement comma separated inputs also if required. Currently changed only the `widgets.py` file. Please let me know if changes are to be reflected in some other file also


chinmay0301:
@ftsamis  can you have a look at it once?

yeganer:
It looks like these changes won't implement the desired functionality.
Have you tried running the tardis_example.yml with the new syntax?

ftsamis:
I'd say @yeganer is correct. yaml implicit resolvers are a very specific thing and unrelated with the GUI code. Make a search and see how they work.

chinmay0301:
But what we essentially want is to decompose a string input into the corresponding key values for the dict element `black_body_sampling`right? So why is it wrong to just decompose it using a python code in gui where we are comparing the predefined yaml file with the input yaml file?

yeganer:
Our config validation isn't done by the gui but by `tardis.io.config_reader`. The gui doesn't work with the latest version anyway.

Instead of getting a dictionary with `'start', 'stop' and 'num'` we want to get a `np.linspace` when loading the config with yaml.

I suggest to test the changes with `spectrum` instead of `blackbody_sampling` as that's currently used.

chinmay0301:
@yeganer how should I store the np array obtained? as in, should it be stored as `object['spectrum'] = np.linspace(start, stop, num)` ?

chinmay0301:
I have'nt implemented the resolver yet, but essentially this is what we want to achieve right? That we can directly pass a `np.linspace` array to the `base.py` file in the montecarlo folder.

scripts/tardis
```diff
@@ -6,15 +6,11 @@ import numpy as np
 
 import logging
 import argparse
-
-
-
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request.

scripts/tardis
```diff
@@ -6,15 +6,11 @@ import numpy as np
...
 tardis_description =\
 """
 TARDIS Supernova Montecarlo Radiative transfer code
 
 """
-
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request.

tardis/gui/widgets.py
```diff
@@ -256,9 +256,9 @@ def __init__(self, yamlconfigfile, parent=None):
                           'no_of_packets':[True, None],
                           'iterations':[True, None],
                           'black_body_sampling':{
-                                                    'start': '1 angstrom',
-                                                    'stop': '1000000 angstrom',
-                                                    'num': '1.e+6',
+                                                   'start': '1 angstrom',
+                                                   'stop': '1000000 angstrom',
+                                                   'num': '1.e+6',
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request.

tardis/io/config_reader.py
```diff
@@ -6,9 +6,9 @@
 
 import tardis
 from tardis.io import config_validator
-from tardis.io.util import YAMLLoader, yaml_load_file
+from tardis.io.util import YAMLLoader, yaml_load_file, quantity_from_str
+from tardis.util import quantity_linspace
 
-pp = pprint.PrettyPrinter(indent=4)
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request.

tardis/io/util.py
```diff
@@ -65,7 +66,18 @@ class YAMLLoader(yaml.Loader):
     A custom YAML loader containing all the constructors required
     to properly parse the tardis configuration.
     """
-
+    def demo_constructor(self, node):
+    	data = self.construct_scalar(node)
+    	return data
```
ftsamis: What is `demo_constructor`? It doesn't seem to serve any purpose to me.

tardis/io/util.py
```diff
@@ -65,7 +66,18 @@ class YAMLLoader(yaml.Loader):
...
+    def spectrum_constructor(self, node):
+    	"""
+    	A constructor to convert spectrum values to nparray
+    	"""
+    	data = self.construct_scalar(node)
+    	data = data.split('%')
+        start = u.Quantity(data[0], data[3])
+        stop = u.Quantity(data[1], data[3])
+        return data
```
ftsamis: - Indentation seems wrong
- `spectrum_constructor` is way too restrictive for a name that constructs ranges in general.
- You create start and stop local variables without using them again.
- Your delimiting scheme seems different from that specified in #687.
- You seem to ignore the `num` property, you don't handle cases where the unit is omitted/not required.
- You return a list and not a linspace.
- The `spectrum_constructor` doesn't seem to be used at all in the end.

tardis/io/util.py
```diff
@@ -90,13 +102,14 @@ def mapping_constructor(self, node):
         return OrderedDict(self.construct_pairs(node))
 
 YAMLLoader.add_constructor(u'!quantity', YAMLLoader.construct_quantity)
+YAMLLoader.add_constructor(u'!demo', YAMLLoader.demo_constructor)
```
ftsamis: We do not have or need any `!demo` properties.

tardis/io/util.py
```diff
@@ -90,13 +102,14 @@ def mapping_constructor(self, node):
...
 YAMLLoader.add_implicit_resolver(u'!quantity',
                                  MockRegexPattern(quantity_from_str), None)
 YAMLLoader.add_implicit_resolver(u'tag:yaml.org,2002:float',
-                                 MockRegexPattern(float), None)
+                                                                          MockRegexPattern(float), None)
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request.

tardis/io/util.py
```diff
@@ -90,13 +102,14 @@ def mapping_constructor(self, node):
...
 YAMLLoader.add_constructor(yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG,
                            YAMLLoader.mapping_constructor)
-
+YAMLLoader.add_implicit_resolver(u'!demo', re.compile(r'^\d+%\d+%\d+%\d+$'))
```
ftsamis: Your `demo` constructor, except from an inappropriate name, doesn't seem to do anything useful.

tardis/montecarlo/base.py
```diff
@@ -359,16 +359,16 @@ def from_config(cls, config):
...
-            config.spectrum.start.to('Hz', u.spectral()),
-            num=config.spectrum.num + 1)
-
-        return cls(seed=config.montecarlo.seed,
+        # spectrum_frequency = quantity_linspace(
+        #     config.spectrum.stop.to('Hz', u.spectral()),
+        #     config.spectrum.start.to('Hz', u.spectral()),
+        #     num=config.spectrum.num + 1)
+        spectrum_frequency = config.spectrum.array       
+	return cls(seed=config.montecarlo.seed,
```
ftsamis: - Don't comment code that you think is not needed, instead just remove it. 
- config.spectrum.array does not occur from any yaml constructor.
- The purpose was to have a linspace created directly from the `YAMLLoader` through e.g. `config.spectrum` but not only for spectrum.

tardis/montecarlo/base.py
```diff
@@ -359,16 +359,16 @@ def from_config(cls, config):
...
                    spectrum_frequency=spectrum_frequency,
                    virtual_spectrum_range=config.montecarlo.virtual_spectrum_range,
                    sigma_thomson=sigma_thomson,
                    enable_reflective_inner_boundary=config.montecarlo.enable_reflective_inner_boundary,
                    inner_boundary_albedo=config.montecarlo.inner_boundary_albedo,
                    line_interaction_type=config.plasma.line_interaction_type,
-                   distance=config.supernova.get('distance', None))
\ No newline at end of file
+                   distance=config.supernova.get('distance', None))
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request

tardis/util.py
```diff
@@ -428,4 +428,4 @@ def quantity_linspace(start, stop, num, **kwargs):
         raise ValueError('Both start and stop need to be quantities with a '
                          'unit attribute')
 
-    return np.linspace(start.value, stop.to(start.unit).value, num, **kwargs) * start.unit
\ No newline at end of file
+    return np.linspace(start.value, stop.to(start.unit).value, num, **kwargs) * start.unit
```
ftsamis: Don't make changes to files or code that are out of scope for your pull request

tardis/io/config_reader.py
```diff
@@ -217,6 +217,10 @@ def from_yaml(cls, fname, *args, **kwargs):
         try:
             yaml_dict = yaml_load_file(fname,
                                        loader=kwargs.pop('loader', YAMLLoader))
+            start = yaml_dict['spectrum']['start'] 
+ 			stop = yaml_dict['spectrum']['stop'] 
+ 			num = yaml_dict['spectrum']['num'] 
+ 			yaml_dict['spectrum']['array'] = quantity_linspace(start, stop, num+1)
```
ftsamis: - Identation seems wrong.
- The bug report you say you are fixing wants the `start`, `stop`, and `num` properties removed, you still use them here and add another one.

tardis/util.py
chinmay0301: I will look into it carefully this time. Sorry for so many errors.

PR #694, opened 2017-01-17
Significantly reduce peak memory usage when vpacket logging is enabled

ftsamis:
@Heringer-Epson attempted a run with 

```yaml
last_no_of_packets: 1.0e+6
no_of_virtual_packets: 5
```

on a system with 16GB of RAM (no swap) and got TARDIS killed by the OS just after the final iteration was finished.

**Fix:**
Instead of allocating numpy arrays for the virt_packet* arrays all at once, copying from the C arrays, and freeing the C arrays, simply transfer ownership of the C array memory to a numpy array object.

With this fix, peak memory usage by virt_packet arrays is reduced to half.

### DIFF
--- a/tardis/montecarlo/montecarlo.pyx
+++ b/tardis/montecarlo/montecarlo.pyx
@@ -21,13 +21,8 @@ ctypedef np.int64_t int_type_t
 cdef extern from "numpy/arrayobject.h":
     void PyArray_ENABLEFLAGS(np.ndarray arr, int flags)
 
-cdef int_array_to_numpy(void *ptr, np.npy_intp N):
-    cdef np.ndarray[int_type_t, ndim=1] arr = np.PyArray_SimpleNewFromData(1, &N, np.NPY_INT64, ptr)
-    PyArray_ENABLEFLAGS(arr, np.NPY_OWNDATA)
-    return arr
-
-cdef double_array_to_numpy(void *ptr, np.npy_intp N):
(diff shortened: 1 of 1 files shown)


tardis/montecarlo/montecarlo.pyx
```diff
@@ -18,6 +18,14 @@ np.import_array()
 
 ctypedef np.int64_t int_type_t
 
+cdef extern from "numpy/arrayobject.h":
+    void PyArray_ENABLEFLAGS(np.ndarray arr, int flags)
+
+cdef c_array_to_numpy(void *ptr, int dtype, np.npy_intp N):
```
yeganer: This is just a really small thing, but as we are using numpy API, I think we should use `np.int_t dtype` instead of `int dtype`.

ftsamis: I'm not sure we should do this. The numpy reference declares this as an `int`: https://docs.scipy.org/doc/numpy/reference/c-api.array.html#c.PyArray_SimpleNewFromData

unoebauer:
Just tested it with ``tardis-example`` and the above stated settings for real and virtual packets in the spectral cycle (compiled with vpacket-logging and open-mp, run with 2 threads). Works like a charm and memory consumption does not increase after the spectral cycle is complete! Great job @ftsamis!

tardis/montecarlo/montecarlo.pyx
yeganer: Alright, then leave it.

wkerzendorf:
very good!

PR #697, opened 2017-01-26
Relativity

chvogl:
Addresses TEP010: Consistent Treatment of relativistic effects.

This PR implements:
* transformation of continuum (e.g. electron scattering) opacity
* consistent treatment of volume based estimators
* angle aberration
* common direction frequency surfaces of relativistic Sobolev theory

Still missing relativistic effects:
* transformation of Sobolev optical depth (only factor 1/gamma)
* j_blues_estimator (only factor 1/gamma) 
* (transformation of deflection angle for reflecting photosphere)

These missing effects are considered non-critical and insignificant.

### DIFF
--- a/tardis/montecarlo/src/cmontecarlo.c
+++ b/tardis/montecarlo/src/cmontecarlo.c
@@ -140,7 +140,7 @@ binary_search (const double *x, double x_insert, int64_t imin,
 }
 
 void
-do_angle_aberration (rpacket_t *packet, const storage_model_t *storage)
+angle_aberration_CMF_to_LF (rpacket_t *packet, const storage_model_t *storage)
 {
   if (storage->full_relativity)
     {
@@ -159,7 +159,7 @@ do_angle_aberration (rpacket_t *packet, const storage_model_t *storage)
  * @return CMF direction cosine
  */
 double
-inverse_angle_aberration (rpacket_t *packet, const storage_model_t *storage, double mu)
+angle_aberration_LF_to_CMF (rpacket_t *packet, const storage_model_t *storage, double mu)
 {
   double beta = rpacket_get_r (packet) * storage->inverse_time_explosion * INVERSE_C;
   return (mu - beta) / (1.0 - beta * mu);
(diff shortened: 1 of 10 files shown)


tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -139,12 +139,55 @@ binary_search (const double *x, double x_insert, int64_t imin,
   return ret_val;
 }
 
+void
+do_angle_aberration (rpacket_t *packet, const storage_model_t *storage)
```
yeganer: Since this is only active during `full_relativity` I'd do the relativity check in this function instead of outside.
That way you can't execute the function by accident and the code would be cleaner.

The disadvantage is you'd have to look at the function implementation to know this is `full_relativity` only. But I think we can live with that.

tardis/montecarlo/src/cmontecarlo.c
```diff
@@ -139,12 +139,55 @@ binary_search (const double *x, double x_insert, int64_t imin,
...
+/** Transform the lab frame direction cosine to the CMF
+ *
+ * @param packet
+ * @param storage
+ * @param mu lab frame direction cosine
+ *
+ * @return CMF direction cosine
+ */
+double
+inverse_angle_aberration (rpacket_t *packet, const storage_model_t *storage, double mu)
```
yeganer: Same as above.
Maybe explicitly add `inline` directive?

tardis/montecarlo/src/cmontecarlo.c
chvogl: I think this is an excellent idea.

tardis/montecarlo/src/rpacket.c
```diff
@@ -18,14 +18,20 @@ rpacket_init (rpacket_t * packet, storage_model_t * storage, int packet_index,
...
-  current_energy =
-    current_energy / (1 -
-                      (current_mu * current_r *
-                       storage->inverse_time_explosion * INVERSE_C));
+
+  double beta = current_r * storage->inverse_time_explosion * INVERSE_C;
+
+  if (storage->full_relativity)
+    {
+      current_nu = current_nu * (1 + beta * current_mu) / sqrt(1 - beta * beta);
```
unoebauer: Couldn't you use the doppler_factor routines for this?

Issue #680, opened 2016-12-16
Many docstrings are inconsistent with the numpydoc format


ftsamis:
Using the numpydoc format allows for type hinting in PyCharm and easy linking to other docstrings. However, in many places the docstrings violate this format, breaking the type hinting for PyCharm.

An example:
```python
def calculate_density_after_time(densities, time_0, time_explosion):
    """
    scale the density from an initial time of the model to the
    time of the explosion by ^-3

    Parameters:
    -----------

    densities: ~astropy.units.Quantity
        densities

    time_0: ~astropy.units.Quantity
        time of the model

    time_explosion: ~astropy.units.Quantity
        time to be scaled to

    Returns:
    --------

    scaled_density
    """
```

Here, the keyword "Parameters" is followed by a colon, immediately breaking type hinting, and furthermore the parameter types start with a "~", which is unnecessary and also breaks class resolving.

There are probably more kinds of docstring errors in other places too, so this needs some careful research.

Note: You can view the formatted documentation of an object with `ctrl + q` in PyCharm.

nitinkgp23:
Hey, I am newcomer here and I would like to work on this issue. Can someone help me in going forward with this?

yeganer:
What questions do you have?

unoebauer:
We may think about making an aggregate issue, collection everything related to coding style.

debajyotidasgupta:
hello,
i am new to this organisation and would like to take this as my first pr
It would be helpful if tell me what all changes should i make?

ImmortalRabbit:
Is this issue still open? It's just something was merged already.

arjunsavel:
@marxwillia should we close this, or did the relevant PR merge not completely fix this?

PR #701, opened 2017-01-28
Make docstrings consistent with numpydoc format

nitinkgp23:
In many places, docstrings violate the numpydoc format, hence breaking
the type hinting in PyCharm. This commit resolves all those
inconsistencies.

Fixes https://github.com/tardis-sn/tardis/issues/680


nitinkgp23:
@ftsamis Can you please review the PR? :)

tardis/model/density.py
```diff
@@ -137,19 +137,19 @@ def calculate_density_after_time(densities, time_0, time_explosion):
     scale the density from an initial time of the model to the
     time of the explosion by ^-3
 
-    Parameters:
+    Parameters
     -----------
 
-    densities: ~astropy.units.Quantity
+    densities: astropy.units.Quantity
```
ftsamis: Let's keep the space before the `:` consistent. There are other instances of this too.

tardis/montecarlo/base.py
```diff
@@ -272,16 +272,16 @@ def calculate_radiationfield_properties(self):
         Parameters
         ----------
 
-        nubar_estimator : ~np.ndarray (float)
+        nubar_estimator : np.ndarray (float)
```
ftsamis: I [found out](https://www.jetbrains.com/help/pycharm/2016.3/type-hinting-in-pycharm.html#legacy) that PyCharm suggests `np.ndarray[float]`. Check the link for more info. Would be nice to also fix any other instances that match the table in the link

tardis/montecarlo/packet_source.py
```diff
@@ -40,14 +40,14 @@ def create_packet_nus(self, T, no_of_packets):
...
-        no_of_packets: ~int
+        no_of_packets: int
 
         Returns
         -------
 
-            : ~numpy.ndarray
+            : numpy.ndarray
             array of frequencies
         """
```
ftsamis: The indentation here is wrong, also no need for the leading `:` if there is no name for the return value but only a type. There are more similar instances.

tardis/plasma/standard_plasmas.py
```diff
@@ -44,15 +44,15 @@ def assemble_plasma(config, model, atom_data=None):
...
+    config: io.config_reader.Configuration
+    model: model.Radial1DModel
+    atom_data: atomic.AtomData
         If None, an attempt will be made to read the atomic data
         from config.
 
     Returns
     -------
-    : ~plasma.BasePlasma
+    : plasma.BasePlasma
```
ftsamis: No need for the leading `:` as previously mentioned.

tardis/simulation/base.py
```diff
@@ -139,7 +139,7 @@ def advance_state(self):
 
         Returns
         -------
-            converged : ~bool
+            converged : bool
```
ftsamis: Over-indentation

Issue #702, opened 2017-01-28
Replace MockRegexPattern with proper regular expressions


ftsamis:
[MockRegexPattern](https://github.com/ftsamis/tardis/blob/152c3f74714ae3515646122be2012ca3b12cd339/tardis/io/util.py#L35-L60) is a class introduced to easily find (match with) strings that represent astropy Quantities in the YAML configuration. 

The main problem with matching Quantities is units. We could either be too coarse, matching everything that looks like `<number><space(s)><string>` to a Quantity, or too strict and require that a string must be passed to a `string_to_quantity` function and throw no exceptions to be considered a valid Quantity.

We have chosen to go with the strict way with the tradeoff of making parsing extremely slower compared to if it used actual regular expressions, significantly affecting unit tests performance too (which load configuration files dozens of times).

Furthermore, since the default regular expression for floats in the pyyaml package didn't work very nicely, not recognizing some forms of floats, thus requiring `-.inf` instead of `-inf`, `5.e+5` instead of `5e5` or `5.e5` we chose to use `MockRegexPattern` for floats too, passing the strings to the python `float` constructor and seeing if it throws any exception.

After some thinking, I'm persuaded we should stick with regular expressions:

- For floats, write a proper regular expression that would match with all valid `float` types in Python. Examples include: `nan`, `inf`, `5e1`, with all their signed or decimal variants (`-inf`, `-nan`, `+5.0e+1`, and so on)
- For quantities create a regular expression that would use the regular expression for floats as mentioned above, in the form of `<float_regex>\s*[a-zA-Z/ ]*` and let the quantity yaml constructor complain if the units are invalid.

Keep in mind that with the current code and in the worst case, every single value in a YAML file could be checked against `MockRegexPattern` for Quantity, fail, then be checked against `MockRegexPattern` for float, fail, and then match against its actual type. That leads to quite some significant overhead, and the order of the resloving is not in our control.

yeganer:
I'm not sure but how the `!object` syntax in yaml works but essentially we could require all quantities to be marked with `!quantity` for example. This would be explicit but it won't be backwards compatible.

ftsamis:
This is actually possible now too. What the implicit resolver (and therefore the regex) does is understand that without the tag. I don't think we should make the tag a requirement and break compatibility since we don't currently have any other property type resolving of which would be ambiguous with the resolving of quantities. It would also take off from simplicity in the syntax.

yeganer:
[...]

PR #703, opened 2017-01-28
Replace MockRegexPattern with real regex patterns for quantity and float parsing in YAMLLoader

ftsamis:
This is a fix for #702. For a detailed description check the issue.

### DIFF
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -1,6 +1,7 @@
 #Utility functions for the IO part of TARDIS
 
 import os
+import re
 import pandas as pd
 import numpy as np
 import collections
@@ -11,6 +12,34 @@ from tardis.util import element_symbol2atomic_number
 
 import logging
 logger = logging.getLogger(__name__)
+
+float_pattern = r"""    # A numeric string consists of:
+    (?P<sign>[-+])?              # an optional sign, followed by either...
+
+    (
+
+        (?=\d|\.\d)              # ...a number (with at least one digit)
+
+        (?P<int>\d*)             # having a (possibly empty) integer part
+
+        (\.(?P<frac>\d*))?       # followed by an optional fractional part
+
+        (E(?P<exp>[-+]?\d+))?    # followed by an optional exponent, or...
+
+    |
+
+        Inf(inity)?              # ...an infinity, or...
+
+    |
+        NaN                      # NaN
+    )
+"""
+
+unit_pattern = r"[a-zA-Z1(][\w\s/^*()+-]*"
+
+float_regex = re.compile("^" + float_pattern + "$", re.VERBOSE | re.IGNORECASE)
+quantity_regex = re.compile("^" + float_pattern + "\s+" + unit_pattern + "$",
+                            re.VERBOSE | re.IGNORECASE)
 
 
 def quantity_from_str(text):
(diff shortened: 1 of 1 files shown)


tardis/io/util.py
```diff
@@ -12,6 +13,34 @@
...
+
+    (
+
+        (?=\d|\.\d)              # ...a number (with at least one digit)
+
+        (?P<int>\d*)             # having a (possibly empty) integer part
+
+        (\.(?P<frac>\d*))?       # followed by an optional fractional part
+
+        (E(?P<exp>[-+]?\d+))?    # followed by an optional exponent, or...
```
yeganer: I think both `E` and `e` should be allowed. This pattern wouldn't recognize `1e10`, right?

tardis/io/util.py
```diff
@@ -12,6 +13,34 @@
...
+
+    |
+
+        Inf(inity)?              # ...an infinity, or...
```
yeganer: I think `inf` should be valid, too.

tardis/io/util.py
```diff
@@ -12,6 +13,34 @@
...
+
+    |
+        NaN                      # NaN
```
yeganer: Please add `nan`.

tardis/io/util.py
ftsamis: This pattern is compiled with re.IGNORECASE, so both E and e are allowed.

tardis/io/util.py
ftsamis: It is, the regex translates to "inf" optionally followed by "inity". If you are talking about the case, we use this pattern with re.IGNORECASE.

wkerzendorf:
@ftsamis you have now programmed into TARDIS things that are built-in to Python. Using the built-in method works - not clear what advantage (other than you learning more advanced regexp) this has?

tardis/io/util.py
ftsamis: Regarding lower-case `nan` look at my previous replies.
Python supports and converts both +nan and -nan to just nan. This proves that there is no good reason to support signed nans, but also there is no good reason not to. :P

ftsamis:
@wkerzendorf Regarding the float regex I'd think of it more like fixing/monkey-patching a [broken regex pattern in pyyaml](https://bitbucket.org/xi/pyyaml/src/10c55a624429f28191ef2bc1d9db44684946eb5d/lib/yaml/resolver.py?at=default&fileviewer=file-view-default#resolver.py-177:184). Python has nothing to do with it other than me basing on their pattern for decimals to be as compatible as possible with Python.

`MockRegexPattern` was not a built-in method, instead it was a hack that did cost time, especially in tests. See #702 for a detailed description and some metrics.

tardis/io/util.py
yeganer: I forgot that we use `re.IGNORECASE`.

tardis/io/util.py
yeganer: Oh right, but I think we should use `e` as afaik most ppl use the lowercase letter.
To be more explicit I'd drop `re.IGNORECASE` and modify the regex slightly.

yeganer:
@wkerzendorf This is a bug in PyYAML and a discussion on [stackoverflow](http://stackoverflow.com/questions/30458977/yaml-loads-5e-6-as-string-and-not-a-number) suggests a fixed regex for floats.
There is also a fork fixing this bug and others because apparently the original project hasn't been updated in a while.

I think this is something worth changing and as we already have a working solution I don't see a reason to not do this.
If you don't trust the above regex, one could use the regex from stackoverflow which matches everything defines as float by the JSON specification and subsequently the YAML spec.

tardis/io/util.py
ftsamis: Capital letter `e` for scientific notation is perfectly valid syntax and used widely. Let's not restrict things for no good reason. If we drop the `re.IGNORECASE` we'd have to make this pattern bigger to catch all the variants of `NaN` (`nan`, `NAN`, ...) and `inf`, or restrict syntax and not allow `Inf` (with a capital `i`) for example, which there is no good reason doing that I can think of.

yeganer: I want to match both `e` and `E` but I prefer `e` in the code, that's all :)

ftsamis:
@yeganer How can one not trust a regular expression? It's not like it is a huge one that would take a team of expert parser authors to understand. Did you notice that the pattern on the stackoverflow page you link as more trustworthy does not match `inf` but still requires `.inf`?

And regarding the JSON specification, we do **not** want to follow the JSON specification (if it's the same as the float automaton shown at http://www.json.org/), since it doesn't acknowledge `inf` or `nan` as floats at all. That also means that the pattern you are linking does not strictly follow the JSON specification and it does break the JSON spec in way more than one ways (json does not allow starting with `+`, it doesn't allow starting with `.` and so on).

If anyone has trouble understanding/trusting the regexes I introduce in this PR and wants me to explain them character by character, let me know.

PR #704, opened 2017-01-28
Fix ignoring model tests

ftsamis:
When converting model to a package and transferring tests to create the `model/tests/test_base.py` I forgot to also make the tests folder a package by creating an `__init__` module. As a result, pytest didn't include the `test_base.py` unit tests.


ftsamis:
The good news is that this is passing!

Issue #699, opened 2017-01-27
[OpenMP] Use OMP_NUM_THREADS by default


yeganer:
If the configuration is missing `montecarlo.nthreads` use the OpenMP default value for parallelization.
On the C level this is already possible by calling `montecarlo_main_loop` with `nthreads=0`.

All that is left to do is change the default value in the yaml schema and verify this is working as expected.

saurvs:
Hello I'm a new contributor here and I'd like to take a stab at this.

unoebauer:
@yeganer -- is this still relevant?

unoebauer:
@yeganer, @wkerzendorf - I just had a look at this issue and came to the following conclusions:

* setting nthreads in the config to 0 triggers the desired behaviour - Tardis OMP uses the number of threads stored in the environment variable ``OMP_NUM_THREADS``
* **However**, if the environment variable is not set, OMP uses all threads available on the system (e.g. on my laptop with an intel i5 and hyperthreading 4). 

Based on these findings, I'd advocate not to change the default value, since it would most likely lead to the undesired outcomes for inexperienced users who have not set ``OMP_NUM_THREADS``. Users with OMP experience who wish to control the Tardis OMP behaviour via an environment variable can still do so by setting nthreads to 1 in their configs.

wkerzendorf:
From my perspective - by default - TARDIS should use the maximum number of threads available. That means the default value should be 0 (not visible in the configuration file), but we should have in our examples nthreads=1 commented out so people see how they can change the number of threads.

unoebauer:
Ok, I think I can live with that. I'll include it in my PR.

unoebauer:
Resolved by #797.

PR #705, opened 2017-02-03
Set default value for nthreads in montecarlo schema to 0

saurvs:
Addresses https://github.com/tardis-sn/tardis/issues/699.

This ensures to that, if `nthreads` isn't specified, the value of the `OMP_NUM_THREADS` environment variable is used by OpenMP instead of `1`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,11 +60,11 @@ if not RELEASE:
 # invoking any other functionality from distutils since it can potentially
 # modify distutils' behavior.
 cmdclassd = register_commands(PACKAGENAME, VERSION, RELEASE)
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@saurvs okay - now go and find something more difficult ;-)

saurvs:
@wkerzendorf  Sure :). I also found some typo's in `setup.py` and put in this PR, if that's okay.

yeganer:
I think changing the typos in `setup.py` doesn't need a separate PR as they are thematically linked

wkerzendorf:
@tardis-sn/tardis-core we'll merge this later

Issue #707, opened 2017-02-05
[Documentation] Fix installation guidelines


yeganer:
I just had a quick look at our documentation and noticed several issues:

- Some places still suggest `pip install tardis-sn` in some way or another. This should be fixed to point to the development version
- We have guides for installing the tardis requirements via `pip-requirements`. Do we want to support pip at all or anaconda all the way?

unoebauer:
I lean towards only supporting anaconda.

unoebauer:
Closing in favour aggregate issue

PR #708, opened 2017-02-10
Add new installation instructions

wkerzendorf:
These are the new installation instructions as referenced in #707

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -5,8 +5,8 @@ Installation
 ************
 
 Before installing TARDIS, please check its :ref:`requirements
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -63,99 +63,16 @@ and after you are done with TARDIS you can deactivate::
 One does not need to recreate the environment, but simply activate it every time
 TARDIS is used.
```
unoebauer: I would be more verbose here: Since TARDIS has reached a mature state, we recommend always installing the latest development version.

Issue #679, opened 2016-12-16
Automatically convert and store ranges as linspaces


ftsamis:
Similar to what we do with quantities, which we now convert to `astropy.unit.Quantity` instances with a yaml resolver and constructor, should we also automatically convert ranges, currently expressed as `{'start': ..., 'stop': ..., 'num': ...}` dicts to `numpy.linspace`s? Most of the classes that take such ranges as parameters, eventually convert the dict to a linspace.

This could be easily done with a yaml constructor for the dict type, but it would require a few changes in other places of the code too.

The change would be seamless, since all values would be accessible:

- start would be linspace[0]
- stop would be linspace[-1]
- num would be range(linspace)

This is the base for #687. I have opened two separate issues for similar things because this one doesn't require us to bump the configuration version (no changes to the configuration file syntax required), while #687 does. 

This originated from #624.

**Example:**
`virtual_spectrum_range` is a range property (this can be verified by looking that in the configuration schema it is consisted from the properties `start`, `stop`, and `num`. 

The `YAMLLoader` should detect these properties and instead of returning a dict with the aforementioned keywords and their values it should return a `numpy.linspace` object with the correct values.

This way, when accessing `conf.montecarlo.virtual_spectrum_range` from a `tardis.io.config_reader.Configuration` object we would get a `np.linspace` directly without the need to create it every time in the various `from_config` classmethods.

mangalutsav:
@ftsamis is anyone else doing this?

ftsamis:
@mangalutsav This needs to be discussed with the TARDIS developers first in order to decide if it is something that we really want to do. For now it is just a proposal.

yeganer:
I think it's worth doing and it shouldn't be a difficult change.

If we want to be really fancy, we can subclass `np.ndarray` and add `start`, `stop` and `num` during initialization.

@mangalutsav Feel free to show us a PR

priyanka303:
@ftsamis is anybody else is doing this ??

ftsamis:
I am not aware of any PR on this. Feel free to demonstrate your approach with a PR.

yeganer:
@ftsamis Can you please edit the Issue description to more clearly reflect the changes required to implement this feature request and to differentiate it from #687.
I think there is some confusion going on between the goals of the issues.

Am I right in thinking the implementation of this will expect a dict with `start, stop, num, (unit)` keys and should return a `np.linspace`?
[...]

PR #710, opened 2017-02-12
Fix issue #679

suyashgupta20:
start, stop, num return as a np.linspace

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -5,8 +5,8 @@ Installation
 ************
 
 Before installing TARDIS, please check its :ref:`requirements
-<requirements_label>`. We provide instructions for multiple ways of installing
-TARDIS here, but we strongly recommend to follow the Anaconda :ref:`guide
+<requirements_label>`. We provide instructions for installing TARDIS using
+    Anaconda :ref:`guide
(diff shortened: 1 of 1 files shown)


tardis/io/util.py
```diff
@@ -87,6 +87,11 @@ def construct_quantity(self, node):
         return quantity_from_str(data)
 
     def mapping_constructor(self, node):
+        mapping_dict = self.construct_mapping(node)
+        if set(mapping_dict.keys()) == {'start', 'stop', 'num'}:
+            return OrderedDict([('linspace', np.linspace(mapping_dict['start'].cgs.value,
+                                                         mapping_dict['stop'].cgs.value,
+                                                         mapping_dict['num']))])
```
ftsamis: That seems like a good start, but you've missed two important points:

1. Why do you return an `OrderedDict` containing one `linspace`, instead of directly the `linspace` itself?
2. What about units? You ignore that information currently.

suyashgupta20:
image

@ftsamis I got this result as you want according to our chat.

ftsamis:
@suyashgupta20 This is not really what was described in the chat. Have a look again!

PR #711, opened 2017-02-24
from_hdf5 function (TEP014)

vg3095:
@wkerzendorf  
This PR is related to first and bonus objective for [TEP014](http://opensupernova.org/gsoc2017/doku.php?id=ideas_page) (Reading Simulation from file).
I have implemented a from_hdf5 function in Simulation Base class, which returns Radial1DModel object.
Presently it is not extended for Plasma and MonteCarlo Runner objects.
homogeneous_density and luminosity_requested are set to None for now.

Example Usage :

```
model = Simulation.from_hdf5('/path/to/model_output.hdf')
print model.time_explosion
print model.v_boundary_inner
print model.v_boundary_outer
```

### DIFF
--- a/tardis/simulation/base.py
+++ b/tardis/simulation/base.py
@@ -454,16 +454,16 @@ class Simulation(object):
         v_boundary_outer = sim_dict['model']['v_outer'][
             len(sim_dict['model']['v_outer']) - 1] * u.cm / u.s
         dilution_factor = np.array(sim_dict['model']['w'])
-        velocity = np.append(v_boundary_inner.value,
+        velocity = np.append(sim_dict['model']['v_inner'],
                              v_boundary_outer.value) * u.cm / u.s
 
-        #Presently homologous_density and luminosity_requested parameters are set to None
+        # Presently homologous_density and luminosity_requested parameters are
+        # set to None
         model = Radial1DModel(velocity, None, abundance, time_explosion,
                               t_inner, None, t_radiative,
                               dilution_factor, v_boundary_inner,
                               v_boundary_outer)
-        
-        #TODO Extend it to plasma and montecarlo objects and return Simulation object
+        # TODO : Extend it to plasma and montecarlo objects and return Simulation object
 
         return model
 
--- a/tardis/simulation/base.py
+++ b/tardis/simulation/base.py
@@ -4,10 +4,11 @@ import numpy as np
 import pandas as pd
 from astropy import units as u
 from collections import OrderedDict
-
+import h5py
 from tardis.montecarlo import MontecarloRunner
 from tardis.model import Radial1DModel
 from tardis.plasma.standard_plasmas import assemble_plasma
+import os
 
 # Adding logging support
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@vg3095 show me `with` syntax ;-)

vg3095:
@wkerzendorf Now , I have used '`with`' statement while handling files using `pd.HDFStore` and `h5py.File` function calls. 
No `logger` warnings are raised now.

tardis/simulation/base.py
```diff
@@ -410,3 +411,138 @@ def from_config(cls, config, **kwargs):
...
+        model = Radial1DModel(velocity, None, abundance, time_explosion,
+                              t_inner, None, t_radiative,
+                              dilution_factor, v_boundary_inner,
+                              v_boundary_outer)
+        # TODO : Extend it to plasma and montecarlo objects and return Simulation object
+
+        return model
+
+    @classmethod
+    def read_model_data(cls, h5_file, path, sim_dict, file_path):
```
ftsamis: This should live in the `Radial1DModel` class and named `from_hdf`.

tardis/simulation/base.py
```diff
@@ -410,3 +411,138 @@ def from_config(cls, config, **kwargs):
...
+            raise IOError("Supplied HDF5 File %s does not exists" % file_path)
+
+        with h5py.File(file_path, 'r') as h5_file:
+            sim_dict = {}  # It will store HDF Objects in a dict form
+            for simulation in h5_file.keys():
+                for key in h5_file[simulation]:
+                    sim_dict[key] = {}
+                    if 'model' in key:
+                        cls.read_model_data(
+                            h5_file, simulation + '/model/', sim_dict[key], file_path)
```
ftsamis: If you create a classmethod under `Radial1DModel` named `from_hdf` you could call this here like `Radial1DModel.from_hdf(hdf_file, ...)` and you'd get a fully populated Model. This way the `from_hdf` concept would be more modular and easier to extend.

tardis/simulation/base.py
```diff
@@ -410,3 +411,138 @@ def from_config(cls, config, **kwargs):
...
+                    if 'plasma' in key:
+                        cls.read_plasma_data(
+                            h5_file, simulation + '/plasma/', sim_dict[key], file_path)
```
ftsamis: Similar here, for `BasePlasma`.

tardis/simulation/base.py
```diff
@@ -410,3 +411,138 @@ def from_config(cls, config, **kwargs):
...
+
+        # Presently homologous_density and luminosity_requested parameters are
+        # set to None
+        model = Radial1DModel(velocity, None, abundance, time_explosion,
+                              t_inner, None, t_radiative,
+                              dilution_factor, v_boundary_inner,
+                              v_boundary_outer)
+        # TODO : Extend it to plasma and montecarlo objects and return Simulation object
+
+        return model
```
ftsamis: I know it's just a demo, but make sure the `Simulation.from_hdf` returns a `Simulation` object, not a model. Essentially, each class with a `to_hdf` should probably have a `from_hdf` classmethod, which returns a populated instance of the class it refers to.

tardis/simulation/base.py
```diff
@@ -4,10 +4,11 @@
 import pandas as pd
 from astropy import units as u
 from collections import OrderedDict
-
+import h5py
```
vg3095: As of now, `pandas.HDFStore()` does not support iterating over hierarchically , while `h5py` does.
The issue is still opened there . [Link](https://github.com/pandas-dev/pandas/issues/10143).

So, I am using `h5py` only for transversing HDF file , and for reading particular attribute (like `time_explosion`) , I am using `pd.HDFStore()`

tardis/simulation/base.py
vg3095: For the renaming part ,` read_model_data` , just stores attributes in a nested dictionary , and it does not return anything. Renaming it to `from_hdf `under `Radial1DModel` class will be misleading  , as it can mean, it returns `Radial1DModel `object .
So , I think , it will be better if I move it to `Radial1DModel `, without renaming.

I cannot return `Radial1DModel` ,just by reading model attributes in HDF file .
Because , some attributes, which are required for generating this object , also resides in `plasma` and `runner `path.

tardis/simulation/base.py
```diff
@@ -410,3 +411,138 @@ def from_config(cls, config, **kwargs):
...
+            raise ValueError("h5file Parameter can`t be None")
+        with pd.HDFStore(file_path, 'r') as data:
+            for key in h5_file[path].keys():
+                sim_dict[key] = {}
+                buff_path = path + '/' + key + '/'
+                sim_dict[key] = data[buff_path]
+
+
+    @classmethod
+    def read_plasma_data(cls, h5_file, path, sim_dict, file_path):
```
vg3095: It will be better if I move it to `BasePlasma` class, without renaming.(as for the same reasons , mentioned above for `Radial1DModel` class)

vg3095:
@ftsamis Thanks , for the detailed review . I have replied to some of the comments . Please take a look. As for the rest , I will update it accordingly.

tardis/simulation/base.py
ftsamis: What I am suggesting is to take the work for reading a `Radial1DModel` from an hdf file away from `Simulation` and into `Radial1DModel`. If you do that, you may find out that you probably don't need a separate `read_model_data` function, but just combining it with `from_hdf` could be fine.

In order to understand better what I'd want it to look like, take a look at `to_hdf`. You'll see that the `simulation.to_hdf`, basicall does nothing more than calling `model.to_hdf`, `runner.to_hdf` and `plasma.to_hdf` and redirecting the responsibility towards the smaller structures.

Regarding your second point, there are two solutions:
 - Either change the `to_hdf`s to store the missing attributes, or
 - Just assume access to the entire hdf file in each class (Model, Plasma, Runner, Simulation, ...) and read attributes from other sections.

For now, doing the second is just fine.

tardis/simulation/base.py
```diff
@@ -410,3 +411,138 @@ def from_config(cls, config, **kwargs):
...
+        None
+        """
+
+        if file_path is None:
+            raise ValueError("File Path can`t be None")
+        if not os.path.exists(file_path):
+            raise IOError("Supplied HDF5 File %s does not exists" % file_path)
+        if not h5_file:
+            raise ValueError("h5file Parameter can`t be None")
+        with pd.HDFStore(file_path, 'r') as data:
```
ftsamis: At least two of these checks are unnecessary, since when you reach the `pd.HDFStore()` call, it would also raise an exception itself.

vg3095:
@ftsamis 
I have made some structural changes as suggested by you.
Reposting the example usage from above , with a minor change

Example Usage (Renamed `from_hdf5` -> `from_hdf`) :

```
model = Simulation.from_hdf('/path/to/model_output.hdf')
print model.time_explosion
print model.v_boundary_inner
print model.v_boundary_outer
```

As of now , it cannot return `Simulation` object , as `convergence_strategy` cannot
be set to `None` , at the time of initialization.

Changes:

- Removed `read_model_data` and `read_plasma_data`
- Added `from_hdf` method in `Radial1DModel` class
- Renamed function `from_hdf5` to `from_hdf` (in Simulation Class)
- Removed some unnecessary checks

Issue #100, opened 2014-03-14
Colorize logger


wkerzendorf:
Currently our log messages are uncolorized which makes it a bit difficult to spot warnings and errors during runtime. There are some implementations floating around on the web which show the general gist.

souravsingh:
I wish to do this.How do I start?

wkerzendorf:
just color the log output. have a look online how other people do that.

souravsingh:
@wkerzendorf I have done the fix.Sorry for asking a very naive question-In which folder should the Script go?

souravsingh:
@wkerzendorf I have added a Pull Request for the script.I request you to have a look and see if it needs any changes-https://github.com/tardis-sn/tardis/pull/429

souravsingh:
I have updated the PR.

unoebauer:
@wkerzendorf, @souravsingh: status of this issue/feature request. The PR is out there and Travis seems to be happy with it...

souravsingh:
@unoebauer We need to make changes in a lot of scripts to make Colorizing actually work.I am working on it right now.

Pratik151:
@souravsingh  If you haven't made any progress on this I would like to give this a try.

souravsingh:
@Pratik151 I have a script for colorizing the logs.Just not able to figure out how to hook it into the project.

vedantrathore:
Can I work on this issue? Can someone point me on where to start with this?

ftsamis:
@vedantrathore Given that you have installed and successfully ran the tardis_example following the directions [here](http://tardis.readthedocs.io/en/latest/installation.html), you can see how TARDIS produces log messages and try to colorize them with a different colour depending on the type of the message (info, warning, error, debug).

Feel free to ask in [gitter](https://gitter.im/tardis-sn/tardis) if having trouble with any step.

wkerzendorf:
@unoebauer - I think this issue is of utmost importance to the project and therefore I'm resurrecting it.

PR #712, opened 2017-02-26
Colorize Logger

vg3095:
This PR is in reference to issue #100 .
It adds the ability to colorize logs output.
Make sure before testing it , you have installed coloredlogs

`pip install coloredlogs`

I have tested it while generating a Simulation object from the QuickStart Guide.
Here is a screenshot attached for the same.

colored_logs


wkerzendorf:
@vg3095 please solve this problem without added dependencies and so many changes in so many files.

vg3095:
I have used a script directly from this stackoverflow thread [#Link](http://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output) 
Screenshot
color_2



I was mistaken to import coloredlogs in every file before. It will work , if we add it to only tardis/__init__.py . (Using coloredlogs module). 
If additional formatting for log output is required, the use of coloredlogs module will simplify it.

tardis/color.py
```diff
@@ -0,0 +1,64 @@
+import logging
```
wkerzendorf: this should go  in a util package.

tardis/__init__.py
```diff
@@ -1,6 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
-
+import color
```
wkerzendorf: always use absolute imports

tardis/color.py
yeganer: As this code is borrowed from stackoverflow, you should mention the source.

vg3095:
Updated with proposed changes.

vg3095:
@yeganer 
From the official documentation . [(Link)](https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles)
1. > When logging was added to the Python standard library, the only way of formatting messages with variable content was to use the %-formatting method. Since then, Python has gained two new formatting approaches: string.Template (added in Python 2.4) and str.format() (added in Python 2.6).

2. > **_Logging(as of 3.2)_** provides improved support for these two additional formatting styles. 

3. >you cannot directly make logging calls using str.format() or string.Template syntax, because internally the logging package uses %-formatting to merge the format string and the variable arguments. There would no changing this while preserving backward compatibility, since all logging calls which are out there in existing code will be using %-format strings.

So, inshort , this means Logging module supports newer string formatting from Python 3.2 onwards.

wkerzendorf:
@vg3095 Should this be merged?

vg3095:
@wkerzendorf According to me , it can be merged.

PR #713, opened 2017-03-04
Automatically conversion and store ranges as linspaces(solution for #679)

suyashgupta20:
image

In the **"black_body_sampling"**, since it's quantity having a unit, so it returns its unit as well whereas the **'spectrum'** is a unitless quantity so its unit is not returning.

@ftsamis please review it once.

### DIFF
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -92,10 +92,10 @@ class YAMLLoader(yaml.Loader):
         if set(mapping_dict.keys()) == {'start', 'stop', 'num'}:
             if hasattr(mapping_dict['start'], 'unit'):
                 return quantity_linspace(mapping_dict['start'], mapping_dict['stop'],
-                                         mapping_dict['num'])
+                                         mapping_dict['num']+1)
             else:
                 return np.linspace(mapping_dict['start'], mapping_dict['stop'],
-                                         mapping_dict['num'])
+                                         mapping_dict['num']+1)
         return OrderedDict(self.construct_pairs(node))
 
 YAMLLoader.add_constructor(u'!quantity', YAMLLoader.construct_quantity)
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -8,6 +8,7 @@ from collections import OrderedDict
 import yaml
 from astropy import constants, units as u
 from tardis.util import element_symbol2atomic_number
+from tardis.util import quantity_linspace
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


tardis/io/util.py
```diff
@@ -87,6 +87,16 @@ def construct_quantity(self, node):
         return quantity_from_str(data)
 
     def mapping_constructor(self, node):
+        mapping_dict = self.construct_mapping(node)
+        if set(mapping_dict.keys()) == {'start', 'stop', 'num'}:
+            if type(mapping_dict['start']) == int:
```
ftsamis: The check here is flawed. What if it is a float? It would be better to check if it is a quantity with `hasattr('unit')` instead of checking if it is a number.

tardis/io/util.py
```diff
@@ -87,6 +87,16 @@ def construct_quantity(self, node):
...
+                return np.linspace(mapping_dict['start'],
+                                                mapping_dict['stop'],
+                                                mapping_dict['num']),
+            else:
+                return u.Quantity(np.linspace(mapping_dict['start'].value,
+                                                mapping_dict['stop'].value,
+                                                mapping_dict['num']), mapping_dict['start'].unit)
```
ftsamis: Use `tardis.util/quantity_linspace` here instead.

ftsamis:
The code looks good now, but `spectrum` is not a unitless quantity. Which file did you use as input to get what you show in the screenshot?

suyashgupta20:
Actually, I just make spectrum unitless because of checking the working of code. Otherwise, I used **tardis.tests.integration_tests.w7.config.yml** to check my code.

ftsamis:
@suyashgupta20 Oh, ok, that makes sense then. So, the next step to solve #679 would be to change the code wherever it uses `property[start]/[stop]/[num]` (which no longer exist with your changes) to handle property as a prepared linspace.

suyashgupta20:
Okay @ftsamis I will try for that.What is the meaning of changing just one or two cases ?
I mean to say that how many cases are there ?

suyashgupta20:
@ftsamis please have a look on that.

suyashgupta20:
@ftsamis please check it..now is it seems good or not?

yeganer:
I'm not an expert with the config implementation but I think we should differentiate between config options that expect a single quantity, like `initial_t_inner` and options that expect/represent an array of quantities, like `spectrum`.

This PR allows us to transform `start, stop, num` entries into linspaces. This change should be reflected in the validation by introducing a new type or using an existing type that ensures that once this change goes live, `spectrum` is always a linspace after config validation.

@ftsamis @wkerzendorf Your thought's on this?
Should we add this to the issue description and discuss there?

Issue #672, opened 2016-11-25
[Integration Tests] Merge 'generate_reference_path' and 'reference_path'


yeganer:
These two options should be merged as reference data will be generated in a subdirectory with the githash anyway.

deep110:
@wkerzendorf @yeganer I would like to work on this issue, can you point me where, to start with it.

unoebauer:
Closing since #715 has been merged

PR #715, opened 2017-03-13
Merge 'generate_reference_path' and 'reference_path' in integration tests (bug #672)

rinklejain:
When *--generate-references* argument is used then path specified in `generate_reference` is used to generate reference data otherwise path of directory already containing the reference data is specified in `reference`. So instead of two fields, they are merged into one.

- **tests/integration_tests/conftest.py**: merged `gen_ref_path` and `reference_filepath` to `reference_filepath`.  
- **tests/integration_tests/test_integration.py**: changes made in test file to comply with changes in conftest.py
- **docs/integration.yml**: updated doc as there is only one field required rather than `reference` and `generate_reference`

### DIFF
--- a/tardis/tests/integration_tests/conftest.py
+++ b/tardis/tests/integration_tests/conftest.py
@@ -40,7 +40,7 @@ def pytest_terminal_summary(terminalreporter):
             terminalreporter.config.getvalue("integration-tests")):
         # TODO: Add a check whether generation was successful or not.
         terminalreporter.write_sep("-", "Generated reference data: {0}".format(os.path.join(
-            terminalreporter.config.integration_tests_config['generate_reference'],
+            terminalreporter.config.integration_tests_config['reference'],
             tardis_githash[:7]
         )))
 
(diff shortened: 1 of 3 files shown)


tardis/tests/integration_tests/conftest.py
```diff
@@ -77,15 +77,18 @@ def plot_object(request):
...
 
     path = {
         'config_dirpath': request.param,
-        'reference_filepath': os.path.join(os.path.expandvars(
-            os.path.expanduser(integration_tests_config['reference'])), hdf_filename
-        ),
-        'gen_ref_path': os.path.join(os.path.expandvars(os.path.expanduser(
-            integration_tests_config['generate_reference'])), tardis_githash[:7]
-        ),
+        'reference_filepath': ref_path,
```
karandesai-96: You might want to think of another name here, because when you generate references, it would be like:
`path/as/mentioned/in/yml/githash` which is not "filepath" per se.

karandesai-96: /ping: @yeganer. Rest all of it looks logically correct to me. Please verify it once :rocket:

rjain111: Sure. I'll make the necessary changes.

unoebauer:
@wkerzendorf, @yeganer - can we merge this?

Issue #673, opened 2016-11-25
[Integration Tests] Add plot_function to 'model_quantities'


yeganer:
The idea is to compare the data and run the plotting function in only one test.

Additionally one could add `model_scalars` as a 2nd fixture for all scalars.

suyashgupta20:
@yeganer, @karandesai-96 I am interested in solving this issue. Till now I have gone through the _test_integration.py_ file and understand how the plots are forming at the server and will make a PR soon for this Issue.

unoebauer:
Closing due to inactivity

PR #716, opened 2017-03-20
plot_function for the 'model_quantities' (issue #673)

suyashgupta20:
Similar to the spectrum I have created a function `plot_model_quantities` which returns the plot corresponding to every `model_quantities` to the test function, `test_model_quantities`

### DIFF
--- a/tardis/tests/integration_tests/test_integration.py
+++ b/tardis/tests/integration_tests/test_integration.py
@@ -53,8 +53,19 @@ quantity_comparison = [
      'plasma.luminosity_inner.cgs.value'),
      ]
 
+scalar_comparison = [
+    ('/simulation/model/scalars',
+     'model.scalars'),
+    ('/simulation/plasma/scalars',
+     'plasma.scalars'),
+     ]
+
 @pytest.fixture(params=quantity_comparison)
 def model_quantities(request):
+    return request.param
+
+@pytest.fixture(params=scalar_comparison)
+def model_scalars(request):
     return request.param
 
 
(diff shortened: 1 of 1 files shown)


tardis/tests/integration_tests/test_integration.py
```diff
@@ -154,6 +156,34 @@ def test_model_quantities(self, model_quantities):
         tardis_quantity = eval('self.result.' + tardis_quantity_name)
         assert_allclose(tardis_quantity, reference_quantity)
 
+    def plot_model_quantities(self, model_quantities):
+        reference_quantity_name, tardis_quantity_name = model_quantities
+        reference_quantity = self.reference[reference_quantity_name]
+        tardis_quantity = eval('self.result.' + tardis_quantity_name)
+
+        gs_mq = (plt.GridSpec(2, 1, height_ratios=[3, 1]))
+        model_ax_mq = (plt.subplot(gs_mq[0]))
```
karandesai-96: You don't have to enclose these calls in parenthesis. Same thing applies to other occurrences as well.

suyashgupta20: okay, I will change this in my next commit.

suyashgupta20:
image
I got these type of plots in my report.

suyashgupta20:
@yeganer  @unoebauer...what are your views on this.

karandesai-96:
As the issue states, there are a certain set of scalars in model quantities which do not make sense to plot, so I think you should first add legends and a pass/ fail badge to the plot, and the next thing to do is to create a fixture parametrized y scalars only.

So you see, there would be `test_model_quantities`, `test_model_scalars`, `plot_model_quantities` but not `plot_model_scalars`.

suyashgupta20:
Okay @karandesai-96  i will make that changes.
actually i got the error for the values like `/simulation/plasma/level_number_density` in every reference fie that is at.h5, paper1.scatter.h5 etc. What we have to do for the values like this. @unoebauer

suyashgupta20:
@karandesai-96 I added fixture for model_scalars but the problem is that it's actual value (tardis_quantity) is not define inside the montecarlo and plasma like other values. so what we have to do for that ?

PR #717, opened 2017-03-21
Mark tardis full test as slow and adds commandline option for the same

rinklejain:
As mentioned on the ideas page for the project `Expanding the Integration-Testing Framework`, *Tardis full test* is marked as slow and commandline option `py.test` is added for it's execution.

- **tardis/conftest.py**: commandline option `py.test` added with boolean flag `store_true`. When `py.test` option is mentioned in the run test command *TARDIS full test* will run.

- **tardis/tests/test_tardis_full.py**: marker `slow` added to class `TestSimpleRun`.

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -75,7 +75,7 @@ def pytest_addoption(parser):
     parser.addoption("--less-packets",
                      action="store_true", default=False,
                      help="Run integration tests with less packets.")
-    parser.addoption("--py.test",dest='py.test',
+    parser.addoption("--full-test",dest='full-test',
                      action="store_true", default=False,
                      help="Run Tardis full test.")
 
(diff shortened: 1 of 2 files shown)


tardis/conftest.py
```diff
@@ -75,6 +75,9 @@ def pytest_addoption(parser):
     parser.addoption("--less-packets",
                      action="store_true", default=False,
                      help="Run integration tests with less packets.")
+    parser.addoption("--py.test",dest='py.test',
+                     action="store_true", default=False,
+                     help="Run Tardis full test.")
 
 
 # -------------------------------------------------------------------------
```
unoebauer: Well :) - the idea was to have a more illustrate command line argument, something like "--full-testing", or "--slow".

rjain111: I have changed the commandline argument as asked. I would like work on the 'Expanding the Integration-Testing Framework' project. How should I proceed @unoebauer?

PR #719, opened 2017-03-24
Updated gui to be compatible with current version.

Heringer-Epson:
Also had to modify the master base.py file to make atom_data a self variable and the MonteCarlo base.py file to make luminosity_inner a self variable.

### DIFF
--- a/tardis/base.py
+++ b/tardis/base.py
@@ -35,7 +35,6 @@ def run_tardis(config, atom_data=None):
         tardis_config = Configuration.from_config_dict(config)
 
     simulation = Simulation.from_config(tardis_config, atom_data=atom_data)
-    simulation.atom_data = atom_data
     simulation.run()
 
     return simulation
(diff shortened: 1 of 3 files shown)


tardis/base.py
```diff
@@ -35,6 +35,7 @@ def run_tardis(config, atom_data=None):
         tardis_config = Configuration.from_config_dict(config)
 
     simulation = Simulation.from_config(tardis_config, atom_data=atom_data)
+    simulation.atom_data = atom_data
```
unoebauer: I don't think you need this - atom-data is stored in the plasma object, which itself is an attribute of the simulation object.

tardis/montecarlo/base.py
```diff
@@ -295,8 +295,9 @@ def calculate_radiationfield_properties(self):
         return t_rad * u.K, w
 
     def calculate_luminosity_inner(self, model):
-        return (4 * np.pi * const.sigma_sb.cgs *
+        self.luminosity_inner = (4 * np.pi * const.sigma_sb.cgs *
                 model.r_inner[0] ** 2 * model.t_inner ** 4).to('erg/s')
+        return self.luminosity_inner
```
unoebauer: I would prefer if you do the assignment in the run function, i.e. undo this particular change and then set self.luminosity_inner similar to self.time_of_explosion.

PR #720, opened 2017-03-24
Update to gui

Heringer-Epson:
The variables in the user interface have been updated to reflect changes on tardis. The outdated version was making use of the 'model' class, which corresponds to an older implementation of tardis. The updated version uses the 'simulation' class.


unoebauer:
Hi @Heringer-Epson - does the Gui now work?

Heringer-Epson:
Yes, it should be working. The tests I ran opened the interface fine.

Note that I only made modifications on widget.py that were sufficient to
get the interface to run. This means that if the code had other functions
that did not generate errors, then they might still be outdated.
The modifications on lines 473-481 need to be carefully checked, to make
sure the updated variables mean the same was the outdated ones.

Best,
Epson

pfreddy:
After the update, the GUI starts indeed. However, the "Show linw info" button doesn't open the new window with the ifo about the last interactions. The error message:

File "/home/barna/anaconda3/envs/tardis/lib/python2.7/site-packages/tardis/gui/widgets.py", line 57, in show_line_info
    self.span.xy[2][0], self.tablecreator))
  File "/home/barna/anaconda3/envs/tardis/lib/python2.7/site-packages/tardis/gui/widgets.py", line 822, in __init__
    self.parent.model)
AttributeError: 'ModelViewer' object has no attribute 'model'

These lines are not affected by this update.

ssim:
If #765 is merged hopefully some of the issues flagged here are resolved and this PR can be closed?

unoebauer:
@pfreddy, @Heringer-Epson - we have updated the GUI (PR #765, already merged into master). Can you check whether these changes solve the problems you experienced?

Heringer-Epson:
Seems to be working. Only thing that is not clear to me is how to set which
wavelength range one wants. The range seems to be 5000-10000 by default.
Anyway, this is not a big deal (for me at least.)

Best, Epson.
p.s. Sorry the delay to reply - I just came back from vacation.

unoebauer:
@Heringer-Epson - if you are referring to the range for extracting additional line information, you can simply drag the boundaries of the red shaded region with the cursor and thus set by hand the wavelength range.

pfreddy:
@unoebauer It works! However, I had to slightly modify tardis/analysis.py, because the required (tardis_env27.yml) pandas 0.20 doesn't use the 'sort' function, but 'sort_values' instead.

line 106 >   self.last_line_in_table.sort_values('count', ascending=False,
inplace=True)
line 110 > self.last_line_out_table.sort_values('count', ascending=False,
inplace=True)

Thanks!

wkerzendorf:
@Heringer-Epson @unoebauer what's going on with this PR?

unoebauer:
I think, we can abandon and close this PR since the GUI was officially fixed with PR #765. Any objections, @Heringer-Epson?

PR #721, opened 2017-03-24
Unit-Testing Project, Prelim Task for GSoC-17 (Replace the comparison values with loaded from file)

suyashgupta20:
## Unit-Testing the Plasma Module
**In this PR:**
as given on the idea page for the unit testing project, the prelim task is to replace comparison value with the values loaded from a file.

- I replace the comparison values with the values coming from HDF file **partition_compare_data.h5** and

- add this file in data of plasma's test folder.


wkerzendorf:
use the `to_hdf` functionality of the plasma.

suyashgupta20:
ok @wkerzendorf, I will change it in my next commit. are the other things correct?

suyashgupta20:
@wkerzendorf  
In this PR I am not writing anything in the HDF file. Are you saying that to write the reference data to the HDF file by `to_hdf` method instead of providing it manually?

wkerzendorf:
@suyashgupta20 nope - but the plasma module has the capability to write to the hdf file - you should use this file format

suyashgupta20:
I have synchronized this PR for generating the reference for plasma and understanding now how can I run it in the command line and use it as a reference data.
@wkerzendorf is this method appropriate now as your requirement ?

suyashgupta20:
My idea first generates the reference and then use it as reference data till we are not generating the another reference.
The second part is pending now, I will complete it very soon.

wkerzendorf:
@suyashgupta20 the build fails - that is already a bad sign

suyashgupta20:
The build is failing because the task is incomplete now. I just want an Inspection from you that it is correct way or not ?

tardis/plasma/tests/test_partition_function.py
```diff
@@ -1,19 +1,80 @@
...
+def partition_compare_data_fname():
+    fname = 'partition_compare_data.h5'
+    return os.path.join(path[0], 'plasma', 'tests', 'data', fname)
+
+@pytest.fixture()
+def partition_compare_data(partition_compare_data_fname):
+    return h5py.File(partition_compare_data_fname, 'r')
+
+@pytest.mark.skipif(not pytest.config.getvalue("partition-test"),
+                    reason="partition tests are not included in this run")
```
unoebauer: According to travis, this is exactly where all the tests fail. The "partition-test" config value is not defined. You'll have to do this. Just check how we did this with the "--atomic-dataset" config value and replicate.

suyashgupta20: Yes, I have understood it now but the problem is to running it on the command line.

suyashgupta20:
@wkerzendorf, @unoebauer my Idea is to generate the reference through the command line and use it as a reference data for test_partition_function until we are not generating the next reference.
Now I am looking at methods for command line by which reference data will generate same as the integration test.

PR #727, opened 2017-05-22
Added from_hdf method under Plasma, Model , MonteCarloRunner and Simulation class (TEP014)

vg3095:
@wkerzendorf  @chvogl 
In this PR , I made the following additions:-

- Added `from_hdf` and `to_hdf` method under `Homologous Density` Class
- Added  `from_hdf` method under `Model` and `Plasma` class

Previously , I thought of a reverse lookup dictionary type structure (ex- `t_rad` -> `TRadiative`), to generate `plasma_properties` list from saved attributes under HDF file , as I mentioned in my proposal . The problem I encountered was that some properties have same key but different values (ex `phi` ->`PhiSahaLTE` , `PhiSahaNebular`) . 
To resolve these conflicts , I would have to save the config properties , such as `ionization` , `excitation` , etc. If I put a check for each such type , basically I would rewrite `assemble_plasma` method .

So, I thought of a workaround to reuse `assemble_plasma` method using the extracted `model` from HDF file and a `config` object created using only Plasma attributes saved in HDF file. 
To create a custom config object from Plasma attributes , I introduced a `plasma_only` parameter in  `from_config_dict` method , whose default value is `False` (meaning it will work as it was working previously), but if I pass this as `True` in combination with `validate=False` , I can get a config object of Plasma only attributes.

`Atomic_data` is required for `Plasma` object initialisation , which is not saved under `Simulation` HDF file . I am reading `atomic_data` from separate file. Default argument is set to the `atomic_data` provided with example file from TARDIS. `atomic_data_uuid`  of file , should be same as that of 
`Simulation` HDF file.

Note -> `pandas.HDFStore()` does not support iterating over hierarchically , while `h5py` does.
That`s why I am using h5py module.

## Example Usage 
You will have to save Simulation object using this PR , as I have added some new properties to be stored under HDF file.
 
```
from tardis import run_tardis
sim = run_tardis('tardis_example.yml')
sim.to_hdf('model_output.hdf')
```

```
sim_hdf = Simulation.from_hdf('model_output.hdf')
#sim_hdf = Simulation.from_hdf('model_output.hdf','kurucz_cd23_chianti_H_He.h5')
print sim_hdf.model.w
print sim_hdf.plasma.f_lu
```


tardis/io/config_reader.py
```diff
@@ -232,7 +232,7 @@ def from_yaml(cls, fname, *args, **kwargs):
             yaml_dict, *args, **kwargs)
 
     @classmethod
-    def from_config_dict(cls, config_dict, validate=True, config_dirname=''):
+    def from_config_dict(cls, config_dict, validate=True, config_dirname='', plasma_only=False):
```
wkerzendorf: what do I need this switch for?

vg3095: I want to create a `custom config object` from saved `plasma attributes` in `HDF` file . I am creating this config , just from plasma attributes (not including `montecarlo` or `model` properties) , so that I can use it in `assemble_plasma` method.

**If I turn this switch** `ON` (`plasma_only=True`) , It will bypass checking `montecarlo` section (It checks for `convergence_strategy` under montecarlo) , which is not required for `assemble_plasma` method . If I do not introduce this switch , this function cannot return a `Configuration` object **based only on**  `Plasma attributes`. 

This is a workaround to reuse `assemble_plasma` method to create `Plasma` object from HDF file, and this switch will be required in this case only. Please read the 3rd paragraph of my description of PR, above for more info. 

Hope this satisfies your question.

vg3095:
@wkerzendorf 
I have updated code , to include these changes : -
- Added `from_hdf` method under `Montecarlo` class
- Proper initialization of `Simulation` object under its `from_hdf` method
- Removed dependency of `h5py` module and removed extra `h5_file` parameter from my methods
- I have saved `Simulation` object variables (such as `nthreads` , `convergence_strategy` , etc) , which are required for its initialization as `consts` group under HDF file
- Reconstructed `Simulation` object using `from_hdf` method could be run/iterated further

As pointed by @yeganer , to remove `h5py` dependency , I have made `hdf_properties` as class attributes , and I have updated my code to initialize using `cls`

### Example Usage
(Reposting from above with minor changes)

You will have to save Simulation object using this PR , as I have added some new properties to be stored under HDF file.

```
from tardis import run_tardis
sim = run_tardis('tardis_example.yml')
sim.to_hdf('model_output.hdf')
```

```
sim_hdf = Simulation.from_hdf('model_output.hdf')
#sim_hdf = Simulation.from_hdf('model_output.hdf','kurucz_cd23_chianti_H_He.h5')
print sim_hdf.model.w
print sim_hdf.plasma.f_lu
print sim_hdf.runner.output_energy
```

wkerzendorf:
@vg3095 I think now we are at a point - where it would be great to add some tests and see how this would work in practice - saving to and loading from hdf.

vg3095:
I have added unit tests.
I am comparing properties of  (1) and (3) 
`Simulation object`(1) ----> Save through `to_hdf`(2) ----> Reconstruct through `from_hdf` using that saved file(3)

Issue #725, opened 2017-05-19
[Feature Request] Give Radial1DModel, MonteCarloRunner and BasePlasma access to the Simulation object


yeganer:
I propose to add a new property to `Radial1DModel`, `MonteCarloRunner` and `BasePlasma` that stores a reference to the `Simulation` instance. I would call it `parent`.
This would, for example, allow an initialized MonteCarloRunner instance to do `self.parent.model.t_inner` to access the termperature of the model.

For the implementation I would use pythons `@property` and `@property.setter`. If the parent is accessed before it is linked to a Simulation, an appropriate Error should be raised.

This change allows easier access to properties not part of the current class.

unoebauer:
Please provide a clear motivation why this feature is needed.

yeganer:
From my point of view, there are two motivations:

1. Conceptually, there is a clear relationship between the each Class and the Simulation. Currently this relationship is only implemented one-way.

2. Allowing the access of the Simulation object and therefore the other parts of the simulation enables us to more easily access important values not part of the current class. Currently this is done by either copying the value and storing it in the class (bad because changing one value does not change all others) or by passing for example the model or plasma as arguments to the functions that heavily use other values.

Concrete example: [MonteCarloRunner stores model.volume](https://github.com/tardis-sn/tardis/blob/d86187c1cb95ffcd46d30bd2dfe62276d9f21035/tardis/montecarlo/base.py#L141)
Here the runner stores the value of `model.volume` which is used later by `calculate_radiationfield_properties`. In my opinion the current implementation contradicts our design decision to keep the physical quantities where they belong (volume is clearly defined by the physical structure of the simulation model). We could fix this in two ways: pass the model as argument to `calculate_radiationfield_properties` (this would work because this method is called by the simulation which has access to the model. If this was called by the plasma for example the calling method in plasma would need access to the model, too.)
The other solution would be to allow the three main classes access to the Simulation object, allowing to access these quantities directly.

In summary I think this is a very small addition that allows for a more elegant implementation of methods because it is not required to pass `model`, `runner` and `plasma` to every single method that accesses one quantity not stored in the current scope.
This won't break anything, however it would allow silly accesses like `self.parent.model.parent.plasma.parent.runner` but I think this would be a trade-off we can live with.

PR #728, opened 2017-06-02
Implement #725

yeganer:
A simple change that allows the three main dataholders (model, runner and plasma) access the parent (simulation) and therefore allow for less clutter in function calling.

It's just a small change which I think would benefit the overall code quality in the future.


wkerzendorf:
@yeganer let's discuss this later - specifically maybe instantiating runner every single time for each run

PR #734, opened 2017-06-06
[WIP] Refactoring from/to_hdf methods 

vg3095:
from/to_hdf Refactoring:
- [x] Spectrum
- [x] Homologous Density
- [x] Model
- [ ] Plasma
- [x] Runner
- [ ] Simulation

Unit Tests for above:
- [x] Spectrum
- [x] Homologous Density
- [x] Model
- [ ] Plasma
- [x] Runner
- [ ] Simulation
- [x] HDFReaderWriter

### DIFF
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -166,10 +166,10 @@ def check_equality(item1, item2):
         return True
 
 
-class HDFReaderWriter():
+class HDFReaderWriter(object):
 
-    @classmethod
-    def to_hdf(cls, path_or_buf, path, elements, complevel=9, complib='blosc'):
+    
+    def to_hdf_util(self, path_or_buf, path, elements, complevel=9, complib='blosc'):
         """
         A function to uniformly store TARDIS data
         to an HDF file.
@@ -225,27 +225,25 @@ class HDFReaderWriter():
                     scalars_series = store[scalars_path].append(scalars_series)
             scalars_series.to_hdf(path_or_buf, os.path.join(path, 'scalars'))
     
-    scalars=['delta_frequency']
-
     @classmethod
-    def from_hdf(cls, path, file_path):
-        pass
-        # homologous_density_path = path + '/spectrum'
-        # homologous_density = {}
+    def from_hdf_util(cls, path, file_path):
+        hdf = {}
 
-        # with pd.HDFStore(file_path, 'r') as data:
-        #     for key in cls.hdf_properties:
-        #         homologous_density[key] = {}
-        #         if key in cls.scalars:
-        #             continue
-                
-        #         buff_path = homologous_density_path + '/' + key + '/'
-        #         homologous_density[key] = data[buff_path]
-        # return cls(*homologous_density)
-        # #Creates corresponding astropy.units.Quantity objects
-        # # density_0 = np.array(homologous_density['density_0']) * u.g / u.cm**3
-        # # time_0 = (homologous_density['scalars']['time_0']) * u.s
-        # # return cls(density_0, time_0)
+        with pd.HDFStore(file_path, 'r') as data:
+            for key in cls.hdf_properties:
+                hdf[key] = {}
+                buff_path = path + '/' + key + '/'
+                if key in cls.quantity_attrs:
+                    hdf[key] = u.Quantity(
+                        data[buff_path], cls.quantity_attrs[key])
+                else:
+                    try:
+                        hdf[key] = data[buff_path]
+                    except:
+                        hdf[key] = data[path + '/scalars'][key]
+                        if hdf[key] is 'none':
+                            hdf[key] = None
+        return hdf
 
 
 
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -256,7 +256,7 @@ class HDFReaderWriter(object):
                     except:
                         buff_path = os.path.join(path, 'scalars')
                         hdf[key] = data[buff_path][key]
-                        if hdf[key] is 'none':
+                        if hdf[key] == 'none':
                             hdf[key] = None
         return hdf
 
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -237,8 +237,8 @@ class HDFReaderWriter(object):
                 buff_path = path + '/' + key + '/'
                 if key in cls.quantity_attrs:
                     try:
-                        if isinstance(data[buff_path], pd.Series):
-                            hdf[key] = np.array(data[buff_path])
+                        if isinstance(data[buff_path], pd.Series) and len(data[buff_path])>1:
+                            hdf[key] = u.Quantity(np.array(data[buff_path]), cls.quantity_attrs[key])
                         else:
                             hdf[key] = u.Quantity(
                             data[buff_path], cls.quantity_attrs[key])
@@ -247,7 +247,7 @@ class HDFReaderWriter(object):
                              data[path + '/scalars'], cls.quantity_attrs[key])
                 else:
                     try:
-                        if isinstance(data[buff_path], pd.Series):
+                        if isinstance(data[buff_path], pd.Series) and len(data[buff_path])>1:
                             hdf[key] = np.array(data[buff_path])
                         else:
                             hdf[key] = data[buff_path]
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -228,7 +228,7 @@ class HDFReaderWriter(object):
             scalars_series.to_hdf(path_or_buf, os.path.join(path, 'scalars'))
     
     @classmethod
-    def from_hdf_util(cls, path, file_path):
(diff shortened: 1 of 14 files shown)


tardis/io/util.py
```diff
@@ -166,7 +166,91 @@ def check_equality(item1, item2):
...
+        #         homologous_density[key] = data[buff_path]
+        # return cls(*homologous_density)
+        # #Creates corresponding astropy.units.Quantity objects
+        # # density_0 = np.array(homologous_density['density_0']) * u.g / u.cm**3
+        # # time_0 = (homologous_density['scalars']['time_0']) * u.s
+        # # return cls(density_0, time_0)
+
+
+
+@classmethod
```
wkerzendorf: `to_hdf` is not a class method as it will write `to_hdf` after the object has been initialized.

tardis/montecarlo/spectrum.py
```diff
@@ -93,7 +93,5 @@ def to_hdf(self, path_or_buf, path='', name=''):
         if not name:
             name = 'spectrum'
         spectrum_path = os.path.join(path, name)
-        properties = ['luminosity_density_nu', 'delta_frequency', 'wavelength',
-                      'luminosity_density_lambda']
-        to_hdf(path_or_buf, spectrum_path, {name: getattr(self, name) for name
-                                            in properties})
+        HDFReaderWriter.to_hdf(path_or_buf, spectrum_path, {name: getattr(self, name) for name
```
wkerzendorf: you do not need to to `HDReaderWriter.to_hdf` as it has already inherited.

tardis/montecarlo/spectrum.py
```diff
@@ -2,14 +2,14 @@
...
+from tardis.io.util import HDFReaderWriter
 
-
-class TARDISSpectrum(object):
+class TARDISSpectrum(object, HDFReaderWriter):
     """
     TARDIS Spectrum object
     """
-
+    hdf_properties = ['luminosity_density_nu', 'delta_frequency', 'wavelength',
```
yeganer: The `hdf_properties` *should* be identical to the arguments of `__init__`. However this does not work with this version of the spectrum class, hence why I suggested using the PR version.

tardis/io/util.py
```diff
@@ -166,7 +166,91 @@ def check_equality(item1, item2):
         return True
 
 
-def to_hdf(path_or_buf, path, elements, complevel=9, complib='blosc'):
+class HDFReaderWriter():
```
wkerzendorf: please inherit from `object`

tardis/io/util.py
```diff
@@ -166,7 +166,91 @@ def check_equality(item1, item2):
...
+
+    @classmethod
+    def to_hdf(cls, path_or_buf, path, elements, complevel=9, complib='blosc'):
```
wkerzendorf: Elements should be attached to the classes as class attributes. Let me know if this is unclear

tardis/io/util.py
```diff
@@ -166,7 +166,91 @@ def check_equality(item1, item2):
...
+        """
+        scalars = {}
+        for key, value in elements.iteritems():
+            if hasattr(value, 'cgs'):
+                value = value.cgs.value
+            if np.isscalar(value):
+                scalars[key] = value
+            elif hasattr(value, 'shape'):
+                if value.ndim == 1:
+                    # This try,except block is only for model.plasma.levels
```
wkerzendorf: Then you can do a `PlasmaHDFReaderWriter` which inherits from this one.

wkerzendorf:
@vg3095 maybe doing this on Homologous density with a test where you can save and then reconstruct is a good way to continue this development from

tardis/io/util.py
vg3095: You are seeing the old commit . I have updated it

tardis/io/util.py
vg3095: I do not understand this point. Can you give an example

tardis/io/util.py
```diff
@@ -166,7 +166,95 @@ def check_equality(item1, item2):
...
+            scalars_series.to_hdf(path_or_buf, os.path.join(path, 'scalars'))
+    
+    @classmethod
+    def from_hdf_util(cls, path, file_path):
+        hdf = {}
+
+        with pd.HDFStore(file_path, 'r') as data:
+            for key in cls.hdf_properties:
+                hdf[key] = {}
+                buff_path = path + '/' + key + '/'
```
wkerzendorf: us `os.path.join()`

vg3095: This path is inside HDF store,

wkerzendorf:
That starts to look very good.

wkerzendorf:
The next step is to do a reader writer exercise in pytest - write to a file and then read from it and see that the objects are the same.

tardis/io/util.py
wkerzendorf: you suggest that this is only relevant for `model.plasma.levels` then maybe it should live there as an addition to the `to_hdf` in this specific class.  But we can discuss this later.

tardis/io/util.py
wkerzendorf: I understand - but `os.path.join` works for these paths as well.

tardis/io/util.py
```diff
@@ -166,7 +166,101 @@ def check_equality(item1, item2):
...
+    def from_hdf_util(cls, file_path, path=''):
+        hdf = {}
+
+        with pd.HDFStore(file_path, 'r') as data:
+            for key in cls.hdf_properties:
+                hdf[key] = {}
+                buff_path = path + '/' + key + '/'
+                if key in cls.quantity_attrs:
+                    try:
+                        if isinstance(data[buff_path], pd.Series) and len(data[buff_path])>1:
```
wkerzendorf: I don't think you need this. I think both `pd.Series` and `pd.DataFrame` have a values attribute. You should also never really use `isinstance` but use ducktyping.

tardis/io/util.py
```diff
@@ -166,7 +166,101 @@ def check_equality(item1, item2):
...
+                            hdf[key] = u.Quantity(np.array(data[buff_path]), cls.quantity_attrs[key])
+                        else:
+                            hdf[key] = u.Quantity(
+                            data[buff_path], cls.quantity_attrs[key])
+                    except:
+                         hdf[key] = u.Quantity(
+                             data[path + '/scalars'][key], cls.quantity_attrs[key])
```
wkerzendorf: `os.path.join`

tardis/io/util.py
```diff
@@ -166,7 +166,101 @@ def check_equality(item1, item2):
...
+                else:
+                    try:
+                        if isinstance(data[buff_path], pd.Series) and len(data[buff_path])>1:
```
wkerzendorf: same as `abovez

tardis/model/density.py
```diff
@@ -78,6 +81,44 @@ def from_config(cls, config):
...
+            Path to transverse in hdf file
+        file_path : 'str'
+            Path of Simulation generated HDF file 
+        Returns
+        -------
+        model : `~HomologousDensity`
+        """
+
+
+        buff_path = path + '/homologous_density'
```
wkerzendorf: `os.path.join`

vg3095: I will change it everywhere , where I have written in this format

tardis/io/util.py
vg3095: The problem is that I want to convert `pd.Series` objects into `np.array` objects , and `pd.DataFrame` can also be converted to `np.array` without raising an error. So, I think using ducktyping , will not help me here.

tardis/io/util.py
vg3095: This `to_hdf` block is just moved from existing code. I have only made one addition , a check if value is `None`.

tardis/plasma/standard_plasmas.py
```diff
@@ -153,3 +166,43 @@ def assemble_plasma(config, model, atom_data=None):
                         property_kwargs=property_kwargs, **kwargs)
 
     return plasma
+
+
+class PlasmaHDFReaderWriter(HDFReaderWriter, BasePlasma):
+
+    @classmethod
+    def from_hdf(cls, file_path, path, model, atom_data_fname=None):
```
vg3095: @wkerzendorf I want to place this `from_hdf` method under `BasePlasma` class , but this function(`from_hdf`) depends on `assemble_plasma` . If I place this function in `BasePlasma` class , it forms a case of `circular imports`. I am confused here , how can I proceed to refactor this??

wkerzendorf:
This PR is getting rather long. Our suggestion is to split it up into one PR that just introduces the `ReaderWriter` and its test. And then seperate PRs for the plasma and other stuff. If you need help with git techniques @yeganer is your friend.

vg3095:
Ok , I will split it

PR #736, opened 2017-06-12
[TEP014] Added HDFReaderWriter class + Unit tests

vg3095:
- `class_properties` parameter is for nested class objects.
(For e.g. `homologous_density` inside `Radial1DModel` class)


tardis/io/tests/test_HDFReaderWriter.py
```diff
@@ -0,0 +1,105 @@
...
+
+class Mock_HDF(HDFReaderWriter, object):
+    hdf_properties = ['property']
+    quantity_attrs = {}
+
+    def __init__(self, property):
+        self.property = property
+
+    def to_hdf(self, file_path, path=''):
+        buff_path = os.path.join(path, 'dummy')
```
wkerzendorf: the path should be the path and a name (you use dummy, but we should use something else). set the `name` kwd to `None` and if it is `None` use the class name as the name

tardis/io/tests/test_HDFReaderWriter.py
```diff
@@ -0,0 +1,105 @@
...
+        self.to_hdf_util(file_path, buff_path, {name: getattr(self, name) for name
+                                                in self.hdf_properties})
+
+    @classmethod
+    def from_hdf(cls, file_path, path=''):
```
wkerzendorf: same here.

tardis/io/tests/test_HDFReaderWriter.py
```diff
@@ -0,0 +1,105 @@
...
+import pandas.util.testing as pdt
+import pytest
+from astropy import units as u
+from astropy.tests.helper import assert_quantity_allclose
+from numpy.testing import assert_almost_equal, assert_array_almost_equal
+
+from tardis.io.util import HDFReaderWriter
+
+
+class Mock_HDF(HDFReaderWriter, object):
```
wkerzendorf: refactor to `MockHDF`

tardis/io/tests/test_HDFReaderWriter.py
```diff
@@ -0,0 +1,105 @@
...
+    expected = Mock_HDF.from_hdf(fname, 'Mock_HDF')
+    assert_array_almost_equal(actual.property, expected.property)
+
+
+arrays = [['L1', 'L1', 'L2', 'L2', 'L3', 'L3', 'L4', 'L4'],
+           ['one', 'two', 'one', 'two', 'one', 'two', 'one', 'two']]
+tuples = list(zip(*arrays))
+mock_multiIndex = pd.MultiIndex.from_tuples(tuples, names=['first', 'second'])
+
+@pytest.mark.xfail
```
wkerzendorf: what's going on here.

tardis/io/util.py
```diff
@@ -166,6 +166,116 @@ def check_equality(item1, item2):
         return True
 
 
+class HDFReaderWriter(object):
+
+    
+    def to_hdf_util(self, path_or_buf, path, elements, complevel=9, complib='blosc'):
```
wkerzendorf: could this be a static method?

tardis/io/tests/test_HDFReaderWriter.py
vg3095: I first created a dummy `MultiIndex` object from this pandas [documentation](https://pandas.pydata.org/pandas-docs/stable/advanced.html) .
And as I mentioned in the description of the PR , it is not handled yet.

yeganer: So when you attempt to do this, what happens? Do you get an error? in that case, you should, for now, test that this error is properly raised.
If the output is just wrong, then you could maybe test that too?

Marking a test as `xfail` just because it is failing is bad practice. Either test that it is failing properly (that means the correct error is thrown) or skip it.

A situation where an `xfail` makes sense would be testing `sqrt` for integers. You expect this to fail for integers below 0.

tardis/io/util.py
```diff
@@ -288,6 +289,27 @@ def from_hdf_util(cls, file_path, path=''):
...
+        if name is None:
+            name = cls.__name__
+
+        buff_path = os.path.join(path, name)
+        data = cls.from_hdf_util(file_path, buff_path)
+        for key, value in cls.class_properties.items():
+            data[key] = value.from_hdf(file_path, buff_path, name=key)
+
+        #Get initialization paramaters from constructor definition of class
+        argspec = inspect.getargspec(cls.__init__).args
```
yeganer: I like this line a lot. I previously was thinking about removing `hdf_properties` and using the arguments of `__init__`.

Well done!

tardis/io/util.py
```diff
@@ -288,6 +289,27 @@ def from_hdf_util(cls, file_path, path=''):
...
+        argspec.remove('self')
+
+        initializer_dict = {name: data[name] for name in argspec}
+
+        try:
+            return cls(**initializer_dict)
+        except:
+            return data
```
yeganer: I think `from_hdf` should always return an instance of class `cls`. If that is not possible (missing data), then we should raise an error detailing which values were missing.

tardis/io/util.py
```diff
@@ -288,6 +289,27 @@ def from_hdf_util(cls, file_path, path=''):
...
 
+    @classmethod
+    def from_hdf(cls, file_path, path='', name=None):
+
+        if name is None:
+            name = cls.__name__
+
+        buff_path = os.path.join(path, name)
+        data = cls.from_hdf_util(file_path, buff_path)
+        for key, value in cls.class_properties.items():
```
yeganer: key -> name, value -> c/klass

tardis/io/util.py
```diff
@@ -288,6 +289,27 @@ def from_hdf_util(cls, file_path, path=''):
                             hdf[key] = None
         return hdf
 
+    @classmethod
+    def from_hdf(cls, file_path, path='', name=None):
+
+        if name is None:
+            name = cls.__name__
```
yeganer: There is a handy shortcut for this 'construct':
`name = name or cls.__name__`

tardis/io/util.py
```diff
@@ -240,6 +238,9 @@ def to_hdf(self, file_path, path='', name=None):
         data = self.get_properties()
         buff_path = os.path.join(path, name)
         self.to_hdf_util(file_path, buff_path, data)
+        for property in self.class_properties.keys():
```
yeganer: I'm thinking, would it be possible to call the `to_hdf` method during `to_hdf_util`? All complex objects we want to save should implement the `to_hdf` method. So during the writing process, you should be able to save the object (use ducktyping there).

tardis/io/util.py
```diff
@@ -215,9 +215,12 @@ def to_hdf_util(path_or_buf, path, elements, complevel=9, complib='blosc'):
                     pd.DataFrame(value).to_hdf(
                         path_or_buf, os.path.join(path, key))
             else:
-                data = pd.DataFrame([value])
-                data.to_hdf(path_or_buf, os.path.join(path, key))
-
+                try:
+                    value.to_hdf(path_or_buf, path, name=key)
+                except:
```
yeganer: Please don't use a blank `except` but only catch the error you expect.

tardis/io/util.py
```diff
@@ -283,34 +283,33 @@ def from_hdf_util(cls, file_path, path=''):
...
                             hdf[key] = data[buff_path]
                     except:
-                        buff_path = os.path.join(path, 'scalars')
-                        hdf[key] = data[buff_path][key]
-                        if hdf[key] == 'none':
-                            hdf[key] = None
+                        try:
+                            c = cls.class_properties.get(key)
+                            hdf[key] = c.from_hdf(file_path, path, name=key)
+                        except:
```
yeganer: Same here, no blank `excepts`.

vg3095:
@wkerzendorf If there are no issues , then this can be merged now.

tardis/io/util.py
```diff
@@ -166,6 +166,184 @@ def check_equality(item1, item2):
...
+        path: str
+            Path inside the HDF store to store the `elements`
+        name: str
+            Group inside the HDF store to which the `elements` need to be saved
+
+        Returns
+        -------
+
+        """
+        name = name or self.__class__.__name__
```
wkerzendorf: so if `name` is `None` can you check for the `hdf_name` attribute (you can set on the class to None) and if that doesn't exist can you make a function that converts the `ClassName` to `class_name`

vg3095: @wkerzendorf Means whenever `name` is `None` , then a function converts `Class name` , like `HomologousDensity` to `homologous_density`. ?

vg3095: @wkerzendorf If I understand this correctly
- Add a `hdf_name` attribute under Class
- If it doesn`t exist convert `ClassName` to `class_name` (like `homologous_density`)

PR #743, opened 2017-06-14
Fix bug in plasma

yeganer:
While having a second look at the LevelNumberDensity module in plasma I found a bug.
Instead of defining the `calculate` function as a method, we assigned an attribute with a reference to a method during the `__init__`. That is a so called `instancemethod` and should only be used in special cases. In our case the `calculate` method is the same for all instances which should be reflected by the class.

If this fix hides implementation details, one could leave the original names and assign the function on the class level instead of in the `__init__`.

### DIFF
--- a/tardis/plasma/properties/level_population.py
+++ b/tardis/plasma/properties/level_population.py
@@ -8,11 +8,13 @@ logger = logging.getLogger(__name__)
 
 __all__ = ['LevelNumberDensity', 'LevelNumberDensityHeNLTE']
 
+
 class LevelNumberDensity(ProcessingPlasmaProperty):
     """
     Attributes:
     level_number_density : Pandas DataFrame, dtype float
-                           Index atom number, ion number, level number. Columns are zones.
+                           Index atom number, ion number, level number.
+                           Columns are zones.
     """
     outputs = ('level_number_density',)
     latex_name = ('N_{i,j,k}',)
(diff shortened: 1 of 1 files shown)


tardis/plasma/properties/level_population.py
```diff
@@ -35,82 +33,48 @@ def _initialize_indices(self, levels, partition_function):
...
         self._ion2level_idx = indexer.ix[levels.droplevel(2)].values
 
-    def _calculate_dilute_lte(self, level_boltzmann_factor, ion_number_density,
-        levels, partition_function):
+    def calculate(
+            self, level_boltzmann_factor, ion_number_density,
+            levels, partition_function):
         """
-        Reduces non-metastable level populations by a factor of W compared to LTE in the case of dilute-lte excitation.
+        Reduces non-metastable level populations by a factor of W
```
chvogl: This is not what this method does.

tardis/plasma/properties/level_population.py
```diff
@@ -8,25 +8,23 @@
...
     outputs = ('level_number_density',)
     latex_name = ('N_{i,j,k}',)
     latex_formula = ('N_{i,j}\\dfrac{bf_{i,j,k}}{Z_{i,j}}',)
 
-    def calculate(self):
-        pass
-
     def __init__(self, plasma_parent):
         """
         Calculates the level populations with the Boltzmann equation in LTE.
```
chvogl: I know that this is not particularly relevant for this PR but this class has nothing to do with either the Boltzmann equation or LTE.  It is used for all excitation treatments including dilute-lte and nlte.

PR #744, opened 2017-06-14
[TEP014] Added HDFWriter class + Unit Tests

vg3095:


vg3095:
@wkerzendorf . Please review

tardis/io/util.py
```diff
@@ -166,6 +167,106 @@ def check_equality(item1, item2):
...
+                if scalars_path in store:
+                    scalars_series = store[scalars_path].append(scalars_series)
+            scalars_series.to_hdf(path_or_buf, os.path.join(path, 'scalars'))
+
+    def get_properties(self):
+        data = {name: getattr(self, name) for name in self.hdf_properties}
+        return data
+    
+    @staticmethod
+    def convert_to_camel_case(s):
```
yeganer: As I understand it, what we are doing here is converting to snake_case. Please rename the function accordingly.

tardis/io/util.py
```diff
@@ -166,6 +167,106 @@ def check_equality(item1, item2):
...
+        s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', s)
```
yeganer: Can you please explain to me why we need two substitutions?
Additionally I don't understand why you added a `[a-z]+` at the end of the capture group. This prevents `AB` from being converted to `a_b`, which I think we want?

tardis/io/util.py
```diff
@@ -166,6 +167,106 @@ def check_equality(item1, item2):
...
+        return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1).lower()
```
yeganer: As mentioned above, I don't understand why we need this second substitution as the first seems to cover everything?

Maybe you can add some simple tests for some strings like `CamelCase, ABC, ABcdefG, snake_Case`?

tardis/io/util.py
vg3095: I used this regex from this thread->[Stackoverflow](https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case)

vg3095:
I have used regex to convert from `CamelCase` to `snake_case` from here --> [Stackoverflow](https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case)

tardis/io/tests/test_HDFWriter.py
```diff
@@ -123,3 +123,16 @@ def test_objects_write(tmpdir, attr):
         fname, key='/test/mock_class/nested_object/property')
     assert_array_almost_equal(
         actual.nested_object.property, nested_property)
+
+
+def test_snake_case():
+    obj = MockHDF(None)
```
yeganer: you don't need this line because `convert_to_snake_case` is a static method. You can simply write `MockHDF.convert_to_snake_case(...)`

tardis/io/tests/test_HDFWriter.py
```diff
@@ -0,0 +1,138 @@
...
+def test_complex_obj_write(tmpdir, attr):
+    fname = str(tmpdir.mkdir('data').join('test.hdf'))
+    actual = MockHDF(attr)
+    actual.to_hdf(fname, path='test')
+    expected = pd.read_hdf(fname, key='/test/mock_hdf/property').values
+    assert_array_almost_equal(actual.property, expected)
+
+arrays = [['L1', 'L1', 'L2', 'L2', 'L3', 'L3', 'L4', 'L4'],
+          ['one', 'two', 'one', 'two', 'one', 'two', 'one', 'two']]
+tuples = list(zip(*arrays))
```
yeganer: I think this is a very complicated way to create these arrays. I'd suggest you use `np.array(...)` and then simply transpose it.

vg3095:
@wkerzendorf This PR consists of both `HDFWriter` class and  its `unit tests`. Please review.

PR #747, opened 2017-06-20
[TEP014] Update Model and Density classes to use HDFWriter + Unit Tests 

vg3095:


vg3095:
@wkerzendorf Please review

tardis/model/base.py
```diff
@@ -6,13 +6,13 @@
...
 from tardis.io.model_reader import read_density_file, read_abundances_file
-from tardis.io.util import to_hdf
+from tardis.io.util import HDFWriter
 from density import HomologousDensity
 
 logger = logging.getLogger(__name__)
 
 
-class Radial1DModel(object):
+class Radial1DModel(HDFWriter, object):
```
yeganer: Remove `object` since it is not needed here.

tardis/model/tests/test_base.py
```diff
@@ -202,3 +204,48 @@ def test_ascii_reader_exponential_law():
...
         assert_almost_equal(model.density[i].value, mdens)
         assert model.density[i].unit ==  u.Unit(expected_unit)
+
+###
+# Save and Load
+###
+
+
+@pytest.fixture(scope="module")
+def hdf_file_path(tmpdir_factory):
```
yeganer: You should define a top level fixture in `conftest.py` that can be used by all test modules. That way you only have to write this code once and if it needs changes, you only have to change one instance of it.

tardis/model/tests/test_base.py
```diff
@@ -202,3 +204,48 @@ def test_ascii_reader_exponential_law():
...
+    path = tmpdir_factory.mktemp('hdf_buffer').join('model.hdf')
+    return str(path)
+
+
+@pytest.fixture(scope="module")
+def actual_model():
```
yeganer: Maybe we can have this as a top level fixture, too? See above.

tardis/model/tests/test_base.py
```diff
@@ -202,3 +204,48 @@ def test_ascii_reader_exponential_law():
...
+    filename = 'tardis_configv1_verysimple.yml'
+    config = Configuration.from_yaml(data_path(filename))
+    model = Radial1DModel.from_config(config)
+    return model
+
+
+@pytest.fixture(scope="module", autouse=True)
+def to_hdf_buffer(hdf_file_path, actual_model):
```
yeganer: This is an interesting way to achieve that the model is saved, however I'd prefer if there was a fixture that either returns the filename or the buffer the data was saved to.
Someone not very familiar with the internal workings of `pytest` might not understand what's happening here.

tardis/model/tests/test_density.py
```diff
@@ -0,0 +1,47 @@
...
+import pytest
+from numpy.testing import assert_almost_equal
+
+from tardis.io.config_reader import Configuration
+from tardis.model.density import HomologousDensity
+
+
+###
+# Save and Load
+###
```
yeganer: See remarks for `test_model.py`. Most apply here aswell

tardis/model/tests/test_base.py
vg3095: @yeganer If I return `hdf buffer path` from this function , then it same as input argument `hdf_file_path`. I think , If I return `hdf_file_path` here , then it is inconsistent .

yeganer:
@vg3095 I haven't had the time to look in depth at the latest changes. Can you please explain, why you had to revert the name changes? I'm confused...

vg3095:
@yeganer , If I change `HDFWriter` to `HDFWriterMixin`, then I have to change it in all places in `HDFWriter Unit Tests` , and I think ,it will out of scope for this PR.

For now , I am importing `HDFWriter` in classes like this -->
`from tardis.io.util import HDFWriter as HDFWriterMixin`
in `Radial1DModel` and `Homologous Density` classes.

I will then, make a seperate PR for this afterwards.

tardis/conftest.py
```diff
@@ -117,3 +119,29 @@ def included_he_atomic_data(test_data_path):
...
+    config = Configuration.from_yaml(path)
+    return config
+
+@pytest.fixture(scope="session")
+def model(hdf_config):
+    model = Radial1DModel.from_config(hdf_config)
+    return model
+
+@pytest.fixture(scope="session")
+def homologous_density(hdf_config):
```
yeganer: I would not say that homologous density is used anywhere else than the corresponding test

tardis/conftest.py
```diff
@@ -117,3 +119,29 @@ def included_he_atomic_data(test_data_path):
...
+# HDF Fixtures
+###
+
+@pytest.fixture(scope="session")
+def hdf_file_path(tmpdir_factory):
+    path = tmpdir_factory.mktemp('hdf_buffer').join('test.hdf')
+    return str(path)
+
+@pytest.fixture(scope="session")
+def hdf_config():
```
yeganer: I'd suggest a rename to `config_verysimple`

tardis/conftest.py
```diff
@@ -117,3 +119,29 @@ def included_he_atomic_data(test_data_path):
...
+    filename = 'tardis_configv1_verysimple.yml'
+    path = os.path.abspath(os.path.join('tardis/io/tests/data/', filename))
+    config = Configuration.from_yaml(path)
+    return config
+
+@pytest.fixture(scope="session")
+def model(hdf_config):
```
yeganer: Can you try to create a `simulation_verysimple` fixture? If that works, then we can get a model by `simulation_verysimple.model` and therefore we don't need additional fixtures.
Do you think, that could work?

yeganer:
@vg3095 You are right, the name change of `HDFWriter` should not be part of this PR but I think changing the name during the import is not the right way to go about this.

I'd suggest you make a small PR containing the name change and then you rebase this PR and other on your name change PR. That way you can implement it correctly right from the start but you can keep the change separate.
If you are not sure what I'm talking about in terms of `git`, have a look at `git rebase` and `git rebase -i`.

vg3095:
@yeganer Coverage is decreasing , due to usage of `common Simulation fixture` for all `hdf tests`, should I change it , or is it ok .

yeganer:
@vg3095 Looks good, however please remove the two commits which introduce the name change and then revert it in the next commit. You can delete, modify or merge commits with `git rebase -i`

tardis/conftest.py
```diff
@@ -117,3 +118,30 @@ def included_he_atomic_data(test_data_path):
...
+@pytest.fixture(scope="session")
+def config_verysimple():
+    filename = 'tardis_configv1_verysimple.yml'
+    path = os.path.abspath(os.path.join('tardis/io/tests/data/', filename))
+    config = Configuration.from_yaml(path)
+    return config
+
+@pytest.fixture(scope="session")
+def simulation_verysimple(config_verysimple, atomic_data_fname):
+    atomic_data = AtomData.from_hdf5(atomic_data_fname)
```
yeganer: I know there is a problem with the atomic data and preparing it multiple times. I think we could solve this by having a session fixture that does `AtomData.from_hdf(...)` and another function scope fixture that takes the first fixture and returns a `copy.deepcopy`. Could that work?

vg3095:
@yeganer I created a separate session based `atomic_dataset fixture` , like you said , and in `kurucz_atomic_data fixture`  , I used `deepcopy` of this session based fixture , but still the `integration tests` are failing.  Look at the last commit.

vg3095:
The issue is fixed now.

tardis/conftest.py
```diff
@@ -136,12 +139,7 @@ def config_verysimple():
...
 
 @pytest.fixture(scope="session")
-def simulation_verysimple(config_verysimple, atomic_data_fname):
-    atomic_data = AtomData.from_hdf5(atomic_data_fname)
-    if atomic_data.md5 != '21095dd25faa1683f4c90c911a00c3f8':
-        pytest.skip('Need default Kurucz atomic dataset '
-                    '(md5="21095dd25faa1683f4c90c911a00c3f8"')
-
-    sim = Simulation.from_config(config_verysimple, atom_data=atomic_data)
+def simulation_verysimple(config_verysimple, atomic_dataset):
```
yeganer: You are using the `session` scoped atomic dataset fixture here. I believe you want to use `kurucz_atomic_data` which is the one making use of the `deepcopy`, right?

vg3095: @yeganer `kurucz_atomic_data` fixture is a `function scope` fixture , that\`s why It can\`t be used for `session based fixture` (It throws `ScopeMismatch Error`) , that\`s why , I used `atomic_dataset` here(`session based fixture`) and used `deepcopy`  to create a new copy for atomic dataset inside the function itself..

vg3095: @yeganer And you are seeing the old commit

wkerzendorf:
@vg3095 can you rebase onto the current master

wkerzendorf:
@yeganer  and @wkerzendorf signed off - will merge when tests pass

PR #748, opened 2017-06-20
[TEP014] Update Runner and Spectrum classes to use HDFWriter + Unit Tests

vg3095:


vg3095:
@wkerzendorf Please review

vg3095:
Don`t review it yet. Will update after PR #747 is merged.

vg3095:
@wkerzendorf Please review.

tardis/montecarlo/tests/test_spectrum.py
```diff
@@ -166,44 +166,26 @@ def compare_spectra(actual, desired):
...
+@pytest.fixture(scope="module")
+def hdf_file_path(tmpdir_factory):
+    path = tmpdir_factory.mktemp('hdf_buffer').join('spectrum.hdf')
+    return str(path)
+
+@pytest.fixture(autouse=True)
+def to_hdf_buffer(hdf_file_path,spectrum):
+    spectrum.to_hdf(hdf_file_path, name='spectrum')
+
+spectrum_attrs = ['_frequency', 'luminosity']
```
yeganer: I think we should use `TARDISSpectrum.hdf_properties` here. That way we don't have to update the tests if we change the `hdf_properties`.

yeganer:
Overall this looks very good.

However you have old commits in there which use `HDFWriter` instead of the mixin. While this is not a major problem due to all changes being merged together it might be a problem if we ever have to use `git bisect` again. So I'd suggest to `git rebase -i upstream/master` and edit the commits in question so that at least running the tardis commandline script is possible without errors.

PR #749, opened 2017-06-21
[TEP014] Change name of HDFWriter to HDFWriterMixin

vg3095:
As yeganer said, in PR #747 , we should change the name of `HDFWriter` to `HDFWriterMixin` as , this would then correctly reflect that `HDFWriter` is only a utility class and not the common Root for all our classes


vg3095:
Coverage fall is  due to removal of `class_properties` under Mock Unit tests for `HDFWriterMixin` . They are actually not required now , as we are not focusing on `from_hdf` implementation.

PR #750, opened 2017-06-21
[TEP005][QOL] Add MontecarloRunner.spectrum_integrated

yeganer:
This is a shortcut for obtaining an integrated spectrum for the
wavelength grid defined in the configuration file. If incompatible
configuration settings are detected, an IntegrationError is raised.

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -127,8 +127,7 @@ class MontecarloRunner(object):
     @property
     def spectrum_integrated(self):
         return self.integrator.calculate_spectrum(
-                self.spectrum_frequency[:-1],
-                1000)
+                self.spectrum_frequency[:-1])
 
     @property
     def integrator(self):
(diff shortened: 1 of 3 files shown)


tardis/montecarlo/base.py
```diff
@@ -125,6 +125,12 @@ def spectrum_virtual(self):
                 self.montecarlo_virtual_luminosity)
 
     @property
+    def spectrum_integrated(self):
+        return self.integrator.calculate_spectrum(
+                self.spectrum_frequency[:-1],
+                1000)
```
unoebauer: The hardcoded 1000 is the resolution of the p-grid, right? If so, I think we should make this also a configuration variable.

yeganer: What name would you suggest for such a configuration option?

wkerzendorf:
@unoebauer I agree that it would be nice to have a config option for the p-values, so I would say that we put this as an item in the project todo - but that we merge this PR.

tardis/montecarlo/base.py
unoebauer: something like ``impact_parameter_bins``

PR #752, opened 2017-06-22
[TEP014] PlasmaWriterMixin

vg3095:


wkerzendorf:
first of all @vg3095 rebase

tardis/io/util.py
```diff
@@ -265,7 +269,51 @@ def to_hdf(self, file_path, path='', name=None):
...
+                data = {name.__class__.__name__: name for name in self.plasma_properties
+                        if isinstance(name, tuple(collection))}
+            else:
+                data = {name.__class__.__name__: name for name in self.plasma_properties}
+                data['atom_data_uuid'] = self.atomic_data.uuid1
+            if 'atomic_data' in data:
+                data.pop('atomic_data')
+        return data
+
+    def to_hdf(self, file_path, path='plasma', collection=None):
```
wkerzendorf: this should be the same as the "normal" `to_hdf`

wkerzendorf: in this case you should also have class properties by using the `property` decorator on the outputs for example.

vg3095: @wkerzendorf I am facing a small problem here , I don`t want `name` parameter here for `Plasma hdf calls` , 
With `name` parameter it will save like ex-> `/plasma/abundance/abundance`
without name parameter `/plasma/abundance/`

vg3095:
@wkerzendorf Please review.

yeganer:
@wkerzendorf @vg3095 I think this PR need some rework before we can merge it. Specifically I don't like mixing functionality of the PlasmaHDFWriter into the normal HDFWriter (the changes to the `to_hdf` method by adding a 2nd kw argument). I'd prefer to either solve this by adding a collection property to the PlasmaHDFWriter or by overwriting the `to_hdf` method (and then calling the original with `super`)

vg3095:
@wkerzendorf @yeganer Now, I have updated this.

wkerzendorf:
@tardis-sn/tardis-core can someone else look over this and approve this for merging?

PR #757, opened 2017-07-04
[TEP007] Decay and merge isotopic abundance dataframe

vg3095:
I have added a `decay` method under `Model` class , which will take `time_explosion` as parameter , and then decay `isotopic abundance frame` , and returns `merged dataframe` .
See this notebook for detailed output  
[Jupyter Notebook](https://github.com/vg3095/decay/blob/master/isotope_decay_exp.ipynb)

### DIFF
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -87,3 +87,33 @@ class IsotopeAbundances(pd.DataFrame):
         df = IsotopeAbundances.from_materials(decayed_materials)
         df.sort_index(inplace=True)
         return df 
+
+    def as_atomic_numbers(self, abundance, t, normalize=True):
+        """
+        Merge decayed Isotope dataframe with abundance passed as parameter 
+
+        Parameters
+        ----------
+
+        t: float or astropy.units.Quantity
+            If float it will be understood as days
+        abundance: pd.DataFrame 
+        normalize : bool
+            If true, resultant dataframe will be normalized along the coloumn
+
+        Returns:
+            : merged abundances
+        """
+        #Drop mass_number coloumn in isotopic_abundance dataframe
+        isotope_abundance = self.decay(t).reset_index(
+            level='mass_number').drop('mass_number', axis=1)
+
+        #Merge abundance dataframes
+        modified_df = pd.concat([isotope_abundance, abundance])
+        modified_df = modified_df.groupby('atomic_number').sum()
+
+        if normalize:
+            norm_factor = modified_df.sum(axis=0)
+            modified_df /= norm_factor
+
+        return modified_df
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -111,7 +111,7 @@ class IsotopeAbundances(pd.DataFrame):
         Returns:
             : merged abundances
         """
-        isotope_abundance = self.merge_isotopes()
+        isotope_abundance = self.as_atoms()
         #Merge abundance and isotope dataframe
         modified_df = isotope_abundance.add(other, fill_value=0)
 
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -33,7 +33,7 @@ class IsotopeAbundances(pd.DataFrame):
             for key, value in material.items():
                 abundances.loc[cls.id_to_tuple(key), i] = value
 
-        return abundances
+        return cls(abundances)
 
 
 
@@ -88,15 +88,22 @@ class IsotopeAbundances(pd.DataFrame):
         df.sort_index(inplace=True)
         return df 
 
-    def as_atomic_numbers(self, abundance, t, normalize=True):
+    def merge_isotopes(self):
         """
-        Merge decayed Isotope dataframe with abundance passed as parameter 
+        Merge Isotope dataframe according to atomic number 
+
+        Returns:
+            : merged isotope abundances
+        """
+
+        return self.groupby('atomic_number').sum()
+
+    def as_atomic_numbers(self, abundance, normalize=True):
+        """
+        Merge Isotope dataframe with abundance passed as parameter 
 
         Parameters
         ----------
-
-        t: float or astropy.units.Quantity
-            If float it will be understood as days
         abundance: pd.DataFrame 
         normalize : bool
             If true, resultant dataframe will be normalized
@@ -104,11 +111,8 @@ class IsotopeAbundances(pd.DataFrame):
         Returns:
             : merged abundances
         """
-        #Drop mass_number coloumn in isotopic_abundance dataframe
-        isotope_abundance = self.decay(t).reset_index(
-            level='mass_number').drop('mass_number', axis=1)
-
-        #Merge abundance dataframes
+        isotope_abundance = self.merge_isotopes()
+        #Merge abundance and isotope abundance
         modified_df = isotope_abundance.add(abundance, fill_value=0)
 
         if normalize:
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -112,7 +112,7 @@ class IsotopeAbundances(pd.DataFrame):
             : merged abundances
         """
         isotope_abundance = self.merge_isotopes()
-        #Merge abundance and isotope abundance
+        #Merge abundance and isotope dataframe
         modified_df = isotope_abundance.add(abundance, fill_value=0)
 
         if normalize:
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -88,7 +88,7 @@ class IsotopeAbundances(pd.DataFrame):
         df.sort_index(inplace=True)
         return df 
 
-    def merge_isotopes(self):
+    def as_atoms(self):
         """
         Merge Isotope dataframe according to atomic number 
 
@@ -98,13 +98,13 @@ class IsotopeAbundances(pd.DataFrame):
 
         return self.groupby('atomic_number').sum()
 
-    def as_atomic_numbers(self, abundance, normalize=True):
+    def merge(self, other, normalize=True):
         """
         Merge Isotope dataframe with abundance passed as parameter 
 
         Parameters
         ----------
-        abundance: pd.DataFrame 
+        other: pd.DataFrame 
         normalize : bool
             If true, resultant dataframe will be normalized
 
@@ -113,7 +113,7 @@ class IsotopeAbundances(pd.DataFrame):
         """
         isotope_abundance = self.merge_isotopes()
         #Merge abundance and isotope dataframe
-        modified_df = isotope_abundance.add(abundance, fill_value=0)
+        modified_df = isotope_abundance.add(other, fill_value=0)
 
         if normalize:
             norm_factor = modified_df.sum(axis=0)
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -99,7 +99,7 @@ class IsotopeAbundances(pd.DataFrame):
             If float it will be understood as days
         abundance: pd.DataFrame 
         normalize : bool
-            If true, resultant dataframe will be normalized along the coloumn
+            If true, resultant dataframe will be normalized
 
         Returns:
             : merged abundances
@@ -109,8 +109,7 @@ class IsotopeAbundances(pd.DataFrame):
             level='mass_number').drop('mass_number', axis=1)
 
         #Merge abundance dataframes
-        modified_df = pd.concat([isotope_abundance, abundance])
-        modified_df = modified_df.groupby('atomic_number').sum()
+        modified_df = isotope_abundance.add(abundance, fill_value=0)
 
         if normalize:
             norm_factor = modified_df.sum(axis=0)
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -190,7 +190,7 @@ class Radial1DModel(HDFWriterMixin):
             stop -= 2
         if not self.raw_isotope_abundance.empty:
             self._abundance = self.raw_isotope_abundance.decay(
-                self.time_explosion).as_atomic_numbers(self.raw_abundance)
+                self.time_explosion).merge(self.raw_abundance)
         abundance = self._abundance.ix[:, start:stop]
         abundance.columns = range(len(abundance.columns))
         return abundance
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -72,6 +72,7 @@ class Radial1DModel(HDFWriterMixin):
         self.homologous_density = homologous_density
         self._abundance = abundance
         self.isotope_abundance = isotope_abundance
+        self.decayed = False
         self.time_explosion = time_explosion
 
         self.raw_abundance = self._abundance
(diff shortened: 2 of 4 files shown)


tardis/model/base.py
```diff
@@ -189,6 +191,33 @@ def abundance(self):
         abundance.columns = range(len(abundance.columns))
         return abundance
 
+    def decay(self, day, normalize=True):
```
yeganer: I think this should happen during the initialization of the model.

The behavior should be:
If there is `isotope_abundance` as parameter, decay these with `time_explosion` and merge them with the normal abundances. If we do this, save the normal abundances and the undecayed abundances somewhere on the model (example `raw_abundance`, `raw_isotope_abundance`)

tardis/model/base.py
```diff
@@ -189,6 +196,33 @@ def abundance(self):
         abundance.columns = range(len(abundance.columns))
         return abundance
 
+    def decay(self, time_explosion, normalize=True):
```
yeganer: I don't think we want a decay function at all. Maybe we need a helper function at some point but in general we don't need this.
We don't need functionality to 'redecay' a model because we will always use `time_explosion` as the time of the decay.

tardis/model/base.py
```diff
@@ -72,7 +72,14 @@ def __init__(self, velocity, homologous_density, abundance, time_explosion,
         self.v_boundary_outer = v_boundary_outer
         self.homologous_density = homologous_density
         self._abundance = abundance
+        self.decayed = False
```
yeganer: This is not needed.

tardis/model/base.py
```diff
@@ -189,6 +196,33 @@ def abundance(self):
...
+
+        if not self.decayed:
+            self.decayed = True
+        else:
+            raise ValueError("Isotopic Abundance already decayed")
+
+        isotope_abundance = self.raw_isotope_abundance.decay(time_explosion)
+
+        #Set atomic_number as index in isotopic_abundance dataframe
+        isotope_abundance = isotope_abundance.reset_index().drop(
```
wkerzendorf: why the many `reset_index` - not sure that you need those.

tardis/model/base.py
```diff
@@ -189,6 +196,33 @@ def abundance(self):
...
+            'mass_number', axis=1).set_index(['atomic_number'])
+
+        #Merge abundance dataframes
+        modified_df = isotope_abundance.append(self.abundance)
+        modified_df = modified_df.reset_index().set_index('atomic_number')
```
wkerzendorf: same here. you do not need to make something an index to have `groupby` work in it.

tardis/model/base.py
```diff
@@ -189,6 +196,33 @@ def abundance(self):
...
+            raise ValueError("Isotopic Abundance already decayed")
+
+        isotope_abundance = self.raw_isotope_abundance.decay(time_explosion)
+
+        #Set atomic_number as index in isotopic_abundance dataframe
+        isotope_abundance = isotope_abundance.reset_index().drop(
+            'mass_number', axis=1).set_index(['atomic_number'])
+
+        #Merge abundance dataframes
+        modified_df = isotope_abundance.append(self.abundance)
```
wkerzendorf: you definitely do not want to append - but add the decayed dataframe to the abundances.

tardis/model/base.py
```diff
@@ -72,7 +72,14 @@ def __init__(self, velocity, homologous_density, abundance, time_explosion,
         self.v_boundary_outer = v_boundary_outer
         self.homologous_density = homologous_density
         self._abundance = abundance
+        self.decayed = False
         self.time_explosion = time_explosion
+
+        self.raw_abundance = self._abundance
```
yeganer: We should discuss and define at some point the names of all `abundance` variables involved.
There is input: `abundance` and `isotope_abundance` (should be attached to model for diagnosis and debugging purposes)
There is output: `Model.abundance` which should only contain the region defined by the velocity boundaries (as it currently does) AND which I think should contain the merged result if `isotope_abundance` is passed as an optional argument.
Then we have `self._abundance` which looks like it's currently the reference to the input (for diagnosis)

So what we basically should decide is, how do we cache the result of the decay (do we even cache it?) and how do we attach the input to the model.

yeganer:
I'd suggest to add a `as_atomic_numbers` or similar to the `IsotopeAbundance` class which returns the dataframe converted to the format `Model.abundances` currently is in.

vg3095:
As of now, these are the variables associated with `abundance`
`raw_abundance` : Normal abundance as passed during initialization
`raw_isotope_abundance` : Isotope abundance as passed during initialization (Undecayed)
`isotope_abundance` : Decayed Isotope abundance (using time_explosion)
`_abundance` : If  `raw_isotope_abundance` is present , then it has combined dataframe , else
normal abundance dataframe.
`abundance` : Contains the region defined by the velocity boundaries. As `_abundance` is updated(when `isotope_abundance` is present), changes will reflect here also, when dataframes are merged.

vg3095:
Travis fail is due to time out in Mac build .

tardis/model/base.py
vg3095: @wkerzendorf I have now used `pd.concat()` in the latest commit . 
- If you mean `pd.add()` , then according to what I could find , `index values` for both `Dataframe` must be equal, then only result is correct, otherwise, values not present in one dataframe , are filled with `NaN` values
- `pd.merge()` and `pd.join()` operation results in addition along columns.
So, I think , it leaves me with `pd.concat()` , and I have also added a test to ensure if it works correctly.

yeganer:
I don't think we need `isotope_abundance`, right? Because it can be created from `raw_isotope_abundance` at all times

tardis/model/base.py
```diff
@@ -189,6 +199,25 @@ def abundance(self):
         abundance.columns = range(len(abundance.columns))
         return abundance
 
+    def as_atomic_numbers(self, normalize=True):
```
yeganer: This method should be part of `IsotopeAbundance`

tardis/io/decay.py
```diff
@@ -86,4 +86,34 @@ def decay(self, t):
...
+
+        Returns:
+            : merged abundances
+        """
+        #Drop mass_number coloumn in isotopic_abundance dataframe
+        isotope_abundance = self.decay(t).reset_index(
+            level='mass_number').drop('mass_number', axis=1)
+
+        #Merge abundance dataframes
+        modified_df = pd.concat([isotope_abundance, abundance])
```
wkerzendorf: I want you to add them here

tardis/model/base.py
```diff
@@ -63,7 +63,7 @@ class Radial1DModel(HDFWriterMixin):
     def __init__(self, velocity, homologous_density, abundance, time_explosion,
                  t_inner, luminosity_requested=None, t_radiative=None,
                  dilution_factor=None, v_boundary_inner=None,
-                 v_boundary_outer=None):
+                 v_boundary_outer=None, isotope_abundance=None):
```
wkerzendorf: move this up

tardis/model/base.py
```diff
@@ -73,6 +73,13 @@ def __init__(self, velocity, homologous_density, abundance, time_explosion,
         self.homologous_density = homologous_density
         self._abundance = abundance
         self.time_explosion = time_explosion
+
+        self.raw_abundance = self._abundance
+        self.raw_isotope_abundance = isotope_abundance
```
wkerzendorf: if it is None - you should make an empty frame

tardis/io/decay.py
vg3095: @wkerzendorf  For `pd.add()` , `index values` for both `Dataframe` must be equal, then only result is correct, otherwise, index values not present in one dataframe , are filled with `NaN` values

tardis/model/base.py
wkerzendorf: @vg3095 I think the decay is extremely quick - maybe just a property would be fine. Actually can you profile how quick?

vg3095:
@wkerzendorf I have changed it . Travis fail is due to timeout in Mac build. I think you will have to re-trigger Mac build, after some time.

tardis/model/base.py
vg3095: @wkerzendorf If by `profile` , meaning time taken by decay function here
I made a dummy dataframe, then used decay on it.
It takes around 0.005 seconds for 10 elements and around 0.01 seconds for 100 elements

vg3095: @wkerzendorf I have now used `cProfile` module, If you want to know , no. of function calls and time taken for decay method.
  
- For 10 elements
6405 function calls (6351 primitive calls) in 0.007 seconds

- For 20 elements
7635 function calls (7581 primitive calls) in 0.008 seconds

- For 50 elements
11325 function calls (11271 primitive calls) in 0.011 seconds

- For 100 elements
17475 function calls (17421 primitive calls) in 0.016 seconds

vg3095:
@wkerzendorf 
Is there anything remaining to change in this PR ?

tardis/io/decay.py
```diff
@@ -86,4 +86,33 @@ def decay(self, t):
         
         df = IsotopeAbundances.from_materials(decayed_materials)
         df.sort_index(inplace=True)
-        return df 
\ No newline at end of file
+        return df 
+
+    def as_atomic_numbers(self, abundance, t, normalize=True):
```
wkerzendorf: thi should be called - `merge_isotopes` and should only perform the groupby operation.

tardis/io/decay.py
```diff
@@ -86,4 +86,33 @@ def decay(self, t):
...
+            If float it will be understood as days
+        abundance: pd.DataFrame 
+        normalize : bool
+            If true, resultant dataframe will be normalized
+
+        Returns:
+            : merged abundances
+        """
+        #Drop mass_number coloumn in isotopic_abundance dataframe
+        isotope_abundance = self.decay(t).reset_index(
```
wkerzendorf: use groupby to sum up over all mass numbers.

tardis/io/decay.py
```diff
@@ -86,4 +86,33 @@ def decay(self, t):
...
+            level='mass_number').drop('mass_number', axis=1)
+
```
wkerzendorf: then return here. The rest should happen somewhere else.

tardis/model/base.py
```diff
@@ -73,6 +73,15 @@ def __init__(self, velocity, homologous_density, abundance, time_explosion,
         self.homologous_density = homologous_density
         self._abundance = abundance
         self.time_explosion = time_explosion
+
+        self.raw_abundance = self._abundance
+        if isotope_abundance is not None:
```
wkerzendorf: if it is None make an empty one.

tardis/model/base.py
```diff
@@ -348,11 +357,16 @@ def from_config(cls, config):
             logger.warning("Abundances have not been normalized to 1."
                            " - normalizing")
             abundance /= norm_factor
-
+        
+        if hasattr(config.model, 'isotope_abundance'):
+            isotope_abundance = config.model.isotope_abundance
```
wkerzendorf: no that's wrong - we have not decided on a config option and that should also come later!

tardis/model/base.py
vg3095: @wkerzendorf I have made one , see line 82-84

tardis/io/decay.py
vg3095: @wkerzendorf `the rest` means normalizing and returning as `model abundance dataframe` format should be moved to other separate function (`as_atomic_numbers`) and only groupby here(`merge_isotopes`)

tardis/model/base.py
yeganer: You can use `self.raw_isotope_abundance = isotope_abundance or pd.DataFrame()`

tardis/model/base.py
```diff
@@ -73,6 +73,15 @@ def __init__(self, velocity, homologous_density, abundance, time_explosion,
         self.homologous_density = homologous_density
         self._abundance = abundance
         self.time_explosion = time_explosion
+
+        self.raw_abundance = self._abundance
+        if isotope_abundance is not None:
+            self.raw_isotope_abundance = isotope_abundance
+            self._abundance = self.raw_isotope_abundance.as_atomic_numbers(
```
yeganer: I would add this line to the `abundance` property.

vg3095: @yeganer If I put this line in under `abundance` property, then this line would be executed _every time_ `model.abundance` is called , and I think it should only be called 1 time during initialization.

yeganer: As your profiling showed, executing this for 100 elements takes less than 0.02 seconds. In my opinion it's okay to call it every time. especially as it should always reflect the current state of the model.

Currently tardis objects are not designed to be changed, nevertheless if one were to change `time_explosion` on the model, the isotopic abundances have to be recalculated. The easiest way to achieve this is by doing this (very fast) computation every time. I think you can count the number of accesses to `model.abundance` during a typical tardis run with one hand.

tardis/io/decay.py
```diff
@@ -86,4 +86,37 @@ def decay(self, t):
...
+        """
+        Merge Isotope dataframe according to atomic number 
+
+        Returns:
+            : merged isotope abundances
+        """
+
+        return self.groupby('atomic_number').sum()
+
+    def as_atomic_numbers(self, abundance, normalize=True):
```
yeganer: I'd call this method something like `merge_with` or only `merge` as that reflects what this function does, merge Isotopes with a normal DataFrame.
Maybe we can call the other argument `other` so we can easily distinguish between `self` and `other`.

tardis/io/decay.py
```diff
@@ -86,4 +86,37 @@ def decay(self, t):
         
         df = IsotopeAbundances.from_materials(decayed_materials)
         df.sort_index(inplace=True)
-        return df 
\ No newline at end of file
+        return df 
+
+    def merge_isotopes(self):
```
yeganer: I'm not happy with the name of this function, maybe we should call it `as_atoms`? That's short and reflects that the output is only atoms instead of isotopes.

vg3095:
@wkerzendorf  I think , in last meeting, I could not convey my message properly.
- As of now `as_atoms` (Renamed from `merge_isotopes` , as per suggestion by @yeganer )-> Does _only groupby to sum up over all mass numbers.
- `merge` (Renamed from `as_atomic_numbers`)-> takes 2 data-frame , then adds and normalizes it .

I think, you confused with the names , as almost the names are reversed , and thought `merge` to be `merge_isotopes`. 
If you want me to revert the names back,
(Make `as_atoms` to `merge_isotopes` and `merge` to `as_atomic_numbers`)  , then a confirmation would be nice.

wkerzendorf:
@vg3095 this is outdated right?

vg3095:
@wkerzendorf Yes, this is outdated.

wkerzendorf:
@vg3095 are you sure that is not in the codebase - as we merged the uniform case?

vg3095:
@wkerzendorf I have rebased it now. You can review it.

unoebauer:
@wkerzendorf, @vg3095 - so, do we still need to merge this or not? I am confused since I thought that all the decay stuff has already been merged.

vg3095:
@wkerzendorf @unoebauer Yes, it needs to be merged. This is the only PR related to decay stuff. Rest all were related to config options for isotope.

PR #740, opened 2017-06-12
[GSoC] Formal integral Project

yeganer:
Continuation of #726

This PR contains all changes related to the formal integral project. This is an aggregation PR. That means future changes will be made as PR to the feature branch `formal_integral`. Once that branch is feature complete, it will be merged into master.

Milestones:

- [x] Basic Implementation
- [x] Dicuss OMP
- [x] Config Integration
- [x] Documentation (See #621 and #745)
- [ ] Verification (Will be done in a future PR)

PR #759, opened 2017-07-04
Formal Integral: including electron scattering

unoebauer:
The goal of this PR is the inclusion of the electron scattering into the Formal Integral scheme, introduced by @yeganer in PR #740.

For this purpose, the mean intensity in the blue and red wings of each line transition have to be calculated (see Eqs. 23, 24 and 28 in Lucy 1999) and passed to the Formal Integral integrator. The most obvious place for this calculation is alongside the determination of the attenuated line source functions ``att_S_ul`` in the routine ``make_source_function`` in ``formal_integral.py``.

The intensities have to be passed to the formal integrator (alongside ``att_S_ul``). In the integrator, the recursive stepping through the line list has to be adjusted to include the electron scattering contribution (see Eqs. 26, 27 and 28 in Lucy 1999). For this purpose, the electron scattering optical depth accumulated along the constant p paths has to be calculated.

Finally, unit tests for these additions have to be devised, the sanity checks revised and the integration with the configuration system checked.

**Note:** Even after this change, the Formal Integral will only work with the downbranching scheme for now.

**Milestones:**
- [x] calculate Jblues_lu and Jreds_lu
- [x] calculate electron scattering optical depths on integration segments
- [x] adjust recursive integration along z for const-p rays
- [ ] properly account for the edges of the line list
- [x] properly treat e-scattering effect across shell interfaces (not only record dtau but dtau * Jkkp)
- [ ] add unit tests
- [x] verification with single and multiline setups
- [x] update documentation

### DIFF
--- a/tardis/montecarlo/formal_integral.py
+++ b/tardis/montecarlo/formal_integral.py
@@ -128,7 +128,7 @@ class FormalIntegrator(object):
         att_S_ul =  ( wave * (q_ul * e_dot_u) * t  / (4*np.pi) )
 
         result = pd.DataFrame(att_S_ul.as_matrix(), index=transitions.transition_line_id.values)
-        att_S_ul = result.ix[atomic_data.lines.index.values].as_matrix()
+        att_S_ul = result.ix[atomic_data.lines.index.values].values
 
         # Jredlu should already by in the correct order, i.e. by wavelength of
         # the transition l->u (similar to Jbluelu)
--- a/tardis/montecarlo/formal_integral.py
+++ b/tardis/montecarlo/formal_integral.py
@@ -116,7 +116,7 @@ class FormalIntegrator(object):
         # functions
         Jbluelu_norm_factor = (const.c.cgs * model.time_explosion /
                                 (4 * np.pi * runner.time_of_simulation *
-                                 model.volume)).to("1/cm^2/s").value
+                                 model.volume)).to("1/(cm^2 s)").value
         # Jbluelu should already by in the correct order, i.e. by wavelength of
         # the transition l->u
         Jbluelu = runner.j_blue_estimator * Jbluelu_norm_factor
@@ -134,4 +134,9 @@ class FormalIntegrator(object):
         att_S_ul =  ( wave * (q_ul * e_dot_u) * t  / (4*np.pi) )
 
         result = pd.DataFrame(att_S_ul.as_matrix(), index=transitions.transition_line_id.values)
-        return result.ix[atomic_data.lines.index.values].as_matrix(), Jbluelu, e_dot_u
+        att_S_ul = result.ix[atomic_data.lines.index.values].as_matrix()
+
+        # Jredlu should already by in the correct order, i.e. by wavelength of
+        # the transition l->u (similar to Jbluelu)
+        Jredlu = Jbluelu * np.exp(-plasma.tau_sobolevs.values) + att_S_ul
+        return att_S_ul, Jredlu, Jbluelu, e_dot_u
--- a/tardis/montecarlo/formal_integral.py
+++ b/tardis/montecarlo/formal_integral.py
@@ -50,12 +50,6 @@ class FormalIntegrator(object):
                     'line_interaction_type == "downbranch"'
                     )
 
-        if self.runner.sigma_thomson.value > 1e-100:
-            return raise_or_return(
-                    'The FormalIntegrator currently only works with '
-                    'disable_electron_scattering turned on.'
-                    )
-
         return True
 
     def calculate_spectrum(self, frequency, points=None, raises=True):
--- a/tardis/montecarlo/formal_integral.py
+++ b/tardis/montecarlo/formal_integral.py
@@ -2,6 +2,7 @@ import warnings
 import numpy as np
 import pandas as pd
 from astropy import units as u
+from astropy import constants as const
 
 from tardis.montecarlo.montecarlo import formal_integral
 from tardis.montecarlo.spectrum import TARDISSpectrum
@@ -111,6 +112,15 @@ class FormalIntegrator(object):
         exptau = 1 - np.exp(- plasma.tau_sobolevs)
         Edotlu = Edotlu_norm_factor * exptau * runner.Edotlu_estimator
 
+        # The following may be achieved by calling the appropriate plasma
+        # functions
+        Jbluelu_norm_factor = (const.c.cgs * model.time_explosion /
+                                (4 * np.pi * runner.time_of_simulation *
+                                 model.volume)).to("1/cm^2/s").value
+        # Jbluelu should already by in the correct order, i.e. by wavelength of
+        # the transition l->u
+        Jbluelu = runner.j_blue_estimator * Jbluelu_norm_factor
+
         upper_level_index = atomic_data.lines.set_index(['atomic_number', 'ion_number', 'level_number_upper']).index.copy()
         e_dot_lu          = pd.DataFrame(Edotlu, index=upper_level_index)
         e_dot_u           = e_dot_lu.groupby(level=[0, 1, 2]).sum()
@@ -124,4 +134,4 @@ class FormalIntegrator(object):
         att_S_ul =  ( wave * (q_ul * e_dot_u) * t  / (4*np.pi) )
 
         result = pd.DataFrame(att_S_ul.as_matrix(), index=transitions.transition_line_id.values)
-        return result.ix[atomic_data.lines.index.values].as_matrix(), e_dot_u
+        return result.ix[atomic_data.lines.index.values].as_matrix(), Jbluelu, e_dot_u
--- a/tardis/montecarlo/src/integrator.c
+++ b/tardis/montecarlo/src/integrator.c
@@ -270,10 +270,12 @@ _formal_integral(
                        ++patt_S_ul)
                     {
                       if (*pline < nu_end)
+                      {
                         // Calculate e-scattering optical depth to grid cell boundary
                         zend = storage->time_explosion / C_INV * (1. - nu_end / nu);
                         dtau += (zend - zstart) * escat_op;
                         break;
+                      }
 
                       // Calculate e-scattering optical depth to next resonance point
                       zend = storage->time_explosion / C_INV * (1. - *pline / nu);
--- a/tardis/montecarlo/src/integrator.c
+++ b/tardis/montecarlo/src/integrator.c
@@ -184,7 +184,9 @@ _formal_integral(
       // Initializing all the thread-local variables
       int64_t offset = 0, i = 0,
               size_z = 0,
-              idx_nu_start = 0;
+              idx_nu_start = 0,
+              direction = 0,
+              first = 0;
 
       double I_nu_b[N],
              I_nu_r[N],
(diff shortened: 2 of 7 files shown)


unoebauer:
Somehow, I broke the integration with the last commit. Any ideas why, @yeganer?

tardis/montecarlo/src/integrator.c
```diff
@@ -258,11 +270,23 @@ _formal_integral(
                        ++patt_S_ul)
                     {
                       if (*pline < nu_end)
+                        // Calculate e-scattering optical depth to grid cell boundary
+                        zend = storage->time_explosion / C_INV * (1. - nu_end / nu);
+                        dtau += (zend - zstart) * escat_op;
                         break;
```
yeganer: This break get's called *every* loop because the if statement is not followed by a scope `{}`

tardis/montecarlo/src/integrator.c
```diff
@@ -258,11 +270,23 @@ _formal_integral(
...
+                      zend = storage->time_explosion / C_INV * (1. - *pline / nu);
+                      dtau += (zend - zstart) * escat_op;
+
                       I_nu_r[p_idx] = I_nu_b[p_idx] * (*pexp_tau) + *patt_S_ul;
                       // In the absence of electron scattering, simple recursion as in Lucy 1991
                       // TODO: e-scattering: Replace by Lucy 1999, Eq. 27
                       I_nu_b[p_idx] = I_nu_r[p_idx];
+
+                      // reset e-scattering opacity 
+                      dtau = 0;
```
yeganer: I'm not sure why you reset `dtau` every iteration.

tardis/montecarlo/src/integrator.c
```diff
@@ -258,11 +270,23 @@ _formal_integral(
                        ++patt_S_ul)
                     {
                       if (*pline < nu_end)
```
yeganer: This is a second loop end condition because it's not easy to do that nicely

tardis/montecarlo/src/integrator.c
```diff
@@ -246,10 +255,13 @@ _formal_integral(
...
 
+
                   // Initialize pointers for inner loop
                   pline = storage->line_list_nu + idx_nu_start;
                   pexp_tau = exp_tau + offset + idx_nu_start;
                   patt_S_ul = att_S_ul + offset + idx_nu_start;
+                  pJred_lu = Jred_lu + offset + idx_nu_start;
+                  pJblue_lu = Jblue_lu + offset + idx_nu_start;
 
                   for (;pline < storage->line_list_nu + size_line;
```
yeganer: maybe I can change `pline < storage->line_list_nu + size_line` to 
`(pline < storage->line_list_nu + size_line) & (*pline < nu_end)`

Alternatively one could add `idx_nu_end` and then change the loop condition to `pline < storage->line_list_nu + idx_nu_end`.

tardis/montecarlo/src/integrator.c
unoebauer: Because dtau measures the accumulated e-scattering optical depth between consecutive resonances. I still have to modify the calculation of ``I_nu_b``, where this quantity is then used. Afterwards it has to be reset.

tardis/montecarlo/src/integrator.c
unoebauer: Yeah - my bad! Coding too much in python lately...

unoebauer:
Seems to work pretty well already:

formal_integral

wkerzendorf:
@unoebauer that looks very good!

tardis/montecarlo/src/integrator.c
```diff
@@ -167,19 +168,26 @@ _formal_integral(
...
-          size_tau = size_line * size_shell;
+          size_tau = size_line * size_shell,
+          finished_nus = 0;
 
   double R_ph = storage->r_inner[0];
   double R_max = storage->r_outer[size_shell - 1];
   double pp[N];
   double *exp_tau = calloc(size_tau, sizeof(double));
-#pragma omp parallel firstprivate(L, exp_tau)
+#ifdef WITHOPENMP
```
yeganer: For the formal integral openmp parallelization I chose a different approach. By providing a dummy header, we can limit `#ifdef WITHOPENMP` to the header part of the document and I would strongly advise against including it anywhere in the body.

I know this implementation is basically a copy/paste of my progress from `cmontecarlo.c` but I think we are now able to provide a more readable solution.

tardis/montecarlo/src/integrator.c
```diff
@@ -167,19 +168,26 @@ _formal_integral(
...
+#pragma omp parallel firstprivate(L, exp_tau, finished_nus)
```
yeganer: If we make `finished_nus` shared we can make the bottom printing much nicer

unoebauer:
Unfortunately, and despite the excellent results obtained with tardis_example, the current formal integral scheme goes completely mad in a simple pure hydrogen atmosphere, in which only the Halpha is active. This is independent of whether electron scattering is active or not.

I'll have a closer look at it tomorrow.

yeganer:
@unoebauer Does this apply to the version in the current master or only to this PR?

yeganer:
Okay, I think I found two bugs.

1. We use `line_search` to determine `idx_nu_start`. The output of `line_search` is the index of the next line to the red, or `no_of_lines` if the value is redder than the reddest line. That means, *pline becomes a pointer to uninitialized memory and **ALL** 2D arrays point to the first line in the next shell.
We should really think about ending the line list with a `0`.

2. If there is no line interaction AT ALL for one p ray, then `I_nu_r` never get's updated, that means the intensity is 0 and not related to the value we initialized `I_nu_b` with.

tardis/montecarlo/src/integrator.c
```diff
@@ -213,62 +221,120 @@ _formal_integral(
...
+                      // advance pointers
+                      direction = shell_id[i+1] - shell_id[i];
+                      pexp_tau  += direction * size_line;
+                      patt_S_ul += direction * size_line;
+                      pJred_lu  += direction * size_line;
+                      pJblue_lu += direction * size_line;
                     }
                 }
-              I_nu[p_idx] *= p;
+              I_nu_r[p_idx] *= p;
```
yeganer: If nu is redder than the reddest line `I_nu_r[p_idx]` will always be `0`. So we have to account for the fact, that there may not be a line interaction. I propose to initialize `I_nu_r` as it physically does make sense in my opinion. `I_nu_r` is the intensity after a interaction. We assume a photosphere and emit radiation there. I think it makes sense to say "After they leave the photosphere there are at the red wing of some transition that happens at the boundary of the photosphere".
For the case that the ray does not intersect the photosphere, we initialize to 0. Here again I think we can apply the argument that we are at the red wing of a hypothetical transition after which our intensity is 0. Then we propagate through matter and apply the electron scattering.

Does this interpretation make sense? It would at least solve the problem and make the code a little bit easier (only Jkk has to be calculated in the first block)

What's left is to decide what happens in terms of electron scattering if there is no resonance.

Swapping the initialization to `I_nu_r` and changing the code in `first` does produce the correct spectrum for disabled electron scattering.

unoebauer:
I've now reintroduced I_nu and thus avoided the problem of an uninitialised integration in case of no resonances on the ray. With this change the simple H-atmosphere problem works again when electron scattering is switched off. Once activated, I still get a terrible match between formal integral and virtual packet spectrum. 

The problem with the ``line_list_search`` is still present, but as I will detail in the next comment, the problems in the simple H-atmosphere with electron scattering are more fundamental and cannot be resolved with the current information we are extracting from the Monte Carlo simulation.

Pragmatically, however, the current state of the Formal Integral simulation should work for all realistic supernova calculations (the formal integral matches the virtual spectrum very well in the tardis example setup, both with and without e-scattering). Since we will typically have a very large line list once a realistic supernova ejecta is set up, we will over the entire spectrum not reach the boundaries of this list and thus never run into the numerical problems associated with the ``line_list_search`` and also not into the conceptual problems outline below.

unoebauer:
In order to estimate the effect of electron scattering on the various rays, we need to now the strength of the diffuse radiation field at the LF frequency of the frequency bin we are interested in. Lucy 1999 estimates the diffuse radiation field by considering the mean intensity in the blue and red wings of the next and previous lines. This is justified as long as the current LF frequency is not too far from the frequencies of the lines.

It is however a problem for applications like the H-atmosphere with only H-alpha. Here, most parts of the spectrum are far beyond the frequency of this line and the diffuse radiation field cannot be accurately estimated by considering the intensity in the blue or red wing of the H-alpha line.

To correctly calculate the e-scattering contribution in such cases a ``J_nu`` estimator would have to be recorded in the Monte Carlo simulation. In principle this is possible (@chvogl has done this already in his Type IIP branch) but associated with additional effort.

unoebauer:
@wkerzendorf, @yeganer, @chvogl: please have a look at this PR. I think this is as far as we can go without introducing additional estimators.

tardis/montecarlo/src/integrator.c
```diff
@@ -223,52 +234,109 @@ _formal_integral(
...
+                      // Calculate e-scattering optical depth to next resonance point
+                      zend = storage->time_explosion / C_INV * (1. - *pline / nu);
+
+                      if (first == 1){
+                        // First contribution to integration
+                        // NOTE: this treatment of I_nu_b (given by boundary
+                        // conditions) is not in Lucy 1999; should be
+                        // re-examined carefully 
+                        escat_contrib += (zend - zstart) * escat_op * (*pJblue_lu - I_nu[p_idx]) ;
+                        I_nu[p_idx] = I_nu[p_idx] + escat_contrib;
```
chvogl: If you want you can move this out of the if statement.

tardis/montecarlo/src/integrator.c
```diff
@@ -223,52 +234,109 @@ _formal_integral(
...
+                        first = 0;
+                      }
+                      else{
+                        // Account for e-scattering, c.f. Eqs 27, 28 in Lucy 1999
+                        Jkkp = 0.5 * (*pJred_lu + *pJblue_lu);
+                        escat_contrib += (zend - zstart) * escat_op * (Jkkp - I_nu[p_idx]) ;
+                        I_nu[p_idx] = I_nu[p_idx] + escat_contrib;
```
chvogl: If you want you can move this out of the if statement.

tardis/montecarlo/formal_integral.py
```diff
@@ -124,4 +128,9 @@ def make_source_function(self):
         att_S_ul =  ( wave * (q_ul * e_dot_u) * t  / (4*np.pi) )
 
         result = pd.DataFrame(att_S_ul.as_matrix(), index=transitions.transition_line_id.values)
-        return result.ix[atomic_data.lines.index.values].as_matrix(), e_dot_u
+        att_S_ul = result.ix[atomic_data.lines.index.values].as_matrix()
```
wkerzendorf: if you just do `.values` that is maybe clearer that you just want the array and not as a real matrix.

unoebauer: actually, this is on you :) You've implemented this ages ago (SOCIS 2016) and I've just shuffled things aroun

PR #761, opened 2017-07-04
Add caching for integrated spectrum: Fix #760

### LINKED ISSUE #760: Formal Integral: spectrum is not cached
When using the new Formal Integral feature, the integration is repeated every time the ``spectrum_integrated`` attribute of the runner is called. This is very time consuming and unnecessary. Instead, the spectrum should be calculated one during the first call of the attribute and then cached.

unoebauer:
This small PR fixes #760 and introduces a caching for ``spectrum_integrated``.

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -43,6 +43,7 @@ class MontecarloRunner(object):
         self.inner_boundary_albedo = inner_boundary_albedo
         self.line_interaction_type = line_interaction_type
         self._integrator = None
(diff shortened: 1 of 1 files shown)


wkerzendorf:
@unoebauer : @yeganer suggests caching this on the integrator - and I think that is probably a good idea. There we should then also use `lazyproperty` (http://docs.astropy.org/en/stable/api/astropy.utils.decorators.lazyproperty.html) which does exactly what do coded here.

PR #762, opened 2017-07-07
[TEP007] Isotope uniform config option

vg3095:
This PR is for adding support to read `uniform isotope abundances`.
For seeing sample output : 
[Notebook](https://github.com/vg3095/decay/blob/master/isotope_uniform_config.ipynb) (**Updated**)
[Config](https://github.com/vg3095/decay/blob/master/isotope.yml) (**Updated**)

For Isotopes , you will need to mention mass-no. in every case.
For example `Ni` will not work , `Ni56` , `Ni58` will work.

It is because ,I think Tardis does not store `mass_no.` of stable isotope elements, so as of now , it supports this type of format( `ElementMass_no`)

### DIFF
--- a/tardis/io/model_reader.py
+++ b/tardis/io/model_reader.py
@@ -10,7 +10,7 @@ import logging
 # Adding logging support
 logger = logging.getLogger(__name__)
 
-from tardis.util import parse_quantity, element_symbol2atomic_number, MalformedElementSymbolError
+from tardis.util import parse_quantity
 
 class ConfigurationError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


yeganer:
I think in the config we should merge abundances and isotope abundances.
So specifying `Ni: 0.2` and `Ni56: 0.1` should result in `Ni: 0.3` at `t=0`.
The same goes for the abundance file.

wkerzendorf:
@vg3095 : the suggestion by @yeganer is what we want to implement.

vg3095:
@wkerzendorf Please review .

tardis/model/base.py
```diff
@@ -317,17 +321,29 @@ def from_config(cls, config):
...
                                      index=pd.Index(np.arange(1, 120),
                                                     name='atomic_number'),
                                      dtype=np.float64)
-
             for element_symbol_string in abundances_section:
                 if element_symbol_string == 'type':
                     continue
-                z = element_symbol2atomic_number(element_symbol_string)
-                abundance.ix[z] = float(abundances_section[element_symbol_string])
+                try:
```
wkerzendorf: maybe this splitting - might be better performed earlier in the process and not in the init of the model (generally keeping logic out of inits is good practice).

tardis/io/model_reader.py
```diff
@@ -98,6 +98,45 @@ def read_abundances_file(abundance_filename, abundance_filetype,
...
+                                     dtype=np.float64)
+
+    for element_symbol_string in abundances_section:
+        if element_symbol_string == 'type':
+            continue
+        try:
+            z = element_symbol2atomic_number(element_symbol_string)
+            abundance.ix[z] = float(
+                abundances_section[element_symbol_string])
+        except MalformedElementSymbolError:
```
wkerzendorf: what happens when this fails as well. Actually you can probably use pyne for both. try that.

wkerzendorf:
@vg3095 I think that looks good - apart from the points I made - good job!!

tardis/io/model_reader.py
vg3095: @wkerzendorf If this fails, then it will throw error raised by `pyne`. 
(Ex- `RuntimeError: Not a Nuclide! NIEFEG --> 0`)

If I use `pyne` then I can replace this by something like this:
```
try:
      find atomic_no and mass_no by pyne
      if mass_no!=0:
           Its a isotope
      else:
           Its a normal element
except RuntimeError as e:
      raise RuntimeError('Abundances are not defined properly in config', e.args)
```
`Pyne` returns `mass_no` zero whenever we pass elements in normal form(without `mass_no` attached), something like this:
```
>>> nucname.anum('O')
0
>>> nucname.anum('Ni')
0
```

wkerzendorf: @vg3095 try to figure out how to pass elements to pine - I'm sure that works.

wkerzendorf:
@vg3095 looks good @tardis-sn/tardis-core can someone cross check before I merge?

PR #763, opened 2017-07-11
[TEP007] Isotope stratified file support

vg3095:
This PR is related to adding support for reading `stratified abundances with isotopes` . This is rebased over  PR #762 (Isotope uniform config)
For sample seeing output:
[Notebook](https://github.com/vg3095/decay/blob/master/isotope_file_config.ipynb)
[abund.dat](https://github.com/vg3095/decay/blob/master/abund.dat)
[Config](https://github.com/vg3095/decay/blob/master/isotope_file.yml)

Just want to clear 2 things,
- I have assumed, only `30 normal elements` (hardcoded in code), and rest isotopes.
- Currently the header is of following format (for Z=1 to 30 , numbers are used),
`1 2 3 4........30 Ni56 Ni58`
_Do we need the format same as above or format below_ (represent Z=1 to 30 , as symbols):
`H He ......Zn Ni56 Ni58`
_or you are thinking of some other format_.

- _Will number of `normal elements` be `variable` ?_ . If we want to make normal elements(Z<=30) variable in file, then I can think of this as of now,
   - if we follow 1st header format as I said above, then header is read as `str` format, because of mixed datatypes (1...30(`int`) then isotopes names(`str`) ), I could use regex here to find the index at which  isotope names(`str`) starts,
  - if we follow 2nd header format, then I can use `element_symbol2atomic_number` function defined in `util.py` , and at the header index at which it raises `MalformedElementSymbolError` ,  isotopes will start.

### DIFF
--- a/tardis/io/model_reader.py
+++ b/tardis/io/model_reader.py
@@ -108,7 +108,7 @@ def read_simple_ascii_isotopes(fname):
 
     #Reshape Numpy Structured arrays
     #File rows are read as tuples here, this line converts it to a Matrix form
-    #First colomn is ignored
+    #First column is ignored
     data = data.view(np.float64).reshape(-1, len(data.dtype))[:, 1:]
 
     #Parse isotopes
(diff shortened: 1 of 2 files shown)


vg3095:
@unoebauer @yeganer Take a look at this PR, whenever you have time.

PR #764, opened 2017-07-12
[TEP007] Isotope stratified file support

vg3095:
This PR is related to adding support for reading `stratified abundances with isotopes` . This is rebased over  PR #762 (Isotope uniform config)

I have added 2 functions -
- `read_simple_isotope_abundances `: To read isotope abundances file in new format
- `convert_abundances_format` : To convert old abundances file format to new one

For sample seeing output:
[Notebook](https://github.com/vg3095/decay/blob/master/isotope_file_config.ipynb)
[iso_abund.csv](https://github.com/vg3095/decay/blob/master/iso_abund.csv)
[Config](https://github.com/vg3095/decay/blob/master/isotope_file.yml)


vg3095:
Travis fail is due to Mac build timeout.

unoebauer:
I am wondering if we can simplify the configuration slighly. I am thinking about the following changes:

1.

vg3095:
- I agree on the 1st point, I think we can remove this. I found out that `pyne` has a dictionary/mapping between `element symbol` and its `atomic number` , and it takes only normal `elements`( `not isotopes`). It is called `pyne.nucname.name_zz` . If present in this dictionary/mapping, it means its a normal element else isotope.

- For second point, I think, if we remove `isotopes` option, then also , we have to somehow specify that `header` is present in the config ,  or I have to check for the first row, if it contains numbers or elements. I don`t know if checking first row is good idea or not. So, for 2nd point , I am not sure.

tardis/io/model_reader.py
```diff
@@ -85,19 +85,66 @@ def read_abundances_file(abundance_filename, abundance_filetype,
     """
 
     file_parsers = {'simple_ascii': read_simple_ascii_abundances,
-                    'artis': read_simple_ascii_abundances}
+                    'artis': read_simple_ascii_abundances,
+                    'isotopes': read_simple_isotope_abundances}
```
wkerzendorf: I would actually call this tardis_model. I think this can be some sort of default model. where you specify things in this way

vg3095:
I have removed the 2nd header line('E' or 'I') and changed filetype name to `tardis_model`. Notebook in the description is updated.

tardis/util.py
```diff
@@ -428,4 +430,14 @@ def quantity_linspace(start, stop, num, **kwargs):
...
+    return np.linspace(start.value, stop.to(start.unit).value, num, **kwargs) * start.unit
+
+
+def convert_abundances_format(fname, delimiter='\s+'):
+    df = pd.read_csv(fname, delimiter=delimiter, comment='#', header=None)
+    #Drop shell index column
+    df.drop(df.columns[0], axis=1, inplace=True)
+    #Assign header row
+    df.columns = [nucname.name(i)
+                  for i in range(1, df.shape[1] + 1)]
```
unoebauer: nucname.name(i) will throw an error, if somebody a wrong specifies an element or isotope identifier, right?

tardis/io/model_reader.py
```diff
@@ -85,19 +85,71 @@ def read_abundances_file(abundance_filename, abundance_filetype,
...
+                    'artis': read_simple_ascii_abundances,
+                    'tardis_model': read_simple_isotope_abundances}
+
+    isotope_abundance = pd.DataFrame()
+    if abundance_filetype == 'tardis_model':
+        index, abundances, isotope_abundance = read_simple_isotope_abundances(
+            abundance_filename)
+    else:
+        index, abundances = file_parsers[abundance_filetype](
+            abundance_filename)
```
unoebauer: Can we improve here? Something like
```python
index, abundances, isotope_abundance = file_parser[abundance-filetype]
```
such that we don't need the if clause?

tardis/util.py
vg3095: Yes, it will throw a `Runtime Error` with the wrong name of element/isotope. Ex- `RuntimeError: Not a Nuclide! XYZ --> 0`, where `XYZ` is wrongly specified element.

tardis/io/model_reader.py
vg3095: If, `read_simple_ascii_abundances` function returns an extra `None` (instead of `(index,abundance)`, returning `index,abundance,None`)) , then we can remove this. But I don`t think , its a good idea.

wkerzendorf:
@vg3095 I overlooked that before - but we definitely need tests and documentation for all of this.

vg3095:
@wkerzendorf I have now updated this PR with tests.

wkerzendorf:
@vg3095 in your examples - you should try to make the abundances add up to one :wink:

tardis/io/model_reader.py
```diff
@@ -256,3 +264,31 @@ def read_simple_ascii_abundances(fname):
     abundances = pd.DataFrame(data[1:,1:].transpose(), index=np.arange(1, data.shape[1]))
 
     return index, abundances
+
+
+def read_simple_isotope_abundances(fname, delimiter=','):
```
wkerzendorf: the default delimiter should be a whitespace. this is a special option in `read_csv`.

tardis/io/tests/data/tardis_model_abund.csv
```diff
@@ -0,0 +1,11 @@
+C,Mg,Si,Ni56,Ni58
```
wkerzendorf: default delimiter should be whitespace.

vg3095:
@wkerzendorf I have now updated csv files

wkerzendorf:
@vg3095 can you rebase this. @tardis-sn/tardis-core looks good from my side - can someone else have a look?

tardis/io/model_reader.py
unoebauer: Ok, fine with me

tardis/util.py
unoebauer: Excellent

Issue #690, opened 2017-01-13
GUI does not work with current TARDIS version


unoebauer:
Since the GUI was originally designed to work with a ``model`` object but since TARDIS now returns a simulation object, it can't start but crashes with the following Error message:

```

In [3]: mdl =  tardis.run_tardis(config)

In [4]: interface.show(mdl)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-8332875ec95a> in <module>()
----> 1 interface.show(mdl)

/home/ulrich/miniconda2/envs/tardis/lib/python2.7/site-packages/tardis_sn-1.5.dev2668-py2.7-linux-x86_64.egg/tardis/gui/interface.pyc in show(model)
     41     tablemodel = SimpleTableModel
     42     win = Tardis(tablemodel)
---> 43     win.show_model(model)
     44 
     45     if importFailed:

/home/ulrich/miniconda2/envs/tardis/lib/python2.7/site-packages/tardis_sn-1.5.dev2668-py2.7-linux-x86_64.egg/tardis/gui/widgets.pyc in show_model(self, model)
   1159         """
   1160         if model:
-> 1161             self.mdv.change_model(model)
   1162         if model.converged:
   1163             self.successLabel.setText('<font color="green">converged</font>')

/home/ulrich/miniconda2/envs/tardis/lib/python2.7/site-packages/tardis_sn-1.5.dev2668-py2.7-linux-x86_64.egg/tardis/gui/widgets.pyc in change_model(self, model)
    567         self.model = model
    568         self.tablemodel.arraydata = []
--> 569         self.tablemodel.add_data(model.t_rad.value.tolist())
    570         self.tablemodel.add_data(model.ws.tolist())
    571 

AttributeError: 'Simulation' object has no attribute 't_rad'
```

_Note: this is a GSoC 2017 project related issue_

vedantrathore:
@unoebauer I would like to work on this, could you give me a few pointers to start with

wkerzendorf:
@vedantrathore I would suggest you stick to one topic and move to a new one - once you have solved it.

vedantrathore:
Ok sure @wkerzendorf

deep110:
@wkerzendorf @unoebauer I have submitted a PR #691 on this issue. Can you look into it.

pfreddy:
I would like to ask if there has been any progress made in this issue?
[...]

PR #765, opened 2017-07-13
Attempt to fix up the GUI

ssim:
This is an attempt to fix up the GUI to work with the current TARDIS version.

Much of the work here was done by Mark Magee, with SAS helping.

The approach to getting things working has been empirical / pragmatic - an attempt to replace cases where information has been moved around. In some cases this may have led to small changes in functionality.

@unoebauer @chvogl It would be appreciated if someone can check how the "lines" objects are been accessed (via iloc) - it appears to work but the indexing was not completely obvious.


tardis/analysis.py
```diff
@@ -77,13 +77,13 @@ def update_last_interaction_filter(self):
...
-            line_in_nu = self.lines.wavelength.ix[self.last_line_interaction_in_id].values
+            line_in_nu = self.lines.wavelength.iloc[self.last_line_interaction_in_id].values
             packet_filter = (line_in_nu > self.wavelength_start.to(u.angstrom).value) & \
                 (line_in_nu < self.wavelength_end.to(u.angstrom).value)
 
                 
-        self.last_line_in = self.lines.ix[self.last_line_interaction_in_id[packet_filter]]
-        self.last_line_out = self.lines.ix[self.last_line_interaction_out_id[packet_filter]]
+        self.last_line_in = self.lines.iloc[self.last_line_interaction_in_id[packet_filter]]
+        self.last_line_out = self.lines.iloc[self.last_line_interaction_out_id[packet_filter]]
```
unoebauer: Actually, we have to use ``ix`` since we want to get the line with a certain ID back and the lines dataframe is indexed by these IDs. The reason why in your trials ``idx`` did not work but ``iloc`` did is because you did not properly mask out the non-interacting or purely e-scattering packets. Thus, many packets have a ``-1`` as line ID (which simply means *no line interaction performed*). ``ix`` fails in this case, since there is no line with ID -1 in the data frame. ``iloc`` simply returns the last element in the data frame (some line with a very long wavelength). But of course, this is not a line which contributes to the wavelength window in question.

Once a proper masking is implemented, ``ix`` can be safely used again

unoebauer:
I'll try to submit a PR with a proper masking

unoebauer:
I've made changes to the GUI to mask out non-line interacting packets. I've tested the GUI with my changes on tardis_example and it works.

ssim:
This is of relevant to #690

wkerzendorf:
@ssim @unoebauer - great work!

PR #770, opened 2017-07-20
Rebase pr692 (needed for new carsus format)

yeganer:


yeganer:
Closed in favor of #786

PR #771, opened 2017-07-20
[TEP007] Add cmfgen2tardis script

vg3095:
This script takes a `CMFGEN file` as an input , and an output path to save the converted CSV file in new TARDIS file format. 
An example of input CMFGEN file ([Link](https://github.com/vg3095/decay/blob/master/DDC15_SN_HYDRO_DATA_0.976d)) and its converted CSV file ([Link](https://github.com/vg3095/decay/blob/master/DDC15_SN_HYDRO_DATA_0.csv))
Example, for how to use-
Format - `cmfgen2tardis /path/to/input_file path/to/output/`
`cmfgen2tardis tardis_example/DDC15_SN_HYDRO_DATA_0.976d tardis_example/`
______
- Now velocity, density, electron_density and temperature are also saved.

- Values are stored in reverse order, to match how Tardis stores things. For ex - velocity values in Tardis are stored in increasing order, while in Cmfgen files , it is in decreasing order .

### DIFF
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -1,1 +1,0 @@
-from scripts.cmfgen2tardis import *
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -1,0 +1,1 @@
+from scripts.cmfgen2tardis import *
--- a/scripts/cmfgen2tardis.py
+++ b/scripts/cmfgen2tardis.py
@@ -1,0 +1,62 @@
+import os
+import sys
+
+import numpy as np
+import pandas as pd
+
+from tardis.atomic import AtomData
+
+atomic_dataset = AtomData.from_hdf5()
+
+
+def get_atomic_number(element):
+    index = -1
+    for atomic_no, row in atomic_dataset.atom_data.iterrows():
+        if element in row['name']:
+            index = atomic_no
+            break
+    return index
+
+
+def convert_format(file_path):
+
+    with open(file_path, 'r') as f:
+        for line in f:
+            items = line.split()
+            n = len(items)
+
+            if 'data points' in line:
+                df = pd.DataFrame(columns=np.arange(int(items[n - 1])),
+                                  index=pd.Index([],
+                                                 name='element'),
+                                  dtype=np.float64)
+
+            if 'mass fraction\n' in line:
+                    abundances = []
+                    element_string = items[0]
+                    atomic_no = get_atomic_number(element_string.capitalize())
+                    element_symbol = atomic_dataset.atom_data.loc[atomic_no]['symbol']
+
+                    #Its a Isotope
+                    if n == 4:
+                        element_symbol += items[1]
+
+                    for line in f:
+                        items = line.split()
+                        if items:
+                            abundances.extend(
+                                np.array(items).astype(np.float64))
+                        else:
+                            break
+
+                    df.loc[element_symbol] = abundances
+        return df.transpose()
+
+
+def parse_file():
+    file_path = sys.argv[1]
+    output_path = sys.argv[2]
+    df = convert_format(file_path)
+    filename = os.path.basename(file_path)
+    save_name = '.'.join((os.path.splitext(filename)[0], 'csv'))
+    df.to_csv(os.path.join(output_path, save_name), index=False, sep=' ')
--- a/scripts/cmfgen2tardis.py
+++ b/scripts/cmfgen2tardis.py
@@ -1,10 +1,16 @@
 import os
 import sys
-
+import argparse
 import numpy as np
 import pandas as pd
 
 from tardis.atomic import AtomData
+
+parser = argparse.ArgumentParser()
+parser.add_argument('input_path', help='Path to a CMFGEN abundance file')
+parser.add_argument(
+    'output_path', help='Path to store converted TARDIS abundance file')
+args = parser.parse_args()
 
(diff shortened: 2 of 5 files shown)


scripts/__init__.py
```diff
@@ -0,0 +1 @@
+from scripts.cmfgen2tardis import *
```
vg3095: @wkerzendorf @yeganer When I leave `__init__.py` file empty, travis fails, because it skips this file.
(`warning: build_scripts: scripts/__init__.py is an empty file (skipping)`). [Travis Report](https://travis-ci.org/tardis-sn/tardis/jobs/255615422).
Can this be avoided.

scripts/cmfgen2tardis.py
```diff
@@ -0,0 +1,66 @@
...
+                            abundances.extend(
+                                np.array(items).astype(np.float64))
+                        else:
+                            break
+
+                    df.loc[element_symbol] = abundances
+        return df.transpose()
+
+
+def parse_file():
```
wkerzendorf: I don't think it works that way. look at the link I sent you

vg3095: @wkerzendorf I think, you are saying to create a `__main__.py` file under scripts, as in the link you sent. If we do , as in the link you sent, we may  be able to do something like this -
```
tardis_scripts cmfgen2tardis --input_file=/path/file --output_file=/tmp/
tardis_scripts get_spectrum --gdb
```
So, first argument tardis_scripts will be fixed, meaning we want to run scripts, second argument will be the script we want to run, and rest will be arguments required for individual scripts.

By my current approach
`cmfgen2tardis --input_file=/path/file --output_file=/tmp/`
will work, and for this we do not require a `__main__.py` file.[Stackoverflow](https://stackoverflow.com/questions/774824/explain-python-entry-points)

So do you want the 1st way or 2nd way ?

unoebauer:
Would be good to have a more detailed description of the scope of this PR... Is it taking CMFGEN config files and then generating appropriate TARDIS inputs or is it supposed to do something else?

vg3095:
@unoebauer I have updated the description for PR now.

wkerzendorf:
@vg3095 I don't think these should live within scripts outside of the package. I think you also misunderstood the entrypoints idea. Essentially it will generate a script based on calling a "`main`" function that takes `args`. so you wrote it like a script but that is not the right approach. Write is as a function that takes `args`.

wkerzendorf:
@vg3095 can you also link a CMFGEN model?

vg3095:
@wkerzendorf I have now added a CMFGEN file in the description

wkerzendorf:
@vg3095 once this works - we also need to read the velocity, temperature, atom density and electron density.

tardis/scripts/cmfgen2tardis.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def parse_file(args):
+    df = convert_format(args.input_path)
+    filename = os.path.basename(args.input_path)
+    save_name = '.'.join((os.path.splitext(filename)[0], 'csv'))
+    df.to_csv(os.path.join(args.output_path, save_name), index=False, sep=' ')
+
+
+def main():
```
wkerzendorf: I think they entry point actually gives args to the function. So we might want to use that.

wkerzendorf:
@vg3095 actually, I think its close. Let's do some documentation and then I think we can merge and work on getting the velocity and other quantities in there.

tardis/scripts/cmfgen2tardis.py
vg3095: @wkerzendorf See these links
- [Stackoverflow](https://stackoverflow.com/questions/2853088/setuptools-not-passing-arguments-for-entry-points). First and second answer. Here it is mentioned  

> setuptools console_scripts entry point wants a function of no arguments. 

- [Blog](http://locallyoptimal.com/blog/2014/03/14/executable-python-scripts-via-entry-points/). Here also entry point function is used without `args`.

vg3095:
@wkerzendorf I have now added doc for the script.

tardis/scripts/cmfgen2tardis.py
```diff
@@ -0,0 +1,70 @@
...
+
+                    for line in f:
+                        items = line.split()
+                        if items:
+                            abundances.extend(
+                                np.array(items).astype(np.float64))
+                        else:
+                            break
+
+                    df.loc[element_symbol] = abundances
```
vg3095: @wkerzendorf As you said in PR #772 , if we are reversing `velocity`, then all other columns should also be reversed, so that means , while saving `abundances` from CMFGEN file, we should also reverse them, -- right?

vg3095:
- Now `velocity`, `density`, `electron_density` and `temperature` are also saved as DAT file.
- Since, we are reversing `velocity`, so reversing all columns along with it.

tardis/scripts/cmfgen2tardis.py
vg3095: If I am missing something very obvious here, which I can't see, then please point it out.
I also saw [runner](https://github.com/tardis-sn/tardis/blob/master/tardis/tests/integration_tests/runner.py) script , here `run` method is used as an entry point function and is used without passing `args`.

vg3095:
Changed script, to store all quantities in a single CSV file.

PR #772, opened 2017-07-25
[TEP007] CMFGEN density parser

vg3095:
A parser to read velocity , density, electron_density from a CMFGEN file. 
Example - 
[Notebook](https://github.com/vg3095/decay/blob/master/density_cmfgen.ipynb)
[CMFGEN file](https://github.com/vg3095/decay/blob/master/DDC15_SN_HYDRO_DATA_0.976d)
[Converted CMFGEN file](https://github.com/vg3095/decay/blob/master/DDC15_SN_HYDRO_DATA_0.csv)


vg3095:
I have overwrited `electron_densities` in one case for now, if my approach is correct, then I will expand it to others.

tardis/plasma/properties/ion_population.py
```diff
@@ -223,31 +224,39 @@ def _calculate_block_ids(phi):
         return calculate_block_ids_from_dataframe(phi)
 
     def calculate(self, phi, partition_function, number_density):
-        n_e_convergence_threshold = 0.05
-        n_electron = number_density.sum(axis=0)
-        n_electron_iterations = 0
-
-        while True:
+        if self._electron_densities is None:
```
wkerzendorf: maybe make ElectronDensities an adiitional module.

vg3095: @wkerzendorf  Okay, If I make a `ElectronDensities` module like this under `plasma_input.py`
```
class ElectronDensities(ArrayInput):
    outputs = ('electron_densities',)
```


- First of all, If we are not supplying `electron_densities` from CMFGEN file (normal cases), then how will I set this value ?
The Logic I saw in `IonNumberDensity` , calculates both `electron_densities` and `ion_number_density`. And if I make a separate module
for it, then I will have to remove `electron_densities` from the output list of `IonNumberDensity`, because then it will giver error while
resolving variables(`_resolve_update_list` method)

- And then I think , it will resulting in changing some tests/fixtures which are dependent on electron_density, if above behaviour is changed.

wkerzendorf: maybe what you did is fine. I was hoping to make it even nicer - but maybe for now it's fine.

tardis/io/model_reader.py
```diff
@@ -235,6 +244,20 @@ def read_artis_density(fname):
     return time_of_model, velocity, mean_density
 
 
+def read_cmfgen_density(fname):
+
+    df = pd.read_csv(fname, comment='#', delimiter='\s+', skiprows=1)
+    velocity = (df['Velocity'].values[::-1] * u.km / u.s).to('cm/s')
```
vg3095: @wkerzendorf  For CMFGEN files , I think the convention for `velocity` is different . For ex-

- In CMFGEN files velocity values are in `decreasing` order. [CMFGEN file](https://github.com/vg3095/decay/blob/master/DDC15_SN_HYDRO_DATA_0.976d) 
`6.9999926E+04   6.9999184E+04 ........   8.7744269E+02   8.7166905E+02`

- In TARDIS files, velocity values  are in `ascending` order. [TARDIS file](https://github.com/tardis-sn/tardis/blob/master/tardis/io/tests/data/density.dat) 
`9000.0000 10500.000 ......................... 21000.000 22500.000`

- So, when I did not reverse velocity array here(for CMFGEN case), it was not parsing due to this part of code( because I think the convention for velocity is different for CMFGEN case) 
```
    v_inner = velocity[:-1]
    v_outer = velocity[1:]

    invalid_volume_mask = (v_outer - v_inner) <= 0
```

- So, I have reversed it here, so that CMFGEN velocity values can be in ascending order as well 

I don't know , if I am right, so can you check this ?

wkerzendorf: @vg3095 if you reverse the velocity - you have to reverse all the other columns as well. They all belong together.

vg3095:
I have now added unit tests.

vg3095:
While saving `cmfgen` files to `tardis_model` file format using cmfgen script, all columns will be stored in reverse order, so that there is no need to reverse it in `read_cmfgen_density` function .

tardis/io/model_reader.py
```diff
@@ -235,6 +244,20 @@ def read_artis_density(fname):
     return time_of_model, velocity, mean_density
 
 
+def read_cmfgen_density(fname):
+
```
wkerzendorf: documentation please

tardis/io/tests/data/tardis_model_density.dat
```diff
@@ -0,0 +1,13 @@
+0.976 day
+Index Velocity Densities ElectronDensities Temperature
```
wkerzendorf: please save without index

wkerzendorf: and everything lower case and singular - and with underscore.

vg3095:
I have now made a single CSV file for abundances, density, etc. Also added docstrings wherever required.

tardis/io/tests/data/tardis_model_format.csv
```diff
@@ -0,0 +1,12 @@
+0.976 day
```
wkerzendorf: maybe add a `t_0:` before that. but other than that - very nice!!

tardis/io/tests/data/tardis_model_format.csv
```diff
@@ -0,0 +1,12 @@
...
+velocity densities electron_densities temperature C O Mg Si Ni56 Ni58
+871.66905 4.2537191E-09 259538070000000 7.6395577 0 0 0 0.6 0.4 0
```
wkerzendorf: electron_densities - should be float. 
and add one more line with units for the different quantities. with 1 for dimensionless units (such as abundance fractions)>

wkerzendorf: the temperatures here are 10^4 k

tardis/io/tests/data/tardis_model_format.csv
wkerzendorf: please write as `t_0: ...`

tardis/io/tests/data/tardis_model_format.csv
```diff
@@ -0,0 +1,13 @@
+t0: 0.976 day       
+velocity temperature densities electron_densities C O Mg Si Ni56 Ni58
+km/s K gm/cm^3 /cm^3 1 1 1 1 1 1
```
wkerzendorf: `gm/cm^3` can become `g/cm^3`

vg3095: @wkerzendorf I have updated it now

wkerzendorf:
@tardis-sn/tardis-core I'm signing off - can someone else look over this please!

PR #773, opened 2017-07-27
Update Documentation

unoebauer:
This PR performs a number of small updates in the Documentation. In particular:

- update of list of works using TARDIS
- update news page: GSOC17
- remove warning about broken GUI since it has been fixed
- update description of spectrum configuration part and add links to description of virtual packet mode and formal integral

### DIFF
--- a/docs/configuration/configuration_old.rst
+++ b/docs/configuration/configuration_old.rst
@@ -249,14 +249,23 @@ The spectrum section defines the
 
     spectrum:
         start : 500 angstrom
-        end : 20000 angstrom
-        bins : 1000
-        sn_distance : lum_density
-        #sn_distance : 10 Mpc
+        stop : 20000 angstrom
+        num : 1000
+        method: virtual
 
-Start and end are given as Quantities with units. If they are given in frequency space they are switched around if
-necessary. The number of bins is just an integer. Finally the ``sn_distance`` can either be a distance or the special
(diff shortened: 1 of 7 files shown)


docs/configuration/configuration_old.rst
```diff
@@ -249,14 +249,23 @@ The spectrum section defines the
...
-Start and end are given as Quantities with units. If they are given in frequency space they are switched around if
-necessary. The number of bins is just an integer. Finally the ``sn_distance`` can either be a distance or the special
-parameter ``lum_density`` which sets the distance to :math:`\sqrt{\frac{1}{4 \pi}}` to calculate the luminosity density.
+        stop : 20000 angstrom
+        num : 1000
+        method: virtual
+
+Start and end are given as Quantities with units. If they are given in
+frequency space they are switched around if necessary. The number of bins is
+just an integer. Finally, the method option selects the final spectral synthesis mode. Currently, there are three options:
```
yeganer: Please add that this is only relevant for the tardis script when saving the output as ascii.

unoebauer: Well, currently this may be true. But eventually, some of these will be mutually exclusive. For example, when the Formal integral method is selected, the virtual packet scheme will be deactivated.

wkerzendorf: maybe we should add this - that it has no effect on the tardis executable (currently).

PR #774, opened 2017-08-03
Replace Comparison values in Plasma Unit Tests with reference HDF file

vg3095:
This PR contains following changes - 
- Compare values from a reference HDF file , rather than from hard coded values
_____

Some new tests for Plasma

- **General Properties**
        - Dilution factor('w')

- **Atomic Properties**
       - nu
       - wavelength_cm
       - f_lu
       - metastability

- **Radiative Properties**
       - transition_probabilites

- **Scalar Properties**
        - helium_treatment
        - link_t_rad_t_electron
        - time_explosion

- **NTLE** 
   - **Ion Population Properties**
        - previous_electron_densities 
    - **Radiative Properties**
         - previous_beta_sobolev


vg3095:
I have currently replaced values for `test_partition_function.py`. If my approach is correct, I will then expand it to other plasma tests.

vg3095:
@wkerzendorf Can you check my approach ?

.travis.yml
```diff
@@ -21,27 +21,29 @@ env:
...
+        - PLASMA_LTE_REFERENCE_URL='https://github.com/vg3095/tardis-refdata/raw/plasma-ref2/plasma_reference/plasma_lte_reference.h5'
         - MINICONDA_URL='http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh'
 
 matrix:
     include:
         - python: 2.7
           env:
             - COMPILER=gcc
-            - SETUP_CMD='test --args="--atomic-dataset=$HOME/kurucz_cd23_chianti_H_He.h5"'
+            - SETUP_CMD='test --args="--atomic-dataset=$HOME/kurucz_cd23_chianti_H_He.h5 --plasma-reference=$HOME/plasma_reference/"'
```
vg3095: @wkerzendorf Here I want to pass path for `plasma-reference` files, but travis doesn't recognise this option.
image
I have added this option in `conftest.py` and this option works fine in my laptop. 
So, what should I do here?

.travis.yml
```diff
@@ -21,27 +21,29 @@ env:
         - SETUP_CMD='test'
         - TEST_MODE='normal'
         - ATOM_DATA_URL='https://github.com/tardis-sn/tardis-atomdata/raw/master/kurucz_cd23_chianti_H_He.h5.zip'
+        - PLASMA_NLTE_REFERENCE_URL='https://github.com/vg3095/tardis-refdata/raw/plasma-ref2/plasma_reference/plasma_nlte_reference.h5'
```
wkerzendorf: @vg3095 can you check out how this refdata is  solved in carsus  and use the same methods

.travis.yml
wkerzendorf: @yeganer has some comments on this - I let him discuss this with you.

yeganer:
@vg3095 as @wkerzendorf pointed out I have some ideas how we could solve the issue with the reference data for tardis. The idea borrows the approach from carsus in some way.
I would propose to move ALL reference data to the `tardis-refdata` repository and then rename the config option `atomic-dataset` to `tardis-refdata` which points to a checkout of said repository.

On the tardis side this has the benefit that we can treat all file names as relative to this 'root'. What do you think? Would this be a good idea?

vg3095:
@yeganer I think its a great idea.

yeganer:
@vg3095 With this approach, we should also test for the version of the `tardis-refdata` repository and fix this version in tardis. So specific tests would need version `> 0.1` for example.

vg3095:
@wkerzendorf You can review this PR now.

wkerzendorf:
@vg3095 is this rebased on #775 ?

vg3095:
@wkerzendorf Not completely. This misses 2 commits, which I added afterwards in PR #775. I have left it deliberately for now, so that `travis.yml` uses PR[#2](https://github.com/tardis-sn/tardis-refdata/pull/2) branch reference files. If you could merge Plasma reference files in PR[#2](https://github.com/tardis-sn/tardis-refdata/pull/2), then I could update this.

vg3095:
@wkerzendorf You can review this PR. I have rebased it. 
I have added some new tests , which I have mentioned in the description of the PR.

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,231 @@
...
+            cls.sim.plasma.to_hdf(cls.reference_file_path)
+            pytest.skip("Reference data saved at {0}".format(
+                cls.reference_file_path))
+
+        cls.plasma = cls.sim.plasma
+
+
+class TestNLTEPlasma(BasePlasmaTest):
+
+    @classmethod
```
wkerzendorf: why is this a classmethod?

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,231 @@
...
+    @pytest.fixture(scope="class", autouse=True)
+    def setup(cls, atomic_data_fname, tardis_ref_path):
+        cls.reference_file_path = os.path.join(
```
wkerzendorf: same code as above - why is this not in the baseclass

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,231 @@
...
+    def setup(cls):
+        pass
+
+    def read_hdf_attr(self, attr):
+        return pd.read_hdf(self.reference_file_path, os.path.join('plasma', attr))
+
+    #GENERAL PROPERTIES
+    def test_beta_rad(self):
+        pdt.assert_almost_equal(self.plasma.beta_rad,
+                                self.read_hdf_attr('beta_rad').values)
```
wkerzendorf: this could be parameterizable - that might make it easier.

.travis.yml
```diff
@@ -70,6 +70,8 @@ before_install:
     - if [[ $TEST_MODE == 'spectrum' ]]; then git lfs install --skip-smudge; fi
     - if [[ $TEST_MODE == 'spectrum' ]]; then git clone $TARDIS_REF_DATA_URL $HOME/tardis-refdata; fi
     - if [[ $TEST_MODE == 'spectrum' ]]; then cd $HOME/tardis-refdata; fi
+    - if [[ $TEST_MODE == 'spectrum' ]]; then git fetch origin pull/2/head:plasma-ref2; fi
```
vg3095: Yes, this is temporary.

vg3095:
@wkerzendorf I have updated this PR.

yeganer:
I don't have specific comments on the PR however, the coverage dropped by 6%. This is not acceptable, even more so for a PR that targets tests and has full coverage of the plasma module as the ultimate goal. Please have a look at the coverage report and fix this.

vg3095:
@yeganer @wkerzendorf I have made a PR #779 regarding this, can you check that?
Currently coverage report is saved, when tests run without `tardis-refdata` parameter. Tests for Plasma, depend on this parameter(`tardis-refdata`), and that's why coverage report does not correctly reflect the status for same.

yeganer:
@vg3095 I agree with that. We had a discussion about that topic sometime last year. If I remember correctly we decided to not test during the run with reference files because we were interested in the coverage of our unit-tests, not the integration tests.

With the addition of the integration tests and our plan to actively use `tardis-refdata` for the reference values of the unit tests, I think it is a sensible change to do the coverage run with reference data.
@wkerzendorf What are your thoughts?

wkerzendorf:
@vg3095 please rebase this and then see how high the coverage is for plasma.

vg3095:
@wkerzendorf I have rebased this. 
- There is a `1%` fall due to changing of `atomic-dataset`. Previously Plasma fixtures were calculated using `chianti_he_db` dataset, now the Plasma config files uses `kurucz_cd23_chianti_H_He` dataset. So, as far as I can understand from the coverage report, lines related to `collision_data` in `tardis/atomic.py` does not get covered using  `kurucz` dataset.

- Coverage of Plasma related files is constant because, [test_hdf_plasma.py](https://github.com/tardis-sn/tardis/blob/master/tardis/plasma/tests/test_hdf_plasma.py) also compares plasma attributes from HDF file.

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,122 @@
...
+from numpy.testing import assert_almost_equal
+from tardis.io.config_reader import Configuration
+from tardis.simulation import Simulation
+
+
+class BasePlasmaTest(object):
+    #Class defining all common tests for different setups of Plasma
+    #This can then be inherited for different Plasma setup
+
+    @pytest.fixture(scope="class", autouse=True)
```
yeganer: I think, we don't need to use `autouse=True` here as you are using the fixture directly in the tests.

wkerzendorf:
@vg3095 the general idea is to get the plasma part close to 100% of coverage - what is missing?

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,122 @@
...
+    def plasma(self, atomic_data_fname, tardis_ref_path, reference_file_path, config_path):
+        self.config = Configuration.from_yaml(config_path)
+        self.config['atom_data'] = atomic_data_fname
+        self.sim = Simulation.from_config(self.config)
```
yeganer: If I read the tests correctly, there is no need to set variables on `self` as they are never checked.

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,122 @@
...
+    combined_properties = general_properties + partiton_properties + atomic_properties + ion_population_properties + \
+        level_population_properties + radiative_properties + \
+        j_blues_properties
+
+    @pytest.mark.parametrize("attr", combined_properties)
+    def test_plasma_properties(self, reference_file_path, plasma, attr):
+        actual = getattr(plasma, attr)
+        expected = pd.read_hdf(reference_file_path,
+                               os.path.join('plasma', attr))
+        assert_almost_equal(actual, expected.values)
```
yeganer: Why are you sometimes using `pdt.assert_almost_equal` and here numpy's `assert_almost_equal`?
In general we should always compare the dataframes/series directly so we don't miss changes to the index.

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,122 @@
...
+    j_blues_properties = ['j_blues']
+
+    combined_properties = general_properties + partiton_properties + atomic_properties + ion_population_properties + \
+        level_population_properties + radiative_properties + \
+        j_blues_properties
+
+    @pytest.mark.parametrize("attr", combined_properties)
+    def test_plasma_properties(self, reference_file_path, plasma, attr):
+        actual = getattr(plasma, attr)
+        expected = pd.read_hdf(reference_file_path,
```
yeganer: I think we should not use `pd.read_hdf` as I think that opens the file for every test. I think it would be better to have a fixture for the opened `pd.HDFStore`.
I think one could use a 'yield-fixture' for the `pd.HDFStore` (see https://docs.pytest.org/en/2.7.3/yieldfixture.html#yieldfixture)

yeganer:
@vg3095 Do you think it is a sensible idea to add `chianti_h_he_db.h5` to `tardis-refdata` and use that database for the plasma tests? That way we don't lose collisional data.

A different solution would be to use `kurucz_atoms_chianti_many.h5`. That database in `tardis-refdata` should be fixed and work. Maybe you should discuss this with @wkerzendorf

tardis/plasma/tests/test_complete_plasmas.py
vg3095: @yeganer  `j_blues, beta_sobolev, lines_upper_level_index, lines_lower_level_index,
t_rad, w, t_electrons, g_electron, beta_rad, stimulated_emission_factor`, these plasma quantities are `Numpy arrays`.

vg3095:
@wkerzendorf @yeganer According to what I understand, these combinations remain uncovered according to coverage report,
When , 
`helium_treatment` is `recomb-nlte` or `numerical-nlte`
`NLTE` species has `coronal_approximation` or `classical_nebular`
`Radiative Rates type` is `blackbody` or `detailed` , 

So, I think, I will create new Plasma configs, using these options, to increase Plasma coverage.

I have doubt regarding testing Plasma using `numerical-nlte` option, as according to what I see in the code, it requires an external `heating rate data file`, and a `specific atomic dataset` . So, I don't know , how to test this.

Regarding to what @yeganer suggested about using `kurucz_atoms_chianti_many.h5` dataset for creating plasma reference files , I am getting a ` LinAlgError: Singular matrix` error using this dataset, for which I think , yeganer has made a PR. And using 'chianti_h_he_db.h5', it goes in infinite loop.

vg3095:
I will try to update this PR by tomorrow evening, with yeganer suggestions and with new plasma configs with the options I said above . My tests are also going on in college.

tardis/plasma/tests/test_complete_plasmas.py
yeganer: I agree, they are numpy arrays. However, I think we should mirror the way these arrays are transformed into series and then compare the resulting series objects.

yeganer:
@vg3095 So far my PRs didn't fix the `LinAlgError`. They only made it possible to run NLTE config files with the current tardis version.

`chianti_h_he_db.h5` should not end up in an infinite loop. It looks like something in that setup went wrong.

yeganer:
@chvogl Do you have a heating rate data file that can be used for one of the configurations in the plasma tests?

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,136 @@
...
+            actual = pd.Series(actual)
+        else:
+            actual = pd.DataFrame(actual)
+        expected = reference.select(os.path.join('plasma', attr))
+        pdt.assert_almost_equal(actual, expected)
+
+    def test_zeta_data(self, plasma, reference):
+        actual = getattr(plasma, 'zeta_data')
+        expected = reference.select(os.path.join('plasma', 'zeta_data'))
+        assert_almost_equal(actual, expected.values)
```
vg3095: @yeganer If I do `pdt.assert_almost_equal(actual, expected)`, I am getting this error
image

I tried these options for changing `dtype` of columns for `actual` zeta data dataframe.
- `actual = actual.apply(pd.to_numeric, errors='coerce')`
- `actual = actual.convert_objects(convert_numeric=True)`
- `actual = actual.astype(np.int64)`

And, None of the above options are working for me, I want to convert `dtype` of `actual` zeta data attribute from `object` to `int64`, to make assertion possible here .

vg3095:
@yeganer Have you seen that jupyter notebook?. I am not able to run that type of plasma configs, without it I will not be able to generate plasma reference for these tests . If, I am making config in the wrong way, then please let me know.

wkerzendorf:
@yeganer has a great idea to try to get the coverage up to 100%. I'll let him work with you on that.

tardis/plasma/tests/test_complete_plasmas.py
yeganer: I'll look into that. This might be fixed with the carsus databases

yeganer:
@vg3095 Let me very quickly explain my idea for testing all setups.

The plasma has a number of different settings (ionization, line_interaction_type, etc.). In general we can assume that they are independent of each other. That means, if `ionization = 'nebular'` fails, it will fail independently of the setting of `line_interaction_type`. We can use this to our advantage.
My first idea was to just parametrize all settings as fixtures and then have a 'config' fixture which combines these. Pytest would then automatically test every possible permutation of these settings. However, because we assume they are all independent of each other, we only need to test the diagonale instead of all possible combinations.
My idea on how to do this would be to parametrize a config fixture and the argument is a dictionary of 'overwrites'. That means, we take a very basic configuration (for now I'd suggest the lte one already present in tardis) and then overwrite everything from the dict. Such a dict would then include `{'ionization': 'nebular'}` for example.
In the end, we have a list of overwrite dictionaries as the values for the parametrization. Does that sound sensible?

The only thing we would need to discuss is how to associate reference values to dictionaries. Any suggestions?

vg3095:
@yeganer Okay, I will try this. Regarding setting HDF `reference_file_path` (if I understand correctly), I think, I will set it in the  `config` object itself , and use this value in `plasma` fixture , which I defined in tests.

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -1 +1,139 @@
...
 
+import pandas as pd
+import pytest
+from pandas.util import testing as pdt
+from numpy.testing import assert_almost_equal
+from tardis.io.config_reader import Configuration
+from tardis.simulation import Simulation
+
+
+setupI = {
```
yeganer: I was more thinking about having a list of all options like this:
```
ionization = [
{ 'ionization': 'nebular'},
{ 'ionization': 'lte'},
]
excitation = [
{ 'excitation' : 'lte'},
{ 'excitation': 'dilute-lte'}
]

@pytest.fixture(params=ionization+excitation)
def config(...):
    for k, v in ...params.items():
        config.plasma[k] = v
    # rest of the code
```

yeganer: That way we test each option individually which makes it easier to see which option causes a failure.

I'm not sure right now but I think we might do this seperately for nlte and lte configurations but that should not be your concern. Your main goal should be to design this system in  such a way that we can easily add the rest.

vg3095: @yeganer Oh, then I cannot think of any way of setting `reference_path`. And, apart from that , saving reference file for all the generated permutations ( even assuming we have 4 options to set(i.e. `ionization`, `excitation`, etc), and having 2 different values that can be set for each option - 2^4 = 16 reference files(the real no. will be much greater than that, considering all options).

vg3095: @yeganer Did you mean , that like for example , `ionization` : `nlte` , that the values will be same, irrespective of other plasma config options?

yeganer: We don't need to test all permutations. That would be too much reference data. But we can test all options 'alone'. I think this would give us ~20 different setups which I think is doable.

yeganer:
For the reference path, one method could be to hash the string representation of the config. The result should be unique and one could use this as the identification of the reference data set.
Additionally I think, we should store ALL data in the same HDF store and use the hash as the key in the HDFStore (instead of saving to `/plasma/`)

wkerzendorf:
@vg3095 that looks great - can you tell us what the test percentage is for plasma - and what and if things are missing.

tardis/io/util.py
```diff
@@ -282,6 +282,8 @@ def get_properties(self):
         data['atom_data_uuid'] = self.atomic_data.uuid1
         if 'atomic_data' in data:
             data.pop('atomic_data')
+        if 'nlte_data' in data:
```
wkerzendorf: can you add a warning here that `nlte_data` can not be saved.

wkerzendorf:
@vg3095 I wrote some comments on gitter.

PR #779, opened 2017-08-16
Save coverage report when tests use tardis-refdata

vg3095:
Currently coverage report is saved, when tests run without `tardis-refdata` parameter. So, the tests which depend on `tardis-refdata` or `atomic-dataset` were skipped, and files which depend on these, are not included in coverage. This PR aims to change that, as many tests , such as tests for Plasma, depend on this parameter(`tardis-refdata`), and coverage report does not correctly reflect the status for same.


vg3095:
Coverage increases by `13.6%` using `tardis-refdata` parameter.

PR #781, opened 2017-08-23
Omit coverage of test related files

vg3095:


wkerzendorf:
@vg3095 can you also exclude `tardis/gui*`, `tardis/stats/*`, `tardis/analysis.py` all `version.py`, `astropy_init` thanks!

PR #786, opened 2017-08-25
The final carsus merge

yeganer:
This PR replaces #770


docs/physics/plasma/plasma_plots/lte_ionization_balance.py
```diff
@@ -1,7 +1,8 @@
 import os
 from matplotlib import pyplot as plt
 from matplotlib import colors
-from tardis import atomic, plasma_array, util
+from tardis import plasma_array, util
```
yeganer: I should fix that, right?

docs/physics/plasma/plasma_plots/nebular_ionization_balance.py
```diff
@@ -1,7 +1,8 @@
 import os
 
 from matplotlib import colors
-from tardis import atomic, plasma_array, util
+from tardis import plasma_array, util
```
yeganer: Same as above, there is no `plasma_array`

docs/physics/plasma/plasma_plots/lte_ionization_balance.py
wkerzendorf: probably

docs/physics/plasma/plasma_plots/nebular_ionization_balance.py
wkerzendorf: yep

tardis/plasma/properties/partition_function.py
```diff
@@ -162,14 +162,14 @@ def _main_nlte_calculation(
                     (number_of_levels**2, len(t_electrons)))
             r_lu_matrix_reshaped[r_lu_index] = B_lus[np.newaxis].T * \
                 j_blues_filtered * beta_sobolevs_filtered
-            if atomic_data.has_collision_data:
+            if atomic_data.collision_data is not None:
                 if previous_electron_densities is None:
                     collision_matrix = r_ul_matrix.copy()
```
wkerzendorf: `collision_matrix = np.zeros_like(r_ul_matrix)`

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -54,21 +55,46 @@
     {'helium_treatment': 'recomb-nlte', 'delta_treatment': 0.5}
 ]
 
+config_list = (
+        ionization + excitation + radiative_rates_type +
+        line_interaction_type + disable_electron_scattering + nlte +
+        initial_t_inner + initial_t_rad + helium_treatment)
+
+
+def idfn(fixture_value):
```
wkerzendorf: can you quickly document this

yeganer:
My verification show that the `macroatom` scheme does not work right now. These are three plots with current master and this PR for `macroatom, downbranch and scatter`.
As you can see, for `macroatom` there are clear differences while `downbranch` and `scatter` are the same.
Running my NLTE tests with `macroatom` results in an infinite loop which further points towards a problem with the macro atom. I'll have a closer look tomorrow.

example
example_downbranch
example_scatter

yeganer:
Found the bug, nevermind.

yeganer:
After fixing the bug, all tests involving `kurucz_cd23_chianti_h_he.h5` produce near identical results, for example here a setup from @talytha :
tomography

However my NLTE test setup (with no NLTE species) does have a slightly different spectrum:
nlte
 
@wkerzendorf How should we proceed?

yeganer:
@wkerzendorf Tests are passing. Looks like this might be ready to merge.

PR #791, opened 2017-10-06
Propagate changes to the line index into the GUI code

### LINKED ISSUE #790: Last interaction info missing
Hi,
a new error appeared with the latest development version: the GUI starts well, but the widget with the last interaction information doesn't open. Error message is attached below.
There was a similar bug earlier, which is connected to the change from model refrence from Simulation model reference (Issue 690 & 691 & 720), but this error message seems somehow different. 

...
  File "/home/barna/anaconda3/envs/tardis2/lib/python2.7/site-packages/tardis/analysis.py", line 104, in update_last_interaction_filter
    'level_number_upper']].ix[last_line_in_count.index]
...
KeyError: "['atomic_number' 'ion_number' 'level_number_lower' 'level_number_upper'] not in index"

yeganer:
This should fix #790

### DIFF
--- a/tardis/analysis.py
+++ b/tardis/analysis.py
@@ -89,25 +89,29 @@ class LastLineInteraction(object):
         self.last_line_out = self.lines.iloc[self.last_line_interaction_out_id[packet_filter]]
 
         if self.atomic_number is not None:
-            self.last_line_in = self.last_line_in[self.last_line_in.atomic_number == self.atomic_number]
-            self.last_line_out = self.last_line_out[self.last_line_out.atomic_number == self.atomic_number]
+            self.last_line_in = self.last_line_in.xs(self.atomic_number, level='atomic_number', drop_level=False)
+            self.last_line_out = self.last_line_out.xs(self.atomic_number, level='atomic_number', drop_level=False)
 
         if self.ion_number is not None:
-            self.last_line_in = self.last_line_in[self.last_line_in.ion_number == self.ion_number]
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@yeganer thank you for your help. I think it is easier if we simply reindex within the gui to line_id as this is specifically the index that is being saved.

yeganer:
I thought about that. Both indices have their benefits in my opinion. Using `atom_data, ion_number, ...` makes several filtering operations a lot easier and straight forward.

unoebauer:
Hi @yeganer - just tested these changes with tardis_example:

```python
config = tardis.yaml_load("tardis_example_new.yml")
mdl = tardis.run_tardis(config)
from tardis.gui import interface
interface.show(mdl)
```
After the gui fires up, I try to show the line info in some wavelength window and it still crashes with the same error.

unoebauer:
The minimal change to fix this error is (and I tested this!) to do
```python
self.lines = lines.reset_index().set_index("line_id")
```
in ``tardis/analysis.py`` on line 27 and

```python
self.lines_data = lines_data.reset_index().set_index('line_id')
```
in ``tardis/gui/widgets.py`` line 962.

yeganer:
@unoebauer This is very strange because it works on my end. If the error is the same, then it looks like the changes made by this PR were not applied.

unoebauer:
@yeganer - turns out that I still had pandas 0.19.1 installed. Updated to 0.20 (as requested in the conda env file) and everything works.

PR #799, opened 2017-10-15
 tardis/plasma/standard_plasmas.py: old database files warning

laudb:
displays a warning when using old atomic database files

### DIFF
--- a/tardis/plasma/standard_plasmas.py
+++ b/tardis/plasma/standard_plasmas.py
@@ -76,7 +76,7 @@ def assemble_plasma(config, model, atom_data=None):
 
         try:
             atom_data = atomic.AtomData.from_hdf(atom_data_fname)
-        except:
+        except TypeError:
             raise ValueError('Error might be from the use of an old-format of the atomic database,\n please see '+ \
             'https://github.com/tardis-sn/tardis-refdata/tree/master/atom_data for the most recent version.')
 
--- a/tardis/plasma/standard_plasmas.py
+++ b/tardis/plasma/standard_plasmas.py
@@ -73,10 +73,14 @@ def assemble_plasma(config, model, atom_data=None):
             raise ValueError('No atom_data option found in the configuration.')
 
         logger.info('Reading Atomic Data from %s', atom_data_fname)
+
         try:
             atom_data = atomic.AtomData.from_hdf(atom_data_fname)
-        except TypeError:
-            logger.warn('Use new configuration for the atomic database')
+        except:
+            logger.warn(
+            'Error might be from the use of an old-format of the atomic database,\n please refer to '+ \
+            'https://github.com/tardis-sn/tardis-refdata/tree/master/atom_data for the most recent version.')
+            raise
 
     atom_data.prepare_atom_data(
         model.abundance.index,
--- a/tardis/plasma/standard_plasmas.py
+++ b/tardis/plasma/standard_plasmas.py
@@ -77,8 +77,8 @@ def assemble_plasma(config, model, atom_data=None):
         try:
             atom_data = atomic.AtomData.from_hdf(atom_data_fname)
         except TypeError as e:
-            print (e, 'Error might be from the use of an old-format of the atomic database, \n' +\
-                'please see https://github.com/tardis-sn/tardis-refdata/tree/master/atom_data' +\
+            print (e, 'Error might be from the use of an old-format of the atomic database, \n' \
+                'please see https://github.com/tardis-sn/tardis-refdata/tree/master/atom_data' \
                 ',for the most recent version.')
             raise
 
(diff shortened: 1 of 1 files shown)


tardis/plasma/standard_plasmas.py
```diff
@@ -73,7 +73,10 @@ def assemble_plasma(config, model, atom_data=None):
             raise ValueError('No atom_data option found in the configuration.')
 
         logger.info('Reading Atomic Data from %s', atom_data_fname)
-        atom_data = atomic.AtomData.from_hdf(atom_data_fname)
+        try:
+            atom_data = atomic.AtomData.from_hdf(atom_data_fname)
+        except TypeError:
+            logger.warn('Use new configuration for the atomic database')
```
wkerzendorf: @laudb you need to reraise the exception because it doesn't work with the old dataset. In this reraised exception you should suggest that the error likely occured due to the use of an old-format atomic database. Then also point to where to download a new one.

laudb: @wkerzendorf okay, I think I understand what you mean, will make changes to the exception and modify the logged message

wkerzendorf:
retainer the exception and put the error message in the exception not in the logger

wkerzendorf:
Sorry reraise

laudb:
@wkerzendorf grateful for the pointers. I made the necessary change, will it suffice?

tardis/plasma/standard_plasmas.py
```diff
@@ -73,7 +73,12 @@ def assemble_plasma(config, model, atom_data=None):
             raise ValueError('No atom_data option found in the configuration.')
 
         logger.info('Reading Atomic Data from %s', atom_data_fname)
-        atom_data = atomic.AtomData.from_hdf(atom_data_fname)
+
+        try:
+            atom_data = atomic.AtomData.from_hdf(atom_data_fname)
+        except:
```
wkerzendorf: you should never have blanket exceptions. Always catch specific exceptions - you want your code to fail if it hits an unknown problem.

laudb: @wkerzendorf okay, noted. I've made this change in the next update.

laudb:
@wkerzendorf  I've been looking at this some more, and I'm also considering this possibility:
`
try:
    atom_data = atomic.AtomData.from_hdf(atom_data_fname) # read data from 

except TypeError: #  the error due to old file shows as TypeError, hence TypeError exception

    # suggests possible problem and solution, "raise TypeError ('message')" hides cause of error
    print("Error might be from the use of an old-format of the atomic database,\n please see '+ \
            'https://github.com/tardis-sn/tardis-refdata/tree/master/atom_data for the most recent version.")

    # still shows what caused the error 
    raise 
`

wkerzendorf:
@laudb you can catch the error and access the message. So catch the error - display the message and add that this error might be due to the wrong dataset. I'm being very particular there because I want you to understand the error handling better. I know it's difficult and you're doing well - push on :wink:

tardis/plasma/standard_plasmas.py
```diff
@@ -76,9 +76,12 @@ def assemble_plasma(config, model, atom_data=None):
 
         try:
             atom_data = atomic.AtomData.from_hdf(atom_data_fname)
-        except TypeError:
-            raise ValueError('Error might be from the use of an old-format of the atomic database,\n please see '+ \
-            'https://github.com/tardis-sn/tardis-refdata/tree/master/atom_data for the most recent version.')
+        except TypeError as e:
+            print (e, 'Error might be from the use of an old-format of the atomic database, \n' +\
```
wkerzendorf: you don't need the plus. every string in parenthesis is automatically concatenated.

laudb: Oh I see, noted, thanks!

tardis/plasma/standard_plasmas.py
```diff
@@ -73,7 +73,15 @@ def assemble_plasma(config, model, atom_data=None):
             raise ValueError('No atom_data option found in the configuration.')
 
         logger.info('Reading Atomic Data from %s', atom_data_fname)
-        atom_data = atomic.AtomData.from_hdf(atom_data_fname)
+
+        try:
+            atom_data = atomic.AtomData.from_hdf(atom_data_fname)
+        except TypeError as e:
```
wkerzendorf: and no `\` 😉

laudb: okay, I'll have another go.. so catch the error, access the message and display it with the helpful suggestion or possible reason

laudb: oh..now I see

wkerzendorf:
@tardis-sn/tardis-core please have a look at this - I'm fine with merging

laudb:
awesome!

PR #801, opened 2017-11-10
Decay of isotopes branch

wkerzendorf:
This is the Isotopic decay branch of @vg3095 that unfortunately never got merged into TARDIS. 
@tardis-sn/tardis-core please take a look.

PR #802, opened 2017-11-10
Fix for PR #801

vg3095:


tardis/plasma/properties/ion_population.py
```diff
@@ -349,20 +349,4 @@ def calculate(self, phi, partition_function, number_density,
                                                       1].sum(axis=0))
             ion_number_density.ix[2].ix[2].update(helium_population_updated.ix[
                                                       2].ix[0])
-            ion_numbers = ion_number_density.index.get_level_values(1).values
```
vg3095: I think this code block is not required when `electron_densities` are read from a seperate file in `IonNumberDensityHeNLTE` class.

wkerzendorf: @vg3095 well often it isn't - we need to be careful here. It needs to calculate them if they are not given. They will often not be given.

vg3095: @wkerzendorf  
See this [line](https://github.com/vg3095/tardis/blob/ac8e67105baa1667b8ab6f8fe74d8dad492a9777/tardis/plasma/properties/ion_population.py#L321), and [here](https://github.com/vg3095/tardis/blob/ac8e67105baa1667b8ab6f8fe74d8dad492a9777/tardis/plasma/properties/ion_population.py#L350), I have removed this code block  from second link (from `else` part --> if `electron_densities` is present).

PR #809, opened 2018-03-16
fix dependencies for tardis conda env

wkerzendorf:
I found today when trying to make a new TARDIS env that there were dependency issues. I've changed the dependencies slightly in the conda environment file.


wkerzendorf:
@tardis-sn/core can someone review and merge?

wkerzendorf:
@yeganer I'm at a loss why it doesn't build - do you have any idea?

Issue #695, opened 2017-01-20
Unused configuration properties in configuration


ftsamis:
Currently the following `convergence_strategy` properties are never used in the code:

When type is `specific`:

- `convergence_strategy.threshold`
- `convergence_strategy.fraction`

When type is either `damped` or `specific`:

- ` convergence_strategy.lock_t_inner_cycles`
- `convergence_strategy.damping_constant`

We should either remove them or use them (probably the first one for threshold and damping_constant since they don't make much sense).

mohit9727:
@ftsamis  I would like to contribute to this, can you tell me where to begin?

ftsamis:
It is not yet clear if we should remove those properties or try to use them, so this issue is currently under discussion. I'd suggest picking another issue to work on.

unoebauer:
What should we do about this, @wkerzendorf, @yeganer? This seems to be an almost "cosmetic" change. Should we add this to 2.0 or abandon it?

unoebauer:
When simplifying the convergence treatment to only one strategy, ``damped``, in PR #821 this issue has been addressed.

PR #821, opened 2018-04-11
Simplify convergence strategies

unoebauer:
Currently, there is very little difference between the two supported convergence strategies, ``damped`` and ``specific``. Also, as noted in issue #695, a number of properties of the ``convergence_strategy`` are unused. This PR tries to simply the convergence strategy handling by

- merging the two existing strategies into one, which will be called ``damped`` (the name describes the current strategy to update the plasma variables best)
- introduce support for currently unused properties, in particular for ``lock_t_inner_cycles``, ``t_inner_update_exponent``
- remove remaining unused properties
- introduce new property that controls whether the simulation stops before the specified number of iterations if the plasma quantities converged or not

### DIFF
--- a/tardis/io/config_reader.py
+++ b/tardis/io/config_reader.py
@@ -30,9 +30,7 @@ def parse_convergence_section(convergence_section_dict):
         dictionary
     """
 
-
-    convergence_parameters = ['damping_constant', 'threshold', 'fraction',
-            'hold_iterations']
+    convergence_parameters = ['damping_constant', 'threshold']
 
     for convergence_variable in ['t_inner', 't_rad', 'w']:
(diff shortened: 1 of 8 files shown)


unoebauer:
@wkerzendorf - what do you think? I still have to check the different tests and probably need to update some reference data...

wkerzendorf:
@unoebauer In general, I agree - it means that we will break backward compatibility (which we have done in the past and which I'm in favor off).

unoebauer:
True - but one only has to change the type from ``specific`` to ``damped``. All other options are still valid under the updated convergence strategy handling

unoebauer:
Ready to be merged from my point of view.

PR #822, opened 2018-04-12
add osx back in - change cython to pip installation

wkerzendorf:
This was suggested by @scoder in cython/cython#2199. @unoebauer that might fix our problem


unoebauer:
Fixed a typo in the requirements file (``=`` instead of ``==``)

Issue #784, opened 2017-08-24
Differences between NLTE calculation results


yeganer:
@chvogl Brought to my attention that the NLTE result I obtained last week does not match the spectrum in figure 10 of the KerzendorfSim2014 paper very well near 3000 angstrom. Today's results don't have that problem, however the region between 4000 and 5000 angstrom does not match as well as before.

The goal of this issue is to document this behavior and provide a centralized platform for discussion, especially whether we need to look more closely at the underlying problem.

I believe this might be related to convergence criteria but I haven't confirmed anything yet.
Here are the plots (please note that for some reason the colors are reversed in both plots):

Carsus:
![nlte_carsus](https://user-images.githubusercontent.com/6164110/29669441-bd24adb6-88e3-11e7-865d-0fdea9015c13.png)

Last week after fixing NLTE:
![nlte_verification](https://user-images.githubusercontent.com/6164110/29669442-bd3d8a16-88e3-11e7-92cb-bae944399a9c.png)

livnehra:
Hi,

I'm getting some weird results with NLTE.
I tried to follow the code, and found that in `partition_function.py` the data is prepared in `r_ul_matrix_reshaped`, but then the rates matrix is built using:
`rates_matrix = r_lu_matrix + r_ul_matrix + collision_matrix`
without using `r_ul_matrix_reshaped`.

Am I missing something?

Thanks,
Ran

yeganer:
@livnehra Hey, I'm currently not able to test this but this might be working. If it's a bug or not depends on whether `r_ul_matrix_reshaped` owns the data or not. AFAIK a reshape of a numpy array only create a different view of the same memory location. That means we use the reshape to more easily assign the data but when we use the data to create the `rates_matrix` the original view with the updated data is used (because both views point to the same memory). Please correct me when I'm wrong but I assume that's what's going on.

Nice catch! This part of the code could be written differently to better reflect what is going on. I hope this explanation helped.

livnehra:
@yeganer Hi, of course you are correct and this is not a bug.

After some more digging, I think I may have found the issue - 
I noticed that turning on NLTE for Si II changed the **ion populations** significantly (5x). 
I made sure this was not due to temperature, n_electrons, W or other plausible effects.

Finally I noticed that this is what happens:
1. The NLTE code computes a new `level_boltzmann_factor` for Si II.
2. The `partition_function` for Si II is updated based on these (and always sums to 1.0).
[...]

PR #831, opened 2018-06-19
Fix NLTE Normalization - issue #784

livnehra:
Fix NLTE normalization according to @ssim suggestion in the discussion in #784

### DIFF
--- a/tardis/io/atomic.py
+++ b/tardis/io/atomic.py
@@ -442,7 +442,7 @@ class NLTEData(object):
                 C_ul_matrix[level_number_lower, level_number_upper, :] = line.values[2:]
                 delta_E_matrix[level_number_lower, level_number_upper] = line['delta_e']
                 #TODO TARDISATOMIC fix change the g_ratio to be the otherway round - I flip them now here.
-                g_ratio_matrix[level_number_lower, level_number_upper] = line['g_ratio']
+                g_ratio_matrix[level_number_lower, level_number_upper] = 1/line['g_ratio']
             self.C_ul_interpolator[species] = interpolate.interp1d(
                     self.atom_data.collision_data_temperatures,
                     C_ul_matrix)
--- a/tardis/plasma/properties/partition_function.py
+++ b/tardis/plasma/properties/partition_function.py
@@ -210,7 +210,7 @@ class LevelBoltzmannFactorNLTE(ProcessingPlasmaProperty):
                 j_blues,
                 beta_sobolevs,
                 general_level_boltzmann_factor,
-                previous_electron_densities)
+                previous_electron_densities, g)
         return general_level_boltzmann_factor
 
     def _calculate_coronal_approximation(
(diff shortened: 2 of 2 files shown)


tardis/plasma/properties/partition_function.py
```diff
@@ -190,7 +190,7 @@ def _main_nlte_calculation(
                     else:
                         raise e
                 general_level_boltzmann_factor[i].ix[species] = \
-                    level_boltzmann_factor
+                    level_boltzmann_factor * g[species][0] / level_boltzmann_factor[0]
```
wkerzendorf: @livnehra can you make this `g.loc[species, 0]`

wkerzendorf:
@livnehra I don't think that should run - but maybe I'm wrong

tardis/plasma/properties/partition_function.py
```diff
@@ -190,7 +190,7 @@ def _main_nlte_calculation(
                     else:
                         raise e
                 general_level_boltzmann_factor[i].ix[species] = \
-                    level_boltzmann_factor * g[species][0] / level_boltzmann_factor[0]
+                    level_boltzmann_factor * g.loc[species][0] / level_boltzmann_factor[0]
```
wkerzendorf: `.loc[species, 0]`

livnehra: That doesn't work...

wkerzendorf: hmm  - that's interesting. I'll check it tomorrow.

wkerzendorf: yes you're right - I screwed things up in my mind. I think the way you do it now is good.

livnehra: Remember, for example species=(14,1) for Si II

wkerzendorf:
@livnehra the tests fail - your change somehow breaks the code - have a look at the continuous integration to see what's wrong.

livnehra:
Hopefully should pass the tests now.

livnehra:
So the fails now are due to NLTE results changing, which makes sense if the previous results were obtained with the erroneous g_ratio_matrices.

wkerzendorf:
@livnehra - okay I'm ready to update the reference data. I still think it would be good to get a spectrum beforehand.

unoebauer:
@wkerzendorf, @livnehra - what's the idea about proceeding with this PR?

tardis/io/atomic.py
```diff
@@ -442,7 +442,7 @@ def _create_collision_coefficient_matrix(self):
                 C_ul_matrix[level_number_lower, level_number_upper, :] = line.values[2:]
                 delta_E_matrix[level_number_lower, level_number_upper] = line['delta_e']
                 #TODO TARDISATOMIC fix change the g_ratio to be the otherway round - I flip them now here.
```
chvogl: It would be possible to remove the TODO now that it has been addressed.

PR #832, opened 2018-06-19
Fixes #830

### LINKED ISSUE #830: Using Tardis CSV input for custom abundances and isotopes
While working with the CSV input option for custom abundance and isotope profiles, I noticed that the relevant parser in ``model_reader.py``, 
```python 
def read_simple_isotope_abundances(fname, delimiter='\s+'):
```
expects that the custom composition file contains information about the ejecta state (velocity, temperature, electron density, ...) in the first four columns.

I assume that this was at some point implemented to work with CMFGEN setups (right @wkerzendorf?), but for most normal use cases these columns are irrelevant. Either the information is not used or specified in the model density profile setup process. In any case, the current documentation does not specify that when using CSV files, four additional columns are needed.

I would suggest that we split the CSV input option into two use cases

- CMFGEN setups, for which keep the current parser
- normal Tardis use cases: for this we should add an additional parser (based on the current one) which does not expect additional ejecta information in the first four columns

In any case, the documentation should be updated to reflect the behaviour of the CSV parser

unoebauer:
This PR implements changes to the parsers of the various composition files to address #830.


unoebauer:
The ``test_model_reader.py`` does fail. Judging from the setup of the tests, the modified parser was indeed intended for CMFGEN setups. 

Before proceeding, we should discuss our options, @wkerzendorf

unoebauer:
I would suggest the following:

```python
def read_density_file(filename, filetype):
    file_parsers = {'artis': read_artis_density,
                    'simple_ascii': read_simple_ascii_density,
                    'cmfgen_model': read_cmfgen_density}
    ...
def read_abundances_file(abundance_filename, abundance_filetype,
    file_parsers = {'simple_ascii': read_simple_ascii_abundances,
                    'artis': read_simple_ascii_abundances,
                    'cmfgen_model': read_cmfgen_composition,
                    'simple_csv': read_csv_composition}

```

```python
def read_csv_isotope_abundance(self, delimiter="\+s", skip_columns=0):
    ...
    for element_symbol_string in df.index[skip_columns:]:
        ...
```

```python
def read_cmfgen_composition(self, delimiter="\+s"):
    return self.read_csv_isotope_abundance(delimiter=delimiter, skip_columns=4)

def read_csv_composition(self, delimiter="\+s"):
    return self.read_csv_isotope_abundance(delimiter=delimiter, skip_columns=0)
```

What do you think, @wkerzendorf ?

unoebauer:
I still think that for now we should go with a quick fix on the lines that I've suggested. The CMFGEN format is quite specific. And standardizing the input is an undertaking that goes beyond the original scope of this PR.

wkerzendorf:
@unoebauer I think we misunderstand each other - the main thing I'm suggesting is to drop the CMFgen name - as it is not true at the moment. I'll try to catch you online to discuss.

wkerzendorf:
@unoebauer very nice PR (feature; docu + tests - dass ich das noch erleben darf). Doesn't pass - you need to add the file in the package template

unoebauer:
I see, I forgo to update the tests in ``test_tardis_model_density_config.py``. I will do that soon.

unoebauer:
Fixed a few outstanding issues - the complete test suite now passes locally. Hopefully Travis is happy now as well.

PR #833, opened 2018-06-21
restructure landing page to use notebooks

wkerzendorf:


wkerzendorf:
@unoebauer I changed everything to toctrees and that also brought down the number of warnings and hopefully allows us to embed notebooks very easily

wkerzendorf:
I'm still working on it - so don't merge now

wkerzendorf:
@unoebauer I would like to merge this before continuing. It's not perfect - but better what was there before (IMHO)

PR #839, opened 2018-07-18
change to new cython version

wkerzendorf:
@unoebauer, @scoder
version 0.28.4 should have the right fix in there. 

cython/cython#2199

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,7 @@ install:
    - source activate tardis
 
 script:
+    - python -c "import Cython; print Cython.__version__"
     - echo CC=$COMPILER python setup.py $SETUP_CMD
     - CC=$COMPILER python setup.py $SETUP_CMD
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,13 +40,14 @@ matrix:
             - TEST_MODE='spectrum'
 
 #trouble with osx building due to segfault at cython (https://github.com/cython/cython/issues/2199)
-        - os: osx
-          language: generic
-          env:
-            - COMPILER=clang
-            - SETUP_CMD='test --args="--tardis-refdata=$HOME/tardis-refdata/"'
-            - TEST_MODE='spectrum'
-            - MINICONDA_URL='http://repo.continuum.io/miniconda/Miniconda2-latest-MacOSX-x86_64.sh'
+# and https://github.com/cython/cython/issues/2505
+#        - os: osx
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@unoebauer - should we merge?

unoebauer:
I think we should - Travis is happy with both Linux and OSX builds, so I don't see a reason why we shouldn't. We might wait for @Heringer-Epson though to check whether this version works on his local installation.

wkerzendorf:
@unoebauer - no I removed the OSX build for now.

wkerzendorf:
I would merge and put a reminder in my phone to check in 3 months to reinclude the OSX build

unoebauer:
Sorry, I missed that the OSX is deactivated. So on OSX the segfault is still triggered? And locally on your OSX build it works?

wkerzendorf:
@unoebauer it doesn't segfault in the debugger even. I think, I will just disable it for now.

Heringer-Epson:
Sorry for the delay. I will check this today.

wkerzendorf:
@Heringer-Epson be sure to delete montecarlo.c (look at it - it will say `generated with Cython VERSION_NUMBER` then rerun the compilation.

.travis.yml
```diff
@@ -87,15 +88,9 @@ install:
    - cd $TRAVIS_BUILD_DIR
    - conda env create -f tardis_env27.yml
    - source activate tardis
-   #trouble with building due to segfault at cython (https://github.com/cython/cython/issues/2199)
-   #remove if we can get normal cython through conda
```
unoebauer: Should we leave this in as a comment? This way, we'll know what to do when some strange cython error appears in the future.

wkerzendorf: That’s why I put it in

unoebauer: I was thinking of keeping the procedure of getting bleeding-edge cython via github and compiling it with custom flags as a comment block and providing some additional information.

Heringer-Epson:
Yes, this worked upon trying to rerun the installation. Note, using the conda env suggestion in tardis_env27.yml triggers an error due to Pyqt not being installed (I believe it's required for the gui interface). Installing Pyqt=4.11 on the local env then caused a bunch of packages to be downgraded.

wkerzendorf:
that's interesting. anyways. I'll merge once I add the manual cython installing instructions in.

PR #843, opened 2018-07-20
Fix issue with custom_composition reader

unoebauer:
Currently, the ``custom_composition`` reader is not properly working since it expected an additional column containing the cell index, similar to the file struture for the ``simple_ascii`` reader. 

This PR addresses this issue by re-inserting the cell index as the first column into the ``custom_composition`` files.

- [x] fix read_csv_isotope_abundances and read_cmfgen_composition
- [x] update tests and reference files
- [x] update documentation


pfreddy:
I changed the line 405 in model_reader.py, it still gives the error message.

unoebauer:
@pfreddy, @wkerzendorf - I think, the PR is ready. It passes our testing suite offline and it create data frames with the correct shape from the test_csv.dat file from @pfreddy. Whenever Tardis is happy and someone reviewed it, we should consider merging.

unoebauer:
Travis is happy!

unoebauer:
Ok, I've revisited the issue and found some indexing problems in the parser which are hopefully fixed now. On my laptop, I can run your setup @pfreddy without problems. Also the testing suite runs through. If Travis is happy, I'll address your stylistic suggestion, @wkerzendorf .

PR #844, opened 2018-07-20
Remove bbsampling from schema file

pfreddy:
This is a short fix for the schema file


wkerzendorf:
Thank you @pfreddy and @unoebauer

PR #846, opened 2018-07-26
update the OPENMP text to be correct

wkerzendorf:
@unoebauer @chvogl can you please check


wkerzendorf:
@orbitfold is that right? @unoebauer did you check?

unoebauer:
Sure, if you compile Tardis with 
```
./setup.py build install --with-openmp
```
you include the OpenMP library and can then run Tardis in parallel mode.

PR #849, opened 2018-07-26
Missing refdata

wkerzendorf:


wkerzendorf:
@unoebauer @chvogl can you sign off?

wkerzendorf:
yep - it crashed. you can see the commit above not passing.

PR #850, opened 2018-07-27
Add full test for formal integral method

unoebauer:
This PR adds another test for the formal integral method, specifically checking the final spectrum determined with this scheme. The setup is very similar to the one used for ``test_tardis_full.py``.

- [x] add new test with modified setup
- [x] test for integrated spectrum
- [x] test formal integral for both macroatom and downbranch modes
- [x] update reference data
- [x] update documentation


unoebauer:
@wkerzendorf - please review (and merge once Travis is happy)

PR #856, opened 2018-07-31
Extend formal integral to line_interaction_type macroatom

chvogl:


tardis/montecarlo/formal_integral.py
```diff
@@ -101,6 +103,29 @@ def make_source_function(self):
...
         runner = self.runner
         atomic_data = self.plasma.atomic_data
+        macro_ref = atomic_data.macro_atom_references
+        macro_data = atomic_data.macro_atom_data
+
+        no_lvls = len(atomic_data.levels)
+        no_shells = len(model.w)
+
+        if runner.line_interaction_type == 'macroatom':
+            ma_int_data = macro_data.query('transition_type >=0')
```
wkerzendorf: never seen query before.

tardis/montecarlo/formal_integral.py
```diff
@@ -101,6 +103,29 @@ def make_source_function(self):
...
+            internal_jump_mask = (macro_data.transition_type >= 0).values
```
wkerzendorf: why do you use both the mask and the query

tardis/montecarlo/formal_integral.py
```diff
@@ -101,6 +103,29 @@ def make_source_function(self):
...
+            internal = plasma.transition_probabilities[internal_jump_mask]
+
+            source_level_index = pd.MultiIndex.from_arrays([
+                    ma_int_data['atomic_number'].values,
+                    ma_int_data['ion_number'].values,
+                    ma_int_data['source_level_number'].values]
+            )
```
wkerzendorf: that exists already in the atomic_data.macro_atom_data. it is called source_level_idx and destination_level_idx (I think the first one might not exits - but we should but it there).

tardis/montecarlo/formal_integral.py
```diff
@@ -118,6 +143,22 @@ def make_source_function(self):
...
         e_dot_lu          = pd.DataFrame(Edotlu, index=upper_level_index)
         e_dot_u           = e_dot_lu.groupby(level=[0, 1, 2]).sum()
+        e_dot_u_src_idx = macro_ref.loc[e_dot_u.index].references_idx.values
+
+        if runner.line_interaction_type == 'macroatom':
+            C_frame = pd.DataFrame(
+                columns=np.arange(no_shells), index=macro_ref.index
+            )
+            q_indices = (source_level_idx, destination_level_idx)
+            for shell in range(no_shells):
```
wkerzendorf: `xrange`

wkerzendorf:
This looks very good. I'm merging

wkerzendorf:
@chvogl

wkerzendorf:
@chvogl there is a comment in atomic about destination_level_idx which should be called reference_idx. Any opinion on that?

chvogl:
@wkerzendorf In my opinion we can remove this comment.

wkerzendorf:
let's do that. it just causes additional confusion. I can make a PR. I also want to log to stdout instead of err - for ease of use in jupyter notebooks and will combine the PRs

PR #859, opened 2018-08-02
Use interpolated values to do formal integral with more shells

chvogl:


wkerzendorf:
@chvogl are the refdata still misbehaving ?

chvogl:
@wkerzendorf Yes, the problem still persists. I have run the tests locally with the same cython that is used on Travis and the tests pass. I am currently at a loss to why it does not work on Travis. It must be related to the interpolation in some way.

wkerzendorf:
@chvogl should we just merge this?

chvogl:
@wkerzendorf In principle, I think we can merge this. Do you think we should resolve #878 first?

PR #881, opened 2019-02-13
Adds description of gcc problem to documentation

ycamacho:


docs/installation.rst
```diff
@@ -100,3 +100,13 @@ Go into the tardis/montecarlo directory and build montecarlo.c by hand::
     cython montecarlo.pyx
 
 Then, ``python setup.py build`` should run without problems.
+
+
+**Problem:** 
+        when trying to set up CC=gcc-5 python setup.py develop --with-openmp the following error popped up:
```
wkerzendorf: can you get rid of the spaces like in the on above?

ycamacho: Done

PR #883, opened 2019-02-21
Update __init__.py

nileshpatra:
added '.' for relative file path , so that there are no conflicts during installation of tardis


wkerzendorf:
I think absolute imports are better - try that.

nileshpatra:
@wkerzendorf , did that for absolute path , import works on my machine . please review

wkerzendorf:
That is not the full absolute path. TARDIS.io.util is

nileshpatra:
Yeah , but when the repository is cloned on the machine , it would take the path for the utility folder in the same directory TARDIS/io where until is present . Is it wrong?😅

wkerzendorf:
that works as long as in your system there is no `util` package. it is safer to spell it out what you mean. but enough with this PR. try to get the code running with the master branch version and get a spectrum out.

PR #871, opened 2018-09-27
Resurrect Integration Tests

unoebauer:
This PR updates various config and python files such that the integration test suite can be run locally. In a next step (future PR), the integration tests will be integrated as a cron job into the Travis continuous integration process.

Verify the following features
- [x] reference generation and validation with few packets
- [ ] reference generation and validation with high statistics
- [x] local reporting
- [ ] (*optional*) remote reporting

PR #887, opened 2019-02-26
Resurrect integration

wkerzendorf:
resurrect of #871


marxwillia:
I don't think this PR is relevant anymore now that we are using Azure.

PR #915, opened 2019-04-12
Parsing STELLA files for TARDIS

sarafina325:
This is a general parser to import STELLA files into TARDIS.


wkerzendorf:
@sarafina325 great first PR.

PR #928, opened 2019-05-01
Added csvy parser

marxwillia:


tardis/io/parsers/csvy.py
```diff
@@ -0,0 +1,21 @@
...
+        yaml_end_ind = -1
+        for i, line in enumerate(fh):
+            if i == 0:
+                assert line.strip() == YAML_DELIMITER, 'First line of csvy file is not \'---\''
+            yaml_lines.append(line)
+            if i > 0 and line.strip() == YAML_DELIMITER:
+                yaml_end_ind = i
+                break
+        else:
+            raise Exception('End YAML_DELIMITER not found')
```
wkerzendorf: try to not raise blanket Exception state what it is like `ValueError` or something like this.

tardis/io/parsers/csvy.py
```diff
@@ -0,0 +1,21 @@
+import yaml
+import pandas as pd
+
+YAML_DELIMITER = '---'
+def load_csvy(fname):
```
wkerzendorf: please add a documentation stub - using the numpydoc format. Pycharm can help you with that.

tardis/io/parsers/csvy_ex
```diff
@@ -0,0 +1,37 @@
+---
+profile: tabular-data-resource
+name: my-dataset
+path: https://raw.githubusercontent.com/csvy/csvy.github.io/master/examples/example.csvy
+title: Example file of csvy 
+description: Show a csvy sample file.
+format: csvy
+mediatype: text/vnd.yaml
```
wkerzendorf: what is that?

wkerzendorf:
Looks great @marxwillia! We need to use our special yaml parser to make sure it converts quantities strings to quantities - but we can work on this the next few days.

marxwillia:
Just updated the csvy_model.yml schema file so that it includes the builtin options for density profiles and uniform abundance.  I think we want the user to be able to run TARDIS this way in addition to the new flexibility of the csvy model file.

In the density and abundance definitions, the user can choose 'csvy' which has no subfields.  My current plan is to make the user responsible for setting up the csv part of the model csvy file correctly, with help from some asserts.

tardis/io/parsers/csvy.py
```diff
@@ -0,0 +1,45 @@
...
+        yaml_end_ind = -1
+        for i, line in enumerate(fh):
+            if i == 0:
+                assert line.strip() == YAML_DELIMITER, 'First line of csvy file is not \'---\''
+            yaml_lines.append(line)
+            if i > 0 and line.strip() == YAML_DELIMITER:
+                yaml_end_ind = i
+                break
+        else:
+            raise ValueError('End YAML_DELIMITER not found')
```
wkerzendorf: **maybe** you meant to do something else

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,147 @@
+$schema: http://json-schema.org/draft-04/schema#
+type: object
+additionalProperties: false
+properties:
+  name: 
+    type: string
+    description: name of the model being run
```
wkerzendorf: maybe we do not want to require them

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,147 @@
...
+
+  description:
+    type: string
+    description: description of the model being run
+
+  tardis_model_config_version:
+    type: string
+    description: Version of the configuration file
+
+  atom_data:
```
wkerzendorf: this should not be there.

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,147 @@
...
+    type: string
+    description: path to the atom data
+
+  datatype:
```
wkerzendorf: why do we call it 'dataype' - please add comment

wkerzendorf: this is not required

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,147 @@
...
+    - $ref: '#/definitions/density/power_law'
+    - $ref: '#/definitions/density/uniform'
+
+  abundance:
+    $ref: '#/definitions/abundance/uniform'
+
+  required:
+  - name
+  - tardis_config_version
+  - atom_data
```
wkerzendorf: that is not required

tardis/io/tests/csvy_test.ipynb
```diff
@@ -0,0 +1,454 @@
+{
```
wkerzendorf: maybe this should go to the documentation?

tardis/io/tests/data/csvy_full.csvy
```diff
@@ -0,0 +1,24 @@
+---
+name: csvy_full
+description: Example csvy config file for TARDIS.
+tardis_model_config_version: v1.0
+atom_data: path/to/atom_data
+datatype:
+  fields:
+    -  name: velocity
+       unit: km/s
+       desc: velocities of shell outer bounderies.
```
wkerzendorf: this is not required, right?

tardis/io/tests/test_csvy_reader.py
```diff
@@ -0,0 +1,27 @@
+from tardis.io.parsers import csvy
+import pytest
+import os
+from astropy import units as u
+import numpy.testing as npt
+
+test_data_directory = os.path.dirname(__file__)
```
wkerzendorf: this is a constant

tardis/io/tests/test_csvy_reader.py
```diff
@@ -0,0 +1,27 @@
...
+
+def data_path(filename):
```
wkerzendorf: this is a fixture and probably should be treated like one.

tardis/io/tests/test_csvy_reader.py
wkerzendorf: this is defined twice

tardis/io/tests/test_csvy_reader.py
wkerzendorf: and defined here

tardis/io/tests/test_csvy_reader.py
```diff
@@ -0,0 +1,27 @@
...
+    data_dir = os.path.dirname(__file__)
+    return os.path.join(data_dir, 'data', filename)
+
+def test_csvy_finds_csv_first_line():
+    yaml_dict, csv = csvy.load_csvy(data_path('csvy_full.csvy'))
+    assert csv['velocity'][0] == 10000
```
wkerzendorf: `assert_almost_equal`

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,147 @@
...
+    - $ref: '#/definitions/density/branch85_w7'
+    - $ref: '#/definitions/density/exponential'
+    - $ref: '#/definitions/density/power_law'
+    - $ref: '#/definitions/density/uniform'
+
+  abundance:
+    $ref: '#/definitions/abundance/uniform'
+
+  required:
+  - name
```
chvogl: Do we want to require a name for the model?

tardis/io/tests/data/csvy_full.csvy
```diff
@@ -0,0 +1,24 @@
+---
```
chvogl: Is there a specific reason to have this at the beginning of the file?

tardis/io/parsers/csvy_ex
marxwillia: This file is removed.

tardis/io/schemas/csvy_model.yml
marxwillia: Name is no longer a required property.

tardis/io/schemas/csvy_model.yml
marxwillia: Added comment: "datatype name used for consistency with astropy ecsv files."

tardis/io/tests/data/csvy_full.csvy
marxwillia: No desc is not required.  It would have to be explicitly listed as a required property.  I did a quick test to verify.

tardis/io/tests/data/csvy_full.csvy
marxwillia: The astropy ecsv header begins with "---".  That being said we are definitely making some different choices already from the ecsv header design, so I would be open to not having the initial YAML_DELIMITER.

tardis/io/tests/test_csvy_reader.py
marxwillia: Fixed to use pytest.fixture

setup.py
```diff
@@ -113,6 +113,7 @@
 package_info['package_data'].setdefault(PACKAGENAME, [])
 package_info['package_data'][PACKAGENAME].append('data/*')
 package_info['package_data'][PACKAGENAME].append('io/schemas/*')
+package_info['package_data'][PACKAGENAME].append('io/tests/data/*')
```
wkerzendorf: use this `https://github.com/tardis-sn/tardis/blob/master/tardis/io/setup_package.py` for adding that.

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,144 @@
+$schema: http://json-schema.org/draft-04/schema#
+type: object
+additionalProperties: false
+properties:
+  name: 
+    type: string
+    description: name of the model being run
+
+  description:
```
wkerzendorf: are these required keys?

tardis/io/schemas/csvy_model.yml
```diff
@@ -0,0 +1,144 @@
...
+    type: string
+    description: description of the model being run
+
+  tardis_model_config_version:
```
wkerzendorf: that should be required

tardis/io/schemas/csvy_model.yml
wkerzendorf: nope - we should not. can the required key live within the node

tardis/io/tests/data/csvy_full.csvy
wkerzendorf: I think its fine.

tardis/io/tests/test_csvy_reader.py
```diff
@@ -0,0 +1,33 @@
+import tardis
+from tardis.io.parsers import csvy
+import pytest
+import os
+from astropy import units as u
+import numpy.testing as npt
+
+
+DATA_PATH = os.path.join(tardis.__path__[0], 'io', 'tests', 'data')
```
wkerzendorf: could be a fixture as well.

tardis/io/schemas/csvy_model.yml
marxwillia: I'm pretty sure that the required key cannot live within the node.

https://json-schema.org/understanding-json-schema/reference/object.html#required

tardis/io/schemas/csvy_model.yml
marxwillia: tardis_model_config_version is now the only property required.

tardis/io/schemas/csvy_model.yml
marxwillia: Not currently.

PR #930, opened 2019-05-10
Fix for conda-forge package

epassaro:
Fix suggested at astropy/astropy-helpers#221 and used in astroquery's MANIFEST.in file.


meta.yaml
```diff
@@ -0,0 +1,98 @@
...
+    git_url: https://github.com/tardis-sn/tardis.git
+    git_branch: master
+
+build:
+    script: python setup.py install --single-version-externally-managed --record record.txt
+
+requirements:
+    host:
+        - setuptools
+        - python=3
```
wkerzendorf: this is bad - we shouldn't manually put this in here.

epassaro: My bad, this is old and unnecessary file. Should be discarded.

PR #932, opened 2019-05-15
Pipeline

Youssef15015:
Renamed and moved pipeline to PiplineA and pipelines directory.
Next step is to identify the pipeline on Azure Devops, which requires a merge first.


wkerzendorf:
@Youssef15015 - please make a new PR that does not have 20 commits for moving one file 😉

pipelines/pipelineA.yml
```diff
@@ -37,6 +37,10 @@ jobs:
           echo "##vso[task.prependpath]$CONDA/bin"
     displayName: Add conda to PATH
 
+  - bash: |
```
wkerzendorf: why is this in there?

Youssef15015: That is a good question.

PR #934, opened 2019-05-21
Relocating pipeline yml files

Youssef15015:
Relocated the azure_pipelines.yml into a new directory called pipelines, and renamed it as pipeline_A


wkerzendorf:
the directory should be called azure-pipelines. Can you also give the pipeline a more descriptive name.

Youssef15015:
Will do, and I'll have to think about the name for a minute.

PR #937, opened 2019-06-10
Deploying docuemntation using Azure

Youssef15015:
Update deployment of documentation through the deploy_docs shell file, using Sphinx instead of doctr, to push to github, and add the azure pipeline yaml file that runs it.


azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,99 @@
...
+    vmImage: $[variables.vm_Image]
+
+#### Basic setup of our virtual machine(vm)
+  strategy:
+    matrix:
+      linux:
+        vm_Image: 'Ubuntu-16.04'
+        conda: '/usr/share/miniconda'
+      mac:
+        vm_Image: 'macOS-10.13'
```
wkerzendorf: i don't think we need to build this twice, right?

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,99 @@
...
+    inputs:
+     hostName: $(gh_host)
+     sshPublicKey: $(public_key)
+     #sshPassphrase: $(Agent.TempDirectory)  # Optional - if it was left empty while generating the key.
+     sshKeySecureFile: 'id_azure_rsa'
+ # You can mask hostName & sshPublicKey values under secret variables (i.e. gh_host & public_key here) in your azure pipeline page.
+ # Use steps at: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables
+
+  - bash: |
+          ls
```
wkerzendorf: do we need this step?

deploy_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+# sphinx-apidoc -f -M -o docs/source/ .
+
+# # Add new docs to the repo
+# git add docs/source --all
+
+# # Commit and push to SOURCE_BRANCH
+# git commit -m "Regenerate API Documentation: ${SHA}" || true
+# git push $SSH_REPO $SOURCE_BRANCH || true
+
+git checkout $TARGET_BRANCH || git checkout --orphan $TARGET_BRANCH
```
wkerzendorf: what does `orphan` do?

deploy_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+cd ..
+
+# Clean out existing contents
+rm -rf out/docs/** || exit 0
+
+# Pull from SOURCE_BRANCH again
+git pull || true
```
wkerzendorf: why `||true`

deploy_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+# Now let's go have some fun with the cloned repo
+cd out
+git config --local user.name "Azure Pipelines"
+git config --local user.email "azuredevops@microsoft.com"
+
+echo "doing git add/commit/push"
+
+# Commit the "changes", i.e. the new version.
+# The delta will show diffs between new and old versions.
+git add --all
```
wkerzendorf: are the files then not in the `out/` directory?

wkerzendorf:
@Youssef15015 I guess you can also disable travis - right?

azure-pipelines/doc-build.yml
Youssef15015: I agree!

azure-pipelines/doc-build.yml
Youssef15015: Nope, I forgot about that.

deploy_docs.sh
Youssef15015: Orphan creates a new branch. This was originally Alice's code. So I suspect it is a fail safe, just incase you do not have gh-pages locally as a branch?

Youssef15015: A temporary new branch

deploy_docs.sh
Youssef15015: This was also from Alice's script that I did not question, but it explains it here: https://unix.stackexchange.com/questions/325705/why-is-pattern-command-true-useful

So that the code doesn't exit if it cannot "Pull from Source_Branch" again. I guess it would be better to not let the code run, instead of building the docs based on the gh-pages branch?
Is this something I should doublecheck with Alice?

deploy_docs.sh
Youssef15015: The files should be in both the out/ directory and the vm(which won't commit them), it does cd once more into out/ before this command at line 64, or did I misinterpet what you meant?

azure-pipelines/doc-build.yml
Youssef15015: Do you think we should leave all the others ? I was going to delete them too. In each step

deploy_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+#    echo "Skipping deploy on $TRAVIS_BRANCH. We only deploy docs automatically from development."
+#    exit 0
+#fi
+
+# Save some useful information
+REPO=`git config remote.origin.url`
+SSH_REPO=${REPO/https:\/\/github.com\//git@github.com:}
+SHA=`git rev-parse --verify HEAD`
+
+# Add rsa keys to the ssh agent to push to GitHub
```
marxwillia: Is the commented out code supposed to be present? Like for users to uncomment (in which case do you want to add flags to the script instead), or is it old code that should be removed?

Youssef15015: The plan is to remove it, but I was waiting for Alice to come back and update her before we fully decide to abandon this method.

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,86 @@
+# Specify which branches you want to trigger for continous deployment and/or applicable for pull requests.
+# Otherwise it triggers all of them.
+trigger:
+- master
+
+pr:
+- master
+
+# Variable we can refernece later.
```
wkerzendorf: typo

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,86 @@
...
+  system.debug: 'true'
+  ref.data.home: '$(Agent.BuildDirectory)/tardis-refdata'
+
+  ref.data.github.url: 'https://github.com/tardis-sn/tardis-refdata.git'
+  tardis.build.dir: $(Build.Repository.LocalPath)
+
+
+jobs:
+
+- job: 'Test'
```
wkerzendorf: name change?

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,86 @@
...
+  pool:
+    vmImage: $[variables.vm_Image]
+
+# Basic setup of our virtual machine(vm).
+  strategy:
+    matrix:
+      linux:
+        vm_Image: 'Ubuntu-16.04'
+        conda: '/usr/share/miniconda'
```
wkerzendorf: is this true?

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,86 @@
...
+  - bash: |
+      sudo chown -R $USER $CONDA
+      conda update -y conda    
+    displayName: updating conda and activating
+  - bash: |
+        sh ci-helpers/install_tardis_env.sh     
+    displayName: 'Install TARDIS env'
+  - bash: |
+        sh ci-helpers/fetch_reference_data.sh
+    displayName: 'Fetch Reference Data'
```
wkerzendorf: do we need this?

deploy_docs.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
```
wkerzendorf: how is this licensed?

deploy_docs.sh
wkerzendorf: what do you mean by the vm

Youssef15015: I meant to say the source branch. A.k.a the master that is being run in the vm, which makes the out directory, which github tells me is a repo, that is also checkout out to the gh-pages branch, that is  pushed to github independent of the original source branch. 
Github gives me this message also:
warning: adding embedded git repository: out
hint: You've added another git repository inside your current repository.

deploy_docs.sh
harpolea: This script comes from a code with the BSD license here: https://github.com/AMReX-Codes/amrex/blob/master/LICENSE. 

My experience of software licensing is limited, but from what I can gather from the internet, it would be sufficient to add this to the end of the existing license with a note along the lines of `deploy_docs.sh was modified from the original script in the AMReX code (https://github.com/AMReX-Codes/amrex), which is covered by the following license`

deploy_docs.sh
harpolea: I'm assuming this is just there as a fail safe in case for some reason the git pull doesn't work. I'm not entirely sure why this would be the case though, so it's probably safe to remove.

deploy_docs.sh
harpolea: https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---orphanltnewbranchgt

It creates a new branch without any history. I believe the idea is to help keep the `gh-pages` branch clean so that it only contains commits related to docs deployment.

deploy_docs.sh
Youssef15015: Removed.

azure-pipelines/doc-build.yml
Youssef15015: Actually specifying the pool earlier would be the basic setup of the Vm, if that what was your question was regarding, but this part is unnecessary unless we are testing multiple operating systems. So I can remove this all together.
There is indeed a miniconda in that path.

azure-pipelines/doc-build.yml
Youssef15015: Change to Build Documentation

azure-pipelines/doc-build.yml
Youssef15015: I did not suspect so, and the documentation is the same with and without it.

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,74 @@
...
+      conda update -y conda    
+    displayName: updating conda and activating
+  - bash: |
+        sh ci-helpers/install_tardis_env.sh     
+    displayName: 'Install TARDIS env'
+  - bash: |
+        source activate tardis
+        bash deploy_docs.sh
+    displayName: 'TARDIS build and deployment to gh-pages'
+# See github.com/Youssef15015/tardis for docuemntation of these shell files.
```
wkerzendorf: misspelling and this is not where the documentation should live.

azure-pipelines/doc-build.yml
```diff
@@ -0,0 +1,74 @@
...
+
+#  - task: PublishTestResults@2
```
wkerzendorf: remove the things that are not needed.

deploy_docs.sh
Youssef15015: Change to git pull origin $SOURCE_BRANCH

deploy_docs.sh
```diff
@@ -0,0 +1,61 @@
...
+   echo "exiting with no docs changes"
+   exit 0
+fi
+
+# Otherwise, commit and push
+git commit -m "Deploy to GitHub Pages: ${SHA}"
+git push $SSH_REPO $TARGET_BRANCH
+cd ..
+
+# See https://github.com/AMReX-Codes/amrex/blob/master/LICENSE for the license
```
wkerzendorf: what LICENSE is this. just add it on top (MIT, BSD3, etc. ) - and say this file was modified from ... and is LICENSed the same

deploy_docs.sh
```diff
@@ -0,0 +1,61 @@
...
+# deploy_docs.sh was modified from the original script in the AMReX code (https://github.com/AMReX-Codes/amrex), 
+# which is covered by the following license
```
harpolea: This should not be here, but in the actual tardis LICENSE file itself. You should copy and paste the last two lines at the bottom of the tardis LICENSE file, then under than copy and paste the text of the LICENSE file from the AMReX code (not just provide a link to it).

deploy_docs.sh
harpolea: The license is the LBNL variant of the BSD3 license: https://spdx.org/licenses/BSD-3-Clause-LBNL.html

harpolea: The original script is this one https://github.com/AMReX-Codes/amrex/blob/master/build_and_deploy.sh

wkerzendorf: I would suggest that we say at the beginning of the file that we take this from  https://github.com/AMReX-Codes/amrex/blob/master/build_and_deploy.sh licensed accoriding to  https://spdx.org/licenses/BSD-3-Clause-LBNL.html - maybe

harpolea: That's probably fine, but better might be to link to the repo's DOI (https://doi.org/10.5281/zenodo.2555438), and say that it's licensed according to BSD-3-Clause-LBNL (https://github.com/AMReX-Codes/amrex/blob/master/LICENSE).

PR #947, opened 2019-07-03
Removed docs other than index for RTD redirect branch

jaladh-singhal:
The main page (index) at https://tardis.readthedocs.io/en/docs-prepare_release shows docs are moved but the other pages are still available like: https://tardis.readthedocs.io/en/docs-prepare_release/installation.html The reason is presence of other rst files that get build at RTD. So I removed them to prevent misleading the users about the location of docs!

Also I've removed `v` in conf.py because `V` is automatically prepended in docs - this will fix `Vv-move-documentation` appearing in title bar at https://tardis.readthedocs.io/en/docs-prepare_release


Youssef15015:
Awesome, we might as well remove the .readthedocs.yml file, right ?

jaladh-singhal:
@Youssef15015 actually this PR is for redirect @wkerzendorf has put on rtd main page: https://tardis.readthedocs.io/en/docs-prepare_release - as you can see, this is controlled by a separate branch: `docs/prepare_release` at TARDIS. So for building the main page successfully, we can't remove `.readthedocs.yml` file

I understand you want to remove RTD files at TARDIS as we're using gh-pages, I also did same at starkit & wsynphot - but for that you should remove them from master branch at TARDIS, you could create a PR for that?

Youssef15015:
Oh, I see. Thanks for clarifying that. I can do that in a seperate PR. Maybe that could also be a general clean up PR to remove all the files we do not use, which we can use the coverage results to help us do. That would be a good final project :)

jaladh-singhal:
Yes, sounds good!
Also make sure to fix the documentation links (replace readthedocs with github page link) mentioned in several files of TARDIS - which I can easily see by searching `readthedocs` in the search bar of tardis repository.

PR #949, opened 2019-07-11
Add issue template

Youssef15015:
Added a template that will appear when writing issues.


wkerzendorf:
that looks like a good idea. I just had a look at https://github.com/tardis-sn/carsus/blob/master/.github/ISSUE_TEMPLATE.md and it suggests to use a newer template. We should update both templates to the newer version

Youssef15015:
The recommended option they suggest requires administrative privileges:
https://help.github.com/en/articles/creating-issue-templates-for-your-repository

PR #950, opened 2019-07-17
Gui

marxwillia:
Converted mappable object to list to make the GUI run with Python 3


wkerzendorf:
@marxwillia they all fail

PR #960, opened 2019-08-22
Save integrated spectrum to hdf if requested

chvogl:
Currently, the integrated spectrum is not saved with the `to_hdf` method. This PR adds this functionality if the spectrum method requested in the config is `integrated`.


chvogl:
We can discuss the creative use of the `__new__` method. If anyone has objections, I will find a different solution.

Issue #800, opened 2017-10-26
Feature request: Add virtual packet diagnostics to hdf dump of runner


unoebauer:
With this issue I'd like to discuss whether it would make sense to add the virtual packet diagnostic arrays to the HDF5 write-out of the runner class whenever the ``--with-vpacket-logging`` flag is active. In particular, I'd like to add the following arrays:
* ``virt_packet_last_interaction_type``
* ``virt_packet_last_line_interaction_in_id``
* ``virt_packet_last_line_interaction_out_id``
* ``virt_packet_last_interaction_in_nu``
* ``virt_packet_nus``
* ``virt_packet_energies``

The main benefit of adding these arrays would be that I could use the native TARDIS ``to_hdf`` functionality for the tardisanalysis tools and could scrap the clunky interface which is currently used.

What do you think, @wkerzendorf?

kaushik94:
@wkerzendorf @unoebauer I would like to take up this task if it is ok?

ImmortalRabbit:
Is this issue still open? It's just something was merged already.

PR #967, opened 2019-09-29
log additional properties in hdf witeout when --with-vpacket-logging is present

kaushik94:
@wkerzendorf this is an attempt at fixing #800 . I am not sure how to proceed from here. Can you please tell which other classes you would like me to add these properties? Or does this satisfy?


wkerzendorf:
@kaushik94 - we'll have a look - thanks

Issue #970, opened 2019-10-22
Spikes in the integrated spectrum


afloers:
#### 
I ran the ddc10 model with 500,000 photon packets. Increasing the number of packets to 2.5e6 didn't solve the problem.

#### Problem description
In some models (e.g. DDC10) the integrated spectrum has extremely high spikes in the UV. The problem is only present in the integrated spectrum. The real spectrum doesn't suffer from this issue. 

#### Expected Output
The integrated spectrum is shown in green, the real spectrum in orange.
<img width="1025" alt="Screenshot 2019-10-22 at 11 36 46" src="https://user-images.githubusercontent.com/33418619/67274202-416be900-f4c0-11e9-8104-b61bf8234675.png">

chvogl:
@afloers, it would be great if you could upload the configuration yaml file as well as the csvy model file; this would allow us the reproduce the problem.

afloers:
[tardis_config.zip](https://github.com/tardis-sn/tardis/files/3755606/tardis_config.zip)
I used these files to compute the model.

marxwillia:
@afloers I can't download the files.

chvogl:
Fixed with merge of #971.

PR #971, opened 2019-10-23
Fix formal integral spikes bug

chvogl:
Fixes a bug in the formal integral that produces high amplitude spikes in the integrated spectrum (see #970).


chvogl:
I have switched from linear interpolation of the tau_sobolevs to an interpolation scheme that produces constant values within a non-interpolated shell; this behavior is closer to what actually happens in the MC simulation and prevents extrapolation to negative values. The scheme only matches the treatment MC simulation exactly if the number of interpolated shells is a multiple of the original number. Should we enforce something like this?

PR #974, opened 2019-10-31
Updating the citation guidelines. 

wkerzendorf:
Inspired by discussion with Stephen Smartt, @ssim et al.

### DIFF
--- a/azure-pipelines/doc-build.yml
+++ b/azure-pipelines/doc-build.yml
@@ -66,7 +66,6 @@ jobs:
   - bash: |
         source activate tardis
         sudo npm install -g github-changes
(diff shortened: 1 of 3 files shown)


marxwillia:
Rerun the tests, things should pass.

jamesgillanders:
The link for the Zenodo citation will continue to get outdated if the version is always fixed like in this case. Consider using the link that always defaults to the most up-to-date version (https://doi.org/10.5281/zenodo.592480), and/or specify that the user should/is encouraged to change the version link depending on the version they used for their work

jamesgillanders:
Also, it would be good to get this merged to improve citation stuff

jamesgillanders (approved):

jamesgillanders (changes requested):
see comments - otherwise looks good

PR #1047, opened 2020-02-24
Numba replace intensity_black_body 

jordi5:
## Description
Replace numexpr.evaluate with numba decorator in function intensity_black_body and added error handling

numexpr.evaluate has a error handling in witch it treats DivideByZero as numpy does. Numba doesn't do this by default, halting the program instead. Adding (error_model='numpy') converts the error value to a NaN as numpy does.

Although cProfiling on the function itself shows significant improvement, doing the same for the whole code to test performance improvement doesn't show improvement since it's a one-time call.

image

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -3,7 +3,6 @@ import os
 import re
 from collections import OrderedDict
 
-import numexpr as ne
 import numpy as np
 import pandas as pd
 import yaml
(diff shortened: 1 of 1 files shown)


jordi5:
I'm testing ways to make the test raise an exception

jordi5:
This change doesn't seem make the code faster, profiling, in fact shows the opposite.
I believe this is not the ideal case for implementing numba to save time.
image

jordi5:
This change doesn't seem make the code faster, profiling, in fact shows the opposite.
I believe this is not the ideal case for implementing numba to save time.
image

jordi5:
The function is called only once so (i think) it doesn't seem like the best place to implement numba and expect a performance increase 

image

wkerzendorf:
@jordi5 - yep not the best place to improve - so now that you know how to profile - have a look for better places :smile:

PR #1049, opened 2020-02-25
Modified intensity_black_body and calculate_luminosity to use Numba

BugsBunny0:


tardis/util/base.py
```diff
@@ -121,7 +122,7 @@ def roman_to_int(roman_string):
         raise ValueError('Can not interpret Roman Numeral {0}'.format(roman_string))
     return result
 
-
+@njit(error_model = 'numpy')
```
wkerzendorf: why do you use this?

tardis/util/base.py
```diff
@@ -3,7 +3,7 @@
 import re
 from collections import OrderedDict
 
-import numexpr as ne
+# import numexpr as ne
```
wkerzendorf: good!

tardis/util/base.py
```diff
@@ -242,7 +243,7 @@ def create_synpp_yaml(radial1d_mdl, fname, shell_no=0, lines_db=None):
     with open(fname, 'w') as f:
         yaml.dump(yaml_reference, stream=f, explicit_start=True)
 
-
+@njit( error_model = 'numpy')
```
wkerzendorf: please profile the code - does this make it faster?

tardis/util/base.py
BugsBunny0: Thanks

marxwillia:
Once you have profiled the code, try to find other functions to optimize.

PR #1052, opened 2020-02-25
modified intensity_black_body using numba

shreyansjain012:
Optimize the `intensive_black_body` speed with numba
## Description

- Imported njit from numba
- Removed numexpr because numba can only work with numpy arrays
- Modified function intensive_black_body with numba
## Motivation and Context

It significantly reduces the processing speed of function

## How Has This Been Tested?

For testing of performance, I ran the commands

`from tardis import run_tardis`
`from tardis.io.atom_data.util import download_atom_data`
`download_atom_data('kurucz_cd23_chianti_H_He')`
`sim = run_tardis('tardis_example.yml')`

and then I profiled this function `run_tardis` so that I can see what other function it is calling and their respective time, from there I searched for the function `intensity_black_body` which gives the current output. I did this thing with and without modifying the function with numba
image

When I tried using `intensity_black_body` with my own input parameters in a separate notebook and profiled it I found that it does indeed optimizes the function but only after the second call, for the first call it takes more time than usual(without numba) so I am guessing it is because of low number of total number of times this function is called(21 times)
Screenshot from 2020-02-25 20-47-53
**without using numba**

Screenshot from 2020-02-25 20-48-09
**when using function with numba**

## Types of changes

## Checklist:

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -3,13 +3,13 @@ import os
 import re
 from collections import OrderedDict
 
-import numexpr as ne
 import numpy as np
 import pandas as pd
 import yaml
 from tardis import constants
 from astropy import units as u
 from pyne import nucname
+from numba import njit
 
 import tardis
 from tardis.io.util import get_internal_data_path
@@ -243,6 +243,8 @@ def create_synpp_yaml(radial1d_mdl, fname, shell_no=0, lines_db=None):
         yaml.dump(yaml_reference, stream=f, explicit_start=True)
 
 
+
+@njit( error_model = 'numpy')
 def intensity_black_body(nu, T):
     """
     Calculate the intensity of a black-body according to the following formula
@@ -265,8 +267,7 @@ def intensity_black_body(nu, T):
     """
     beta_rad = 1 / (k_B_cgs * T)
     coefficient = 2 * h_cgs / c_cgs ** 2
-    intensity = ne.evaluate('coefficient * nu**3 / '
-                            '(exp(h_cgs * nu * beta_rad) -1 )')
+    intensity = coefficient * nu**3 / ( np.exp(h_cgs * nu * beta_rad) -1 )
     return intensity
 
 


tardis/util/base.py
```diff
@@ -10,6 +10,8 @@
 from tardis import constants
 from astropy import units as u
 from pyne import nucname
+from numba import njit
+from tqdm.autonotebook import tqdm
```
wkerzendorf: why do you import this?

wkerzendorf:
okay - try to profile the whole code and see how much of a difference it makes

tardis/util/base.py
shreyansjain012: actually I ran this on jupyter notebook and without this it was giving some warnings.

wkerzendorf: but you do not use tqdm in this function, right?

shreyansjain012: Yes I didn't. I will remove it.

shreyansjain012:
> okay - try to profile the whole code and see how much of a difference it makes

Okay, thanks. Let me learn about code profiling and I' ll get back to you

shreyansjain012:
@wkerzendorf `intensity_black_body` function is not called inside base.py where am I supposed to use cprofile?

wkerzendorf:
follow the quickstart example - have a look what the functions are that call this.

shreyansjain012:
@wkerzendorf 
When running j_blues.py for profiling `intensity_black_body` using numba on the contrary decreased the performance a little bit

performance when using numba
with numba

performance when not using numba
withoutNumba

wkerzendorf:
you are just importing the module - not really running code in there. Have a look at the quickstart notebook - run the whole TARDIS simulation and then look from there where else to change code

shreyansjain012:
@wkerzendorf I followed the quickstart notebook. Ran whole tardis simulation and got this stats

image

am I still doing something wrong?

wkerzendorf:
what function did you profile?

shreyansjain012:
> what function did you profile?

run_tardis

PR #1053, opened 2020-02-25
Use numba instead of numexpr in tardis.util.base.intensity_black_body

svineet:
`intensity_black_body` in `tardis.util.base` has been modified to use numba instead of numexpr

A simple profile done in IPython terminal, with `intensity_black_body_old` representing the numexpr version
```python
In [1]: %timeit intensity_black_body(np.random.rand()*100, 100*np.random.rand())                                                                     
2.23 µs ± 90.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [2]: %timeit intensity_black_body_old(np.random.rand()*100, 100*np.random.rand())                                                                 
34.4 µs ± 130 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Numpy is used to generate random numbers above so that we can get results over random numbers and do not obtain profiling results specific to given `nu` and `T`.

About 15x speed up is observed for repeated execution. However as numba does just in time compilation, it will be slower on the first call of the function, therefore if this function is not supposed to be called repeatedly, it would not be wise to optimise using numba.

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -3,13 +3,15 @@ import os
 import re
 from collections import OrderedDict
 
-import numexpr as ne
+# import numexpr as ne
 import numpy as np
 import pandas as pd
 import yaml
 from tardis import constants
 from astropy import units as u
 from pyne import nucname
+
+from numba import njit
 
 import tardis
 from tardis.io.util import get_internal_data_path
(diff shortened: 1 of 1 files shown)


wkerzendorf:
okay - have a look if you can improve other parts.

epassaro:
Tests are failing

PR #1054, opened 2020-02-26
Integration testing - task

not4win:
Completed the integration testing assignment with minor tweaking.
The tests folder has 2 folders unit-test and integration-test
The integration-test folder has the result of integration testing. No errors. Ran successfully
The reference data generated was not included here as it is too big.


marxwillia:
Now try generating your own reference data and comparing it with the reference data on github.

not4win:
Okay! Will do

not4win:
@marxwillia Done with it! Have a look. Also guide me into what I should do next!

PR #1060, opened 2020-02-28
Modify intensity_black_body using numba

lucbece:
## Description
Modify function intensity_black_body to be compiled with numba

Added <code>@jit</code>. decorator to <code>intensity_black_body</code> using <code>nopyhon=True</code> and <code>error_model='numpy'</code> as arguments. The first one is set to run the function without the involvement of the Python interpreter, while the latter is set to handle division by zero by returning a <code>NaN</code>, an implicit feature of <code>numexpr.evaluate</code>. The <code>numexpr</code> import was removed, as no other function calls it

## Motivation and Context
Improve performance using the numba library

## How Has This Been Tested?
Used <code>cProfile</code> to test the entire code. Contrary to what was intended, the results show an increase in time, given that numba's performance boost relies on calling the decorated function multiple times, which is not the case with <code>intensity_black_body</code>

results

## Types of changes

## Checklist:

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -3,7 +3,6 @@ import os
 import re
 from collections import OrderedDict
 
-import numexpr as ne
 import numpy as np
 import pandas as pd
 import yaml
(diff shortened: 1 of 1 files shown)


marxwillia:
What other functions would benefit from optimization with numba?

PR #1064, opened 2020-02-29
Quantity-ify TARDIS first objective

PIYUSHgoyal16:
## Description
This PR creates a class DataTable which is capable of storing units of attributes and provide capability to pass these to another object of same class.

## Motivation and Context
Pandas DataFrame isn't capable to store the units of the attributes. The other issue is that this information about units of attributes gets further vanished when we try to copy one dataframe into another. 

##Implementation
I have tried to solve this problem by encapsulating the dataframe and units of attributes into a single class. Using this class we can not even store the units of attributes but also pass them into another object of same class by just passing the class members of first object to other.

## How Has This Been Tested?
I have tested the code by importing the library. I have also added short example about how to use the class in comments of the class.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/datatable.py
+++ b/scripts/datatable.py
@@ -1,0 +1,75 @@
+"""
+=====================
+TARDIS data_table module
+=====================
+created on Mar 1, 2020
+
+"""
+
+import numpy as np
+import pandas as pd
+
+class DataTable:
+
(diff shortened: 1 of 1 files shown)


wkerzendorf:
I think subclassing is a better option. Can you try that

PR #1066, opened 2020-03-01
Carsus Atomic Data first objective

PIYUSHgoyal16:
## Description
This PR creates a class Atomic_Data which stores data contained in 'si2_osc_kurucz' downloaded from http://kookaburra.phyast.pitt.edu/hillier/web/CMFGEN.htm.  

## Motivation and Context
The given file contained data but the data was not consistent. It contained two tables and text in between. The number of columns were different in different tables. Also some of the rows have values of different attributes not separated by space which makes it further inconsistent.

## Implementation
I have tried to solve this problem by first loading the data. Then splited the data based on space. Replaced some attributes value to "NaN". Then Making a column named "NaN" and counting the number of values that are Nan in that particular row. By observing the number of values of NaN in a particular row helps in separating the Tables from the text. 
The Nan values in a particular columns also help us to select the number of attributes. In all this work the only assumption used is the number of tables, which too can be generalised further. 

## How Has This Been Tested?
I have tested the code by importing the library. I have also added short example about how to use the class in comments of the class. Relevant Screenshots of the tables have been attached.

Screenshot from 2020-03-01 12-06-17
Screenshot from 2020-03-01 12-07-31

## Types of changes

## Checklist:

### DIFF
--- a/scripts/atomic_data.py
+++ b/scripts/atomic_data.py
@@ -1,0 +1,83 @@
+"""
+=====================
+TARDIS Atomic_data module
+=====================
+created on Feb 29, 2020
+"""
+
+import numpy as np
+import pandas as pd
+
+class Atomic_Data: 
+    """ Will serve as base class for all atomic data contained in 
+    'si2_osc_kurucz' in order to provide consistent Tabular data.
+    
+    Class Members
+        ----------
+        data : dataframe
+            raw unprocessed data
+            
+        first_table : dataframe
+            Contains Energy Levels and statistical weights for Si II
+            
+        second_table : dataframe
+            Oscillator strengths for LLL 
+            
+    Example to use
+         ----------
+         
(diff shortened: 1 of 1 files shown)


PIYUSHgoyal16:
@wkerzendorf  @marxwillia @epassaro kindly review my PR. The only assumption made during this code is that the given file contains 2 tables, which too can be generalised by simply running my method in a loop. i haven't implemented it yet because the file was given. If the file is a variable, then it can be implemented.

wkerzendorf:
@PIYUSHgoyal16 have a look at the carsus package - this is where this belongs. Then you should also write a test and some documentation. good luck!

PIYUSHgoyal16:
@wkerzendorf I am unable to locate carcus package. Can you please direct me by sharing a link for the same.

PR #1072, opened 2020-03-02
Extending Pandas enabling them to store units

PIYUSHgoyal16:
Created a class DataTable that extends the pandas dataframe and enables it to store the units of the attributes. Moreover I have also created copy function that allows this information to copy in another object of same class.

## Description
I have created class DataTable inherited from pd.DataFrame which contains metadata about units. I have done this by subclassing Dataframe as told by @wkerzendorf. I have also created docstrings inside the code that allows the user to see how to actually use the class.

## Motivation and Context
The motivation of this change was the incapability of pandas Dataframe to store units of attribute. It was even harder to conserve this information while copying Dataframe.

## How Has This Been Tested?
I have imported this library and created objects of DataTable class. I have cerified each and every function of the class. Codes for Unit testing are also being written by me that I will upload shortly after confirmation that my method are implemented in correct manner. 

## Future Aspect
I am going to write the documentation and unit testing code according TARDIS standards shortly. Then more functionality in the class can be added in order to increase the feasibility for the user. These function may include things like directly writing information to file or plotting a graph.

## Types of changes

## Checklist:


marxwillia:
Good work.  Now write tests to make sure that the metadata is kept during operations like array slicing, arithmetic, etc.

marxwillia:
Tests are failing.  Looks like it might have something to do with your new test.  Please check the detailed test report, and see how other unit tests are added to tardis.

marxwillia:
Very good work with the tests!  Another thing that we want to support is multiplication of datatables by scalars with units, and dot products of series, each with units.  Can you implement this so that units are updated?

PIYUSHgoyal16:
Sure Marc. I will try my best and report you shortly after completing above task.

marxwillia:
@PIYUSHgoyal16 we want to use the Astropy Quantities for the units in order to do actual dimensional analysis.  Also you should be able to use the pandas dot product and overwrite it, instead of manually implementing it.

PIYUSHgoyal16:
Marc I have used the astropy units in place of strings. This enables to do the actual dimension. I have also added the scalar multiplication, series multiplication(overwritten dot product), add, delete and replace function.

PIYUSHgoyal16:
@marxwillia I have also added the join DataTable feature. Can you please suggest some more features that I can implement.

marxwillia:
@PIYUSHgoyal16 i think your class functions can be drastically simplified by using the pandas implementations of dot product.  Remember that you have extended the DataFrame class.  You should write new class functions that leverage the DataFrame class functions, but also do dimensional analysis.

PIYUSHgoyal16:
@marxwillia I have done that only. Actually the "__private_dot" is a function whose code is been taken from pandas library code. This is due to the fact that I want to use their algorithm so I just took the code and made it a function and called it in the overwritten "dot" function. I made "__private_dot" private so that it can only be accessed within the class. 
It may be that I am unable to do what you actually want to do. But I can assure you the time complexity of this code is as low as it could be. If you want me to implement it in a different way, you can send me some resources related to it. Hope I will be able to do the respective job satisfactorily.

PIYUSHgoyal16:
@marxwillia  can you please explain me what is the vision of this project, what we actually aim to achieve. I wanted to know this stuff so that I can plan my schedule of completing the job within the time and write the same in the proposal.

PIYUSHgoyal16:
Marc I have also figured one more way to implement the dot product. It uses the decorator. Still it requires me to write the original code for dot product. So should I give it a try?

marxwillia:
@PIYUSHgoyal16 you should never need to copy paste code from pandas.  Since your new class subclasses the pandas DataFrame, you can call the original pandas DataFrame functions.  Look up tutorials on function overriding.

PIYUSHgoyal16:
@marxwillia I am unable to access the overridden function of the base class, i.e the dot function. Can u help me in figuring out how it can be accessed.

PIYUSHgoyal16:
@marxwillia thanks for guiding. I was able to figure it out and implemented the same. Hope this time I was able to perform what was actually required.

marxwillia:
@PIYUSHgoyal16 excellent job!  That's exactly the sort of thing that you want.  Check through the rest of your code to make sure you are using the Pandas functions wherever possible.

tardis/datatable.py
```diff
@@ -0,0 +1,314 @@
...
+        Returns
+        ----------
+        object of class DataTable
+        
+        """
+        new_units = []
+        for i in self.units:
+            new_units.append(i*unit)
+
+        dataframe = super(DataTable,self).dot(series)
```
marxwillia: This is exactly what I wanted.

tardis/datatable.py
```diff
@@ -0,0 +1,314 @@
...
+        attr: array of string or int
+            list of columns to which scalar is multiplied
+        
+        """        
+        new_units = []
+        for i in range(len(self.units)):
+            if( list(self.columns) [i] in attr):
+                new_units.append(self.units[i]*unit)
+            else:
+                new_units.append(self.units[i])
```
marxwillia: I don't think this for loop is necessary.  Try to simplify it.

tardis/datatable.py
```diff
@@ -0,0 +1,314 @@
...
+        
+        """
+        ind = list (self.columns).index(attr)
+        if len((self.units)) > max (0 ,ind) :
+            return self.units[ind]
+        else :
+            print( "The unit for the following attribute is not set." )
+            
+    
+    def scalar_mul(self, const, unit, attr):
```
marxwillia: I think this function can be simplified.

tardis/datatable.py
PIYUSHgoyal16: Thanks @marxwillia . Finally, I was able to crack. Thanks for making me think in this direction. Really learned a cool stuff :)

tardis/datatable.py
PIYUSHgoyal16: I have simplified the function "scalar_mul" and even made a PR for the same. Kindly review it.

PIYUSHgoyal16:
@marxwillia what next I could do? Kindly also state the vision of the project so that I could incorporate the plan to implement the same in my GSoC 2020 proposal.

Issue #962, opened 2019-09-04
Require absolute paths for unit tests


marxwillia:
When the unit tests are run using pytest, relative paths defined in fixtures cause problems.

Chinmay4400:
@marxwillia I think I ran into this problem while running test for plasma module which required the simulation to be run for the atomic data used in quickstart notebook. However, when i moved the yml file to data folder, it started looking for this data in that folder instead of where it was downloaded. Is this the same issue? If so, I'd like to work on this.

not4win:
The best solution to this problem is to pass only absolute file path to tardis-refdata argument and to raise an error when absolute file path to tardis-refdata is not provided.

andrewfullard:
Resolved by regression data class

PR #1074, opened 2020-03-02
addressed issue caused by relative file path.

not4win:
Addressing issue #962 . As relative filepath to tardis-refdata cause problems, making the user to add in only the absolute filepath to tardis-refdata seems to solve this issue.


wkerzendorf:
can you rebase - I think the capmanager is already in the code

PR #1078, opened 2020-03-03
Enable the Dask framework for parallel execution for TARDIS

KaairaGupta:
## Description

This is my first attempt at the objective task 'Enable the Dask framework for parallel execution for TARDIS' 
I created a file `run_tardis_with_dask` to test running `run_tardis()` in parallel. But a type error occurred on running `x.result()` where init was missing 12 positional arguments.
I used python debugger to find out that the problem was with `x.result()`. On reading the dask source code I found that once dask completes a task, `result()` function ultimately `yields` the result where a new object of all classes with `HDFWriterMixin` as parent is formed.
At first I deduced that dask was trying to create a Simulation object without passing the 12 required parameters, so I passed them as kwargs and args myself in the class. But then I realised that it was trying to create a new object of all classes with `HDFWriterMixin` as parent and each of their kwargs and args cannot be passed this way. 
Hence, I think my approach is wrong and I am trying to find another way.
Please review my code and help me with it.

## Types of changes

## Checklist:

### DIFF
--- a/run_tardis_with_dask.py
+++ b/run_tardis_with_dask.py
@@ -1,0 +1,14 @@
+from tardis import run_tardis
+from tardis.io.atom_data.util import download_atom_data
+from dask.distributed import Client
+
+# the data is automatically downloaded
+download_atom_data('kurucz_cd23_chianti_H_He')
+
+if __name__ == "__main__":
+    client = Client()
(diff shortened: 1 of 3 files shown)


marxwillia:
I think this should be possible without having to make major changes to the TARDIS code itself.

PR #1080, opened 2020-03-04
Running parallel instances of TARDIS using Dask

ryzbaka:
## Description 
This approach for running parallel TARDIS instances uses Dask to run multiple simulation scripts simultaneously.

A simulation script in this context is a python script that uses TARDIS' `run_tardis` function to run a simulation on a particular dataset.

I've executed the same simulation script in parallel for the purpose of testing this approach.

-Gitter: @ryzbaka

### File Descriptions
* `start_cluster.py` is a python script for configuring and starting a local dask cluster (dask-scheduler and dask-worker nodes).
* `simulation.py` is a script implementing a simulation from the TARDIS quickstart page with the [added functionality](https://github.com/tardis-sn/tardis/blob/b47ae85624559cf2c6e0470a9bc168ac1b1436f4/ryzbaka/simulation.py#L31-L46) of being able to save Simulation objects as pickle files.
* `Dask_Approach_1.ipynb` is a notebook demonstrating how to use the dask cluster to run multiple TARDIS instances which can be viewed [here](https://github.com/ryzbaka/ryzbaka.github.io/blob/master/Dask_Approach_1.ipynb).
* Remaining files are needed by the dask worker nodes for parallel processing.
## Motivation and Context
* The proposed method of running TARDIS processes in parallel can help speed up comparative supernova spectra analysis by simply connecting to a local cluster.

## How does it work?
*  Run `start_cluster.py`.
    *  Number of processes and their thread pool size can be configured at this step.
    * [This](https://github.com/tardis-sn/tardis/blob/b47ae85624559cf2c6e0470a9bc168ac1b1436f4/ryzbaka/start_cluster.py#L8-L22) is the method that helps us do that.
*  Connect to the Dask scheduler using the Client method in the `dask.distributed` module.
*  Use the `os` module's `system` method to run the python simulation scripts.
*  Map this function to a list of simulation script file names using the `Client.map` method or use the `Client.submit` method if only script has to be run.
* Use the `Client.gather` method to actually run the processes (Dask follows lazy execution so no actual work is done by the scheduler and worker nodes of the cluster till this step is executed).
* The Dask cluster will then execute all simulation scripts in parallel and save each resulting Simulation object as a pickle file whose name indicates the time of creation in Unix time.

## How Has This Been Tested?
* The approach was tested by executing multiple simulations scripts using the cluster and comparing execution times.
* Information regarding the testing can be found in the jupyter notebook `Dask_Approach_1.ipynb` [here](https://github.com/ryzbaka/ryzbaka.github.io/blob/master/Dask_Approach_1.ipynb).
* All instances were run on Linux Mint 19.2 Cinnamon.

## Next Steps
* Use native Dask data structures in the codebase (replacing NumPy arrays with Dask arrays, pandas DataFrames with Dask DataFrames, etc) so as to optimize the code for parallel execution on a local cluster.
* Use dask.delayed to create lazy evaluation computation graphs to further increase parallelism in code execution.
* Identify and parallelize [pure functions](https://stackabuse.com/functional-programming-in-python/). A simple example of this can be found [here](https://github.com/ryzbaka/ryzbaka.github.io/blob/master/Parallelizing%20tardis.run_tardis.ipynb), where I attempted to parallelize the `run_tardis` function which creates, runs and returns a Simulation object.


wkerzendorf:
ok - pleaae clean your PR of unnecessary files. You seem to execute another. process to start TARDIS - this exercise does not need that. Try to change your notebook

ryzbaka:
Hi, for the process to execute TARDIS, are you referring to the `simulation.py` file?

ryzbaka:
Added another notebook demonstrating an attempt to optimize helper functions in `tardis.run_tardis` without breaking its functionality using `delayed` which is a low level interface for parallel function execution.
The notebook can be found [here](https://github.com/ryzbaka/ryzbaka.github.io/blob/master/Parallelizing%20tardis.run_tardis.ipynb)

ryzbaka:
@wkerzendorf , from my understanding the goal of this project is to completely parallelize TARDIS. Should I move on to parallelizing methods in the library using Dask?

Issue #948, opened 2019-07-11
TARDIS documentation restructure (I of many)


wkerzendorf:
TARDIS is not only compatible with Python 2.7
typo in recommend on installation
update the link to tardis_env3. yml

change the running tardis to tardis configuration file
move formal integral to configuration

move stella to helper scripts

VGZELDA:
hello @wkerzendorf . Can you please elaborate the issue, I would love to work on it.
Thanks

wkerzendorf:
@VGZELDA it's a closed issue.

wkerzendorf:
I wrote on gitter - have a look there.

VGZELDA:
Sorry, I failed to notice it. I am trying to get familiarized with Tardis. Hope to make a PR soon.

PR #1084, opened 2020-03-07
Updated the Document for better readability

TripleR47:
Updated anaconda activation and deactivation command and added a note for the same.

For conda 4.6 and later versions we use conda activate and conda deactivate for activation and deactivation respectively, and for conda versions prior to 4.6 we use source activate or source deactivate commands for activation and deactivation respectively. I added a note for the same and updated the commands. I also made a few changes for better readability. 

This will help those who are new to the Anaconda. Many people face this problem because different version supports different commands.  
#948

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -55,15 +55,18 @@ To create the environment, change to the directory that you downloaded the envir
 
 Then to activate this environment simply do::
 
-    source activate tardis
-
-or the new method::
-
     conda activate tardis
 
 and after you are done with TARDIS you can deactivate::
 
-    source deactivate
+    conda deactivate
+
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -157,12 +156,12 @@ may encounter the following error::
...
-              The command for the same is :
+              The command for the same is
 
-    pip install sphinx==1.5.6
+			pip install sphinx==1.5.6
     
     or
     
-    conda install sphinx==1.5.6
+				conda install sphinx==1.5.6
```
jaladh-singhal: What's the purpose of these extra spaces?

PR #1086, opened 2020-03-09
Partially resolves the issue #680

Phantsure:
## Description

Made some changes to formatting and docs to match with numpydocs format.

## Changes made to files

- [x] tardis.plasma.base
- [x] tardis.plasma.properties.base
- [x] tardis.montecarlo.base

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -21,6 +21,7 @@ import numpy as np
 logger = logging.getLogger(__name__)
 TARDIS_PATH = TARDIS_PATH[0]
 
+
 class MontecarloRunner(HDFWriterMixin):
     """
     This class is designed as an interface between the Python part and the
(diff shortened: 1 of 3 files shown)


Phantsure:
@jaladh-singhal Yeah, you are right. I was using Pycharm, didn't notice the changes it made. Should i change the name or change the files to match issue? Because I suppose PEP8 guidelines if followed makes a good code, right?

Phantsure:
I removed most PEP8 changes and a few unnecessary import statements too

tardis/plasma/base.py
```diff
@@ -58,8 +56,10 @@ def _build_graph(self):
         """
         Builds the directed Graph using network X
 
-        :param plasma_modules:
-        :return:
+        Parameters
+        ----------
+        ~plasma_properties : list
```
jaladh-singhal: Why is `~` needed here?

tardis/montecarlo/base.py
```diff
@@ -197,11 +197,17 @@ def run(self, model, plasma, no_of_packets,
         Parameters
         ----------
         model : tardis.model.Radial1DModel
+            # UNSURE of description of below params
```
jaladh-singhal: If you don't know about this, better leave it as it is - no need of comment. Similarly, remove `.` from following parameters.

tardis/plasma/base.py
Phantsure: I thought it is used for parameters that are not passed as arguments, but i didn't find any docs to support that. Should i remove it?

tardis/montecarlo/base.py
Phantsure: okay

Phantsure:
done the changes. will keep on updating/adding docs to other files.

PR #1097, opened 2020-03-15
GUI Version 0.1

Pabla007:
Tardis GUI for interaction and Widgets

## Description
This is the written code of interaction & plotting in **Sim.py** from where one can import and use the functions that can plot rather than showing them in jupyter notebook otherwise it will be messy.

https://nbviewer.jupyter.org/gist/Pabla007/6a1e9f26ec4f993b6ab8c8d531d336f9

## Motivation and Context
GUI is dead so it is the best i can do without seeing how the actual code is working.
https://github.com/tardis-sn/tep/blob/master/TEP012_gui_overhaul.rst

 Screenshots (if appropriate):
There are three Widgets and the functionality of each widget :-
1 - Plot
that will plot the line plot using bqplot

2 - Model_Paramters
will give a brief about the model parameters

3 - Temp vs shell plot
will plot Rad Temperature Vs Shell plot

Widgets
______________________________________________________________________________________________________
For interactivity brush selector is used from bqlot so user can select the range and get the values from the plot.

Brush+Selector
______________________________________________________________________________________________________
6 more widgets :-
4 widget are regarding fetching data and show that in form of table
(i.e information about individual iterations)

2 widget are for heatmap
(i.e heatmap for radiation temperature in each cell & electron density in each cell)
Heat_Data
_____________________________________________________________________________________________________

Generation of offline documentation through sphinx
screen

Types of changes

Checklist:

### DIFF
--- a/docs/quickstart/Sim.py
+++ b/docs/quickstart/Sim.py
@@ -1,0 +1,236 @@
+from bqplot import *
+from bqplot.interacts import (BrushSelector)
+from bqplot import LinearScale, Axis, Lines, Scatter, Figure
+from ipywidgets import VBox,HTML
+from bqplot import pyplot as plt
+from ipywidgets import widgets
+from ipywidgets import *
+from ipywidgets import Layout
+
+def line_Plot(mdl):
+    mdl=mdl
+    x_sc = LinearScale()
+    y_sc = LinearScale()
+
+
+    x_data = mdl.runner.spectrum.wavelength
+    y_data = mdl.runner.spectrum.luminosity_density_lambda
+    z_data = mdl.runner.spectrum_virtual.luminosity_density_lambda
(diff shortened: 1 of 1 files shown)


jaladh-singhal:
@Pabla007 Also, the GIFs you have added in notebook, should be here in PR comments with a clear description of what functionality each GIF intends to show.

Pabla007:
Currently working to integrate hover tool and reload button, then will replot after the previous plot in the same area.
Is it possible to use **`bokeh`** if yes then we can make better GUI and provide more functionality in that.

jaladh-singhal:
That's great! Yeah bokeh is capable of handling panning & zooming by default, that's why we expect to see more GUI features implemented!

But I suggest you to start working on your proposal, otherwise you may run short of time!

Pabla007:
@jaladh-singhal I will start working on the proposal then thanks fro the advice ^_^

Issue #1101, opened 2020-03-18
Unset QT_API raises errors in tests


wkerzendorf:
#### Code Sample, a copy-pastable example if possible

```bash

pytest tardis
```
#### Problem description

Currently testing requires pyqt being set
```python
_______________ ERROR collecting tardis/gui/tests/test_gui.py ________________
ImportError while importing test module '/Users/wkerzend/python/tardis/tardis/gui/tests/test_gui.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tardis/gui/tests/test_gui.py:6: in <module>
    from tardis.gui import interface
tardis/gui/interface.py:7: in <module>
    raise ImportError('QT_API was not set! Please exit the IPython console\n'
E   ImportError: QT_API was not set! Please exit the IPython console
E    and at the bash prompt use :
E
E    export QT_API=pyside
E    or
E    export QT_API=pyqt
E
E    For more information refer to user guide.
```
#### Expected Output

We should skip the tests in test_gui if QT_API is not set.

wkerzendorf:
@jordi5 can you look at that.

jordi5:
Should we close this?

PR #1102, opened 2020-03-18
Add skip condition if QT_API is no set (bug #1101)

jordi5:
`pytest tardis` fails on import of gui interface and widgets if QT_API variable is not set

## Description
 Since I the issue appears on import and I don't understand how skipimport would work when the error is not that the module is not found, but that there is an undefined variable within the module, I changed the call so that the import wouldn't check inside the module until needed for the test by importing the whole module. Following this some of the objects inside the function have to be called more explicitly.

I added the skipif decorator from pytest but **I'm not sure the issue is fixed.** The test now skips when the variable QT_API is not defined, but I'm not sure how the `pytest tardis` should proceed when the variable is defined. When I define QT_API either way on my master branch the test still skips, which I suppose is the expected behaviour because I didn't modify anything there. 

On a different commit I modified the functions interface and widgets to change how the Iphyton library calls specific functions, as can be seen this changes a qt4 to a qt5 and is purely for consistency, given that the functions are the same.

## Motivation and Context
Adresses #1101 

## How Has This Been Tested?
Run `pytest tardis` and the test now skips when QT_API is not declared.
Run again for the Iphyton calls change.
Giving( to save as reference)
263 passed, 1119 skipped, 1 xfailed in 24.38 seconds

## Types of changes

All input is appreciated.

### DIFF
--- a/azure-pipelines/simple_test_framework.yml
+++ b/azure-pipelines/simple_test_framework.yml
@@ -59,7 +59,9 @@ jobs:
         source activate tardis
         conda install -y pytest-cov
         pip install git+https://github.com/tonybaloney/pytest-azurepipelines.git
+        export QT_API=pyqt
         pytest tardis --tardis-refdata=$(ref.data.home) --cov=tardis --cov-report=xml --cov-report=html
+        export QT_API=pyqt
     displayName: 'TARDIS test'
 
 
--- a/docs/running/gui.rst
+++ b/docs/running/gui.rst
@@ -4,32 +4,32 @@
 Graphical User Interface
 ************************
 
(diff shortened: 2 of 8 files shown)


Chinmay4400:
@jordi5 We can define QT_API if not declared within the test, that way tests won't have to be skipped. `os.environ['QT_API'] = 'pyqt'`

wkerzendorf:
I think that looks great.

jordi5:
I found how to fix the savefig and subfigs buttons, aparently is an issue of matplotlib backend libraries

jordi5:
This is a mess.

PR #1105, opened 2020-03-19
Add skip condition if QT_API is no set (issue #1101)

jordi5:
`pytest tardis` fails on import of gui interface and widgets if QT_API variable is not set

## Description
 The issue appears on import (and I still don't how skipimport would work but I'll revisit at some point). I changed the call so that the import wouldn't check inside the module until needed for the test by importing the whole module. Following this some of the objects inside the function have to be called more explicitly.

I added the skipif decorator from pytest. The test now skips when the variable QT_API is not defined, but I'm not sure how the `pytest tardis` should proceed when the variable is defined. When I define QT_API either way on my master branch the test still skips, which I suppose is the expected behaviour because I didn't modify anything there. 

On a different commit I modified the functions interface and widgets to change how the Iphyton library calls specific functions, as can be seen this changes a qt4 to a qt5 and is purely for consistency, given that the functions are the same.

## Motivation and Context
Adresses #1101 

## How Has This Been Tested?
Run `pytest tardis` and the test now skips when QT_API is not declared.
Run again for the Iphyton calls change.
Giving( to save as reference)
263 passed, 1119 skipped, 1 xfailed in 24.38 seconds

## Types of changes

All input is appreciated.

### DIFF
--- a/tardis/gui/tests/test_gui.py
+++ b/tardis/gui/tests/test_gui.py
@@ -3,7 +3,8 @@ import pytest
 from tardis.io.config_reader import Configuration
 from tardis.simulation import Simulation
 import astropy.units as u
-import tardis.gui as gui
+from tardis.gui.widgets import Tardis 
+from tardis.gui.datahandler import SimpleTableModel
 from PyQt5 import QtWidgets
 
 
--- a/tardis/gui/tests/test_gui.py
+++ b/tardis/gui/tests/test_gui.py
@@ -3,9 +3,7 @@ import pytest
 from tardis.io.config_reader import Configuration
 from tardis.simulation import Simulation
 import astropy.units as u
-from tardis.gui import interface
-from tardis.gui.widgets import Tardis 
-from tardis.gui.datahandler import SimpleTableModel
+import tardis.gui as gui
 from PyQt5 import QtWidgets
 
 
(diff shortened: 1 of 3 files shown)


jordi5:
@wkerzendorf Could you please reopen this? 
I don't have permissions.
Thanks

jordi5:
Ok, so this is ready to merge I think

tardis/gui/tests/test_gui.py
```diff
@@ -3,9 +3,7 @@
 from tardis.io.config_reader import Configuration
 from tardis.simulation import Simulation
 import astropy.units as u
-from tardis.gui import interface
-from tardis.gui.widgets import Tardis 
-from tardis.gui.datahandler import SimpleTableModel
+import tardis.gui as gui
```
sashank27, line 6:
```suggestion
from tardis.gui.widgets import Tardis 
from tardis.gui.datahandler import SimpleTableModel
```

tardis/gui/tests/test_gui.py
```diff
@@ -39,11 +37,12 @@ def simulation_one_loop(
...
-
+    
+@pytest.mark.skipif('QT_API' not in os.environ, reason="enviroment variable QT_API is not set")
 def test_gui(simulation_one_loop):
     simulation = simulation_one_loop
     app = QtWidgets.QApplication([])
-    tablemodel = SimpleTableModel
-    win = Tardis(tablemodel)
+    tablemodel =gui.datahandler.SimpleTableModel
+    win = gui.widgets.Tardis(tablemodel)
```
sashank27, line 46:
```suggestion
    tablemodel = SimpleTableModel
    win = Tardis(tablemodel)
```

tardis/gui/widgets.py
```diff
@@ -1066,13 +1066,13 @@ def __init__(self, tablemodel, config=None, atom_data=None, parent=None):
         """
 
         #assumes that qt has already been initialized by starting IPython
-        #with the flag "--pylab=qt"
+        #with the flag "--pylab=qt"gut
```
jordi5, line 1069: Fix this typo later

PR #1106, opened 2020-03-19
Optimization using Numba

ryzbaka:
## Description
This is my implementation of optimizing the `intensity_black_body` method using the Numba Just In Time compiler.

## Optimized Method
```
@jit(float64(float64,float64),nopython=True,parallel=True)
def intensity_black_body(nu, T):
    """
    Calculate the intensity of a black-body according to the following formula
    """
    beta_rad = 1 / (k_B_cgs * T)
    coefficient = 2 * h_cgs / c_cgs ** 2
    intensity=(coefficient* nu**3)/(np.exp(h_cgs * nu * beta_rad)-1)
    return intensity
```

## How Has This Been Tested?
```
from tardis.util.base import intensity_black_body
import time

if __name__=='__main__':
	intensity_black_body(nu=1.0,T=1.0) 
	# numba LLVM compiles on first call
	# subsequent calls to the optimized function take full advantage of JIT.
	start=time.time()
	result=intensity_black_body(nu=2*10**5,T=275)
	end=time.time()
	print(f"Time taken after numba optimization :{end-start}")
	print(f"result:{result}")

```
Result:
```
Time taken after numba optimization :1.9073486328125e-06
result:3.379593654879596e-24
```
The time taken by the function after being optimized by Numba was `1.907e-06` whereas it was `0.0030432`.


wkerzendorf:
you are committing 8 files - are they all really necessary for testing numba?

wkerzendorf:
and try to do one idea per PR - this one has the dask stuff in as well.

ryzbaka:
Deleted unnecessary files.

shreyansjain012:
Try profiling whole tardis program instead of giving your own input arguments

ryzbaka:
@shreyansjain012 , could you elaborate on the same please?

shreyansjain012:
learn about profiling

PR #1114, opened 2020-03-23
Optimised intensity_black_body function from Cython to Numba

chinmay-bhat:
## Description

Optimised intensity_black_body function from Cython to Numba using njit decorator.
Renamed older cython function to cython_intensity_black_body.
The difference between runtimes before and optimisation can be found by running the code below.

## Motivation and Context

This is the first objective for GSOC idea 'Optimisation with Numba'.
Improves performance of function

## How Has This Been Tested?

Following code displays runtime difference

```python
from tardis.util.base import intensity_black_body, cython_intensity_black_body
import time

start = time.time()
old = cython_intensity_black_body(1000,273.15)
end = time.time()
print('before optimising, time taken : ', end - start)

start = time.time()
newfn = intensity_black_body(1000,273.15)
end = time.time()
print('after optimising, time taken : ', end - start)
```

Also ran `python setup.py test` with no errors

## Result
```
before optimising, time taken :  0.0012388229370117188
after optimising, time taken :  1.1920928955078125e-05
```

## Types of changes

## Checklist:

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -14,6 +14,8 @@ from pyne import nucname
 import tardis
 from tardis.io.util import get_internal_data_path
 
+from numba import njit, float64
+
 k_B_cgs = constants.k_B.cgs.value
 c_cgs = constants.c.cgs.value
 h_cgs = constants.h.cgs.value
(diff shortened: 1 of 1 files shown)


tardis/util/base.py
```diff
@@ -247,7 +248,7 @@ def create_synpp_yaml(radial1d_mdl, fname, shell_no=0, lines_db=None):
         yaml.dump(yaml_reference, stream=f, explicit_start=True)
 
 
-def intensity_black_body(nu, T):
+def cython_intensity_black_body(nu, T):
```
wkerzendorf: are you working on that?

chinmay-bhat: No I'm working on the intensity_black_body function, which is the numba implementation.
The cython_intensity_black_body function is the old function, used to compare the cython performance with numba implementation.

chinmay-bhat: I am profiling my code. So far the montecarlo method seems to take a lot of time. I'm searching for more functions which can be optimised.
@wkerzendorf I'll be updating this PR with the profile output soon

chinmay-bhat: There was only a difference of 1-2 seconds so I'll optimise other functions.
Using cython
not-optimal

Using numba
optim

PR #929, opened 2019-05-04
Numba montecarlo

wkerzendorf:
## Things to check if they are done ##

Things to do 

 - [x] line scattering
 - [x] electron scattering
 - [x] virtual packets
 - [x] macroatom

- [x] packet logging (last_packet_interaction)

Line estimators are slowing the process down - maybe a useful optimization strategy

## Final Testing ##

There are 4 code versions. There is `upstream/master` there is `numba_montecarlo` both of which have different seeding strategies and will never be numerically equal (those we will summarize as main). There are two versions that have additional code that will allow logging - those we will call `c_compare` and `numba_compare` (and will call the compare branch). 
 - [x] virtual - only compare - sica
 - [x] virtual - only compare - ddc10
 - [ ] integrated - main only - sica
 - [ ] integrated - main only - ddc10
 - [x] estimators - compare only -sica
 - [x] estimators - compare only -ddc10
 - [x] iterations - (compare/main) -sica
 - [x] iterations - (compare/main) -ddc10

 - [x] Need to fix negative boundary distance issue for vpackets: specific case occurs when a packet reaches a boundary with a close line so distance_trace_line = 0 and distance_boundary can be negative (shell ID does not advance)
The models that shall be run is low velocity Si/Ca and DDC10. 

The last step is to get numba to produce a new refdata and look at this in the notebook. 

If all these models run then we agree that numba montecarlo can be merged. 

## Merging Strategy ##

1. talk to TARDIS core 
2. if all tests pass - update the version number to 4.0 with the merged numba montecarlo
3. merge

## Post-merge ##

Start writing unit tests for the numba version
Improve speed.

PR #1115, opened 2020-03-23
Optimisation of Montecarlo scripts using numba 

Debjoy10:
## Description

Improved from previously made pr, https://github.com/tardis-sn/tardis/pull/929 by @wkerzendorf, for the integration of scripts for Monte-Carlo radiative-transfer with python and numba.

### Changes Introduced:
**1. Enabled parallel computation**
Used `@njit(parallel=True)` with `prange` instead of `for` loop over `range` wherever possible.

**2. Used faster `@vectorize`**
In functions involving scalar computation and numerical arguments and single numerical output, `@njit` was replaced with eagerly compiled `@vectorize`.

**3. Improvised for numerical functions with **multiple** return parameters.**
Function `calculate_distance_boundary()` in `r_packet.py`, having 2 return values was changed to function `temp_calculate_distance_line()` which returns one value out of the two computed depending upon the value of temporary variable flag. A wrapper function calls `temp_calculate_distance_line()` twice with `flag` value 1(for distance) and 2(for delta_shell).
```
# Numerical function -
@vectorize([float64(float64, float64, float64, float64, int64)])           
def temp_calculate_distance_boundary(r, mu, r_inner, r_outer, flag):
    distance, delta_shell = ...
    if flag == 1:                  # Determining return variable using value of flag
        return distance
    else:
        return delta_shell


# Wrapper function - 
@njit(**njit_dict)
def calculate_distance_boundary(r, mu, r_inner, r_outer):
    # Calls numerical function twice with different flag values.
    return temp_calculate_distance_boundary(r, mu, r_inner, r_outer, 1), int(temp_calculate_distance_boundary(r, mu, r_inner, r_outer, 2))
```
**4. Improvised for non-numeric inputs.**
Multiple functions imput as arguments, objects of specific classes(Most commonly, `r_packets` and `estimators`). Function conversion to `@vectorize` was similarly obtained using a wrapper function like in above where the numeric values of object varibles were passed to actual function. Changes were made to function `calculate_distance_line()` in `r_packet.py`.
```
# Numerical function - 
@vectorize([float64(float64, float64, float64, float64, float64, float64, int64)])
def temp_calculate_distance_line(mu, nu, r, comov_nu, nu_line, time_explosion, flag):
        distance = ...
        return distance

# Wrapper function -
@njit(**njit_dict)
def calculate_distance_line(r_packet, comov_nu, nu_line, time_explosion, montecarlo_configuration):
    # Extracting useful params    
    nu = r_packet.nu        
    mu = r_packet.mu
    r = r_packet.r

    # Convert Bool value to numeric
    flag = 0                      
    if montecarlo_configuration.full_relativity:
        flag = 1

    # Calling numeric function will proper arguments.
    ret = temp_calculate_distance_line(mu, nu, r, comov_nu, nu_line, time_explosion, flag)
    return ret
```

## Motivation and Context

Eagerly compiled `@vectorize` was observed to execute much faster than lazily compiled `@njit`. However, it can only take scalar/vector numeric inputs and perform scalar operations. This pr provides a way to exploit those functions.

## How Has This Been Tested?

Obtained return values were obtained correct(identical to original functions).
Running `python setup.py test` gave no errors.
Time of executions(obtained using `rapid_test.py`) decreased from - 
(2). 0.078 to 6.65e-05
(3). 0.238 to 0.04
(4). 0.24 to 0.066

## Additional comments
1. In function `calculate_distance_line( )` in script `r_packet.py`, line 100, included **erroneous** operation `r_packet**2`, squaring of a class object. Replaced with `r_packet.r`, for now.
2. Changes 3 and 4, though clearly improving performance, reduce the readability of code. Therefore, one function each was modified in this commit, until further approval.
3. `rapid_test.py` is a temporary script, not supposed to be included in final merge.

## Types of changes

## Checklist:


Debjoy10:
These improvements are introduced at the cost of code-readability, due to introduction of multiple functions for performing a single task.
@wkerzendorf, please review and suggest whether to proceed with this and/or any possible changes.
Also, please look at the possible error mentioned in **Additional Comments** above.

wkerzendorf:
you need to make sure that the result are the same.

Phantsure:
Screenshot_2020-03-25 Screenshot from 2020-03-25 00-16-13 png (PNG Image, 1920 × 1080 pixels) - Scaled (85%)
This is an example. I ran few tests, can't really understand if it is necessary to vectorize the functions as it really doesn't provide much improvement. I am not pointing anything but maybe vectorizing doesn't improve performance

Debjoy10:
Hey @Phantsure,
My bad, I could reproduce your results using `time_it` in jupyter notebook. 
As I noticed while using `time`, 
test1
However, for multiple iterations, the difference almost vanishes.
Screenshot from 2020-03-25 08-33-24
`@njit` takes more time during the first function call, but quite similar time to `@vectorize` for all subsequent calls.
Thanks for pointing that out. :)

Debjoy10:
However, according to testing on my device, for say, up to 100 function calls, the **Total** time for execution for `@vectorize` has a clear advantage over `@njit` given the inefficient first function call.
test2

Any suggestions/corrections are welcome.

shreyansjain012:
Try using `perf_counter` instead of `time.time` it is recommended nowadays. You can refer this... https://stackoverflow.com/questions/7370801/measure-time-elapsed-in-python

PR #1117, opened 2020-03-24
Debugging Numba montecarlo 

shreyansjain012:
Debugging the errors in code. 
## Description

- Missing input arguments
- Outdated numba version
- Non supported data types in nopython mode

Numba version used by tardis does not support ***args in jitclass** see https://github.com/numba/numba/issues/2661 for reference so I use a workaround, for now, later we can update the numba version
Now I am getting this error, I never encountered it while using normal master branch running quickstart example, Looking into it.

image

## Types of changes

## Checklist:


shreyansjain012:
@wkerzendorf Please have a look once, I don't know about this error mentioned above.

PR #1119, opened 2020-03-24
First Objective: Numba Optimization

shailesh1028:
Description
  First Objective: Optimization using Numba

Motivation and Context
Numba is required for faster processing of the code. I have added the jit command over the function intensity_black_body. It considers the datatype in their account. It uses the error_model of NumPy because the operation of division can be handled properly without throwing any error. It also cached the function for further use.

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -3,6 +3,7 @@ import os
 import re
 from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


marxwillia:
Tests are failing.  See if you can figure out why and fix the problem.

PR #1150, opened 2020-05-15
Fix the Plasma tests

wkerzendorf:
In Progress ... Do not merge
There are currently several tests for the plasma calculation. Time to take inventory and clean up

## Description
Fix up PEP8
Try to delete some tests and see if the coverage changes

## Motivation and Context
There seem to be many tests and it is unclear which one them are still needed. We will delete and see if coverage changes.

## How Has This Been Tested?
N/A
 - [ ] remove unnecessary tests
 - [ ] getting codecov to work again


epassaro:
Seems fine. Maybe would be a good idea to change the title of the PR to something more informative (about codecov and deleting tests) to find this in the future (if necessary).

PR #1153, opened 2020-05-18
update doc page to a notebook

sashank27:
Updated [physical quantitities](https://tardis-sn.github.io/tardis/physics/physical_quantities.html#physical-quantities) page to a jupyter notebook


sashank27:
@arjunsavel 

> Does the information exist in radial1d or Simulation? Could they be accessed directly from radial1d?

The physical quantities are mostly either in `Radial1DModel` or `BasePlasma`, both under `Simulation`. So I guess it would be better to use `Simulation` as a base to access these objects and then the subsequent params.

> Similarly regarding the change from accessing tardis_config to accessing model

As mentioned by @chvogl in the chats, `tardis_config` is now removed from the `Simulation` and now these params are inside model

> One thing that I would add is mentioning the use of AstroPy Quantities for simulation.model.v_inner.cgs and related quantities.

Okay, I will update that

PR #1160, opened 2020-05-21
Revert "Add skip condition if QT_API is no set (issue #1101)"

epassaro:
Reverts tardis-sn/tardis#1105.

Not necessary. Should be fixed by #1173.

### DIFF
--- a/tardis/gui/interface.py
+++ b/tardis/gui/interface.py
@@ -9,8 +9,8 @@ else:
         ' export QT_API=pyqt \n\n For more information refer to user guide.')
 import sys
 try:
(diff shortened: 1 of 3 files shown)


jordi5:
Looks good to me!

epassaro:
PR #1101 PR made these changes:

- Switched all classes and functions from PyQt4 to PyQt5. This isn't necessary to make the GUI work since those calls fallback to PyQt5 by default (but looks nicer).

- Removed the `interface` import.

- Introduces the `skipif` condition in `test_gui.py`.

PR #1164, opened 2020-05-26
Fix callbacks in run_tardis

Rodot-:
run_tardis allows users to specify a list of callbacks to be run at each iteration of the simulation.  The current version does not allow a user to specify any extra arguments or access any attributes of the simulation itself.
## Description

when run_tardis allocates callbacks, it now adds them using the *unpacking syntax to allow for arguments to be passed along with the callback function.
Callback assignment now sets the instance of the simulation as the first argument of the callback function.  This allows users to specify functions that can act on the simulation object itself.
## Motivation and Context

The ability to use callbacks with the tardis model was severely limited.  Now callbacks can give live feedback on the simulation as it runs.
## How Has This Been Tested?
I ran through the test modules and none of the failed.  I also wrote a few functions to perform various simple tasks like printing the iteration number at each iteration.

## Types of changes

## Checklist:

### DIFF
--- a/tardis/montecarlo/src/integrator.c
+++ b/tardis/montecarlo/src/integrator.c
@@ -169,7 +169,8 @@ _formal_integral(
   double R_max = storage->r_outer_i[size_shell - 1];
   double pp[N];
   double *exp_tau = calloc(size_tau, sizeof(double));
-#pragma omp parallel firstprivate(L, exp_tau)
+//#pragma omp parallel firstprivate(L, exp_tau)
+#pragma omp parallel 
     {
 
 #pragma omp master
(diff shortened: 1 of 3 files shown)


tardis/montecarlo/src/integrator.c
```diff
@@ -336,8 +337,8 @@ _formal_integral(
           }
         }
       // Free everything allocated on heap
-      free(exp_tau);
       printf("\n");
     }
+      free(exp_tau);
```
wkerzendorf, line 342: I think this is an unintended consequence - did you use the newest upstream/master?

PR #1181, opened 2020-06-02
Draft of new Radial1DModel

jordi5:
I have a couple of ideas and I your input:
Calling Radial1DModel would create a UserModel to store the user parameters along with the proper csvy or yaml model. Then I thinks the 'raw' variables should disappear
There should be a function to set all the attributes (everything that now is under __init__()) or leave it in the generic class and call __super__ on the subclasses.

Remove short variables: w, t_rad.

Why are there private variables line _t_radiative?

```
if t_radiative is None:
            lambda_wien_inner = constants.b_wien / self.t_inner
            self._t_radiative = constants.b_wien / (lambda_wien_inner * (
                1 + (self.v_middle - self.v_boundary_inner) / constants.c))
else:
            self._t_radiative = t_radiative
@property
def t_radiative(self):
    if len(self._t_radiative) == self.no_of_shells:
        return self._t_radiative

    return self._t_radiative[self.v_boundary_inner_index + 1:self.v_boundary_outer_index + 1]

@t_radiative.setter
def t_radiative(self, value):
    if len(value) == len(self._t_radiative):
        self._t_radiative = value
    elif len(value) == self.no_of_shells:
            self._t_radiative[self.v_boundary_inner_index+1:self.v_boundary_outer_index+1] = value
    else:
        raise ValueError("Trying to set t_radiative for unmatching number"
                             "of shells.")
```

seems excesive, but i don't know, and it's repeated with the dilution factor

### DIFF
--- a/Generic_class.ipynb
+++ b/Generic_class.ipynb
@@ -1,0 +1,143 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from astropy import units as u\n",
+    "import numpy as np\n",
+    "from abc import ABC, abstractmethod\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class GenericRadial1DModel():\n",
+    "#In accordance with the factory method, this class should include everything\n",
+    "#needed to construct a product even if the method and atributes are empty     \n",
+    "#TODO: store _0 variables in case we change the time\n",
+    "#set_time would multiply the _0 and return a new variable\n",
+    "    \n",
+    "#These 3 are needed to start a grid, I think the rest of the variables should be in the subclasses __init__()\n",
+    "    def __init__(self, v_inner, v_outer,time_explosion): \n",
+    "        self.v_inner = v_inner\n",
+    "        self.v_outer = v_outer\n",
+    "        \n",
+    "        \n",
+    "        \n",
+    "    @property\n",
+    "    def no_of_shells(self):\n",
+    "        return len(self.v_inner)\n",
+    "\n",
+    "\n",
+    "    @property\n",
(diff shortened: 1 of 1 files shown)


jordi5:
Right now I'm trying to figure out why some parameters are first calculated as private and the declared as properties

harpolea:
It looks to me like the purpose of the private variable `_t_radiative` is to have some way of storing an intermediate value that can then be passed on the `@property` function for `t_radiative`, where it undergoes further processing. Unless `_t_radiative` is used elsewhere in the code, then it could be good if you could come up with a way of eliminating the need for this extra member variable, as the fewer member variables an object has to carry around, the better!

jordi5:
@harpolea How about a lazy property? It's calculated the first time it's called, and then stores the value. The python cookbook discusses it in p.267

harpolea:
So I wasn't actually thinking about the overhead of recomputing properties, but that is almost certainly a good idea for some of these properties whose values will not change over the course of the simulation. My concern was more that (at least from looking at the code pasted above) I don't see why there needs to be a `_t_radiative` at all and why the entire calculation can't just be done in the `t_radiative` property function. That might just be because I need to look at the rest of the code though.

marxwillia:
I think the private variables like _t_radiative is currently an extremely hacky way of preserving the user specified grid properties.  The original input is saved in the private variable, which nothing in the code has access to, and so it will never be corrupted.  The getter then only ever returns the t_radiative values for the shells after the v_boundary cuts are applied.  Unfortunately, we also seem to allow a setter for t_radiative which actually changes the private variable, which i don't think is ideal...

I think by separating the generic user specified model from the "TARDIS" model we will make this much simpler.  Right now we appear to be trying to fit both sets of functionality into one class.  Also, by creating a separate TARDIS model class with the cuts to the grid applied, I think we can eliminate the EXTREMELY CONFUSING v_boundary_inner_index and v_boundary_outer_index which keep track of the indices in the user specified velocity grid of the user specified velocity cuts.  @wkerzendorf do you agree?

PR #1183, opened 2020-06-03
Create code_style_design

wkerzendorf:
Documentation about Code Style and Design


docs/development/code_style_design
```diff
@@ -0,0 +1,5 @@
+#### Code Style and Design Approaches ####
+
```
wkerzendorf, line 2:
```suggestion
We all adhere to PEP8 - all the time
```

PR #1185, opened 2020-06-11
Upgrade pytest version

epassaro:
## Description

- Unpinned `pytest` from `tardis_env3.yml`.
- Installing `pytest-azurepiepelines` from PyPI instead of GitHub.
- Using non boolean skip condition in integration test.
- Using `pytestconfig` instead of `pytest.config`.

## Motivation and Context

@jaladh-singhal reported tests stopped working on Azure. Maybe it's time to move to *pytest* 5.

`pytest.config` it's not strictly deprecated, but its use is discouraged and doesn't work anymore for us (see links below). Instead we are using `pytestconfig` and non-boolean skip conditions.

**See also:**
- https://github.com/pydata/xarray/issues/1353
- https://github.com/pytest-dev/pytest/issues/472
- https://docs.pytest.org/en/latest/historical-notes.html
- https://docs.pytest.org/en/stable/reference.html#pytestconfig

## How Has This Been Tested?

All tests passes on Azure.

## Types of changes

## Checklist:

### DIFF
--- a/tardis_env3.yml
+++ b/tardis_env3.yml
@@ -52,7 +52,7 @@ dependencies:
 
   #Test/Coverage requirements
   - git-lfs
-  - pytest=4
+  - pytest
   - pytest-html
   - requests
   - coverage
--- a/tardis_env3.yml
+++ b/tardis_env3.yml
@@ -52,7 +52,7 @@ dependencies:
 
   #Test/Coverage requirements
   - git-lfs
-  - pytest
+  - pytest=5
   - pytest-html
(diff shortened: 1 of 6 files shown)


tardis_env3.yml
```diff
@@ -52,7 +52,7 @@ dependencies:
 
   #Test/Coverage requirements
   - git-lfs
-  - pytest=4
+  - pytest
```
wkerzendorf, line 55: please pin this.

PR #1188, opened 2020-06-15
fix dotfile creation for plasma graph

sashank27:
Fix the dotfile creation of the plasma graph.

For the tardis quickstart simulation, the graph obtained is as under:

file


wkerzendorf:
I think we can implement this - including the labels in matplotlib and then make it part of a notebook - that goes into the documentation. A zoom/panable svg would be awesome.

wkerzendorf:
@sashank27 how are we progressing on this.

Issue #1125, opened 2020-03-27
Add note that TARDIS only works on linux/mac


wkerzendorf:
Add this in the documentation

Pabla007:
@wkerzendorf I have added that in my documentation already .
https://github.com/tardis-sn/tardis/pull/1061

Phantsure:
Is there any way that we can make it work on Windows too? Maybe using linux subsystem or something like that. Because in my opinion it would be great if we are able to do that.
I can't work on this right now as I am not using Windows.

epassaro:
Closed by #1189

PR #1191, opened 2020-06-17
Add note to tell users which OS are supported (issue #1125)

jordi5:
TARDIS only works on MacOs and Linux distributions and we need to let the users know.
## How Has This Been Tested?
`python setup.py build_docs`  works!

## Screenshot
Screenshot from 2020-06-17 12-36-53

## Types of changes

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -18,7 +18,7 @@ simple TARDIS calculations.
         
     - TARDIS is only compatible with Python >=3.6
     - TARDIS only supports Linux and MacOS.
-    - We strongly recommend installing TARDIS within an Anaconda environment and to always use the latest github development version.
+    - We strongly recommend installing TARDIS within an Anaconda environment and to always use the latest GitHub development version.
 
 Requirements
 ============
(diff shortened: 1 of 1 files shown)


vamads:
The s in MacOS should be capitalized 
Python is compatible with >=3.6

docs/installation.rst
```diff
@@ -14,13 +14,11 @@ simple TARDIS calculations.
...
-
-    TARDIS is only compatible with Python >3.6
-
 .. note::
-    We strongly recommend installing TARDIS within an Anaconda environment and
-    to always use the latest GitHub development version.
+        
+    - TARDIS is only compatbile with Python >3.6
+    - TARDIS only supports Linux and MacOs.
+    - We strongly recommend to install TARDIS within an Anaconda environment and to always use the lastest github development version.
```
vamads, line 21: -Latest is misspelled 
- perhaps could use installing instead of to install 
-Do you think you should use the capitalized version of GitHub?

jordi5: Thanks! Great suggestions, I fixed all of them.

epassaro:
Capitalize `github` -> `GitHub`

PR #1210, opened 2020-06-28
[MRG] Using Mamba in Pipelines

kaushik94:
Do not merge this PR, this is a WIP. Checking if mamba installation is successful on the Azure pipelines.


kaushik94:
Looks like I have to add mamba to PATH.

epassaro:
> Looks like I have to add mamba to PATH.

I think the problem is you're calling `conda` before installing Miniconda.

epassaro:
Remember the `mamba` binary lives in the Miniconda `base` environment.

azure-pipelines/simple_test_framework.yml
```diff
@@ -29,6 +29,10 @@ jobs:
       maxParallel: 4
 
     steps:
+      - bash: |
+          echo "##vso[task.setvariable variable=shellopts]errexit"
+        displayName: "Force exit on error (bash) "
+
```
kaushik94, line 35: @epassaro Is this correct?

epassaro: Yes, but we must try it. For example, a new bash step including:

```
touch -j
echo "hello world"
```

... should fail.

azure-pipelines/simple_test_framework.yml
```diff
@@ -30,33 +30,56 @@ jobs:
...
+      - bash: |
+              echo "##vso[task.prependpath]$CONDA/bin"
         displayName: Add conda to PATH
 
       - bash: |
-          sudo chown -R $USER $CONDA
-          # conda update -y conda
+              sudo chown -R $USER $CONDA
```
epassaro, line 41: Split this step in two steps. The first one does the `chown` thing and must run exclusive for macOS. The second one does the update, but comment it out.

azure-pipelines/simple_test_framework.yml
```diff
@@ -30,33 +30,56 @@ jobs:
...
-          source activate tardis
-          pip install pytest-azurepipelines
-          pytest tardis --tardis-refdata=$(ref.data.home) --cov=tardis --cov-report=xml --cov-report=html
-
+              source activate tardis
+              pip install pytest-azurepipelines
+              pytest tardis --tardis-refdata=$(ref.data.home) --cov=tardis --cov-report=xml --cov-report=html
         displayName: "TARDIS test"
+
       - script: |
```
epassaro, line 83: Use `bash` instead of `script`.

azure-pipelines/simple_test_framework.yml
```diff
@@ -30,33 +30,56 @@ jobs:
...
+              conda install mamba=0.4.0 -c conda-forge -y
+        displayName: "Install mamba"
+
+      - bash: |
+              mamba env create -f tardis_env3.yml
         displayName: "Install TARDIS env"
 
       - bash: |
-          sh ci-helpers/fetch_reference_data.sh
+              # sh ci-helpers/fetch_reference_data.sh
```
epassaro, line 54: Remove all the comments and the last `cd` line.

azure-pipelines/simple_test_framework.yml
```diff
@@ -30,33 +30,56 @@ jobs:
...
+              # Use the following to get the ref-data from the master;
+              git fetch origin
+              git checkout origin/master
+              # Use the following to get the ref-data from a specific pull request
+              #git fetch origin pull/20/head:update-ref
+              #git checkout update-ref
+              git lfs pull --include="atom_data/kurucz_cd23_chianti_H_He.h5" origin
+              git lfs pull --include="atom_data/chianti_He.h5" origin
+              git lfs pull --include="plasma_reference/" origin
+              git lfs pull --include="unit_test_data.h5" origin
```
epassaro, line 67: Add `git lfs pull --include="packet_unittest.h5" origin`

azure-pipelines/simple_test_framework.yml
```diff
@@ -65,6 +65,7 @@ jobs:
               git lfs pull --include="atom_data/chianti_He.h5" origin
               git lfs pull --include="plasma_reference/" origin
               git lfs pull --include="unit_test_data.h5" origin
+              git lfs pull --include="packet_unittest.h5" origin
               echo MD5 `md5sum unit_test_data.h5`
```
epassaro, line 69: add a colon after `MD5`

azure-pipelines/simple_test_framework.yml
kaushik94: Can you explain how to do this? Because the jobs don't run if the previous job fails. So how to use this variable?

epassaro: It's only a test. I want to make it fail on purpuse. 

With the normal Azure Pipelines behavior a step like

```bash: |
            touch -j
            echo "hello world"
```

Returns a green check and keep running the pipeline. That behavior is not desired because we want to know if something is failing (and stop the pipeline). 

If the `#vso...` piece of code works as we want, then the pipeline should fail if any bash command fails. But we don't know, nobody made this before.

See: https://github.com/microsoft/azure-pipelines-agent/issues/1803

kaushik94: @epassaro would it be a good idea to add each bash command as a step in the `steps:`. This way we can describe that each step is dependent on the previous steps and then when any of the steps fail, the pipeline fails right away.

kaushik94:
>Also try to roll back the changes in both helpers scripts, since we are not using them anymore. Probably will be deleted in the future.

@epassaro so I believe you already did that with #1242 ?

epassaro:
I made some changes to the fetch refdata bash script. Probably is the one we will keep, the other two files are not needed anymore.

Keep only the changes in the YAML file in this PR.

PR #1201, opened 2020-06-24
Code style black

wkerzendorf:
Switching code-style to black - one `subpackage` at a time.

PR #1213, opened 2020-06-30
Re-applied Black formatting to tardis after excluding undesired directories

jaladh-singhal:
This PR aims to exclude undesired directories from black formatting. When running black, there are still 10 files which were formatted blacked (which were not covered earlier run, because it was run not on root of project but inside `tardis/`).

A brief summary of changes in each file are:
- `pyproject.toml` - Added directories to exclude from black (including `astropy_helpers` which was being formatted by black)
- `tardis_env3.yml` - Added black as a code quality dependency, so that it is already installed in a new tardis environment
- 10 files formatted by black are distributed as follows:
  - `tardis/plasma/` - 2 files - they are actually one line changes, made by us @wkerzendorf when editing strings - those strings were quite long, so black line-breaked them as expected :white_check_mark: 
  - `docs/` - There are 4 python files inside it
  - `asv/` - 1 file
  - `ah_bootstrap.py`, `setup.py`, `ez_setup.py` are remaining 3 at root directory.

## Motivation and Context
Related #1201 

## How Has This Been Tested?

They're only formatting changes so ideally they shouldn't break the code. And if test pipeline pass, they are good to merge.

### DIFF
--- a/asv/benchmarks/benchmarks.py
+++ b/asv/benchmarks/benchmarks.py
@@ -6,17 +6,21 @@ from tardis.tests import montecarlo_test_wrappers as montecarlo
 
 LINE_SIZE = 10000000
 
+
 class TimeSuite:
     """
     An example benchmark that times the performance of various kinds
     of iterating over dictionaries in Python.
     """
+
     def setup(self):
         self.line = np.arange(LINE_SIZE, 1, -1).astype(np.float64)
 
     def time_binarysearch(self):
         for _ in range(LINE_SIZE):
-            montecarlo.binary_search_wrapper(self.line, np.random.random() * LINE_SIZE, 0, LINE_SIZE - 1)
+            montecarlo.binary_search_wrapper(
+                self.line, np.random.random() * LINE_SIZE, 0, LINE_SIZE - 1
+            )
 
     def time_compute_distance2outer(self):
         for _ in range(1000000):
@@ -25,7 +29,7 @@ class TimeSuite:
             montecarlo.compute_distance2outer_wrapper(0.3, 1.0, 1.0)
             montecarlo.compute_distance2outer_wrapper(0.3, -1.0, 1.0)
             montecarlo.compute_distance2outer_wrapper(0.5, 0.0, 1.0)
-    
+
     def time_compute_distance2inner(self):
         for _ in range(1000000):
             montecarlo.compute_distance2inner_wrapper(1.5, -1.0, 1.0)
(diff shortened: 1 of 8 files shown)


ah_bootstrap.py
```diff
@@ -609,12 +649,15 @@ def _check_submodule_using_git(self):
             # which only occurs with a malformatted locale setting which can
             # happen sometimes on OSX.  See again
             # https://github.com/astropy/astropy/issues/2749
-            perl_warning = ('perl: warning: Falling back to the standard locale '
-                            '("C").')
+            perl_warning = (
+                "perl: warning: Falling back to the standard locale " '("C").'
```
jaladh-singhal, line 653:
```suggestion
                'perl: warning: Falling back to the standard locale ("C").'
```

docs/conf.py
```diff
@@ -89,27 +89,28 @@
 # -- Project information ------------------------------------------------------
 
 # This does not *have* to match the package name, but typically does
-project = u'TARDIS'
-author = u'TARDIS collaboration'
-copyright = u'2013, ' + author
+project = "TARDIS"
```
jaladh-singhal, line 92: Unicode string doesn't make sense in Python 3.6+, so black removes them:
> if your code is already Python 3.6+ only or it’s using the unicode_literals future import, Black will remove u from the string prefix as it is meaningless in those scenarios.

ez_setup.py
```diff
@@ -32,23 +32,30 @@
...
+
+
+vars(subprocess).setdefault("check_call", _check_call_py24)
+
 
 def _install(tarball, install_args=()):
     # extracting the tarball
     tmpdir = tempfile.mkdtemp()
-    log.warn('Extracting in %s', tmpdir)
+    log.warn("Extracting in %s", tmpdir)
```
jaladh-singhal, line 58:
```suggestion
    log.warn(f"Extracting in {tmpdir}")
```

ez_setup.py
```diff
@@ -59,13 +66,13 @@ def _install(tarball, install_args=()):
         # going in the directory
         subdir = os.path.join(tmpdir, os.listdir(tmpdir)[0])
         os.chdir(subdir)
-        log.warn('Now working in %s', subdir)
+        log.warn("Now working in %s", subdir)
```
jaladh-singhal, line 69:
```suggestion
        log.warn(f"Now working in {subdir}")
```

ez_setup.py
```diff
@@ -76,7 +83,7 @@ def _install(tarball, install_args=()):
 def _build_egg(egg, tarball, to_dir):
     # extracting the tarball
     tmpdir = tempfile.mkdtemp()
-    log.warn('Extracting in %s', tmpdir)
+    log.warn("Extracting in %s", tmpdir)
```
jaladh-singhal, line 86:
```suggestion
    log.warn(f"Extracting in {tmpdir}")
```

ez_setup.py
```diff
@@ -87,45 +94,53 @@ def _build_egg(egg, tarball, to_dir):
         # going in the directory
         subdir = os.path.join(tmpdir, os.listdir(tmpdir)[0])
         os.chdir(subdir)
-        log.warn('Now working in %s', subdir)
+        log.warn("Now working in %s", subdir)
```
jaladh-singhal, line 97:
```suggestion
        log.warn(f"Now working in {subdir}")
```

jaladh-singhal:
Excluded `ah_bootstrap.py, setup.py, ez_setup.py` and `docs/conf.py`, rest 6 files formatted by black looks good to me!

PR #1222, opened 2020-07-07
Remove legacy model definition from csvy schema and tests

jordi5:
## Description
In this PR we want to remove the ability for the user to still use configurations similar to the ones in `Radial1DModel.from_config()`.
This is an ongoing effort to have clearly distinct ways to define a model with which TARDIS will work. See [project 15](https://github.com/tardis-sn/tardis/projects/15) and issue  #1239.  
 - Removed references to density, velocity and abundances from `io/schemas/csvy_model.yml` 
 - Removed files in tests that tried to use this feature, and use them in a new test `test_unsupported_field` that is expected to fail.
 - Rename function `filename` to `filename_tuples` for a little more clarity

## Motivation and Context

We first need to merge this PR to be able to change `model/base.py` .

## How Has This Been Tested?
`pytest` tests and xfail tests pass, note that we are still comparing results from the `from_config` method and those test work fine.

## Types of changes

## Checklist:


marxwillia:
@jordi5 The tests that you removed are actually still useful.  They are good examples of CSVY files that we should disallow.  Leave one of the tests in and make sure that it catches the disallowed CSVY fields and raises an exception with appropriate message.

wkerzendorf:
@marxwillia @harpolea what is going on with this?

PR #1225, opened 2020-07-08
add initial contributor guidelines

wkerzendorf:
adding new contributor guidelines

### DIFF
--- a/docs/development/contributing.rst
+++ b/docs/development/contributing.rst
@@ -1,0 +1,28 @@
+**********************
+Contributor guidelines
+**********************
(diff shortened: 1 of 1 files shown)


jaladh-singhal:
What other things would we like to be added - I think some pointers will be great to have @wkerzendorf @ycamacho @marxwillia?

arjunsavel:
Should we point to [this part of the documentation](https://tardis-sn.github.io/tardis/development/git_workflow.html)?

jaladh-singhal:
This pull request can be closed since #1237 already makes these changes!

Issue #1228, opened 2020-07-08
Radial1DModel.from_csvy() returns wrong abundances 


jordi5:
#### Code Sample
download or copy this 2 files [gist](https://gist.github.com/jordi5/d7e828184342e482fea354292ac27bc4)

And then, in a Jupyter Notebook, run this code
  ```
from astropy import units as u
from tardis.io.config_reader import Configuration
from tardis.io.atom_data.base import AtomData
from tardis import run_tardis
from tardis.model import Radial1DModel

tardis_config = Configuration.from_yaml('tardis_csvy_example.yml')
model = Radial1DModel.from_csvy(tardis_config)
```
and look at `model.raw_abundances` and `model.raw_isotope_abundances` they should have the same number of columns but they don't. And the elemental abundances doesn't match those that where used as input. 
#### Problem description
When reading csv data in, the first line is discarded by TARDIS. This is not done by the reader, it's done after reading the abundances and isotope abundance(when they are converted to dataframes and those rows become columns). The step that removes the first column from the isotope abundances is missing, so both dataframes end up with mismatching number of columns. These dataframes are further processed to check that the abundances add up to 1 (and normalize if they are not), but having different number of shells will change all of them, unless they are uniform, in which case only the first and last shell are affected.

#### Expected Output
`model.raw_abundances` and `model.raw_isotope_abundances` should have the same number of columns. and they should match the one submitted in the file

This issue only affects Radial1DModel.from_csvy() and not Radial1DModel.from_config()

PR #1229, opened 2020-07-09
Fix from_csvy_reader abundances drataframes (Issue #1228)

jordi5:
## Description
- Add lines that remove the first line of the isotope_abundances dataframe
- Add test to check that both dataframes have the same number of columns

## Motivation and Context
See Issue #1228 for a full explanation.

## How Has This Been Tested?
`pytest tardis` test passes 

## Types of changes


marxwillia:
The bug fix and current test look good.  I think it would be wise to have a test where you do the decay calculation by hand for a very simple model and compare the results of the radial1DModel abundances and isotope abundances to what we would expect.

tardis/model/tests/data/csvy_vfilter.csvy
```diff
@@ -16,15 +16,13 @@ datatype:
        desc: fractional H abundance
     -  name: He
        desc: fractional He abundance
+    -  name: Ni56
+       desc: fractional Ni56 abundance
 
-v_inner_boundary: 11000 km/s
-v_outer_boundary: 14500 km/s
```
jordi5, line 21: V-inner_boundary changed because I plan on using this file for other tests

jordi5:
I added checks for the fataframes before the decay. the decay test doesn't have the precision I want ([docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html) says that it checks 3 or 5 decimal numbers but the test doesn't pass if I try to check 5 which it should), I will see if I can improve it

PR #1245, opened 2020-07-24
Include user-input seed when setting rpacket_seed

arjunsavel:
## Description

The user-input `montecarlo seed` is now used to generate an array of seeds for packets, which are set within `montecarlo_main_loop`. This seed is added with the current simulation iteration number before generating the packet seeds to preserve randomness across different iterations. The packet seeds are then attached to each associated r_packet.

To accommodate this change, functions in packet_source and the call to create the packet seeds now utilize a default random generator from NumPy v1.19.0, which is currently being tested on the main branch. The current environment being tested is used to limit rebase issues later on.

## Motivation and Context

This ensures reproducibility (by seed-setting) while also recovering the ability for users to set their own seeds. This allows us, for example, to explore the effects of Poisson error on a given spectrum.

## How Has This Been Tested?

N/A (as of now, but happy to test if there are suggestions!)

## Types of changes

## Checklist:


tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -56,14 +56,22 @@ def montecarlo_main_loop(packet_collection, numba_model, numba_plasma,
...
-                           i)
-
-        # We want to set the seed correctly per user; otherwise, random.
-        np.random.seed(i)
+            r_packet = RPacket(numba_model.r_inner[0],
+                               packet_collection.packets_input_mu[i],
+                               packet_collection.packets_input_nu[i],
+                               packet_collection.packets_input_energy[i],
+                               i)
+            np.random.seed(i)
```
chvogl, line 64: I am a little bit lost here. As far as I can tell, the purpose of the `single_packet_seed` is to re-run a packet for debugging purposes. The index of this original packet is the single_packet_seed. So why do we not use the same seed as for the original packet (i.e. `seed = packet_seeds[i]`)? However, my interpretation of the purpose and implementation of the `single_packet_seed` might be completely wrong.

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -56,14 +56,22 @@ def montecarlo_main_loop(packet_collection, numba_model, numba_plasma,
...
+        else:
+            r_packet = RPacket(numba_model.r_inner[0],
```
chvogl, line 66: At first glance, this looks identical to lines 59-63. Is there a reason why this cannot be moved outside of the if-else clause?

tardis/montecarlo/montecarlo_numba/base.py
arjunsavel: This way, we can set `single_packet_seed` to be whatever we'd like — so if we run a simulation and everything works well except for packet 725, we can just input that packet number directly and run a simulation without any other packets.

tardis/montecarlo/montecarlo_numba/base.py
arjunsavel: I think you're right — I'm pretty sure this was for debugging purposes a while back. I'll change this!

tardis/montecarlo/base.py
```diff
@@ -128,6 +129,10 @@ def _initialize_geometry_arrays(self, model):
         self.v_inner_cgs = model.v_inner.to('cm/s').value
 
     def _initialize_packets(self, T, no_of_packets):
+        np.random.seed(self.seed)
+        mc_config_module.packet_seeds = np.random.randint(0,
+                                                          1e9,
```
wkerzendorf, line 134: why 1e9?

tardis/montecarlo/montecarlo_configuration.py
```diff
@@ -2,4 +2,6 @@
 single_packet_seed = -1
 temporary_v_packet_bins = None
 number_of_vpackets = 0
-line_interaction_type = None
\ No newline at end of file
+montecarlo_seed = 0
```
wkerzendorf, line 5: are these constants?

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -55,15 +55,17 @@ def montecarlo_main_loop(packet_collection, numba_model, numba_plasma,
 
     for i in prange(len(output_nus)):
         if montecarlo_configuration.single_packet_seed != -1:
-            i = montecarlo_configuration.single_packet_seed
+            seed = packet_seeds[montecarlo_configuration.single_packet_seed]
```
wkerzendorf, line 58: I think it is best if this is attached to the packet object

tardis/montecarlo/packet_source.py
```diff
@@ -108,8 +111,25 @@ class BlackBodySimpleSource(BasePacketSource):
...
-        nus = self.create_blackbody_packet_nus(T, no_of_packets)
-        mus = self.create_zero_limb_darkening_packet_mus(no_of_packets)
+        self.T = T
+        self.seeds = mc_config_module.packet_seeds
+        nus = self.create_nus()
+        mus = self.create_mus()
         energies = self.create_uniform_packet_energies(no_of_packets)
+        return nus, mus, energies
+
+    def create_nus(self):
```
wkerzendorf, line 121: we should talk about that.

tardis/montecarlo/packet_source.py
```diff
@@ -108,8 +111,25 @@ class BlackBodySimpleSource(BasePacketSource):
...
+        xis = np.zeros((5, len(self.seeds)))
+        for i, seed in enumerate(self.seeds):
+            xis[:, i] = self.create_array(seed)
+        return self.create_blackbody_packet_nus(self.T, xis)
+
+    @staticmethod
+    @njit
+    def create_array(seed):
+        np.random.seed(seed)
+        return np.random.random(5)
```
wkerzendorf, line 131: why 5?

tardis/montecarlo/packet_source.py
```diff
@@ -108,8 +111,25 @@ class BlackBodySimpleSource(BasePacketSource):
...
+
+    def create_nus(self):
+        xis = np.zeros((5, len(self.seeds)))
+        for i, seed in enumerate(self.seeds):
+            xis[:, i] = self.create_array(seed)
+        return self.create_blackbody_packet_nus(self.T, xis)
+
+    @staticmethod
+    @njit
+    def create_array(seed):
```
wkerzendorf, line 129: what array?

tardis/montecarlo/base.py
arjunsavel: I picked a value that was a few orders of magnitude greater than the largest number of packets I've seen; should I base it instead on `no_of_packets`?

tardis/montecarlo/montecarlo_configuration.py
arjunsavel: Once they're set by the configuration file they remain constants afterward, yes

tardis/montecarlo/packet_source.py
arjunsavel: I'm not really sure — I was matching the shape of the original `xis` shape within the blackbody function. I'll take a closer look at the reference material!

tardis/montecarlo/packet_source.py
arjunsavel: I'll name it more specifically!

tardis/montecarlo/packet_source.py
arjunsavel: Had to with the sampling method; I've added further documentation that should clear it up.

tardis/montecarlo/base.py
```diff
@@ -127,11 +134,21 @@ def _initialize_geometry_arrays(self, model):
...
         self.v_inner_cgs = model.v_inner.to('cm/s').value
 
-    def _initialize_packets(self, T, no_of_packets):
+    def _initialize_packets(self, T, no_of_packets, iteration):
+        # the iteration is added each time to preserve randomness
+        # across different simulations with the same temperature,
+        # for example.
+        np.random.seed(self.seed + iteration)
+        if no_of_packets > MAX_SEED_VAL:
+            raise ValueError('Too many packets to seed.')
```
wkerzendorf, line 143: change this description

tardis/montecarlo/base.py
```diff
@@ -127,11 +135,23 @@ def _initialize_geometry_arrays(self, model):
...
-    def _initialize_packets(self, T, no_of_packets):
+    def _initialize_packets(self, T, no_of_packets, iteration):
+        # the iteration is added each time to preserve randomness
+        # across different simulations with the same temperature,
+        # for example. We seed the random module instead of the numpy module
+        # because we call random.sample, which references a different internal
+        # state than in the numpy.random module.
+        random.seed(self.seed + iteration)
+        if no_of_packets > MAX_SEED_VAL:
+            raise ValueError(f"""Numpy's random generator cannot create
```
wkerzendorf, line 146: ... that means one can not specify more packets than that.

tardis/montecarlo/base.py
```diff
@@ -127,11 +135,27 @@ def _initialize_geometry_arrays(self, model):
...
 
-    def _initialize_packets(self, T, no_of_packets):
+    def _initialize_packets(self, T, no_of_packets, iteration):
+        # the iteration is added each time to preserve randomness
+        # across different simulations with the same temperature,
+        # for example. We seed the random module instead of the numpy module
+        # because we call random.sample, which references a different internal
+        # state than in the numpy.random module.
+        seed = self.seed + iteration
+        if no_of_packets > MAX_SEED_VAL:
```
wkerzendorf, line 145: this is no longer true, right?

arjunsavel: We still call np.random.seed in `montecarlo_main_loop`, so if an rpacket has a seed higher than `MAX_SEED_VAL`, the error will be thrown then.

This can be fixed if I change all the random calls to reflect the new NumPy random module update; I was planning to do that in a separate PR

tardis/montecarlo/base.py
```diff
@@ -24,10 +24,17 @@
     configuration_initialize)
 
 import numpy as np
+import random
```
wkerzendorf, line 27: this is not needed anymore, right?

arjunsavel: that's correct!

PR #1249, opened 2020-07-27
Remove decode method

epassaro:
## Description

- Remove `.decode("ascii")` method when reading atom data attributes. 
- In `tardis/tests/fixtures/atom_data.py` changed `DEFAULT_ATOM_DATA_UUID` to bytestring.

## Motivation and Context
It's not strictly necessary to encode strings before storing them inside the HDF5 metadata (unicode is default in Python 3). Carsus is currently signing atomic files **without** ASCII encoding, so we must prevent TARDIS from trying to decode these strings, otherwise will fail.

For older atomic files made with previous versions of Carsus (for example atomic data in `tardis-refdata`) the MD5 and UUID is stored as bytestrings.

> If we decide not to merge this PR, then we should re-add `.encode('ascii')` method in Carsus signing function.

## How Has This Been Tested?
```
 pytest tardis/io/tests --tardis-refdata=$HOME/Desktop/tardis-sn/tardis-refdata --remote-data
```

## Types of changes

## Checklist:

### DIFF
--- a/tardis/io/atom_data/base.py
+++ b/tardis/io/atom_data/base.py
@@ -166,12 +166,12 @@ class AtomData(object):
             atom_data = cls(**dataframes)
 
             try:
-                atom_data.uuid1 = store.root._v_attrs["uuid1"].decode("ascii")
+                atom_data.uuid1 = store.root._v_attrs["uuid1"]
             except KeyError:
                 atom_data.uuid1 = None
 
             try:
-                atom_data.md5 = store.root._v_attrs["md5"].decode("ascii")
+                atom_data.md5 = store.root._v_attrs["md5"]
             except KeyError:
                 atom_data.md5 = None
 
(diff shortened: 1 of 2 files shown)


epassaro:
It's really strange but I don't know why removing this method skips all tests in `tardis/plasma/tests/test_hdf_plasma.py`. I'll mark this as draft for now.

wkerzendorf:
Now that I see that - I actually do not think that this is a good idea. We should be saving these things as strings. byte streams are very confusing and a pain in the ass.

wkerzendorf:
my apologies

epassaro:
Just for clarification:

- Old Carsus atomic files have MD5 and UUID1 stored as bytestrings, so TARDIS must decode them.
- The actual version of Carsus stores MD5 and UUID as strings, no need to decode.

The "problem" is TARDIS must handle these two kinds of files (the actual patch we are discussing). Or just modify Carsus again and store bytestrings as we used to.

epassaro:
Not needed. Closed by https://github.com/tardis-sn/carsus/pull/191

PR #1187, opened 2020-06-15
Add line info widget to TARDIS: demo notebook

jaladh-singhal:
This PR aims to share a notebook for experimentation and code that will be used to create line info widgets.

> NOTE: This is not intended for merging.

## Motivation and Context
This a part of GUI project to create line info widget for TARDIS.

## How Has This Been Tested?
All the code in notebook works fine by executing cells.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #1264, opened 2020-08-05
Add line info widget to widgets subpackage

jaladh-singhal:
This PR aims to put line info widget which was being prepared in a notebook (#1187) as a module within the widgets subpackage so that it can be generated easily by constructing the widget object.

The main changes made are:
- Edited `analysis.py` to make it work with line inf widget:
  - Classmethod `from_model` did not take `packet_filter_mode` as argument but it is required for initializing LastLineInteraction (as `__init__()` takes it), so added it
  - Added filter mode `packet_in_nu` and renamed `packet_nu` to `packet_out_nu` since it only concerned with `output_nu`
  - output_nu and last_interaction_in_nu didn't always had "Hz" unit, so made sure that they are converted to Quantity 1st

- Added `line_info.py` module with the class to create line info widget
- Renamed `base.py` which contained only shell info widget code to `shell_info.py` and similarly done this renaming in tests and `__init__.py`
- Added `util.py` as a helper functions module to current widgets modules
  - `create_table_widget` was common in both `line_info.py` and `shell_info.py`, so removed it from `shell_info.py` and moved it there
  - Added a `TableSummaryLabel` class for creating summary label targetting a table widget (required for handling creation/updation of Total Packets label in last line counts table)
- Added `test_line_info.py` for testing line info module. Note, there's no test file for util.py because it doesn't have any back-end logic but only front-end logic (of creating widgets) and since line_info module uses it so it's already covered.

## To-Dos
- [x] Improvement in line info widget as [described here](https://github.com/tardis-sn/tardis/projects/5#card-42841884)
- [x] Add all the functions from notebook and make the OO interface work
- [x] Add docstrings in module
- [x] Create unit-tests
- [x] Add documentation (implemented in #1280 since it makes changes to entire widgets docs not only line info)


epassaro:
Noticed there are some Pandas `ix` method in `analysis.py` . This method is already deprecated and should be replaced by `loc` or `iloc`.

jaladh-singhal:
> Noticed there are some Pandas `ix` method in `analysis.py` . This method is already deprecated and should be replaced by `loc` or `iloc`.

@epassaro Yes that file is very old, I'm actually using only very few methods of `LastLineInteraction` class in `line_info.py` and none of them are using `ix`. Now that you have mentioned it, perhaps we can scrape this file altogether and mimic only the functionality I require from it within the `line_info.py` itself!

tardis/analysis.py
```diff
@@ -37,9 +40,12 @@ def __init__(
         self.last_line_interaction_shell_id = last_line_interaction_shell_id[
             mask
         ]
-        self.last_line_interaction_angstrom = output_nu.to(
-            u.Angstrom, equivalencies=u.spectral()
-        )[mask]
+        self.last_line_interaction_out_angstrom = u.Quantity(
+            output_nu[mask], "Hz"
+        ).to(u.Angstrom, equivalencies=u.spectral())
+        self.last_line_interaction_in_angstrom = u.Quantity(
```
MarkMageeAstro, line 46: It might be better to rename output_nu and last_interaction_in_nu so that they are consistent. 
e.g. output_nu and input_nu

tardis/widgets/line_info.py
```diff
@@ -0,0 +1,749 @@
+"""Class to create and display Line Info Widget."""
```
MarkMageeAstro, line 1: Might be good to say what the line info widget is

tardis/widgets/line_info.py
```diff
@@ -0,0 +1,749 @@
...
+from tardis.widgets.util import create_table_widget, TableSummaryLabel
+
+
+class LineInfoWidget:
+    """
+    Widget to explore atomic lines that produced features in the simulated spectrum.
+
+    It allows selection of a wavelength range in the spectrum to display a 
+    table giving the fraction of packets that experienced their last
+    interaction with each species. Using toggle buttons, user can specify
```
MarkMageeAstro, line 23: ...users can specify...

tardis/widgets/line_info.py
```diff
@@ -0,0 +1,749 @@
...
+        ----------
+        lines_data : pd.DataFrame
+            Data about the atomic lines present in simulation model's plasma
+        line_interaction_analysis : dict of tardis.analysis.LastLineInteraction
+            Dictionary in which keys are the FILTER_MODES and values are the
+            LastLineInteraction objects initialized with corresponding modes
+        spectrum_wavelength : astropy.Quantity
+            Wavelength values of a real spectrum, having unit of Angstrom
+        spectrum_luminosity_density_lambda : astropy.Quantity
+            Luminosity density lambda values of a real spectrum, having unit
```
MarkMageeAstro, line 62: I think we discussed before, but the units should probably be written as 'erg / s' rather than 'Hz erg'

MarkMageeAstro:
This is a pretty minor comment that likely doesn't affect many people, but I've updated my notebooks to have a dark background. This makes it difficult to read some of the text, as it's black on a dark background. This is shown in the attached image. Maybe adding a white background or changing the colour to something like red would fix this?

Screenshot 2020-09-09 at 16 36 27

tardis/widgets/line_info.py
```diff
@@ -0,0 +1,749 @@
...
+        ----------
+        wavelength_range : list-like or None
+            A list of two float values to specify the wavelength range - first 
+            for the range start and second for the range end. None specifies
+            that no wavelength range is selected and will return empty dataframe
+        filter_mode : str, optional
+            Filter mode of the LastLineInteraction object to use for filtering
+            the selected wavelength range (more details in Notes section).
+            Allowed values are given by the class variable :code:`FILTER_MODES`
+            (default value is :code:`FILTER_MODES[0]`)
```
MarkMageeAstro, line 156: Should probably specify here what FILTER_MODES[0] actually is

tardis/widgets/line_info.py
```diff
@@ -0,0 +1,749 @@
...
+        Produce a plotly figure widget by plotting the spectrum of model.
+
+        Parameters
+        ----------
+        wavelength : astropy.Quantity
+            Wavelength values of a real spectrum, having unit of Angstrom
+        luminosity_density_lambda : astropy.Quantity
+            Luminosity density lambda values of a real spectrum, having unit
+            of (Hz erg)/Angstrom
+        virt_wavelength : astropy.Quantity
```
jamesgillanders, line 462: here's the unit thing Mark mentioned above again - erg/s/Angstrom

MarkMageeAstro:
Some tests are failing. Looks like it's related to the integrated spectrum. Some other recent PRs aren't failing, so could need rebasing?

tardis/widgets/tests/test_line_info.py
```diff
@@ -0,0 +1,345 @@
...
+        ([3000, 4000], "packet_in_nu"),
+        ([16200, 16300], "packet_out_nu"),
+        (None, "packet_in_nu"),
+    ],
+)
+class TestLineInfoWidgetData:
+    """Tests for methods that handles data in LineInfoWidget."""
+
+    def test_get_species_interactions(
+        self, line_info_widget, wavelength_range, filter_mode
```
MarkMageeAstro, line 28: Can we add a description of exactly what this test is looking for and what constitutes a fail?

MarkMageeAstro (changes requested):
Some comments left for minor changes. Mostly about adding comments, etc.

jamesgillanders (changes requested):
See comments above from myself and others - if these are addressed then I am happy to merge

tardis/widgets/line_info.py
```diff
@@ -0,0 +1,749 @@
...
+                ),
+            ],
+            layout=go.Layout(
+                title="Spectrum",
+                xaxis=dict(
+                    title=self.axis_label_in_latex(
+                        "Wavelength", wavelength.unit
+                    ),
+                    exponentformat="none",
+                    rangeslider=dict(visible=True),
```
ycamacho, line 502: I am assuming that this line is for the plot slider under the spectrum plot. If so, it seems like you do not chose the wavelength range of this plot. Is this correct?
Screen Shot 2020-09-10 at 9 53 45 PM

The reason I ask is because the current plot has a wavelength range that extends past the spectrum (the red circle in the picture). There is a range key in rangeslider that might fix this (https://plotly.com/python/reference/layout/xaxis/#layout-xaxis-rangeslider-range). If this isn't an easy fix please don't worry - it's a minor issue.

ycamacho (changes requested):
This is a great PR @jaladh-singhal! We have finished our review of it and have asked for minor changes. Please look them over - if anything is taking too long to figure out we can make it an issue to fix for later. This PR is nearly ready to be merged!

tardis/widgets/line_info.py
jaladh-singhal: It's right next in following line.

tardis/widgets/line_info.py
jaladh-singhal: Yes I agree with @ycamacho and in built api docs, the class properties are also enlisted so it will be clear for readers!

tardis/widgets/line_info.py
jaladh-singhal: Yes this was written way earlier before I changed the units, good catch - fixed it now!

tardis/widgets/line_info.py
jaladh-singhal: Yes I didn't chose the range. The problem is actually not with rangeslider but because the hidden scatter point which bring box selection lies on rightes edge of plot due to using wavelength[0] and luminosity[0]. Plotly tends to add padding (extra space) around it, hence it looks odd as if it's due to extended range of spectrum. There's an already [open issue](https://github.com/plotly/plotly.js/issues/1775) for it and it's not yet have been worked upon.

Though I've managed to fix it by a workaround i.e. calculating an index which brings this single hidden scatter point in exact middle so that extra padding around has no effect.

ycamacho (approved):
Looks ready to merge to me!

MarkMageeAstro (approved):
Also looks good to me

jamesgillanders (approved):

PR #1297, opened 2020-09-10
Track Packet Last Interaction R

marxwillia:
Multiple properties of packet last interactions are tracked by TARDIS including energy, frequency, type of interaction, etc.  Here I add the ability to track the radius from the center of the last interaction.

## Description
Numerous small additions are made to the C montecarlo implementation to track the distance from the center of the last packet interaction.  These changes are modeled after how the frequency of the last interaction is stored.

## Motivation and Context
Line identification is complicated not only by line blending in a given shell by different elements, but also by lines from different elements at different velocities.  Different blueshifts can cause some high velocity features to overlap with low velocity features of other lines.  One way to address this issue is to track where the last interaction happens for the photon packets.

## How Has This Been Tested?
I ran a simulation and then plotted the distribution of where packet last interactions happened for S, Na, and He.  The results are reasonable and shown in the figure below.
image

## Types of changes

## Checklist:

- [ ] I have assigned/requested two reviewers for this pull request.

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -26,6 +26,7 @@ class MontecarloRunner(HDFWriterMixin):
     hdf_properties = ['output_nu', 'output_energy', 'nu_bar_estimator',
                       'j_estimator', 'montecarlo_virtual_luminosity',
                       'last_interaction_in_nu',
+                      'last_interaction_in_r',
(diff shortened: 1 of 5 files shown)


wkerzendorf:
This is useful - but should be implemented in the numba part.

PR #1300, opened 2020-09-14
Remove pip install option from TARDIS installation docs

### LINKED ISSUE #1211: Can't install TARDIS with `pip`
## Description

```
pip install git+https://github.com/tardis-sn/tardis
```
... does not work anymore!.

```
ModuleNotFoundError: No module named 'pyne'
```

Please, try to reproduce this bug in a fresh environment with latest TARDIS release and fix.

jaladh-singhal:
`pip install git+https://github.com/tardis-sn/tardis` seems not to be working as mentioned in #1211 and as Anaconda [suggests](https://www.anaconda.com/blog/using-pip-in-a-conda-environment) that pip should better be avoided with conda, so it needs to be removed from TARDIS installation docs to avoid any confusion for new users.

The only recommended way in our case is to use `python setup.py install` which is already mentioned in docs.

## Motivation and Context

Avoid pip in conda and `pip install` not working but is still mentioned in docs

## Types of changes

## Checklist:

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -68,16 +68,11 @@ TARDIS is used.
 
 For TARDIS development purposes please follow the steps :ref:`here <forking>`
 until the step to install TARDIS in the development mode
-``python setup.py develop``.
+``python setup.py develop``. Development guidelines for
+TARDIS can be found `here <https://tardis-sn.github.io/tardis/development/index.html>`_.
 
-You can also install TARDIS for the latest development version
-(but this is only recommended for pure users)::
-
-    pip install git+https://github.com/tardis-sn/tardis
-
-Development guidelines for TARDIS can be found `here <https://tardis-sn.github.io/tardis/development/index.html>`_.
-
-Alternatively, you can manually clone our repository and install TARDIS by::
+To install TARDIS, it is recommended to first clone our repository and
+then install TARDIS, as follows::
 
     git clone https://github.com/tardis-sn/tardis.git
     cd tardis


harpolea (approved):

epassaro:
I think the **Installation** guide now looks a bit redundant. If you are cloning the repo at some point then you don't need to get the environment file through `curl` or `wget`:

1. Clone the TARDIS repository
2. Install the dependencies with the YAML file in the repository
3. `python setup.py install`

Then, I don't think this closes #1211. Anaconda discourages the use of `pip` and that's fine. But it was possible to install TARDIS through `pip` and we break that at some point (and we don't know exactly why) so the bug still exists.

epassaro:
> I think the **Installation** guide now looks a bit redundant. If you are cloning the repo at some point then you don't need to get the environment file through `curl` or `wget`:
> 
> 1. Clone the TARDIS repository
> 2. Install the dependencies with the YAML file in the repository
> 3. `python setup.py install`
> 
> Then, I don't think this closes #1211. Anaconda discourages the use of `pip` and that's fine. But it was possible to install TARDIS through `pip` and we break that at some point (and we don't know exactly why) so the bug still exists.

what do you think about this @jaladh-singhal ?

wkerzendorf:
@jaladh-singhal the tests fail

jaladh-singhal:
@wkerzendorf It's only for Mac pipeline due to that unrelated test

jaladh-singhal:
> Then, I don't think this closes #1211. Anaconda discourages the use of pip and that's fine. But it was possible to install TARDIS through pip and we break that at some point (and we don't know exactly why) so the bug still exists.

@epassaro this makes sense, I agree it shouldn't closed #1211 - I've removed that from description

> I think the Installation guide now looks a bit redundant. If you are cloning the repo at some point then you don't need to get the environment file through curl or wget:

This is true @epassaro but we still need to make sure that `pip install` isn't recommended until we fix it. So either I can change the wording to "Clone TARDIS Repo -> Create env (without curl/wget ofc) -> `setup.py install`. Or the other way is we leave it as it is and mark it "not working currently" or something and put `setup.py install` section (without curl/wget) before it, marking as "recommended way as of now". Let me know which one do you prefer?

epassaro:
> > Then, I don't think this closes #1211. Anaconda discourages the use of pip and that's fine. But it was possible to install TARDIS through pip and we break that at some point (and we don't know exactly why) so the bug still exists.
> 
> @epassaro this makes sense, I agree it shouldn't closed #1211 - I've removed that from description
> 
> > I think the Installation guide now looks a bit redundant. If you are cloning the repo at some point then you don't need to get the environment file through curl or wget:
> 
> This is true @epassaro but we still need to make sure that `pip install` isn't recommended until we fix it. So either I can change the wording to "Clone TARDIS Repo -> Create env (without curl/wget ofc) -> `setup.py install`. Or the other way is we leave it as it is and mark it "not working currently" or something and put `setup.py install` section (without curl/wget) before it, marking as "recommended way as of now". Let me know which one do you prefer?

I prefer the first option.

PR #1366, opened 2020-12-04
Fix duplicate appending of scalars in HDF writer and add an overwrite option

jaladh-singhal:
When saving a simulation object in HDF using `to_hdf()` to same file path, the scalars are appended instead of being overwritten causing duplicate values in scalrs series of HDF. This is unlike what happens to other properties like numpy arrays, dataframes, etc. which are always overwritten. 

Besides making scalars overwrite (instead of append), this PR makes sure to explicitly prompt users when overwriting a pre-existing HDF file (which earlier used to happen implicitly). For this, a boolean `overwrite` option is added to `to_hdf()` which is by default `False`.

## Changes made
- ` tardis/io/util.py` - fixed scalars appending, added overwrite option & improved related docstrings
- **Several test files using `to_hdf()`** - added `overwrite=True` since parameterization and session scoped fixtures use same hdf file path and earlier implementation implicitly overwrote the written HDFs.

## How Has This Been Tested?

Local tests pass, hopefully Azure pipelines pass too!

### DIFF
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -199,9 +199,7 @@ class HDFWriterMixin(object):
     def to_hdf_util(
         path_or_buf, path, elements, overwrite, complevel=9, complib="blosc"
     ):
-        """
-        A function to uniformly store TARDIS data
-        to an HDF file.
+        """A function to uniformly store TARDIS data to an HDF file.
 
         Scalars will be stored in a Series under path/scalars
         1D arrays will be stored under path/property_name as distinct Series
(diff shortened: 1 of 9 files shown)


tardis/io/util.py
```diff
@@ -209,32 +209,44 @@ def to_hdf_util(path_or_buf, path, elements, complevel=9, complib="blosc"):
...
-            Path inside the HDF store to store the `elements`
+            Path inside the HDF file to store the `elements`
         elements : dict
             A dict of property names and their values to be
             stored.
-
-        Returns
-        -------
+        overwrite: bool
+            If the HDF file path already exists, whether overwrite it or not
```
andrewfullard, line 220: If the HDF file path already exists, whether to overwrite it or not

tardis/io/util.py
```diff
@@ -209,32 +209,44 @@ def to_hdf_util(path_or_buf, path, elements, complevel=9, complib="blosc"):
...
+
+        Notes
+        -----
+        `overwrite` option doesn't have any effect when `path_or_buf` is an
+        HDFStore because it depends on user that in which mode they've opened
+        it.
```
andrewfullard, line 226: `overwrite` option doesn't have an effect when `path_or_buf` is an HDFStore because the user decides on the mode in which they have opened the HDF (path or HDFStore).

andrewfullard (commented):
Just a couple of minor documentation adjustments. Otherwise looks good.

tardis/io/util.py
jaladh-singhal: Umm...actually I meant, file opening mode ('r', 'w', 'a') - when specifying HDFStore user has control over the mode in their hands, so `overwrite` option won't have any effect.

jaladh-singhal: Rephrased it - let me know if it still needs changes!

andrewfullard (approved):

PR #1442, opened 2021-01-29
Fixed distance parameter not producing accurate SDEC plots

### LINKED ISSUE #1440: Fix distance parameter's functionality in generate_plot of SDECPlotter
`plotter.generate_plot_mpl(distance=100 * u.Mpc)` results in this:
image

As per docstring, it expects cm as units but even with that the result looks almost same as above.

Besides, `distance = 0 * u.cm` shows a strange inverted plot. Should we throw an error instead, since that's not practically feasible I think?

jaladh-singhal:
Added lum_to_flux division factor in emission and absorption elements' histogram calculations. And modified distance docstring to use any length unit since calculations automatically convert it to CGS cm.

## Motivation and Context

Fixes #1440 

## How Has This Been Tested?
- [ ] Testing pipeline
- [x] Other (please describe)

Plot looks as expected:
image

## Types of changes

## Checklist:

### DIFF
--- a/tardis/widgets/sdec_plot.py
+++ b/tardis/widgets/sdec_plot.py
@@ -441,7 +441,7 @@ class SDECPlotter:
             [lower_lambda, upper_lambda] * u.AA
         distance : astropy.Quantity
             Distance used to calculate flux instead of luminosities in the plot.
-            Preferrably having units of cm.
+            It should have a length unit like m, Mpc, etc.
 
         Notes
         -----
@@ -660,6 +660,7 @@ class SDECPlotter:
                 group["nus"],
                 bins=self.plot_frequency_bins,
                 weights=group["energies"]
+                / self.lum_to_flux
                 / self.data[packets_mode].time_of_simulation,
             )
 
@@ -736,6 +737,7 @@ class SDECPlotter:
                 group["last_line_interaction_in_nu"],
                 bins=self.plot_frequency_bins,
                 weights=group["energies"]
+                / self.lum_to_flux
                 / self.data[packets_mode].time_of_simulation,
             )
 
@@ -805,7 +807,7 @@ class SDECPlotter:
             [lower_lambda, upper_lambda] * u.AA. Default value is None
         distance : astropy.Quantity or None, optional
             Distance used to calculate flux instead of luminosities in the plot.
-            Preferrably having units of cm. Default value is None
+            It should have a length unit like m, Mpc, etc. Default value is None
         show_modeled_spectrum : bool, optional
             Whether to show modeled spectrum in SDEC Plot. Default value is
             True
@@ -994,7 +996,7 @@ class SDECPlotter:
             [lower_lambda, upper_lambda] * u.AA. Default value is None
         distance : astropy.Quantity or None, optional
             Distance used to calculate flux instead of luminosities in the plot.
-            Preferrably having units of cm. Default value is None
+            It should have a length unit like m, Mpc, etc. Default value is None
         show_modeled_spectrum : bool, optional
             Whether to show modeled spectrum in SDEC Plot. Default value is
             True
--- a/tardis/widgets/sdec_plot.py
+++ b/tardis/widgets/sdec_plot.py
@@ -502,7 +502,10 @@ class SDECPlotter:
         if distance is None:
             self.lum_to_flux = 1  # so that this term will have no effect
         else:
-            self.lum_to_flux = 4.0 * np.pi * (distance.to("cm")) ** 2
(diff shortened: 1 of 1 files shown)


MarkMageeAstro (approved):

jamesgillanders (commented):
Overall, looks good. Do we need a clause at line 501 for distances of 0, or negative values? Also, can we change the unit order for the flux axis to be erg/s/cm2/AA (just to be in line with convention)? Finally, can we make the units inside math mode not italicised?

marxwillia:
This looks good to me.  @jaladh-singhal I believe the reason this happened in the first place is that in my version of the code, I adjusted the weights in the calculation stage, and my code had the plotting and calculation done in the same place.  When you separated the two I think the lum_to_flux got lost.

jaladh-singhal:
> Do we need a clause at line 501 for distances of 0, or negative values?
@jamesgillanders good point! Added one to raise ValueError

> Also, can we change the unit order for the flux axis to be erg/s/cm2/AA (just to be in line with convention)?
Done!

> Finally, can we make the units inside math mode not italicised?
In code, I checked that we're doing nothing to make units look italicized - I think this is how matplotlib renders latex in labels by default. Do you have any example in notice where it does not italicize, we can check its code & use that if you want?

tardis/widgets/sdec_plot.py
```diff
@@ -502,7 +502,10 @@ def _calculate_plotting_data(
         if distance is None:
             self.lum_to_flux = 1  # so that this term will have no effect
         else:
-            self.lum_to_flux = 4.0 * np.pi * (distance.to("cm")) ** 2
+            if distance <= 0:
+                raise ValueError("distance passed must be greater than 0")
+            else:
+                self.lum_to_flux = 4.0 * np.pi * (distance.to("cm")) ** 2
```
jamesgillanders, line 509: This is better, but is it possible to either have a message explaining that if D=0, this doesn't correspond to the luminosity case, and if the luminosity case is desired, then the user needs to remove the distance flag, or set to None. Alternatively, a simple workaround could be to allow D=0 to correspond to the Luminosity case, or would that confuse people? Maybe the first option is better...

tardis/widgets/sdec_plot.py
```diff
@@ -867,7 +872,7 @@ def generate_plot_mpl(
         self.ax.set_xlabel(r"Wavelength $[\AA]$", fontsize=12)
         if distance:  # Set y-axis label for flux
             self.ax.set_ylabel(
-                r"$F_{\lambda}$ [erg/s/$\AA/cm^{2}$]", fontsize=12
+                r"$F_{\lambda}$ [erg/s/cm^{2}/$\AA$]", fontsize=12
             )
```
jamesgillanders, line 876: good - to remove the italicised units, you could try something along the lines of:
plt.ylabel('Luminosity (erg $\mathrm{s^{-1}}$ $\mathrm{\AA^{-1}}$)') 
This is an example where the labels won't be italicised, since they are in \mathrm{}

jamesgillanders (commented):
see my comments. Otherwise, looks good

tardis/widgets/sdec_plot.py
jaladh-singhal: I agree, 1st option is better. Will mention that in error message.

tardis/widgets/sdec_plot.py
jaladh-singhal: Ah, sure I can try that for both luminosity and flux labels - but I think I should do this italics fix in a separate PR to adhere to "one PR, one idea" philosophy.

jamesgillanders: sounds good - can you open an issue for it then?

tardis/widgets/sdec_plot.py
jaladh-singhal: @jamesgillanders done! Let me know if it needs to be reworded for more clarity

tardis/widgets/sdec_plot.py
jaladh-singhal: it's in #1445

tardis/widgets/sdec_plot.py
jamesgillanders: this is perfect, thanks

tardis/widgets/sdec_plot.py
jamesgillanders: nice

jamesgillanders (approved):
All issues now resolved - happy to merge

PR #1460, opened 2021-02-16
Joss paper I

wkerzendorf:
First JOSS paper - not sure if that hits the right direction but this is a first draft.


jaladh-singhal:
@wkerzendorf do we really want to create this within repository? 

It won't be "used" by users so we shouldn't probably it here. Instead I'll suggest to add this in [tardis GitHub wiki](https://github.com/tardis-sn/tardis/wiki) which even makes collaboration easier!

jaladh-singhal (commented):
@wkerzendorf do we really want to create this within repository?

It won't be "used" by users so we shouldn't probably add it here. Instead I'll suggest to add this in tardis GitHub wiki which even makes collaboration easier!

epassaro (commented):
`paper/` folder or `papers/` ? I understand it's just 1 paper for JOSS, right?

Issue #1463, opened 2021-02-19
sim.runner.spectrum_integrated doesn't work when spectrum: integrated: interpolate_shells: > 0


ycamacho:
#### Problem description

When interpolate_shells: > 0 in input yml file sim.runner.spectrum_integrated results in following error

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-4d322c8a718a> in <module>()
----> 1 sim24.runner.spectrum_integrated

/Users/yssavo/Documents/GitHub/tardis/tardis/montecarlo/base.py in spectrum_integrated(self)
    233         if self._spectrum_integrated is None:
    234             self._spectrum_integrated = self.integrator.calculate_spectrum(
--> 235                 self.spectrum_frequency[:-1], **self.integrator_settings
    236             )
    237         return self._spectrum_integrated

/Users/yssavo/Documents/GitHub/tardis/tardis/montecarlo/formal_integral.py in calculate_spectrum(self, frequency, points, interpolate_shells, raises)
     94         self.interpolate_shells = interpolate_shells
     95         frequency = frequency.to("Hz", u.spectral())
---> 96         luminosity = u.Quantity(formal_integral(self, frequency, N), "erg") * (
     97             frequency[1] - frequency[0]
     98         )

/Users/yssavo/Documents/GitHub/tardis/tardis/montecarlo/montecarlo.pyx in tardis.montecarlo.montecarlo.formal_integral()

AttributeError: 'numpy.ndarray' object has no attribute 'values'

#### Expected Output

Formal integral should work when interpolate_shells: > 0

Issue is occurring because sim.runner.electron_densities_integ is a numpy array instead of a pandas series. Fix needs to occur in tardis/tardis/montecarlo/formal_integral.py line 253

ycamacho:
@chvogl This issue seems to have occurred during your PR 1358 https://github.com/tardis-sn/tardis/pull/1358

chvogl:
Yes, this was a relatively stupid error on my part. Thanks for fixing it @Rodot. The problem should be resolved through the merge of PR #1464.

chvogl:
I am closing this for now.

PR #1464, opened 2021-02-19
Fixed the array types when interpolate_shells > 0.  Now returns the p…

Rodot-:
When interpolate_shells > 0, the formal integral broke with an attribute error claiming that numpy arrays had no attribute "values"
Issue #1463 

I've corrected the return types of one of the functions so that they are no longer numpy arrays.

## Description

When interpolate_shells > 0, `interpolate_integrator_quantities` would change the type of electron_densities_integ, tau_sobolev_integ, and Jredlu from pandas dataframes or series to numpy arrays due to the fact that `scipy.interp1d` always returns a numpy array.  This was problematic because the formal_integral Cython wrapper access the `.values` attributes of these arrays assuming they were pandas objects.  I've corrected this so now these arrays are set to have the correct type.
## Motivation and Context
Now interpolate_shells can have values greater than 0 without failing.

## How Has This Been Tested?
- [x] Testing pipeline
- [ ] Other (please describe)

## Types of changes

## Checklist:

### DIFF
--- a/tardis/montecarlo/formal_integral.py
+++ b/tardis/montecarlo/formal_integral.py
@@ -250,26 +250,26 @@ class FormalIntegrator(object):
 
         r_middle_integ = (r_integ[:-1] + r_integ[1:]) / 2.0
 
-        runner.electron_densities_integ = interp1d(
+        runner.electron_densities_integ = pd.Series(interp1d(
             r_middle,
             plasma.electron_densities,
             fill_value="extrapolate",
             kind="nearest",
-        )(r_middle_integ)
+        )(r_middle_integ))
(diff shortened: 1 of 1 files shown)


Rodot-:
Should also be of note: This bug really demonstrates the need for formal integral tests

chvogl (approved):
Thanks for fixing my blunder ;)

chvogl:
I also agree that we should include a test with interpolate_shells > 0, which is the setting people should be using.

chvogl:
I took the liberty to merge this without a second review. The changes are minimal and fix an urgent problem.

PR #1478, opened 2021-03-14
add the symlink to visualization examples

yuyizheng1112:
## Description

**Changes made:**
Add the configuration file _tardis_example.yml_ to the directory: docs/using/visualization.
## Motivation and Context

In _generating_widgets.ipynb_ and _sdec_plot.ipynb_, the file path of function `run_tardis()` is _'tardis_example.yml'_, which directs to the file in the same directory as notebooks. Therefore, the configuration file should be added to the directory.

## How Has This Been Tested?
- [x] Testing pipeline
- [ ] Other (please describe)

截屏2021-03-14 上午10 46 51

## Types of changes

## Checklist:

- [ ] I have assigned and requested two reviewers for this pull request

### DIFF
--- a/docs/using/visualization/tardis_example.yml
+++ b/docs/using/visualization/tardis_example.yml
@@ -1,0 +1,1 @@
+../components/models/examples/tardis_example.yml


marxwillia:
This seems reasonable to me.  The one thing we might want to consider is that the tardis_example.yml file is going to be living in multiple places in the repository.  We should think about other ways to make our common example files easily accessible, while minimizing copies.  @jaladh-singhal any thoughts?

smithis7:
We use symbolic links for that file. The only real copy of `tardis_example.yml` is under `docs/using/components/models/examples/`-- this way, if we need to change it we only need to change one file.

@marxwillia is right though about accessibility. Perhaps we move the file somewhere easier to find, like just under `docs/`-- that would require changing all of the symlinks, so it would be a little bit of a task.

andrewfullard:
Converting to draft until comments are addressed

jaladh-singhal:
> We use symbolic links for that file. The only real copy of `tardis_example.yml` is under `docs/using/components/models/examples/`-- this way, if we need to change it we only need to change one file.

I second this. @yuyizheng1112 can you create a symlink to above path (as of now for this PR) instead of copying config file?

> @marxwillia is right though about accessibility. Perhaps we move the file somewhere easier to find, like just under `docs/`-- that would require changing all of the symlinks, so it would be a little bit of a task.

@marxwillia  @andrewfullard where do you think is the best place to keep `tardis_example.yml` within repo? @yuyizheng1112 once this is decided, you can dig all existing `tardis_example.yml` files & symlinks and change them as @smithis7 mentioned in another PR.

jaladh-singhal (changes requested):
@yuyizheng1112 I think you're close, can you complete this PR with requested change (in comment above)?

yuyizheng1112:
> @yuyizheng1112 I think you're close, can you complete this PR with requested change (in comment above)?

Sorry that I just saw your comment and thanks for your guidance!

yuyizheng1112:
> > We use symbolic links for that file. The only real copy of `tardis_example.yml` is under `docs/using/components/models/examples/`-- this way, if we need to change it we only need to change one file.
> 
> I second this. @yuyizheng1112 can you create a symlink to above path (as of now for this PR) instead of copying config file?
> 
> > @marxwillia is right though about accessibility. Perhaps we move the file somewhere easier to find, like just under `docs/`-- that would require changing all of the symlinks, so it would be a little bit of a task.
> 
> @marxwillia @andrewfullard where do you think is the best place to keep `tardis_example.yml` within repo? @yuyizheng1112 once this is decided, you can dig all existing `tardis_example.yml` files & symlinks and change them (as @smithis7 mentioned) in another PR.

I am willing to take this task if the place of `tardis_example` would be changed.

smithis7 (approved):

jaladh-singhal (approved):

PR #1514, opened 2021-03-25
FIRST OBJECTIVE - 'Notebook Logging Framework'

parikshit14:
As a part of the first objective, this PR is created to add the functionality to remove the unwanted warning that generally pops up during simulation which makes the output look complex and confusing to the end-user.

## Description
Added a `verbosity_suppression` option that suppresses warnings by default. It is added as a parameter in the `run_tardis` function. It is a boolean variable with default value **True** which by default hides the warnings, whereas when verbose is set to **False** , it will show the warnings though it won't be suppressing errors. All the changes that are made are in _base.py_ file.

## Motivation and Context
While execution of the code various warnings pop-up and gets displayed with the main output. So this is to make things easier for the end user. The user is provided with the choice to suppress warnings if they wish, but the errors will be displayed as it is.

## How Has This Been Tested?
- [ ] Testing pipeline
- [x] Other (please describe)
1. The above changes are tested on the _tardis_example.yml_ file and works perfectly fine. 
2. No extra library/extension is installed and works completely in the tardis virtual environment.
3. It removes all the warning (if present).

run_tardis-before-suppression
warnings-in-graph
run_tardis-after-suppression
graph-without-warning

## Types of changes

## Checklist:
@jamesgillanders  @marxwillia  kindly review this pull request


tardis/base.py
```diff
@@ -7,6 +7,7 @@ def run_tardis(
     packet_source=None,
     simulation_callbacks=[],
     virtual_packet_logging=False,
+    verbose_suppression=True,  # CHANGE1!! adding a paramenter to suppress the verbosity
```
jamesgillanders, line 10: comments in the code are fine, and encouraged in places where the code may be particularly dense or confusing. However, comments referring to changes in the code will quickly become obsolete after the code is updated. Also, we can easily see the proposed changes in your PRs using GitHub's diff feature, that will highlight the modifications made. So... I would say that a comment detailing what the flag does is okay, but not also a comment numbering the change

tardis/base.py
```diff
@@ -26,7 +27,9 @@ def run_tardis(
     virtual_packet_logging : bool
         option to enable virtual packet logging
         [default=False]
-
+    verbose_suppression : bool
+        option to suppress the warnings
+        [default=True]
     Returns
```
jamesgillanders, line 33: good succinct explanation of the new flag - after seeing this, perhaps the entire comment on line 10 could be removed

tardis/base.py
```diff
@@ -35,6 +38,13 @@ def run_tardis(
     from tardis.io.atom_data.base import AtomData
     from tardis.simulation import Simulation
 
+    # CHANGE2!! warning removal's code
+    import warnings
```
jamesgillanders, line 42: no need for comments on changes

tardis/base.py
```diff
@@ -35,6 +38,13 @@ def run_tardis(
...
+
+    if verbose_suppression == True:
+        warnings.filterwarnings("ignore")
+    #########
+
```
jamesgillanders, line 47: remove #####

jamesgillanders (commented):
neat solution to suppress the warnings from the code

tardis/base.py
parikshit14: yes, it actually makes sense what you are trying to say, will update accordingly. thanks for rectifying

tardis/base.py
parikshit14: Sure will do that

MarkMageeAstro:
Looks good to me, nice job. One comment is that I would avoid moving the example file and adding new notebooks to the tardis folder. It should go somewhere else, maybe documentation or quickstart?

parikshit14:
ok sure, I didn't know where to place it so I kept it in the main _tardis_ folder. 
No problem I will update it, thanks for the feedback.

PR #1517, opened 2021-03-26
Add catch for missing elements in SDEC plot

### LINKED ISSUE #1516: Potential issue with elements not in absorbed/emitted regions for SDEC plot
In the latest SDEC code, the elements to be coloured in the plot are determined based on the elements emitted or absorbed within a given wavelength range. It's possible that the elements emitted within a wavelength range are not the same as those absorbed. Therefore, when doing the plot there could be a situation where an error is thrown because a given element was not present in the emitted/absorbed dataframe.

MarkMageeAstro:
This fixes  #1516.

The elements coloured in the plot are determined via the absorption and emission. It's possible however, that these elements are not the same, e.g. if you specify a limited wavelength range. For example, you could have a small wavelength region where Fe is absorbed, but not emitted. This could give an error if you try to plot an element not in the emission dataframe, for example. This adds a try statement to catch this case. 

## How Has This Been Tested?
- [ ] Testing pipeline
- [X] Other (please describe)
I've tried artificially adding new elements to the elements list that are not included in the model. This produces the following:

Screenshot 2021-03-26 at 16 26 49

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] I have assigned and requested two reviewers for this pull request

### DIFF
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -1016,20 +1016,25 @@ class SDECPlotter:
         elements_z = self.elements
         # Contribution from each element
         for i, atomic_number in enumerate(elements_z):
-            lower_level = upper_level
-            upper_level = (
-                lower_level
-                + self.emission_luminosities_df[atomic_number].to_numpy()
-            )
+            # Add a try catch because elemets_z comes from the total contribution of absorption and emission.
+            # Therefore it's possible that something in elements_z is not in the emission df
+            try:
+                lower_level = upper_level
+                upper_level = (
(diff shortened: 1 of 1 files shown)


jamesgillanders (approved):
I ran some checks and you were right about the original code throwing an error in the narrow wavelength regime if there were different emission and absorption elements, so good spot! Ran some tests with the proposed fix, and it works perfectly. The output message on the exception is a particularly nice feature. Good job!

jaladh-singhal (approved):
Looks good :+1:

PR #1529, opened 2021-04-09
Mohit gsoc abundance Widget Challenges

mohith01:
## Description
Hello guys, i have first run the [quickstart](https://tardis-sn.github.io/tardis/quickstart/quickstart.html) to understand how tardis works in Example-1.ipynb.This needs tardis_example.yml to work.Then i did the challenge for Abundance Widget which i have titled "Challenge Problem.ipynb" . The Challenge Problem.ipynb needs testconfig.yml,testdensity.txt and abund.dat... There's a test folder which is just for reference and my understanding.

My idea for solving abundance widget involves using [gradio](https://www.gradio.app/). Basic idea is to integrate tardis and gradio to form a widget which users can run any model compositions and the plot will be shown

## Checklist:

- [ ] I have assigned and requested two reviewers for this pull request


luharvey:
Can you produce the plot in plotly too? Perhaps also think about using different linestyles and/or markers for the different elements to make it clearer.

mohith01:
> Can you produce the plot in plotly too? Perhaps also think about using different linestyles and/or markers for the different elements to make it clearer.

Yeah sure i will try and push it up

mohith01:
> Can you produce the plot in plotly too? Perhaps also think about using different linestyles and/or markers for the different elements to make it clearer.

I have replaced matplotlib with plotly in "Challenge Problem.ipynb".Gotta say plotly is easy and aesthetically pleasing.
image

luharvey:
Having the x axis be in km/s would be better I reckon because they tend to be the units we talk about with supernovae. It'll also fix the labels on the x axis, I'm assuming B = 'Billion' because 10^9 is about right. One more thing might be to set the y scale to go between 0.0 and 1.0 as it's mass fraction we're looking at.

mohith01:
> Having the x axis be in km/s would be better I reckon because they tend to be the units we talk about with supernovae. It'll also fix the labels on the x axis, I'm assuming B = 'Billion' because 10^9 is about right. One more thing might be to set the y scale to go between 0.0 and 1.0 as it's mass fraction we're looking at.

km/s is fine but if i make y axis 0 to 1 i think the graph will become too small in y direction

luharvey:
It being small in the y direction isn't really a concern for now because when introducing interactivity to the plot we'll want the y range to be fixed and for that it'll need to show all possible mass fractions so 0.0 to 1.0.

mohith01:
> It being small in the y direction isn't really a concern for now because when introducing interactivity to the plot we'll want the y range to be fixed and for that it'll need to show all possible mass fractions so 0.0 to 1.0.

yup i have pushed this...this is the screenshot now
image

mohith01:
@jaladh-singhal  pls check the [document](https://docs.google.com/document/d/1VQZaCzDtnXPh2SYBF1eqTu6H9bLgH641EKZwbNbs6Jo/edit?usp=sharing ) once.I have made the changes u told me to do.

PR #1550, opened 2021-04-25
Updating Enum class for last_interaction_type 

### LINKED ISSUE #1396: Give name to last_interaction_type values
Currently, kromer plot and other viz tools uses `sim.runnner.last_interaction_type` extensively and it's values are hard-coded (i.e. -1, 1, 2) throughout the comparisons & calculations. Instead they should be named to what they actually represents (i.e. No Interaction, E-scattering, Line Interaction respectively).

Probably a suitable way to achieve this is to create an Enum class in `montecarlo_numba/base.py` to store this mapping of values to names which can then be used by Kromer Plot or other modules whenever required.

DhruvSondhi:
This PR aims to implement & fix #1396 
Implements & Updates `Enum` class for `last_interaction_type` :)

**Description**
Updated the `InteractionType` enum class for `last_interaction_type` as stated through the issue.
Restructured this class to the `base.py` from the `r_packet.py` module. Updated required references :)

**Motivation and context**
Updating the enumeration `InteractionType` class will allow for easier usage of the parameters for calculation as well as comparison. 
Values have been given the correct parameter name :)

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Type of change**

- [x] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/r_packet.py
+++ b/tardis/montecarlo/montecarlo_numba/r_packet.py
@@ -30,12 +30,11 @@ from tardis.montecarlo.montecarlo_numba.numba_config import (
     ENABLE_FULL_RELATIVITY,
 )
 
-
 class InteractionType(IntEnum):
-    BOUNDARY = 1
+    NO_INTERACTION = -1
+    BOUNDARY = 0
+    ESCATTERING = 1
     LINE = 2
-    ESCATTERING = 3
-
 
 class PacketStatus(IntEnum):
     IN_PROCESS = 0
@@ -189,7 +188,7 @@ def trace_packet(r_packet, numba_model, numba_plasma, estimators):
             (distance_electron < distance_trace)
             and (distance_electron < distance_boundary)
         ) and distance_trace != 0.0:
-            interaction_type = InteractionType.ESCATTERING
+            interaction_type = InteractionType.ESCATTERING  # E-SCATTERING
             # print('scattering')
             distance = distance_electron
             r_packet.next_line_id = cur_line_id
@@ -211,7 +210,7 @@ def trace_packet(r_packet, numba_model, numba_plasma, estimators):
             tau_trace_combined > tau_event
             and not montecarlo_configuration.disable_line_scattering
         ):
-            interaction_type = InteractionType.LINE  # Line
+            interaction_type = InteractionType.LINE  # LINE
             r_packet.last_interaction_in_nu = r_packet.nu
             r_packet.last_line_interaction_in_id = cur_line_id
             r_packet.next_line_id = cur_line_id
--- a/tardis/montecarlo/montecarlo_numba/r_packet.py
+++ b/tardis/montecarlo/montecarlo_numba/r_packet.py
@@ -32,9 +32,10 @@ from tardis.montecarlo.montecarlo_numba.numba_config import (
 
 class InteractionType(IntEnum):
     NO_INTERACTION = -1
-    BOUNDARY = 0
-    ESCATTERING = 1
+    BOUNDARY = 1
     LINE = 2
+    ESCATTERING = 3
+
 
 class PacketStatus(IntEnum):
     IN_PROCESS = 0
@@ -179,7 +180,7 @@ def trace_packet(r_packet, numba_model, numba_plasma, estimators):
             (distance_boundary <= distance_trace)
             and (distance_boundary <= distance_electron)
         ) and distance_trace != 0.0:
-            interaction_type = InteractionType.BOUNDARY  # BOUNDARY
+            interaction_type = InteractionType.BOUNDARY  
             r_packet.next_line_id = cur_line_id
             distance = distance_boundary
             break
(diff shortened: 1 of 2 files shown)


MarkMageeAstro:
Does it matter that the interaction type numbers have changed around a bit? i.e. boundary was 1, is now 0. e-scattering was 3, is now 1. I guess not but it looks like parts of the code already expect e-scattering is 1. I think it would be good to change the parts of the code too. 

e.g.
        mask_escatter = (
            self.data[packets_mode].packets_df["last_interaction_type"][
                self.packet_nu_range_mask
            ]
            == 1
        ) & (
            self.data[packets_mode].packets_df["last_line_interaction_in_id"][
                self.packet_nu_range_mask
            ]
            == -1
        )

becomes

        mask_escatter = (
            self.data[packets_mode].packets_df["last_interaction_type"][
                self.packet_nu_range_mask
            ]
            == InteractionType.ESCATTERING
        ) & (
            self.data[packets_mode].packets_df["last_line_interaction_in_id"][
                self.packet_nu_range_mask
            ]
            == -1
        )


This is already done for some type checks in the code, but not all

MarkMageeAstro (changes requested):
I should have added my previous comment as part of the review instead of a separate comment...

DhruvSondhi:
@MarkMageeAstro I don't think changing the `Enum` values may have a bigger impact on the codebase ... the main change that would be introduced would be that now condition may be checked in a different way ... else everything remains the same ie the executions procedure &/or checking for the condition ( == InteractionType.Type_of_Interaction)

tardis/visualization/tools/sdec_plot.py
```diff
@@ -1250,7 +1254,7 @@ def _plot_emission_mpl(self):
             except:
                 # Add notifications that this species was not in the emission df
                 if self._species_list is None:
-                    longer.info(
+                    logger.info(
```
jaladh-singhal, line 1257: Good catch, don't know how this got committed - can you please do these changes (longer -> logger) in a separate PR instead? 

~ One PR should work on one idea so that code history remains relevant :)

tardis/montecarlo/montecarlo_numba/r_packet.py
```diff
@@ -30,12 +30,11 @@
     ENABLE_FULL_RELATIVITY,
 )
 
-
 class InteractionType(IntEnum):
-    BOUNDARY = 1
+    NO_INTERACTION = -1
+    BOUNDARY = 0
+    ESCATTERING = 1
```
jaladh-singhal, line 36: I'm a bit skeptical about this if changing these values leads to unintended side-effects (though tests pass) - can you double-check @andrewfullard (since you've recently worked with the numba montecarlo)?

jaladh-singhal (changes requested):
I'll advise being extra careful about this change since we currently have no tests written for the SDEC plot. So can you re-run the SDEC plot notebook on this branch, to see if anything has not broken?

DhruvSondhi:
@jaladh-singhal I can confirm that the code in the notebook runs without any change ... if needed I can commit the notebook with the output for the same :)

tardis/visualization/tools/sdec_plot.py
DhruvSondhi: Done, I have dropped the commit for the same ... I will wait for this PR to get merged ... then I will send in another one fixing this ^^

tardis/montecarlo/montecarlo_numba/r_packet.py
andrewfullard: Yeah I personally don't see a need for this change to the enum values

marxwillia: I agree with @andrewfullard.  Let's not change enum values if it's not necessary, especially since we don't have great tests for the sdec plot.

marxwillia:
@DhruvSondhi Can you please upload a picture showing a side by side comparison of the SDEC plot generated before and after your changes?  Thanks.

DhruvSondhi:
Hello @marxwillia, These are some of the plots that are generated by my PR :
**Virtual Packet Mode**
Before Changes -
image
After Changes -
image

**Real Packet Mode**
Before Changes -
image
After Changes -
image

**Virtual Plotly Graph**
Before Changes -
image
After Changes -
image

tardis/montecarlo/montecarlo_numba/r_packet.py
DhruvSondhi: It has been reverted ... a small note, like suggested above by @MarkMageeAstro, the following condition :
```
 mask_escatter = (
        self.data[packets_mode].packets_df["last_interaction_type"][
            self.packet_nu_range_mask
        ] == 1
```
becomes 
```
 mask_escatter = (
        self.data[packets_mode].packets_df["last_interaction_type"][
            self.packet_nu_range_mask
        ]
        == InteractionType.BOUNDARY
```
These changes are not that important as nothing is broken and associated with the enum values (operations or functionality) but it was a small point to make 🤔

jaladh-singhal: @marxwillia @andrewfullard The change "ESCATTERING = 1" **needs to be done** to give semantic name to `1` that is used throughout SDEC plot for comparison. I was just concerned if this change (which also requires changing BOUNDARY to some other value since it is currently 1) will cause any side-effects for rest of the codebase?

DhruvSondhi: Yes @jaladh-singhal, I have reverted it to that particular condition ... These values are very difficult to map as it has different values in different contexts.

andrewfullard (approved):

andrewfullard:
I'm going to close this for now and we can approach it a more comprehensive way later.

Issue #1343, opened 2020-11-13
Config parser should verify that parameters are physically realistic


Rodot-:
#### Problem description

tardis will currently run when given negative abundances without complaining.  Other physically unrealistic parameters like negative densities are also run without complaint resulting in odd behavior of TARDIS (freezing up for example).  We should implement some sort of checks in the config parsers (or when we run tardis) to make sure that the config represents a physically realistic system

aribalam:
@Rodot- Would like to take this up. Could you point me in the right direction?

DhruvSondhi:
@andrewfullard this issue can be closed !!

PR #1554, opened 2021-04-26
Adding Config Value Validation for Different Parameters via Config File

DhruvSondhi:
This PR aims to implement **config validation** for _values_ that are passed through the `.yml` file.
This Implements #1343 

**Description**
This PR aims to implement validation for values that can be passed through the tardis configuration file. 
These are aimed at Density & Abundances in the current iteration.

**Motivation and context**
The tardis configuration file allows for non-realistic values to be passed without raising any errors. This PR aims to fix this & implement validation for the values so as to allow only real physics values to be added

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Examples**
None at current status :)

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

    - [x] I will update the documentation accordingly.

### DIFF
--- a/tardis/io/tests/test_config_reader.py
+++ b/tardis/io/tests/test_config_reader.py
@@ -1,5 +1,6 @@
 # tests for the config reader module
 import os
+from attr import validate
 import pytest
 import pandas as pd
 from numpy.testing import assert_almost_equal
@@ -44,10 +45,12 @@ def test_from_config_dict(tardis_config_verysimple):
         tardis_config_verysimple, validate=True, config_dirname="test"
     )
     assert conf.config_dirname == "test"
+
     assert_almost_equal(
         conf.spectrum.start.value,
         tardis_config_verysimple["spectrum"]["start"].value,
     )
+
     assert_almost_equal(
         conf.spectrum.stop.value,
         tardis_config_verysimple["spectrum"]["stop"].value,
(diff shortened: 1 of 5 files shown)


DhruvSondhi:
I will be adding the tests for the Density as well ... But I have a query @Rodot-, Do we need to validate other parameters or other ways of parsing data such as the `csvy` file, abundance files?

andrewfullard:
These tests should be part of the JSON Schema, not if statements in code. See e.g. https://json-schema.org/understanding-json-schema/reference/numeric.html

The schema are located in /tardis/io/schemas

andrewfullard (commented):
Looks good so far. I think @Rodot- can provide more parameters that should have additional validation.

Rodot-:
A couple I can think of: Any physical quantity should be positive (temperatures, velocities, times, densities, etc). time_explosion can't be zero.  For various density profiles, make sure that the start velocity is lower than the stop velocity.

DhruvSondhi:
Hello, @andrewfullard @Rodot- ... While going through the code for validating the different parameters in the `YAML` file, I think I will need to make changes to the `config_reader.py` & `config_validator.py` (I am not absolutely sure about the second one but could be dependent) for checking the valid conditions ... This is due to the fact the there is an `astropy quantity` based object being used to take in the input. This is not present in the normal JSON Schema & hence, we need to make changes in alternative ways. What do you suggest about the same?  🤔

DhruvSondhi:
Another thing that is troubling me, `allowed_value` & `allowed_type` are not being validated in the schema... I have tried many combinations for the same but the schema validation just ignores this property... if these properties work as expected then altogether the above approach can be removed 🤔

Rodot-:
Could you add some tests where you make sure that it fails on invalid config files in all of the places you added checks for?

DhruvSondhi:
Hello @Rodot-, I have added the tests for all the required sections in the `tardis_config` **YAML** file. Please do let me know if further changes or additions are required :)

tardis/io/tests/test_config_reader.py
```diff
@@ -68,3 +71,78 @@ def test_config_hdf(hdf_file_path, tardis_config_verysimple):
     actual = pd.read_hdf(hdf_file_path, key="/simulation/config")
     expected = expected.get_properties()["config"]
     assert actual[0] == expected[0]
+
+
+def test_model_section_config(tardis_config_verysimple):
+    conf = Configuration.from_config_dict(
```
Rodot-, line 77: Please add a docstring describing what this test does

tardis/io/tests/test_config_reader.py
```diff
@@ -68,3 +71,78 @@ def test_config_hdf(hdf_file_path, tardis_config_verysimple):
...
+            < conf.supernova.luminosity_wavelength_end.value
+        ):
+            raise ValueError(
+                "End Limit must be greater than Start Limit for Luminosity"
+            )
+    assert ve.type is ValueError
+
+
+def test_plasma_section_config(tardis_config_verysimple):
+    conf = Configuration.from_config_dict(
```
Rodot-, line 126: Please add a docstring describing this test

tardis/io/tests/test_config_reader.py
```diff
@@ -68,3 +71,78 @@ def test_config_hdf(hdf_file_path, tardis_config_verysimple):
...
+    with pytest.raises(ValueError) as ve:
+        if (conf.plasma.initial_t_inner.value >= -1) and (
+            conf.plasma.initial_t_rad.value >= -1
+        ):
+            raise ValueError("Initial Temperatures are Invalid")
+    assert ve.type is ValueError
+
+
+def test_spectrum_section_config(tardis_config_verysimple):
+    conf = Configuration.from_config_dict(
```
Rodot-, line 140: Please add a docstring describing this test

Rodot-:
@DhruvSondhi This is looking good, we just need some docstrings added to the tests then we should be good to merge

tardis/io/tests/test_config_reader.py
DhruvSondhi: Sure, I didn't know that we had to add docstrings for Tests too ... though I saw the `docstring_coverage` in the red :)

tardis/io/tests/test_config_reader.py
DhruvSondhi: Sure :)

tardis/io/tests/test_config_reader.py
DhruvSondhi: Sure, changes are on its way :)

DhruvSondhi:
@Rodot- @andrewfullard Should I squash all the commits?

andrewfullard (approved):

Rodot- (approved):

Issue #1561, opened 2021-05-10
Moving Config File Under `docs/`


smithis7:
**Is your feature request related to a problem? Please describe.**
As mentioned in #1478 the example configuration file is fairly difficult to find. It is currently in `docs/using/components/models/examples/tardis_example.yml`. We only keep one real copy of the file, and the rest of the time we use symlinks to reference the file.

**Describe the solution you'd like**
We would like to have to have the real copy in an easier location to access, perhaps just under `docs/`. This involves:
- Moving the real copy of the file.
- Adding a symlink to `docs/using/components/models/examples/`.
- Updating all of the other symlinks in the documentation to reflect the new location of the real copy.

parikshit14:
Can I work on this as well?

PR #1562, opened 2021-05-11
added symlinks for tardis_example.yml

parikshit14:
Added Symlinks for tardis_example wherever they were needed

**Description**
Replaced the original file with symlinks:
1. _tardis/docs/using/components/models/examples/tardis_example.yml_

placed the original **tardis_example.yml** in _tardis/docs_

also added a symlink in _tardis/docs/quickstart/_ as it is generally used in quickstart.ipynb

**Motivation and context**

solves issue #1561 

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

@smithis7 @wkerzendorf kindly verify the changes


docs/quickstart/tardis_example.yml
```diff
@@ -0,0 +1 @@
+/home/parikshit/Desktop/tardis/docs/tardis_example.yml
```
atharva-2001, line 1: I guess you should use relative imports here. In **this** case, because this is a symlink for `docs/tardis_example.yml`, just `../tardis_example.yml` should work.

parikshit14: thanks for rectifying the mistakes I really don't know how I missed them. Thanks again

smithis7:
Good start! A few things:
- `docs/research/code_comparison/plasma_compare/tardis_example.yml` is actually a different YAML file from the one elsewhere in the documentation, so we want to leave it how it was. I can see where having it named `tardis_example.yml` could cause confusion, and that may be something to rename in the future.
- There are several other occurrences of the YAML file throughout the documentation. They are all currently symlinks that direct to `docs/using/components/models/examples/tardis_example.yml` (so in this PR, you should only need to replace the actual file with a symlink once, and everything else will just be changing symlinks like you did with quickstart). To go about finding them all, you will want to build the documentation locally, and see where the notebooks fail to run (which will be the result of a broken symlink), and fix the symlink. You'll want to repeat this until the docs build properly. @epassaro once showed me a more efficient way to locate all broken symlinks, but I don't recall what it was (maybe he can comment how to do it).

parikshit14:
> Good start! A few things:
> 
> * `docs/research/code_comparison/plasma_compare/tardis_example.yml` is actually a different YAML file from the one elsewhere in the documentation, so we want to leave it how it was. I can see where having it named `tardis_example.yml` could cause confusion, and that may be something to rename in the future.

ok, I thought they were the same. Just checked and found the difference, no problem I will revert this back to normal.

> * There are several other occurrences of the YAML file throughout the documentation. They are all currently symlinks that direct to `docs/using/components/models/examples/tardis_example.yml` (so in this PR, you should only need to replace the actual file with a symlink once, and everything else will just be changing symlinks like you did with quickstart). To go about finding them all, you will want to build the documentation locally, and see where the notebooks fail to run (which will be the result of a broken symlink), and fix the symlink. You'll want to repeat this until the docs build properly. @epassaro once showed me a more efficient way to locate all broken symlinks, but I don't recall what it was (maybe he can comment how to do it).

ok, so all the existing symlinks(present originally in this repository) are reference from `docs/using/components/models/examples/tardis_example.yml` and not from `docs/research/code_comparison/plasma_compare/tardis_example.yml`. correct me if I am wrong! because this can again lead to a lot of confusion.

smithis7:
> ok, so all the existing symlinks(present originally in this repository) are reference from `docs/using/components/models/examples/tardis_example.yml` and not from `docs/research/code_comparison/plasma_compare/tardis_example.yml`. correct me if I am wrong! because this can again lead to a lot of confusion.

Right. The issue I posted is only concerned with the file that was under `docs/using/components/models/examples/tardis_example.yml`, so you just need to change symlinks that point to `docs/using/components/models/examples/tardis_example.yml` so that they point to `docs/tardis_example.yml`.

smithis7:
> I tried to test the changes you recommended in the second part but apparently the symlinks are not broken i.e. the symlinks which are now pointing to the symlink of `docs/tardis_example.yml` (symlink -> symlink -> docs/tardis_example.yml) are working completely fine. I checked them using the notebooks given with them respectively and they work the same as they were working before. 
> 
> But still, if you require I can update them to the current situation

I see why that is. Technically, it would be fine, as the symlinks point to another symlink that points to the proper document, but I think it would be better to change them so they point directly to the proper place. I think you can find all of them by searching for tardis_example.yml in the TARDIS code.

parikshit14:
Updated as suggested!

smithis7:
Looks really good! This is my fault as I did not put this in the original issue, but as I was reviewing this I noticed that there is a link in `docs/quickstart/quickstart.ipynb` (3rd code cell) that points to the old location. That's the last link that points to the old location that I found, so if you can change that one I will approve the PR. Sorry again for missing that link in the original issue.

parikshit14:
> Looks really good! This is my fault as I did not put this in the original issue, but as I was reviewing this I noticed that there is a link in `docs/quickstart/quickstart.ipynb` (3rd code cell) that points to the old location. That's the last link that points to the old location that I found, so if you can change that one I will approve the PR. Sorry again for missing that link in the original issue.

no problem, I also missed it while I was testing quickstart.ipynb !!

smithis7 (approved):
Perfect!

andrewfullard (approved):

PR #1565, opened 2021-05-12
Fix for close line test.  

Rodot-:
Appears that we don't actually want to look for close lines as much as check for numerical errors when computing the comoving frequency.  The test now occurs in calculate_distance_line.  The packet specs have been updated to remove the close line flags and the tests for the close lines have been removed

Removed all close line tests.  calculate_distance_line now checks if the nu_diff is within CLOSE_LINE_THRESHOLD (Which is a configurable parameter, a better value should be chosen, anyone have any ideas?)

**Motivation and context**
The close line checks were producing incorrect results before.  

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 
Since these changes change the way the spectrum is computed, the ref-data needs to be regenerated after this is merged.  For the time being, testing has been done by tracking packets manually and running the quickstart notebook.

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/base.py
+++ b/tardis/montecarlo/montecarlo_numba/base.py
@@ -151,7 +151,6 @@ def montecarlo_main_loop(
     virt_packet_last_line_interaction_in_id = []
     virt_packet_last_line_interaction_out_id = []
 
-    print("Running post-merge numba montecarlo (with C close lines)!")
     for i in prange(len(output_nus)):
         if montecarlo_configuration.single_packet_seed != -1:
             seed = packet_seeds[montecarlo_configuration.single_packet_seed]
--- a/tardis/montecarlo/montecarlo_numba/base.py
+++ b/tardis/montecarlo/montecarlo_numba/base.py
@@ -200,7 +200,7 @@ def montecarlo_main_loop(
         ).astype(np.int64)
         # if we're only in a single-packet mode
         # if montecarlo_configuration.single_packet_seed == -1:
-        #     break
+        #    break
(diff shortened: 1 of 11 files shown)


Rodot-:
`test_montecarlo_main_loop` in `
tardis/montecarlo/montecarlo_numba/tests/test_base.py` must now be rewritten.  The current version compares the run to the results that would have been produced by the old C version of tardis to make sure functionality was not broken when we switched over to numba.  This PR break that since the changes to the close line handling now cause tardis to diverge from the older C version.  This test must now be rewritten to update reference data appropriately and compare only to the current version of tardis.

Rodot-:
@andrewfullard @epassaro it seems that running the compare-refdata after updating the refdata works, but the tests appear to be still using the old ref-data.  Am I correct on this or am I missing something?

epassaro:
1. The reference data has been updated correctly according to the latest comparison (the refdata from the repo is the same as the one generated by the tip of this PR).
2. Tests are using the updated reference data because started failing in the master branch (as other PRs as well).
3. The latest fix for the TARDIS env passed on Sunday https://dev.azure.com/tardis-sn/TARDIS/_build/results?buildId=4220&view=results
4. There's a slight chance that some package updated recently and broke the tests and it's not your fault. Save your env by doing `conda env export > env_jack.yml` and post it on Slack.

andrewfullard (approved):

wkerzendorf (approved):
looks good

Issue #1572, opened 2021-05-15
Add pre-commit hook


Molkree:
**Is your feature request related to a problem? Please describe.**
As part of #1170, creating separate issue for discussion specific to the pre-commit.

**Describe the solution you'd like**
Add pre-commit hook that will run Black + other tools of choice. Black definitely will be there as it's already in the GH check and recommended in tardis docs, what else should be used?

isort seems like a good [idea](https://github.com/tardis-sn/tardis/issues/1170#issuecomment-635947440).

flake8 will require more configuration but also can be added.

**Describe alternatives you've considered**
Everyone writes however they want and PRs get messier from "blackify" commits more often. Anarchy.

**Additional context**
I think this should be a staggered approach:
- first add pre-commit with only Black, this should be easy.
- then add `isort`, will need to add config as well.
- flake8 (or pylint? or something else?) will require significantly more effort and should be done later.

---

`pre-commit` requires to pin to rev, like this:
```yml
-   repo: https://github.com/psf/black
    rev: 21.5b1
```
which means that config should be updated from time to time.

How to do that?
"Dumb" solution: add a workflow that runs `pre-commit autoupdate` weekly and commits changes.
Maybe an "easier" solution: use [pre-commit.ci](https://pre-commit.ci/). It will update pre-commit config every week AND it will check all PRs and fix them. This will even allow to remove the existing Black check from GitHub Actions (at least for PRs). They say it is free for open source repositories but paid for "organization" repositories. I suppose tardis-sn is the former?
If we don't use pre-commit.ci, we'll also need to add isort+linter checks to GitHub Actions manually.

Issue #1563, opened 2021-05-11
Procedure for Addition of New Dependecies to TARDIS


Rodot-:
**Is your feature request related to a problem? Please describe.**

Addition of new packages to TARDIS comes with the risk of introducing potential long-term instability of our package.  To mitigate this, we present a procedure for the review of pull requests which introduce new packages to TARDIS.

**Describe the solution you'd like**

* Addition of new packages to TARDIS should be done through a pull request whose only change is the addition of the new package in the TARDIS environment file.
* 4 reviewers must approve of said pull request
    * At least 3 reviewers must be members of the [TARDIS Core Team](https://tardis-sn.github.io/tardis/team_and_governance/team.html)
    * At least 1 reviewer must be a TARDIS contributor who has made recent contributions to components of TARDIS that this package addition would affect.

PR #1405, opened 2020-12-30
Restructured development workflow and added code quality & documentation guidelines in it

KevinCawley:
This PR creates the Coding and Docstring Guidelines for TARDIS, as well as how to create docs locally.  It also addressed some of the syntax errors that occurred when docs were compiled locally.

## Description
The first file that is added is Code_Quality_Guidelines.rst. This file addresses the code standard of TARDIS. The main section is the use of Black to format code files to create consistency in the code that is submitted. 
The next section of this file addresses the handling of exception and raising errors. This is made so that invalid input can be dealt with and the user can be alerted to their error, and the code should be written so that edge cases are caught so unknown and strange errors are avoided. 
The third section deals with Docstrings. This section creates the guideline for how docstrings should be formatted, and gives clear directions as to the required syntax for docstrings, with a lot of influence from numpydocs. 
The final section is a brief section about the naming conventions used in TARDIS, and heavily draws from the PEP8 naming convention. 

The next change is in git_workflow.rst. This section gives directions for how to build and compile docs locally, and the importance and benefits of compiling docs locally before being submitted in a commit or merge. This is an alternative way to look at the TARDIS docs, and a faster method than the this one ( https://tardis-sn.github.io/tardis/development/documentation_preview.html). Both methods accomplish the same task, there is no difference in the docs created. 

The remaining changes are to the index.rst file to include my Code_Quality_Guidelines.rst file. Another changes is the removal of a branch of the index.rst in tardis/docs that is outdated (and does not compile as the files no longer exist). All of the other changes are changes to resolve syntax errors that occur when docs are compiled locally, most of them involving the lack of a newline when the documentation syntax expects there to be one. 

## Motivation and Context
This was done to address three Issues (1170, 1252, and 1253). This will create a consistent method and guideline for how to code in TARDIS and how to write docstrings. This will reduce time made changing commits in the future to adhere to the guideline, as it will be written out and a lot easier to follow.

Partially covers: https://github.com/tardis-sn/tardis/issues/1170 (enforcing by pre-commit hook will be in another PR)
Fixes https://github.com/tardis-sn/tardis/issues/1252
Fixes https://github.com/tardis-sn/tardis/issues/1253

## How Has This Been Tested?
- [ ] Testing pipeline
- [x] Other (please describe)

The changes were tested by compiling docs locally, and also online when my version of TARDIS was broken due to untracked files. This also allowed both methods (online and local) to be tested. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
[...]

PR #1398, opened 2020-12-21
add black check workflow

epassaro:
## Description

A simple action to run `black --check tardis` on every push to a PR.

## Motivation and Context

Enforce PEP8.

## How Has This Been Tested?
- [x] Testing pipeline
- [ ] Other (please describe)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] None of the above (please describe)

## Checklist:

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have built the documentation on my fork following [these instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html)
- [x] I have assigned and requested two reviewers for this pull request

PR #1573, opened 2021-05-15
Add pre-commit hook with Black

Molkree:
**Description**
- Added `pre-commit` to the conda environment file.
- Added `pre-commit` config with Black to the root. Install with `pre-commit install`.
- Added `pre-commit` badge to Readme.

Opening this early but I have encountered one issue. When I commit something that shouldn't pass Black it correctly gets changed. I have tested line longer than 80 chars but shorter than default 88 chars to check that `pre-commit` Black sees Black config. But at the same time,  it still tries to modify files that it shouldn't. I have changed something inside `tardis/montecarlo` which should be excluded from the check but when I try to commit it formats there as well 🤔 Not sure if it's my mistake or Black/pre-commit limitation, help is appreciated here.

**Motivation and context**
This part of #1572.

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. Tested it working on my local machine.

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

    Not yet, will add it next. I guess only need to mention it in [Git workflow](https://tardis-sn.github.io/tardis/development/git_workflow.html) and Code quality Black [section](https://tardis-sn.github.io/tardis/development/code_quality.html#black).


epassaro:
Is this related to your issue? https://github.com/psf/black/issues/1667

Molkree:
@epassaro, no, I don't think it's related. In that issue people got confused and thought that Black config doesn't get passed when used by pre-commit. However in our case it definitely gets passed.
Simple test: in tardis you set `line-length = 80` when default is 88. I just created `t.py` in the root with this content:
```python
print("long", "long", "long", "long", "long", "long", "long", "long", "long", "long")
```
This is 85 chars long, so should be formatted according to our config. And it does get formatted when I try to commit.
However when I move this file to `tardis/montecarlo` which should be excluded from Black formatting, it still gets formatted when I try to commit.

So it somehow picks up line length config from .toml but ignores excluded files config 🤔

I'm just going to ask in pre-commit/Black community now.

Molkree:
Alright, I think it is an unavoidable limitation because pre-commit basically runs Black against individual files (if you do the same with Black directly it will also format). So exclude files/dirs should be duplicated in pre-commit config for Black :/

See this [comment](https://github.com/pre-commit/pre-commit/issues/1723#issuecomment-738776280) for a similar question about `flake8`.

Molkree:
By default Black excludes files in .gitignore and a [bunch of others](https://github.com/psf/black/blob/main/src/black/const.py#L2). If you use `exclude` it just overwrites that list with what you pass, but there's `extend-exclude` option that adds to that list. I've changed to this option and removed dirs that are ignored by default.

Then I just copied that explicit exclude list to `pre-commit` config. And added a bunch of dirs to the .gitignore so they don't get checked out.

Molkree:
Should I split it into two PRs according to #1563?

epassaro:
@jaladh-singhal implemented `black`, I think he's more qualified to review this.

.gitignore
```diff
@@ -75,3 +73,27 @@ distribute-*.tar.gz
...
+# Unit test / coverage reports
+htmlcov
+.coverage
+.pytest_cache/
+.tox/
+.nox/
+
+
+# Environments
+.venv
```
andrewfullard, line 99: Why are there additional entries added to .gitignore?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    -   id: black
+        exclude: |
+            (?x)^(
+                (tardis/model
+                | tardis/montecarlo).*
+                | setup.py
+                | docs/conf.py
+                | ez_setup.py
+                | ah_bootstrap.py
+            )$
```
andrewfullard, line 14: Why are the above files excluded from black?

tardis_env3.yml
```diff
@@ -63,6 +63,7 @@ dependencies:
   - docopt
 
   # Code quality
+  - pre-commit
```
andrewfullard, line 66: We do not currently have a robust method for determining new dependencies to be added. You can expect this to take some time.

.gitignore
Molkree: All entries added to .gitignore are from existing Black [configuration](https://github.com/tardis-sn/tardis/blob/351cc8f7b546f9eddf94608dbba2908ed333281b/pyproject.toml#L4). But these folders are just generic cache/build/temporary things, it's a good idea to just ignore them universally. To back it up, all of them are in various .gitignore templates from [gitignore repo](https://github.com/github/gitignore) and there are no such folders in tardis repo so we won't ignore something we shouldn't.

.pre-commit-config.yaml
Molkree: I don't know, they are excluded in the existing Black [config](https://github.com/tardis-sn/tardis/blob/351cc8f7b546f9eddf94608dbba2908ed333281b/pyproject.toml#L4) so I just copied them to the pre-commit Black config.

tardis_env3.yml
Molkree: Determining? Do you mean approving?
If it's the former I'm not sure what to determine here, `pre-commit` is the package required to install/run pre-commit hooks, it's available on [conda-forge](https://anaconda.org/conda-forge/pre-commit), it installs and works fine.
If it's the latter I thought #1563 was about it. Then I suppose I'd need to move this to separate PR, would love to hear if it's the case.

andrewfullard: Yeah sorry, approving is what I meant. Like I said, not a problem with this PR it will just need to be discussed with tardis core.

docs/development/code_quality.rst
```diff
@@ -24,6 +24,8 @@ Black
 
 A better method is to run Black automatically - first `integrate it within the code editor <https://black.readthedocs.io/en/stable/editor_integration.html>`_ you use and then enable the "format on save" or "format on type" option in your editor settings.
 
+If you've installed pre-commit hooks (`pre-commit install`) Black will run before each commit.
```
jaladh-singhal, line 27:
```suggestion
Alternatively, Black will run automatically before each commit if you've installed pre-commit hooks (as mentioned in git worklfow <add link to L154 there, using sphinx ref>)
```

jaladh-singhal (commented):
Everything looks great - thanks for effort @Molkree. This will surely save users from enabling black on their own but I'm not sure if it's advantageous enough that we should add pre-commit hooks for black, why:
1. If contributor's code hasn't followed black formatting, black checking workflow will fail on PR and prevent us from merging it unless it is formatted.
2. This will add `pre-commit` as new dependency and except black, we don't need it as of now for anything else - even for black it's optional (as mentioned in 1st point).

Molkree:
@jaladh-singhal
1. True, but moving checks earlier is good as it is.
2. Right now it's only Black but there's always potential there will be more in future (isort, flake8, etc).

This was already discussed in #1170 and I haven't seen any objections. And in #1405 you mentioned "enforcing by pre-commit hook will be in another PR" :O What has changed since January?

docs/development/code_quality.rst
Molkree: Thanks, I've changed it and added the ref.

andrewfullard (approved):

Molkree:
Rebased on master to solve merge conflict in README.rst

epassaro:
I think `pre-commit` shouldn't be a TARDIS dependency, just leave the choice of installing by yourself adding the `--no-update-deps` flag to `conda`. I think other hooks can be useful too in the future, but always optional!.

jaladh-singhal:
> This was already discussed in #1170 and I haven't seen any objections. And in #1405 you mentioned "enforcing by pre-commit hook will be in another PR" :O What has changed since January?

Molkree:
> #1398 made it possible to enforce black

That's true but in my opinion "please install pre-commit once if you want to contribute" is better than "please run Black before each commit if you want to contribute". Just a matter of convenience, everyone can forget to run formatter (just look at PRs, 2 out of the last 10 had commits fixing Black).

Molkree:
> just leave the choice of installing by yourself adding the `--no-update-deps` flag to `conda`

@epassaro 
I'm not super familiar with `conda` but I think `--no-update-deps` will still install `pre-commit`, it just won't touch existing installations. So this will still add a dependency. Maybe you wanted `--no-deps`?

Also is this your opinion or request for change as a member of Tardis?

andrewfullard:
This needs rebasing

Molkree:
@andrewfullard, rebased

Molkree:
Hm, I see there's a failing check, should I add myself to .mailmap?

andrewfullard:
> 
> 
> Hm, I see there's a failing check, should I add myself to .mailmap?

Yes please, that should fix the failure.

epassaro:
We recently moved to the new Astropy package template. I want to keep their `.gitignore` file untouched at least for a while @Molkree

Molkree:
@andrewfullard, mailmap fixed
@epassaro, gitignore reverted

andrewfullard:
@epassaro I'd like your review on this.

PR #1566, opened 2021-05-12
better handling of pickle5 backport

epassaro:
**Description**

- `pyne` macOS deprecation is still giving us problems (now with @KevinCawley and the numba integration pipeline). I don't think this patch is enough to fix our problems, but provides a better handling of the hack.

- Moved `pickle5` from `pip` section of the environment file (solves #1564).

**Motivation and context**
See #1535. 

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [ ] My change requires a change to the documentation.
    - [ ] I have updated the documentation accordingly.
    - [ ] (optional) I have built the documentation on my fork following [the instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html).
- [x] I have assigned and requested two reviewers for this pull request.

PR #1578, opened 2021-05-20
minor changes to compare refdata pipeline

epassaro:
**Description**

- Move `bokeh` installation step
- Use exact `python` version in `pickle5` patch - [see here](https://www.python.org/downloads/release/python-383/)

**Motivation and context**

**How has this been tested?**
- [x] Testing pipeline.

**Examples**

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/__init__.py
+++ b/tardis/__init__.py
@@ -34,9 +34,9 @@ logger.addHandler(console_handler)
 logging.getLogger("py.warnings").addHandler(console_handler)
 
 # ----------------------------------------------------------------------------
-# pyne holds Python 3.7 on macOS, but refdata is pickled with protocol 5 (3.8)
+# pyne holds Python 3.7 on macOS, but refdata is pickled with protocol 5 (3.8.3)
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 8, 3):
     import pickle5
 
     sys.modules["pickle"] = pickle5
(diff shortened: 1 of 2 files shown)


andrewfullard (approved):

tardis/__init__.py
```diff
@@ -34,9 +34,9 @@
 logging.getLogger("py.warnings").addHandler(console_handler)
 
 # ----------------------------------------------------------------------------
-# pyne holds Python 3.7 on macOS, but refdata is pickled with protocol 5 (3.8)
+# pyne holds Python 3.7 on macOS, but refdata is pickled with protocol 5 (3.8.3)
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 8, 3):
```
Molkree, line 39: @epassaro, oh, I think this check is not needed at all. pickle5 is only available on < 3.8 through conda, the whole tardis environment now resolves to Python 3.7 after #1566 because of this (macOS AND Linux).

epassaro: The data is already pickled with 3.8, so yes, we need this

Molkree: I think I'm missing something, can you point at the workflow with Python 3.8, please?

PR #1597, opened 2021-05-31
add value error to ensure active virtual packet logging in SDECPlotter

yuyizheng1112:
This PR allows SDECPlotter to raise error if `virtual_packet_logging` is not set to `True`.

**Motivation and context**

Before this change, SDECPlotter can work even when `virtual_packet_logging` is not active. In this case, an index error is raised by function in `matplotlib` while calling the function `plotter.generate_plot_mlp()`, and the output is confusing. 
Since SDECPlotter requires `virtual_packet_logging`, it should ensure the option is set to true.
**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Examples**

Before change:
image

After change:
截屏2021-06-21 下午4 25 57

**Type of change**

- [x] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -423,9 +423,11 @@ class SDECPlotter:
                 )
             )
         else:
-            raise ValueError(
-                "Virtual packet logging is inactive. "
-                "Please set config option virtual_packet_logging: True."
+            return cls(
+                dict(
+                    virtual=None,
+                    real=SDECData.from_simulation(sim, "real"),
+                )
             )
 
     @classmethod
@@ -572,6 +574,11 @@ class SDECPlotter:
             raise ValueError(
                 "Invalid value passed to packets_mode. Only "
                 "allowed values are 'virtual' or 'real'"
+            )
+
+        if packets_mode == "virtual" and self.data[packets_mode] is None:
+            raise ValueError(
+                "SDECPlotter doesn't have any data for virtual packets population and SDEC plot for the same was requested.\nEither set virtual_packet_logging: True in your configuration file to generate SDEC plot with virtual packets, or pass packets_mode=\"real\" in your function call to generate SDEC plot with real packets."
             )
 
         # Store the plottable range of each spectrum property which is
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -415,12 +415,18 @@ class SDECPlotter:
         -------
         SDECPlotter
         """
-        return cls(
-            dict(
-                virtual=SDECData.from_simulation(sim, "virtual"),
-                real=SDECData.from_simulation(sim, "real"),
+        if sim.runner.virt_logging:
+            return cls(
+                dict(
+                    virtual=SDECData.from_simulation(sim, "virtual"),
+                    real=SDECData.from_simulation(sim, "real"),
+                )
             )
-        )
+        else:
+            raise ValueError(
+                "Virtual packet logging is inactive. "
+                "Please set config option virtual_packet_logging: True."
+            )
 
     @classmethod
     def from_hdf(cls, hdf_fpath):
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -578,7 +578,7 @@ class SDECPlotter:
 
         if packets_mode == "virtual" and self.data[packets_mode] is None:
             raise ValueError(
-                "SDECPlotter doesn't have any data for virtual packets population and SDEC plot for the same was requested.\nEither set virtual_packet_logging: True in your configuration file to generate SDEC plot with virtual packets, or pass packets_mode=\"real\" in your function call to generate SDEC plot with real packets."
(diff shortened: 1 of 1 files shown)


jamesgillanders:
I don't understand why virtual_packet_logging needs to be set to True - I thought we left it so that we could make SDEC plots with either real or virtual packets?

jamesgillanders:
@jaladh-singhal do you know why this isn't working?

wkerzendorf:
@yuyizheng1112 can you write tests for the PR?

yuyizheng1112:
@jamesgillanders When calling `generate_plot_mlp()` in the case of `virtual_packet_logging: False`, the function will raise the error above, and I think this can be confusing for people unfamiliar with tardis. 

I also think there is no need to set it to true for those who plot real packets only. But the instruction in `sdec_plot.ipynb` indicates that the option must set to true in order to generate SDEC plot. I did this to consist with the instruction & avoid the error raised by `generate_plot_mlp()`.

Maybe I can put this exception in function `generate_plot_mlp` instead of when generating plotter. What do you think about it?

jamesgillanders:
So can I clarify something? There should be functionality to allow the SDEC plot to generate with real packets, but that does not currently work. And this PR will print an error message to explain that the real packets mode don't work. Is that right? Then at some point, after this fix has been implemented, will we fix the real packets mode so that it does work?

jamesgillanders:
If all of that is the case, then we will have to come back and remove this fix, so that the code will allow a plot to be generated using real packets. Or am I misunderstanding the problem being addressed here?

yuyizheng1112:
@jamesgillanders  The default value for `packets_mode` is `"virtual"` in this function, so in this case, we are plotting using virtual packets. If virtual packet is not loaded in, it will raise the confusing exception above. As for the real mode, it works well no matter the option is active or not.
Does it make sense? :)

jaladh-singhal:
> @jaladh-singhal do you know why this isn't working?

@jamesgillanders becuase there are currently no tests for SDEC plot hence test coverage is failing

jaladh-singhal:
@yuyizheng1112 @jamesgillanders I think I understand both of your reasoning. I see an intersecting solution for this problem:
- First, inside `SDECPlotter.from_simulation`, if `sim.runner.virt_logging` is False then return dictionary with `virtual=None`, otherwise keep it as it is i.e. `virtual=SDECData.from_simulation(sim, "virtual")`. 

- Then, in `_calculate_plotting_data()`, if `packets_mode` is `virtual` and `data[packets_mode]` is None then raise ValueError that "SDECPlotter doesn't have any data for virtual packets population and SDEC plot for the same was requested. Either set `virtual_packet_logging: True` in your simulation's configuration file to generate SDEC plot for virtual packets population, or pass `packets_mode="real"` in your function call to generate SDEC plot for real packets population."

This way even when virtual packets logging is not enabled, SDEC plot can still be obtained for real packets mode (what @jamesgillanders is saying) and will only give error when virtual packets mode is requested by user (what @yuyizheng1112 is trying to do).

jaladh-singhal:
> But the instruction in sdec_plot.ipynb indicates that the option must set to true in order to generate SDEC plot. I did this to consist with the instruction & avoid the error raised by generate_plot_mlp().

@yuyizheng1112 Good point! The instruction should actually be changed to "The virtual packet logging capability must be active in order to produce SDEC Plot **for virtual packets population**. Thus ..." and it should be moved right after 5th cell's output.

jaladh-singhal (changes requested):
@yuyizheng1112 can you please change this as per my comments?

yuyizheng1112:
@jaladh-singhal Sure. I've changed both ValueError and instruction.

tardis/visualization/tools/sdec_plot.py
```diff
@@ -568,6 +576,11 @@ def _calculate_plotting_data(
                 "allowed values are 'virtual' or 'real'"
             )
 
+        if packets_mode == "virtual" and self.data[packets_mode] is None:
+            raise ValueError(
+                'SDECPlotter doesn\'t have any data for virtual packets population and SDEC plot for the same was requested.\nEither set virtual_packet_logging: True in your configuration file to generate SDEC plot with virtual packets, or pass packets_mode="real" in your function call to generate SDEC plot with real packets.'
```
jaladh-singhal, line 581: It's better to break long string to multiple lines following 80 char limit. Also if you use `"..."` for your string you will not need to escape `'` used inside your string - see how L575-L576 does this

jaladh-singhal (changes requested):
Thanks for the changes @yuyizheng1112 - I had a one minor comment.

Also, I hope you've tested this for config files with vpacket logging not enabled - can you post screenshots of that?

jaladh-singhal (approved):
Excellent!

docs/io/visualization/sdec_plot.ipynb
```diff
@@ -328,7 +328,7 @@
    "source": [
```
jaladh-singhal, line 328: <p>There's an extra newline here that shouldn't be present</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/1597/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

KevinCawley (approved):

Issue #1282, opened 2020-08-21
Create an option to build docs without executing notebooks


jaladh-singhal:
Currently, whenever docs are build, the notebooks which are marked to be executed by nbsphinx make build time very long  - since each of them runs simulation model. This is fine for production but for development this becomes really annoying.

So there must be some option or flag while running `make html` or `python -m sphinx build` to explicitly disable notebooks' execution by nbsphinx on run-time. 

Based on my very brief search, nbsphinx doesn't provide any such option for execute control during run-time but only for conf.py and individual notebook's metadata. But I believe, there should be a way to implement this, by adding some if-else logic in conf.py for nbsphinx execute setting which listens to flag specified to build command or something like that.

epassaro:
The default way to execute cells while rendering notebooks with `nbsphinx` is to leave all cells blank. If some cell is already executed then the notebook is **not executed**.

I understand the TARDIS documentation relies on notebook metadata to **always** execute the notebook and ignore the above method. Using this method and running once the Quickstart guide should be enough to build the documentation more quickly.

I don't know about a fancier way to do this.

epassaro:
Closed by #1615

PR #1615, opened 2021-06-03
add flag to disable nbsphinx

epassaro:
**Description**

I thought about a way to fast render the documentation. Made a small patch to `conf.py` and now prepending the variable `DISABLE_NBSPHINX=1` to the build command skips the notebook rendering.

Example:

```
cd docs/
DISABLE_NBSPHINX=1 make html
```

Then `sphinx` just render the `.rst` files.

Also I removed hardcoded metadata in notebooks. If we want to change this values we should do it through `conf.py` not harcoding them (by the way, the hardcoded values were the default ones). Using harcoded metadata leads to hard to debug problems.

**Motivation and context**

@jaladh-singhal's issue no. #1282

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

Locally. Please checkout to this PR and try it too.

**Examples**

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -20,11 +20,9 @@ Besides this, the functions and classes in your code must always contain **docst
 Building documentation locally
 ==============================
 
-To build TARDIS documentation locally, you can use either of the following commands: ::
+To build TARDIS documentation locally, use the following commands:
 
-    python setup.py build_docs
-    
-or ::
+.. code ::
 
     cd docs
     make html
@@ -32,7 +30,7 @@ or ::
 .. note :: 
 
     - If you're working on a fresh local copy of the TARDIS repository, you might need to do ``python setup.py develop`` before executing these commands.
-    - Prepend ``DISABLE_NBSPHINX=1`` to the build command to disable notebook rendering (fast mode).
+    - Use ``DISABLE_NBSPHINX=1 make html`` to disable notebook rendering (fast mode).
 
 After running this command, you can find the built docs (i.e. HTML webpages) in ``docs/_build``. Open the ``index.html`` in your browser to see how the documentation looks like with your edits. Navigate to page where you made changes or file that you added to check whether it looks as intended or not.
 
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -29,7 +29,10 @@ or ::
     cd docs
     make html
 
-.. note :: If you're working on a fresh local copy of the TARDIS repository, you might need to do ``python setup.py develop`` before executing these commands.
+.. note :: 
+
+    - If you're working on a fresh local copy of the TARDIS repository, you might need to do ``python setup.py develop`` before executing these commands.
+    - Prepend ``DISABLE_NBSPHINX=1`` to the build command to disable notebook rendering (fast mode).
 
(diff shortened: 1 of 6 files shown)


smithis7:
Two things:
- Does this work with `python setup.py build_docs` as well? And would you just do `DISABLE_NBSPHINX=1 python setup.py build_docs`?
- You may want to add that example you made in the pull request comment into the documentation (i.e. say that you do `cd docs` and then `DISABLE_NBSPHINX=1 make html`).

epassaro:
> Two things:
> 
> * Does this work with `python setup.py build_docs` as well? And would you just do `DISABLE_NBSPHINX=1 python setup.py build_docs`?
> * You may want to add that example you made in the pull request comment into the documentation (i.e. say that you do `cd docs` and then `DISABLE_NBSPHINX=1 make html`).

- Yes it works.
- It's already in the documentation. Is "prepend to the build command" is clear enough?

smithis7:
> > Two things:
> > 
> > * Does this work with `python setup.py build_docs` as well? And would you just do `DISABLE_NBSPHINX=1 python setup.py build_docs`?
> > * You may want to add that example you made in the pull request comment into the documentation (i.e. say that you do `cd docs` and then `DISABLE_NBSPHINX=1 make html`).
> 
> * Yes it works.
> * It's already in the documentation. Is "prepend to the build command" clear enough?

In my opinion, it could be more clear. Writing out the examples I think would be helpful.

docs/development/documentation_guidelines.rst
```diff
@@ -29,7 +29,10 @@ or ::
     cd docs
     make html
 
-.. note :: If you're working on a fresh local copy of the TARDIS repository, you might need to do ``python setup.py develop`` before executing these commands.
+.. note :: 
+
+    - If you're working on a fresh local copy of the TARDIS repository, you might need to do ``python setup.py develop`` before executing these commands.
+    - Prepend ``DISABLE_NBSPHINX=1`` to the build command to disable notebook rendering (fast mode).
```
andrewfullard, line 35: I'm not entirely sure what you mean by prepend here. Would it be `DISABLE_NBSPHINX=1 python setup.py develop` as the whole command?

epassaro: yes, It's basically an env variable. 

Also you can use `export DISABLE_NBSPHINX=1` and then run the build command.

wkerzendorf: maybe line 75 on the disable nbsphinx to say `never`

epassaro: done!

smithis7 (commented):
Looks good, except that I think you should put the examples of how to do this (both with build docs and with make html) in the documentation.

epassaro:
> Looks good, except that I think you should put the examples of how to do this (both with build docs and with make html) in the documentation.

`python setup.py build_docs` will be deprecated soon so I would not encourage people to learn or use it. Maybe I'll remove the command from all documentation.

epassaro:
> > Looks good, except that I think you should put the examples of how to do this (both with build docs and with make html) in the documentation.
> 
> `python setup.py build_docs` will be deprecated soon so I would not encourage people to learn or use it. Maybe I'll remove the command from all documentation.

I removed the `python setup.py build_docs` command and used the full command to explain the flag @smithis7.

smithis7 (approved):

andrewfullard (changes requested):
Needs a rebase

epassaro:
> Needs a rebase

Done

andrewfullard (approved):

andrewfullard (approved):

Issue #1614, opened 2021-06-03
Virtual Logging & Nthreads


marxwillia:
**Describe the bug**
Kernel crashes when `nthreads` > 1 and `virtual_packet_logging` set to True.

**To Reproduce**
Run a tardis simulation with most up to date repo version using the attached yml file.

**Screenshots**

**System**
 - OS:
    - [ ] GNU/Linux
    - [x] macOS

 - Environment (`conda list`):

fresh install of TARDIS environment, using numba 0.53.1

**Additional context**
TARDIS simulation runs with no errors when virtual logging is True and nthreads=1.  Likewise, TARDIS simulation runs with no errors when nthreads>1 and virtual logging is False.  There is a specific issue with having both nthreads > 1 and virtual logging True.

marxwillia:
yml file below:

`# Example YAML configuration for TARDIS
tardis_config_version: v1.0

supernova:
  luminosity_requested: 9.44 log_lsun
  time_explosion: 13 day

atom_data: kurucz_cd23_chianti_H_He.h5

model:
  structure:
    type: specific
    velocity:
      start: 1.1e4 km/s
      stop: 20000 km/s
      num: 20
    density:
      type: branch85_w7

  abundances:
    type: uniform
    O: 0.19
    Mg: 0.03
    Si: 0.52
    S: 0.19
    Ar: 0.04
    Ca: 0.03

plasma:
  disable_electron_scattering: no
  ionization: lte
  excitation: lte
  radiative_rates_type: dilute-blackbody
  line_interaction_type: macroatom

montecarlo:
  seed: 23111963
  no_of_packets: 4.0e+4
  iterations: 1
  nthreads: 5

  last_no_of_packets: 1.e+5
  no_of_virtual_packets: 3

  convergence_strategy:
    type: damped
    damping_constant: 1.0
    threshold: 0.05
    fraction: 0.8
    hold_iterations: 3
    t_inner:
      damping_constant: 1.0

spectrum:
  start: 500 angstrom
  stop: 20000 angstrom
  num: 10000
  virtual:
    virtual_packet_logging: True
`

PR #1616, opened 2021-06-03
Virtual Packet Logging Nthreads Crash Fix

marxwillia:
**Description**
PR to address issue #1614 

**Motivation and context**
TARDIS crashes when virtual packet logging is True and nthreads > 1.

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/base.py
+++ b/tardis/montecarlo/montecarlo_numba/base.py
@@ -82,25 +82,19 @@ def montecarlo_radial1d(model, plasma, runner):
     runner.last_line_interaction_out_id = last_line_interaction_out_id
 
     if montecarlo_configuration.VPACKET_LOGGING and number_of_vpackets > 0:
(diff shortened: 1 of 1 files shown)


marxwillia:
I thought the issue was main_loop returning a list of arrays, so I now flatten the list of arrays inside main loop.  Kernel still crashes.  My method of using np.append is probably very slow, but I had issues getting concatenate to work with numba because it requires a tuple of arrays. @stuartarchibald any advice on best way to flatten a list of variable length np.arrays in numba?  You can see my `_flatten_arr_list` function, which seems kind of hacky.

andrewfullard:
Closing this since we have a full fix on the way

PR #1626, opened 2021-06-07
Custom Abundance Widget demo notebook

yuyizheng1112:
This PR demonstrates my work on GSoC project.

## Screen shot for the latest interface
GUIdemo_Jupyter_Notebook-2


MarkMageeAstro:
Can you add some examples and screenshots to show how it works?

yuyizheng1112:
@MarkMageeAstro I've made a notebook to show how it works. Please check it in this PR.

MarkMageeAstro:
I might be missing something, but it looks like some of the output in the notebook didn't render? I can't see the parts showing the stuff you put in the slack channel. I mean the parts with the dropdowns and changing abundances, etc.

yuyizheng1112:
@MarkMageeAstro The widgets won't be visible if you are viewing this notebook as an html page. You can run the codes in it so that output can be generated.

MarkMageeAstro:
This looks like a really good start, @yuyizheng1112, nice work. I have a few comments below.

1. It would be nice if there was also a plot showing the abundances. You can have the abundances vs. velocity and each element could be a different colour. Then you would be able to see what the composition is like visually and use that to help guide any changes you want to make in the model.
2. I think the additions we talked about last week would be good to add because that's definitely functionality that any user would want. The ability to add shells is probably a bit more complicated and something we should discuss at the next meeting, but the other stuff like adding new elements, a lock function, and the ability to edit multiple shells at a time would be good. 

This isn't really related to the widget, but in your custom csvy file the density is very high. That's partially why the W in the model is >1. It should actually be around 0.5. I would try dividing the density in each cell by 10 and trying that - it'll make the model a lot faster too.

jaladh-singhal (commented):
This looks great @yuyizheng1112 - I second what @MarkMageeAstro said. I know you are working on lock functionality. After that your next steps should be:
1. Producing the plot
2. Add more elements
3. Try if your logic/design works for making edits to mutiple shells at same time

MarkMageeAstro:
Very nice work @yuyizheng1112, looks great. Some minor comments below.

1. I see in the screenshot that the units have been fixed, but in my version that I just pulled they're still off by a factor of 10. There are some cells in the notebook dividing by 10e5, when it should be 1e5 or 10e4. Also, I think we should stick to unit conversions to avoid issues like this. So if the velocities are read in as cm/s, then use astropy units to convert to km/s. You should also set it so that it doesn't use scientific notation for the velocities, i.e. it should be 12,000 and not 1.2e4.
2. The x axis label can just be 'Velocity (km/s)'. Putting it as 'velocity of outer boundary' could be a bit confusing because each cell does have an outer boundary, but we usually refer to the end of the model as 'outer boundary'.
3. Looks like there's a bug with the last cell. If I run it multiple times, it keeps adding new lines to the plot, so I end up with 2+  lines for each element. That should be fixed so that it'll just refresh the plot and not add the same data multiple times.
4. I think we talked about changing the formatting in the abundance boxes to scientific notation with 2 or 3 decimal places. That'll be cleaner than just standard float notation with 8+ decimal places.

MarkMageeAstro (changes requested):
Good job. I left a comment with a few cosmetic tweaks. Apart from that, I think you can move on to adding the new functionality.

luharvey:
Great work @yuyizheng1112, the widget is looking fantastic and I really like the normalisation functionality.

One small bug I noticed was to do with the 'add element' function. If I add an element with a higher atomic number it adds it in no problem and everything works nicely. When I add an element with a smaller atomic number however, the element slider appears in the right place and the colours of the lines shift but the updating of the legend has the new element at the bottom, meaning that the legend doesn't have the correct colours for each element.

docs/GUIdemo/my_yml.yml
```diff
@@ -0,0 +1,41 @@
+# Example YAML configuration for TARDIS
```
smithis7, line 1: Can you make this a symbolic link that points to `docs/tardis_example.yml`, and then it is best if the link is called `tardis_example.yml` for consistency.

docs/GUIdemo/tardis_example.yml
```diff
@@ -0,0 +1,57 @@
+# Example YAML configuration for TARDIS
```
smithis7, line 1: Can you make this a symbolic link that points to `docs/tardis_example.yml`?

smithis7:
On the documentation side, on top of a the other comment that I made about `tardis_example.yml`, can we rename your directory from `GUIdemo` to something more descriptive of your specific GUI, as we have multiple GUIs that are included in the documentation? Also, I think it would be good to move that directory into `docs/io/visualization/` (this will require a rebase). Finally, make sure that before this PR is merged that you add your notebook into the toctree of `docs/io/visualization/index.rst` so that the page will show up in the documentation properly!

MarkMageeAstro:
I noticed a bug in the demo notebook. When I put my cursor over the mass fractions, the velocity doesn't show up when it's <16,400 km/s. It just says e.g. (, 0.52) for Si at 12,000 km/s

Whenever you're ready to convert to a full PR, can you also update the first post to use the PR template.

docs/GUIdemo/tardis_example.yml
yuyizheng1112: This pr is a demo for GUI and will not be merged. I'll keep things you mentioned in my mind when I make the guiding notebook. Thanks for your review!

yuyizheng1112:
@MarkMageeAstro Sure. As for the bug you mentioned, I've also noticed & fixed it. Thanks!

jamesgillanders:
Can we close this now?

yuyizheng1112:
@jamesgillanders Definitely. Since the module is merged.

PR #1606, opened 2021-06-02
Last interaction radius

marxwillia:
**Description**
Adding a new visualization tool which allows the user to see the location of the last line interaction of all packets in their TARDIS simulation.  This allows the user to see exactly where in the ejecta certain lines transitions are predominantly occurring.  This information can be used to adjust models to better fit observations.

**Motivation and context**
This new tool will allow users to better understand their models and fit to observed data more easily.

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**
Below is an example plot showing the location distributions for a variety of elements/ions.  This new tool will implement both matplotlib and plotly versions.

interaction_dist

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [x] My change requires a change to the documentation.
    - [ ] I have updated the documentation accordingly.
    - [ ] (optional) I have built the documentation on my fork following [the instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html).
- [x] I have assigned and requested two reviewers for this pull request.

PR #1628, opened 2021-06-07
Add Plotter class to plot_util module

yuyizheng1112:
This PR aims to solve the code overlap between  `SDECPlotter` and `InteractionRadiusPlotter`. Here I add `Plotter` class to `plot_util.py` as a composition of plotter tools, which encapsulates some common functions used by both plotters.

**Motivation and context**

Cope with code overlap in pr #1606 

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**


jamesgillanders (approved):
Seems like a good idea to prevent duplication of code. Is this definitely where we want to put it? @MarkMageeAstro @marxwillia - if so, I'd be happy to merge, pending a second review

MarkMageeAstro:
@yuyizheng1112, can you provide more information on the tests? Does making this change still behave as expected? I haven't run the code myself.

@jamesgillanders, I suppose it would go here or in the analysis folder. Since this doesn't really return anything, but is just used for plotting functions, it's probably fine to put here.

andrewfullard:
This needs rebasing

tardis/visualization/plot_util.py
```diff
@@ -60,3 +73,255 @@ def get_mid_point_idx(arr):
...
+                        full_species_list.append(species)
+
+                # full_species_list is now a list containing each individual species requested
+                # e.g. it parses species_list = [Si I - V] into species_list = [Si I, Si II, Si III, Si IV, Si V]
+                self._full_species_list = full_species_list
+                requested_species_ids = []
+                keep_colour = []
+
+                # go through each of the requested species. Check whether it is
+                # an element or ion (ions have spaces). If it is an element,
```
wkerzendorf, line 138: this is a dangerous assumption IMHO

tardis/visualization/plot_util.py
```diff
@@ -60,3 +73,255 @@ def get_mid_point_idx(arr):
...
+                # add all possible ions to the ions list. Otherwise just add
+                # the requested ion
+                for species in full_species_list:
+                    if " " in species:
+                        requested_species_ids.append(
+                            [
+                                species_string_to_tuple(species)[0] * 100
```
wkerzendorf, line 145: why `* 100`

jamesgillanders: so that the IDs are all 3 or 4 digit numbers - e.g. Ca II = 2001

tardis/visualization/plot_util.py
jamesgillanders: This should be fine as it is well documented that if the user doesn't include a space, the code will fall over. And it was the easiest way we could think of at the time for differentiating between a mixed list of ions and elements in a list.

tardis/visualization/plot_util.py
jamesgillanders: FYI, these are all in the current implementation of the SDEC plotter - all this PR does is move these functions to a separate file to prevent duplication.

jaladh-singhal (commented):
I think it's better to do these changes within #1606 because it will be easier to see why  so many of the methods have been moved? 

Also we need to consider if moving these methods don't make writing tests for SDEC plot even harder. And how have you tested if it doesn't break SDEC plot by running the notebook?

yuyizheng1112:
@jaladh-singhal I think it makes sense to do this in another PR because #1606 and this PR are editing different modules.

For the second question, do you mean if it is harder to write tests for SDEC Plot together with `Plotter` class? 
And I ran the SDEC plot in the notebook. The plot works as usual.

jaladh-singhal:
> For the second question, do you mean if it is harder to write tests for SDEC Plot together with Plotter class?

Not exactly harder but we might need to change the way we're planning to write tests now.

andrewfullard:
Can you fix the merge conflict @yuyizheng1112 ? Then we can merge.

wkerzendorf:
See #1838 as it pertains to this issue.

wkerzendorf:
@visualization team - what is going on with this?

jamesgillanders:
Closing this for now. The idea is sound, but we need to update the InteractionRadiusPlotter tool first

PR #1630, opened 2021-06-07
Fix FileExistsError in to_hdf()

### LINKED ISSUE #1624: to_hdf() failed for Simulation object
**Describe the bug**

`to_hdf()` doesn't work when called by a Simulation object. A `FileExistsError` is raised though there was no such file before running.

The bug might be here:
https://github.com/tardis-sn/tardis/blob/8ac89320d50561647399ee35c84d6ec8f3b55cf0/tardis/io/util.py#L233-L247
`buf = pd.HDFStore(path_or_buf, complevel=complevel, complib=complib)`in `try` block will create the HDF file. Therefore, FileExistsError will be definitely raised if `overwrite=False` while running `else` block.

**To Reproduce**

Follow the instruction in the documentation [here](https://tardis-sn.github.io/tardis/using/interaction/to_hdf_notebook.html#to-hdf-notebook).

**Screenshots**

image

**System**
 - OS:
    - [ ] GNU/Linux
    - [x] macOS

yuyizheng1112:
Fix #1624.

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Examples**
Before the change:
image

After the change:
image

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/util.py
+++ b/tardis/io/util.py
@@ -230,21 +230,21 @@ class HDFWriterMixin(object):
         HDFStore because the user decides on the mode in which they have
         opened the HDFStore ('r', 'w' or 'a').
         """
-        try:  # when path_or_buf is a str, the HDFStore should get created
-            buf = pd.HDFStore(path_or_buf, complevel=complevel, complib=complib)
-        except TypeError as e:
-            if e.message == "Expected bytes, got HDFStore":
-                # when path_or_buf is an HDFStore buffer instead
-                buf = path_or_buf
-            else:
-                raise e
-        else:  # path_or_buf was a str
-            if os.path.exists(path_or_buf) and not overwrite:
-                buf.close()
-                raise FileExistsError(
-                    "The specified HDF file already exists. If you still want "
-                    "to overwrite it, set option overwrite=True"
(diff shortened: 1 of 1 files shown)


jaladh-singhal (approved):
Excellent catch @yuyizheng1112 - thanks for fixing it!

tardis/io/util.py
```diff
@@ -230,21 +230,27 @@ def to_hdf_util(
...
-                    "The specified HDF file already exists. If you still want "
-                    "to overwrite it, set option overwrite=True"
+        if (
+            isinstance(path_or_buf, str)
+            and os.path.exists(path_or_buf)
+            and not overwrite
+        ):
+            raise FileExistsError(
+                "The specified HDF file already exists. If you still want "
+                "to overwrite it, set option overwrite=True"
```
jaladh-singhal, line 240:
```suggestion
                "to overwrite it, set function parameter overwrite=True"
```

KevinCawley (approved):

PR #1636, opened 2021-06-08
Adding Convergence Plots

atharva-2001:
This PR adds convergence plots to TARDIS. 

**Description**

Convergence plots will allow the user to see the convergence of the simulation in plots and study the state of the simulation at different iterations. Convergence plots are plotted by default as the simulation runs. One can configure the plots by providing arguments in the `run_tardis` function. This includes overriding the default plotly layout of the graphs and options to change color scale and show or not show the plots.

**Motivation and context**

**How has this been tested?**
- [X] Testing pipeline.
- [ ] Other. 

**Examples**

[Link](https://atharva-2001.github.io/tardis/branch/convergence_plots/io/visualization/convergence_plot) to the convergence plots in the documentation.
[Link](https://nbviewer.jupyter.org/github/atharva-2001/tardis/blob/convergence_plots/docs/io/visualization/convergence_plot.ipynb ) to the notebook.
[Link](https://atharva-2001.github.io/tardis/branch/convergence_plots/api/tardis.visualization.tools.convergence_plot.html) to the API
cplots-new2

Convergence Plots along with progress bars
Peek 2021-07-09 18-15

**Type of change**

- [ ] Bug fix. 
- [X] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [X] My change requires a change to the documentation.
    - [X] I have updated the documentation accordingly.
    - [X] (optional) I have built the documentation on my fork following [the instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html).
- [X] I have assigned and requested two reviewers for this pull request.


MarkMageeAstro:
This looks like a great start. I have a few stylistic comments:

Plasma updates
1. Can you swap the plot round so that temperature is on the left? There are no labels for temperature or velocity, is this the latex rendering issue?
2. I think the first example you have with iterations and then just the numbers looks the best. It's hard to tell exactly what's going on in the last one because the lines aren't connected and that also makes it difficult to see the colours on the right hand side
3. It would be best to change the display formatting and add units. The full temperature should be printed out, not just rounded to the nearest 1000. For example, use 10,000 K and not 10k K. Similarly, the x-axis should be in km/s but at the minute is in cm/s
4. Personally, I think a different colour scheme would be good, like rainbow or jet. If we stick with this one then I would at least swap it round so that the lighter yellow is the last iteration.


Temperature plots
1. We need units here too. Luminosity should be in erg / s
2. A bar chart isn't really the best way to present this data, so I would remove it. The line plot is much better. It should also have a legend to say what each colour is and to say what the horizontal line is. I would use some more contrasting colours as well
3. I think the labels are the wrong way round. In the line plot, the final emitted luminosity is listed as 4e42, but the simulation output says this is the luminosity absorbed. 
4. Might also be nice to show a residual difference between the emitted and requested luminosities for each iteration. This can just be the difference between the two given as a percentage of the luminosity requested or something like that. Could be a subpanel to the main plot, just below

jaladh-singhal (commented):
Apparently, plots didn't show up for me when running simulation - do I have to enable them by passing option? (I thought we were keeping plots by default on and logging turned off)

atharva-2001:
I'm sorry, I had done that to see if I was able to pass the tests. Could you please try once more? The plots are on now by default.

atharva-2001:
Thanks, @MarkMageeAstro! 
I'm sorry for not changing the `t_inner` damping constant. I changed it now.  I also changed the height of the plots and removed the whitespace. Could you please tell me if that looks good to you? 
The `requested_luminosity` line, among other lines in the luminosity plot, do not show marker points by default. 
I just changed it to demonstrate that the default configuration can be overridden. 
image
I'm afraid the link to the notebook in the documentation is generated automatically due to code in `docs/conf.py`. The link will be fixed once the PR is merged. 
I guess the less number of ticks was due to the small width of the plot. It's showing 4 ticks for me now. Should I increase the number of ticks?  Do you mean adding ticks like in [this](https://stackoverflow.com/questions/38064175/major-and-minor-tickmarks-with-plotly)?

atharva-2001:
I'm afraid there isn't a direct way to add minor ticks to plotly charts, but there are workarounds. We would have to specify the names of the ticks or, `ticktext`, as well as the values of the ticks or, `tickvals`. Leaving the `ticktext` of some of the ticks as an empty string would create minor ticks. But we would have to automate finding `tickvals` and `ticktext` according to the simulation result. I'm still working on it, though.

tardis/simulation/base.py
```diff
@@ -162,6 +167,16 @@ def __init__(
                 f"- input is {convergence_strategy.type}"
             )
 
+        if cplots_kwargs["show_plots"] is not False:
```
jaladh-singhal, line 170:
```suggestion
        if cplots_kwargs["show_plots"]:
```
or ` if cplots_kwargs["show_plots"] is True:`

Why not keep it simple :smile:

tardis/simulation/base.py
```diff
@@ -553,6 +605,17 @@ def from_config(
...
             )
 
+        cplots_config_options = [
+            "plasma_plot_config",
+            "luminosity_plot_config",
+            "colorscale",
+            "show_plots",
+            "export_cplots",
+        ]
+        cplots_kwargs = defaultdict(dict)
```
jaladh-singhal, line 615: Is this for creating an empty dictionary?

If yes, then why not do `= {}`?

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+import matplotlib.colors as clr
+import plotly.graph_objects as go
+from IPython.display import display
+import matplotlib as mpl
+import ipywidgets as widgets
+from contextlib import suppress
+from traitlets import TraitError
+
+
+def transition_colors(name="jet", iterations=20):
```
jaladh-singhal, line 12: I don't understand why you're using a default value of iterations, shouldn't it always come from config file?

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+
+        if "colorscale" in kwargs:
+            self.colorscale = transition_colors(
+                kwargs["colorscale"], iterations=self.iterations
+            )
+        else:
+            self.colorscale = transition_colors(iterations=self.iterations)
+
+        if "plasma_plot_config" in kwargs:
+            if kwargs["plasma_plot_config"] != {}:
```
jaladh-singhal, line 62: Better to `and` the both. Also `if x != {}` is equivalent to `if x` - we're checking truthy/falsy values

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+                self.plasma_plot_config = kwargs["plasma_plot_config"]
+
+        if "luminosity_plot_config" in kwargs:
+            if kwargs["luminosity_plot_config"] != {}:
```
jaladh-singhal, line 66: Same as above - and both

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+                automargin=True,
+                side="top",
+                tickformat="g",
+                exponentformat="e",
+                tickmode="auto",
+                nticks=4,
+            ),
+            yaxis2=dict(
+                exponentformat="e",
+                title=r"$\mbox{Luminosity}~(erg~sec^{-1})$",
```
jaladh-singhal, line 194:
```suggestion
                title=r"$\mbox{Luminosity}~(erg~s^{-1})$",
```

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+        if hasattr(self, "plasma_plot_config"):
+            self.override_plot_parameters(fig, self.plasma_plot_config)
+        self.plasma_plot = fig
+
+    def create_luminosity_plot(self):
+        """
+        Creates an empty luminosity plot.
+        The default layout can be overridden by passing luminosity_plot_config dictionary in the run_tardis function.
+        """
+        line_colors = ["orangered", "lightseagreen", "indigo"]
```
jaladh-singhal, line 134: Is it documented how can one override these colors?

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+                    self.override_plot_parameters(trace, value)
+            else:
+                if type(value) == dict:
+                    self.override_plot_parameters(fig[key], value)
+                else:
+                    fig[key] = value
+
+    def build(self, display_plot=True):
+        """
+        Calls the create_plasma_plot and the create_luminosity_plot to build plots.
```
jaladh-singhal, line 239: A docstring of function should never be concerned with what functions it calls - think about it, a user reading your don't know inner mechanics of your code so just writing what functions this function calls isn't helpful. Instead write what it does including the what the functions it call does.

For e.g. something like: "Create both plasma plot and luminosity plot and display them together."

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+        if display_plot:
+            display(
+                widgets.VBox(
+                    [self.plasma_plot, self.luminosity_plot],
+                )
+            )
+
+    def update_plasma_plots(self):
+        """
+        Updates the plasma plots using the data collected using the fetch_data function.
```
jaladh-singhal, line 252: Again no need to tell `..using the fetch_data function` instead explain what it does

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+            col=2,
+            legendgroup=f"group-{self.current_iteration}",
+            name=self.current_iteration,
+            customdata=customdata,
+            hovertemplate="<b>Y</b>: %{y:.2f} at <b>X</b> = %{x:,.0f}%{customdata}",
+        )
+
+    def update_luminosity_plot(self):
+        """
+        Updates the plasma plots using the data collected using the fetch_data function.
```
jaladh-singhal, line 299: Fix this too

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+            if last:
+                if hasattr(self, "plasma_plot_config"):
+                    if "data" in self.plasma_plot_config:
+                        self.override_plot_parameters(
+                            self.plasma_plot, self.plasma_plot_config
+                        )
+
+        self.current_iteration += 1
+        if export_cplots:
+            with suppress(TraitError):
```
jaladh-singhal, line 348: Good that you supressed only TraitError. Did you figure out that why this error arises (becuase supressing should be our last resort)?

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+                self.luminosity_plot.data[
+                    -1
+                ].hovertemplate = "Inner Body Temperature: %{y:.2f} at X = %{x:,.0f}<extra></extra>"
+
+    def update(self, export_cplots=False, last=False):
+        """
+        Calls functions used to build and update convergence plots.
+        """
+        if self.iterable_data != {}:
+            if self.current_iteration == 1:
```
jaladh-singhal, line 334: `and` both conditions

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,360 @@
...
+
+                self.luminosity_plot.data[-1].x = x
+                self.luminosity_plot.data[-1].y = self.value_data["t_inner"]
+                self.luminosity_plot.data[
+                    -1
+                ].hovertemplate = "Inner Body Temperature: %{y:.2f} at X = %{x:,.0f}<extra></extra>"
+
+    def update(self, export_cplots=False, last=False):
+        """
+        Calls functions used to build and update convergence plots.
```
jaladh-singhal, line 331: Again same thing: something like "Build and update convergence plots for each iteration"

tardis/visualization/widgets/shell_info.py
```diff
@@ -4,7 +4,7 @@
     atomic_number2element_symbol,
     species_tuple_to_string,
 )
-from tardis.simulation import Simulation
+
```
jaladh-singhal, line 7: Was Simulation not being used in the module?

tardis/visualization/tools/tests/test_convergence_plot.py
```diff
@@ -0,0 +1,43 @@
...
+        name="iterable", value=range(3), item_type="iterable"
+    )
+    convergence_class.fetch_data(name="value", value=0, item_type="value")
+    assert len(convergence_class.iterable_data["iterable"]) == 3
+
+
+def test_build(convergence_class):
+    convergence_class.build(display_plot=False)
+    assert type(convergence_class.plasma_plot) == go.FigureWidget
+    assert type(convergence_class.luminosity_plot) == go.FigureWidget
```
jaladh-singhal, line 43: This is a good start for tests - you might also want to compare values besides types?

jaladh-singhal (changes requested):
This is an excellent work @atharva-2001 - good code structuring.

There are few comments and questions I've left. The main thing is to write better docstrings - descriptions of several are not very helpful and in some, you didn't documented parameters. Also I think there's need of adding code comments (which describe what a particular part of your code is doing) - they're almost negligible. Please give time to read SDEC plot and line info widget modules, you'll understand both things!

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,389 @@
...
+        The default layout can be overridden by passing plasma_plot_config dictionary in the run_tardis function.
+        """
+        fig = go.FigureWidget().set_subplots(rows=1, cols=2, shared_xaxes=True)
+        fig.add_scatter(row=1, col=1)
+        fig.add_scatter(row=1, col=2)
+
+        fig = fig.update_layout(
+            xaxis={
+                "tickformat": "g",
+                "title": r"$Shell~~Velocity$",
```
MarkMageeAstro, line 106: Change "Shell Velocity" to just "Velocity" and add the units

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,389 @@
...
+            },
+            xaxis2={
+                "tickformat": "g",
+                "title": r"$Shell~~Velocity$",
```
MarkMageeAstro, line 110: Same here

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,389 @@
...
+        fig.add_scatter(
+            name="Residual<br>Luminosity",
+            row=3,
+            col=1,
+            marker_color="cornflowerblue",
+            mode="lines",
+        )
+
+        fig.add_scatter(
+            name="Inner<br>Boundary<br>Temperature",
```
MarkMageeAstro, line 169: Can you put the temperature legend first, since that is the first plot

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,389 @@
...
+                "tickformat": "g",
+                "title": r"$T_{rad}\ [K]$",
+                "tickmode": "auto",
+                "nticks": 15,
+            },
+            yaxis2={
+                "tickformat": "g",
+                "title": r"$W$",
+                "tickmode": "auto",
+                "nticks": 15,
```
MarkMageeAstro, line 123: Is it possible to add some of these things, like nticks, height, etc., as kwargs? Or some method of having this be user configurable

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,389 @@
...
+                    [self.plasma_plot, self.luminosity_plot],
+                )
+            )
+
+    def update_plasma_plots(self):
+        """
+        Updates plasma convergence plots every iteration.
+        """
+        x = self.iterable_data["velocity"].value.tolist()
+        x = [item / 100000 for item in x]
```
MarkMageeAstro, line 275: Should these be a unit conversion instead?

MarkMageeAstro (changes requested):
Another nice PR, @atharva-2001. I've left a few specific comments. Another general request from me would be to add some more code comments. There are some docstrings, but having comments alongside some lines of code would be very useful too.

tardis/visualization/tools/convergence_plot.py
atharva-2001: I'm sorry, my bad, the number of iterations always come from the config file, I just forgot to remove `iterations=20`

tardis/visualization/tools/convergence_plot.py
atharva-2001: image
This is what I have done. Is this what you had in mind?

MarkMageeAstro: Yes, this is much better

tardis/visualization/tools/convergence_plot.py
atharva-2001: The TraitError arises because it was not accepting None as a return value, but `fig.show()` returns None.
```
TraitError: The 'children' trait of a VBox instance contains an Instance of a TypedTuple which expected a Widget, not the NoneType None.
```
No error arises when displaying the plots naturally:
```py 
display(widgets.VBox([go.FigureWidget(), go.FigureWidget()])) # No error
```
But only when one uses `fig.show()` to display the plots (we are doing this only when we are exporting plots)
```py
display(widgets.VBox([go.FigureWidget().show(), go.FigureWidget().show()])) # TraitError
```

tardis/visualization/tools/convergence_plot.py
jamesgillanders: are these colours colour-blind friendly? the defaults should be suitable for all users. I think red/green is a no-go, and here we have orangered/green?

tardis/visualization/tools/convergence_plot.py
atharva-2001: I'm afraid the second `if` statement builds the empty plots and is only executed if it's the first iteration. Using `and` won't allow building the plots only once.

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,388 @@
...
+                "tickformat": "g",
+                "title": r"$\mbox{W}$",
+                "tickmode": "auto",
+                "nticks": 15,
+            },
+            height=450,
+            legend_title_text="Iterations",
+            margin=dict(l=10, r=135, b=25, t=25, pad=0),
+        )
+
```
jamesgillanders, line 130: Label units should not be in italics. Also, the large capitals in T_rad and W should be italicised, but the subscripts should not be (same logic applies to all parameters, e.g. T_inner later).

tardis/visualization/tools/convergence_plot.py
jamesgillanders: ^

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,388 @@
...
+                nticks=4,
+            ),
+            legend_tracegroupgap=0,
+            height=630,
+            hoverlabel_align="right",
+            legend_title_text="Luminosity",
+            hoverlabel_font_color="white",
+            margin=dict(b=25, t=25, pad=0),
+        )
+
```
jamesgillanders, line 224: again, remove italicisation of units here

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,388 @@
...
+            y=self.iterable_data["w"],
+            line_color=self.colorscale[self.current_iteration - 1],
+            row=1,
+            col=2,
+            legendgroup=f"group-{self.current_iteration}",
+            name=self.current_iteration,
+            customdata=customdata,
+            hovertemplate="<b>Y</b>: %{y:.2f} at <b>X</b> = %{x:,.0f}%{customdata}",
+        )
+
```
jamesgillanders, line 314: Is this where the values for Luminosity are generated for the hover box? If so, can you increase the number of significant figures output by default? 1 decimal place truncates a lot of useful info; e.g. see screenshot here. I would suggest 3 decimal places


Screenshot 2021-06-29 at 15 34 11

jamesgillanders: this applies to the other plots for T_rad too

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,388 @@
...
+            + "Emitted Luminosity: "
+            + f'{self.value_data["Absorbed"][-1]:.2g}'
+            + "<br>"
+            + "Requested Luminosity: "
+            + f'{self.value_data["Requested"][-1]:.2g}'
+            + "<br>"
+            + "Absorbed Luminosity: "
+            + f'{self.value_data["Requested"][-1]:.2g}'
+        ]
+
```
jamesgillanders, line 287: Not too sure where the legend stuff is defined, but why is there a Luminosity header for all the lines? This doesn't work if the first line in there is for temperature! Either find a way to separate the temp. line out and then have the Luminosity header for the rest of the lines (and if you do this, remove the luminosity second words from all those labels - e.g. 'emitted luminosity' would become 'emitted') or just remove 
the header. I recall seeing you were finding it difficult to produce 2 separate legends, so perhaps just remove the header.

Screenshot 2021-06-29 at 15 40 58

jamesgillanders (commented):
Code and plots looks good @atharva-2001 - good job! Most of the major things have already been pointed out by the other reviewers, so here are my minor nitpicky things. If/when you get these resolved, I think we'd bee ready to think about merging!

marxwillia (changes requested):

tardis/visualization/tools/convergence_plot.py
atharva-2001: I'm afraid the colors are not color blind friendly. I will change that soon. These colors can be changed by passing a list of them or the name of the color in the `luminosity_colorscale` option. 
Something like this:
```py
sim = run_tardis(
    "tardis_example.yml",
    luminosity_colorscale = "viridis", 
)
```
Or this:
```py
sim = run_tardis(
    "tardis_example.yml",
    luminosity_colorscale = [
        'rgb(229, 134, 6)',
        'rgb(93, 105, 177)',
        'rgb(82, 188, 163)',
        'rgb(153, 201, 69)',
        'rgb(204, 97, 176)',
    ],
)
```

tardis/visualization/tools/convergence_plot.py
atharva-2001: The `height` and many other layout properties can be changed using the `plasma_plot_config` and the `luminosity_plot_config` options, but I'm afraid I didn't get the desired results using that method only for `nticks`. Apart from `nticks`, I believe most of the properties can be overridden. Setting `nticks` to say 10 will sometimes not create 10 `nticks`, but it does increase them, arbitrarily. I am working on it and will create an issue in plotly if I am able to recreate it.

tardis/visualization/tools/convergence_plot.py
atharva-2001: done in https://github.com/tardis-sn/tardis/commit/b652101a3ddb66107fe4b73019f63f7834a2a4d7

tardis/visualization/tools/convergence_plot.py
atharva-2001: Functionality added in https://github.com/tardis-sn/tardis/commit/38bcb465adf12322816f3194badb0fd3ca525a11, added documentation in https://github.com/tardis-sn/tardis/commit/d26efd54ea35a0cc121ab96612e0df1d8bd8f94e

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,388 @@
...
+    Class to create and update convergence plots for visualizing convergence of the
+    simulation.
+
+    Parameters
+    ----------
+    iteration : int
+        iteration number
+    """
+
+    def __init__(self, iterations, **kwargs):
```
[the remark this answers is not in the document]
atharva-2001: done in https://github.com/tardis-sn/tardis/commit/d26efd54ea35a0cc121ab96612e0df1d8bd8f94e

tardis/visualization/widgets/shell_info.py
atharva-2001: I'm afraid not

MarkMageeAstro:
I think this may need rebased because the analysis/opacities script is showing up as a new file added here.

jaladh-singhal (changes requested):
This looks a lot better than before - tests are good. You've fixed docstrings but there are still some problems. And I can see you've added code comments but they're still missing from some modules.

jaladh-singhal (commented):
Excellent work @atharva-2001 - most of the docstrings look very good now! Thanks for the changes.

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,439 @@
...
+            with suppress(TraitError):
+                display(
+                    widgets.VBox(
+                        [
+                            self.plasma_plot.show(
+                                renderer="notebook_connected"
+                            ),
+                            self.t_inner_luminosities_plot.show(
+                                renderer="notebook_connected"
+                            ),
```
jaladh-singhal, line 436: Is it possible to not call show instead just return FigureWidget - I think then display will just show it and you won't need to supress trait error?

docs/io/visualization/convergence_plot.ipynb
```diff
@@ -0,0 +1,846 @@
+{
```
jaladh-singhal, line 1: <p>Highlight the fact that they are stored in <code>cplots</code>  attribute of simulation object <code>sim</code> . And also phrase it out that we have two plots - plasma and t_inner &amp; luminosites within it</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/1636/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

jaladh-singhal, line 1: <p>Is changing the layout here necessary? I think this cell should only focus on show them up since you're already describing how to change layout below</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/1636/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

jaladh-singhal, line 1: <p>This is excellent, as well as description of changing colors section</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/1636/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,439 @@
...
+    **kwargs : dict, optional
+        Additional keyword arguments. These arguments are defined in the Other Parameters section.
+
+    Other Parameters
+    ----------------
+    plasma_plot_config : dict, optional
+        Dictionary used to override default plot properties of plasma plots.
+    t_inner_luminosities_config : dict, optional
+        Dictionary used to override default plot properties of the inner boundary temperature and luminosity plots.
+    plasma_cmap : str, default: 'jet', optional
```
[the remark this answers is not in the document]
atharva-2001: I'm afraid I think what I did on L23 was incorrect? Please correct me if I am wrong. The [examples](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) in the `NumPy` style guide have mentioned default values next to types.  [This](https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-3-parameters) example also has default values next to types.

tardis/visualization/tools/convergence_plot.py
atharva-2001: Actually, this exact thing happens in the `build` function. The plots show up locally without having to suppress any errors, but in order to display them on platforms like nbviewer, we need to have the [`notebook_connected`](https://plotly.com/python/renderers/) renderer. Unfortunately `fig.show()` returns None which causes the error we need to suppress. 

If it would have worked, we would have not needed the `export_cplots` option but would have seen the plots two times online as we see now locally, if we had run the simulation with `export_cplots=True`. One due to the build function and the other due to the `notebook_connected` renderer. The build function fails to display the plots online actually.

docs/io/visualization/convergence_plot.ipynb
atharva-2001: Thank you!!

tardis/visualization/tools/convergence_plot.py
jaladh-singhal: OK - thanks for explanation!

tardis/visualization/tools/convergence_plot.py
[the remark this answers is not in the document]
jaladh-singhal: OK but do that everywhere in your module. Each optional parameter must contain `default: xyz` in your parameter type and description should explain what that does.

tardis/visualization/tools/convergence_plot.py
```diff
@@ -0,0 +1,440 @@
...
+                    fig[key] = value
+
+    def build(self, display_plot=True):
+        """
+        Create empty convergence plots and display them.
+
+        Parameters
+        ----------
+        display_plot : bool, optional
+            Displays empty plots. Defaults to True.
```
jaladh-singhal, line 297: `default: True` should go in type instead of description - like you did everwhere else

docs/io/visualization/convergence_plot.ipynb
jaladh-singhal: @atharva-2001 I believe only this is left?

atharva-2001: I guess I fixed this in https://github.com/tardis-sn/tardis/pull/1636/commits/4401372fd30d2c9846db172d9f546e200e26eca0
Could you please check if it's up to the mark? 
[documentation](https://atharva-2001.github.io/tardis/branch/convergence_plots/io/visualization/convergence_plot)

jamesgillanders (approved):
This all looks great @atharva-2001! All requested changes seem to have been addressed, and the code itself works really well. Good job!

marxwillia (approved):

jaladh-singhal (approved):
Brilliant work! Thank you for addressing all of my nitpicky comments :smile:

jaladh-singhal:
@atharva-2001 this needs rebasing now that #1723 is merged

PR #1645, opened 2021-06-09
configure codecov

epassaro:
**Description**
- Minimum `codecov` configuration.
- Rewrote `.coveragerc` based on Coverage sample configuration file.

I found tests from `visualization` and `montecarlo_numba` were taken into account to measure coverage, that's not right!. Tests always are 100% covered (all tests lines are executed during tests, lol) so now we have a lower but more realistic coverage. This was discovered by using better wildcards. 

Summarizing, `coverage` moved from 73 to 60 files thanks to modifying `.coveragerc`:

- 1 `conftest.py` at the root of the package
- 1 `cmfgen2tardis.py` script
- 2 tests from `visualization`
- 8 tests from `montecarlo_numba` + 1 `conftests.py`

**Motivation and context**
Codecov is failing too often even while PRs are not changing source files.

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above. 

**Checklist**

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -19,6 +19,7 @@ exclude_lines =
 omit =
     *__init__*
     *_astropy_init*
+    *version*
     *tests*
     *conftest*
     *setup_package*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,5 @@
+# .coveragerc to control coverage.py
+
 [report]
 # Regexes for lines to exclude from consideration
 exclude_lines =
--- a/.coveragerc
+++ b/.coveragerc
@@ -22,10 +22,11 @@ omit =
     *tests*
     *conftest*
     *setup_package*
-   
-    tardis/gui/*
-    tardis/stats/*
+
     tardis/analysis.py
+    tardis/gui/*
+    tardis/scripts/*
+    tardis/stats/*
    
 ignore_errors = True
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,48 +2,36 @@
 source = tardis
 
 [report]
+# Regexes for lines to exclude from consideration
 exclude_lines =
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -1,49 +1,35 @@
...
 omit =
+    *__init__*
+    *_astropy_init*
+    *version*
+    *tests*
+    *conftest*
+    *setup_package*
+
+    tardis/analysis.py
+    tardis/gui/*
```
andrewfullard, line 28: I assume this is the deprecated GUI

andrewfullard (approved):
Ironic that this fails codecov itself

.coveragerc
epassaro: Yes it is. I don't know who made the decision of leaving the GUI outside the coverage, because it has some tests. But I think visualization team was planning to deprecate it soon.

epassaro: Also, Is a good decision to leave out the `scripts` folder? IIRC `cmfgen2tardis.py` wasn't working correctly.

DhruvSondhi (approved):

.coveragerc
andrewfullard: Perhaps not, even if something is not working it should probably be checked for test coverage!

epassaro: Ok, I removed it. Do you know why analysis and stats are out of coverage?

andrewfullard: analysis should definitely be in coverage, stats seems to be a deprecated or otherwise unfinished part

epassaro: Ok, solved!

PR #1632, opened 2021-06-07
Formatting Logging Output for Simulation

DhruvSondhi:
This PR aims to change the Formatting of the current Logging done in the Notebook.
Implements decimal up to 3 places in Plasma Stratification values
Better formatted output for `luminosities` values as well as `t_inner` values 

**Description**

Some formatting changes have been done to the logging output for the simulation. Screenshots below differentiate the output, before & after the changes. 

**Motivation and context**

Allows for a better, more digestible formatting for the logging output. Data is more readable & accessible to the User.

**Screenshots**

*Original Logger Output:*
image

*New Formatted Logger Output:*
image
*Jupyter Notebook*

image
image
*Terminal*

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [ ] My change requires a change to the documentation.
    - [ ] I have updated the documentation accordingly.
    - [ ] (optional) I have built the documentation on my fork following [the instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html).
- [x] I have assigned and requested two reviewers for this pull request.

PR #1650, opened 2021-06-14
Adding functionality to Detect Running Environments for TARDIS Simulations

DhruvSondhi:
This PR adds the functionality to check the environment in which the TARDIS Simulation is run. 
Needs to be merged before the #1632 

**Description**

This PR aims to add a new function `check_simulation_env()` which checks if the simulation is being run in a IPython environment or any other environment such as a Terminal, etc. 
This function returns `True` in the former option & `False` in the latter. 

**Motivation and context**

This function is vital for the implementation of the Logging Formatting which is being done in #1632. It would allow us to control the output of the logging & would allow for better formatting for the logging output generated while running the simulation.

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other.  Local Testing

**Examples**

Check PR #1632 

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -13,6 +13,7 @@ from pyne import nucname
 
 import tardis
 from tardis.io.util import get_internal_data_path
+from IPython import get_ipython
 
 k_B_cgs = constants.k_B.cgs.value
 c_cgs = constants.c.cgs.value
@@ -539,3 +540,25 @@ def convert_abundances_format(fname, delimiter=r"\s+"):
     # Assign header row
     df.columns = [nucname.name(i) for i in range(1, df.shape[1] + 1)]
     return df
+
+
+def check_simulation_env():
+    """
+    Checking the environment where the simulation is run
+
+    Returns
+    -------
+    True : if the environment is IPython Based
+    False : if the environment is Terminal or anything else
+    """
+    try:
+        shell = get_ipython().__class__.__name__
+    except NameError:
+        return False
+
+    if shell == "ZMQInteractiveShell":
+        return True
+    elif shell == "TerminalInteractiveShell":
+        return False
+    else:
+        return False


tardis/util/base.py
```diff
@@ -539,3 +540,24 @@ def convert_abundances_format(fname, delimiter=r"\s+"):
...
+    False : if the environment is Terminal or anything else
+    """
+    try:
+        shell = get_ipython().__class__.__name__
+        if shell == "ZMQInteractiveShell":
+            return True
+        elif shell == "TerminalInteractiveShell":
+            return False
+        else:
+            return False
```
andrewfullard, line 561: Only the command that throws the exception should be in the `try` statement

andrewfullard (approved):

tardis/util/base.py
```diff
@@ -539,3 +540,25 @@ def convert_abundances_format(fname, delimiter=r"\s+"):
...
+    try:
+        shell = get_ipython().__class__.__name__
+    except NameError:
+        return False
+
+    if shell == "ZMQInteractiveShell":
+        return True
+    elif shell == "TerminalInteractiveShell":
+        return False
+    else:
```
wkerzendorf, line 563: when does this else statement happen?

DhruvSondhi: The following needs to be considered for the conditional check:
- If we are in IPython Kernel environment, ie Jupyter Notebook, Jupyter Lab, etc, then we are in the `ZMQInteractiveShell`. Please check [this](https://ipython.readthedocs.io/en/stable/config/options/kernel.html).
- If we are in the Terminal IPython environment, ie IPython Kernel running in Terminal, then we are in the `TerminalInteractiveShell`. Please check [this](https://ipython.readthedocs.io/en/stable/config/options/terminal.html).
- If we are in any other environment, ie Google Colab or Normal Shell based Terminal, etc, then we are in the else condition. Hence, the main `else` condition allows us to print the values as a str in #1632

DhruvSondhi: image
*This is Jupyter Notebook*

image
*This is IPython in terminal*

image
*This is normal Python from command line

DhruvSondhi:
I think the `Black` tests are failing because of the comments present in the code 🤔

Rodot-:
I would recommend instead of getting the class name from the dunder methods (not really part of their API), you should compare the shell directly using `isinstance` built-in function and comparing to ZMQInteractiveShell and InteractiveShell directly, which can be found through:

```from IPython.kernel.zmq.zmqshell import ZMQInteractiveShell```
and 

```from IPython.core.interactiveshell import InteractiveShell``

Rodot-:
Additionally, I do feel like the name `check_simulation_environment` is a little ambiguous given that it returns a bool rather than the simulation environment.  Consider renaming it to something like `is_ZMQInteractiveShell` since that is the only condition for which it returns `True`.  Or something like `is_notebook` if the intent is to check if the environment is a notebook (though, I don't know if ZMQInteractiveShell is always the interactive shell for notebooks) 

 Have you tested this in something like jupyter-lab or vscode?

DhruvSondhi:
Yes, Jupyter based Environments have the class name is always **ZMQInteractiveShell**. It is purely based on the IPython kernel being run or not. As for the screenshots above, you can see the different types of output that are returned when we run the `get_ipython().__class__.__name__`.

DhruvSondhi:
>Additionally, I do feel like the name check_simulation_environment is a little ambiguous given that it returns a bool rather than the simulation environment. Consider renaming it to something like is_ZMQInteractiveShell since that is the only condition for which it returns True. Or something like is_notebook if the intent is to check if the environment is a notebook (though, I don't know if ZMQInteractiveShell is always the interactive shell for notebooks

Absolutely, I would change it in the logging formatting PR and hence we can make it little more clearer 😄

DhruvSondhi:
> `from IPython.kernel.zmq.zmqshell import ZMQInteractiveShell`

Just a small update on this import. This has been depricated in IPython 4 version. Newer import is the following:
```from ipykernel.zmqshell import ZMQInteractiveShell```

PR #1506, opened 2021-03-22
Add ability to plot only top contributing elements to SDEC plot

MarkMageeAstro:
Port of the nelements functionality that is already present in the tardisanalysis repo. 

## Description
Add a new argument 'nelements'. The total contribution of each element to the absorbed and emitted luminosities is ranked and the top 'nelements' are selected. These nelements are shown with unique colours. All other elements are shown in silver.

## Motivation and Context
Updates the visualization SDEC plot to bring back functionality that was lost in the port.

## How Has This Been Tested?
- [ ] Testing pipeline
- [ ] Reference Data Comparison following these [instructions](https://tardis-sn.github.io/tardis/development/continuous_integration.html)
- [X] Other (please describe)
Comparisons with older versions of code.

## Screenshots (if appropriate):
Using the following command:
plotter.generate_plot_mpl(packet_wvl_range=[3000, 9000] * u.AA, nelements=2);
Generates this plot:
Screenshot 2021-03-22 at 23 06 37

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] None of the above (please describe)

## Checklist:
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have built the documentation on my fork following [these instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html)
- [X] I have assigned and requested two reviewers for this pull request

PR #1665, opened 2021-06-24
Demonstrating the `nelements` and the `species_list` functionality in the SDEC plot notebook

atharva-2001:
This PR adds documentation for `nelements` and `species_list` options in the SDEC plot notebook. 

**Description**

The `nelements` functionality was added in PR https://github.com/tardis-sn/tardis/pull/1506
The `species_list` functionality was added in PR https://github.com/tardis-sn/tardis/pull/1539
This PR demonstrates these two functionalities in the notebook. 

**Motivation and context**

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

Here is the [link](https://atharva-2001.github.io/tardis/branch/sdec_docs/io/visualization/sdec_plot.html) to the notebook in my documentation
[Link](https://nbviewer.jupyter.org/github/atharva-2001/tardis/blob/sdec_docs/docs/io/visualization/sdec_plot.ipynb) to the notebook on nbviewer

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [X] My change requires a change to the documentation.
    - [X] I have updated the documentation accordingly.
    - [X] (optional) I have built the documentation on my fork following [the instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html).


jaladh-singhal (changes requested):
Thanks for making this PR - it was long due! 

Few suggestions:
- Add these two sections also on matplotlib section (besides plotly) because that is shown first and is preferred in research papers
- Why have you used real packets_mode? SDEC plot looks better with virtual packets mode as you might have noticed (due to noise reduction) and it's also default, please change that. I guess it's possibly because you're using `tardis_example.yml` that have virt packet logging disabled but SDEC plot and line info widget expects opposite - create an issue for that and work on it if you like :)

jaladh-singhal:
Also please check what happens when both nelments and species_list are specified - AFAIR one takes precedence over other and it should be documented as well

MarkMageeAstro (changes requested):
Good job, @atharva-2001. My only comment is that I think for the nelements/ion plots we should set the wavelength range. It'll make it easier to see the different colours.

jaladh-singhal (commented):
Thanks for the changes. One minor but thing: it'll be better if you put 3 cells you added in matpltolib section, before "plotting flux instead of luminosity" since it and hiding spectrum are not as important as nelments and species_list is

Also I see you're using virtual packets now - that's good but don't forget creating issue for this:
> I guess it's possibly because you're using tardis_example.yml that have virt packet logging disabled but SDEC plot and line info widget expects opposite - create an issue for that and work on it if you like :)

MarkMageeAstro (changes requested):
I can't leave comments on the notebook, so I'm just putting them all here. A few small grammar changes.

1. There are a few instances throughout where the plot is just referred to as 'SDEC plot', but there should be either 'an' or 'the' in front of it. These are two examples:
- "Now, import the plotting interface for SDEC plot..." should be "Now, import the plotting interface for the SDEC plot..."
- "By default, SDEC plot is produced..." should be "By default, an SDEC plot is produced..."
2. "Since packets_mode is 1st argument..." -> "Since packets_mode is the first argument..."
3. "By default, modeled spectrum is shown in SDEC plot." -> "By default, the model spectrum is shown in SDEC plot."
4. "This method takes exact same arguments" -> "This method takes the exact same arguments"
5. "In similar manner" -> "In a similar manner"
6. " thereby providing you more" -> " thereby providing you with more"

jaladh-singhal (approved):
Thanks for all the changes. Looks good to me!

MarkMageeAstro (approved):

PR #1685, opened 2021-07-02
Implemented the Simulation Logger as a Class

DhruvSondhi:
Implemented the Simulation Logger as a Class instead of the existing functional implementation

**Description**

**Motivation and context**

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above.  Restructured for clearer implementation

**Checklist**

### DIFF
--- a/tardis/__init__.py
+++ b/tardis/__init__.py
@@ -133,7 +133,7 @@ class SimulationLogger:
                 logger.removeFilter(filter)
 
     @classmethod
-    def logging_state(self, tardis_config, log_state=None, specific=None):
+    def logging_state(self, tardis_config, log_state, specific):
         """
         Function to set the logging configuration for the simulation output
         Called from within run_tardis()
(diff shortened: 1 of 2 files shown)


epassaro:
Sorry about not knowing too much about this. Why we use `log_state` instead of `log_level`? Also why we use `Info` in title case and not lowercase or uppercase?

DhruvSondhi:
Hey @epassaro, `log_state` was actually the name of the variable given by me in the initial implementation of the logger. That time it used to have 3 states: `TRUE`, `FALSE` & `Log_level` which could be a valid string for the log level. After changing the implementation, the variable name was left as it was. 
For the title case, we actually convert the whole string into Uppercase & hence it is just for a check. Also, where did you find `Info` in the code? I am not able to find it 🤔

epassaro:
Not in the code, in the Quickstart guide: `sim = run_tardis('tardis_example.yml', log_state = "Info")`

DhruvSondhi:
Oh, as explained in the previous message, this was just for testing out that it was working & to show that the `log_state` parameter is independent of the case of the input 😉

PR #1498, opened 2021-03-19
Feature request: add virtual-packet-logging option to run_tardis()

yuyizheng1112:
## Description

This PR is to resolve #1356.
## Motivation and Context

The virtual packet data is not enabled by default. Currently, we have to enable `virtual_packet_logging` by editing the configuration file to use visualization tools that require virtual packet data. Although offering instruction in the documentation can partly solve the problem, it is still inconvenient to edit configuration file every time.

The new feature allows us to enable `virtual_packet_logging` option while running TARDIS simulation by setting the parameter `virtual_packet_logging=True`. The default value of the parameter is `False`. Meanwhile, it still supports specifying the option in configuration files.
## How Has This Been Tested?
- [x] Testing pipeline
- [ ] Reference Data Comparison following these [instructions](https://tardis-sn.github.io/tardis/development/continuous_integration.html)
- [ ] Other (please describe)

## Screenshots (if appropriate):
The output of `virt_packet_last_interaction_type` indicates whether the packet has been enabled.
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] None of the above (please describe)

## Checklist:

- [ ] My code follows the code style of this project
- [x] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have built the documentation on my fork following [these instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html)
- [ ] I have assigned and requested two reviewers for this pull request

PR #1693, opened 2021-07-07
Enabling virtual_packet_logging by default in the visualization folder

atharva-2001:
This PR removes the symbolic link to tardis_example.yml in the visualization folder, and adds it with `virtual_packet_logging` enabled.

**Description**

To build the SDEC plot and the Line Info widget, one needs to enable `virtual_packet_logging` from the `run_tardis` function or edit the tardis_example.yml file in the docs folder. This enables `virtual_packet_logging` by default just for the visualization folder, easing the process for the user.

**Motivation and context**

https://github.com/tardis-sn/tardis/pull/1665#pullrequestreview-697583309

**How has this been tested?**
- [X] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [X] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/docs/io/visualization/generating_widgets.ipynb
+++ b/docs/io/visualization/generating_widgets.ipynb
@@ -391,7 +391,7 @@
    ],
    "source": [
     "from tardis import run_tardis\n",
-    "sim = run_tardis('tardis_example.yml')"
+    "sim = run_tardis('tardis_example.yml', virtual_packet_logging=True)"
    ]
   },
   {
@@ -595,7 +595,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
@@ -609,7 +609,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.8"
+   "version": "3.7.10"
   },
   "notify_time": "5",
   "toc": {


smithis7:
Because of how we do things with symlinks, I think you should either:

- rename that file to something else besides `tardis_example.yml`

- add virtual packet logging to `docs/tardis_example.yml` and keep the file in `docs/io/visualization` as a symlink

MarkMageeAstro:
We can probably change to something like tardis_example_SDEC? I guess this would also mean some notebooks need updated too

MarkMageeAstro (changes requested):
See previous comment...

yuyizheng1112:
just a suggestion: I used to add `virtual_packet_logging` option to `run_tardis()` function (see [this pr](https://github.com/tardis-sn/tardis/pull/1498)). It seems that few people know this... Maybe you can set this option to `True` instead of adding new file.

atharva-2001:
That sounds good to me too, maybe we could just set `virtual_packet_logging` to True in both notebooks in https://github.com/tardis-sn/tardis/pull/1665?

jaladh-singhal (changes requested):
I think instead of adding new file, enabling virtual packet logging in run_tardis call will be better as suggested by @yuyizheng1112 .

And if you want to keep the new config file, name it as `tardis_example_virt_logging.yml` instead since it's not only for SDEC but also for Line info widget.

andrewfullard:
Is this PR still relevant?

jaladh-singhal (approved):

jamesgillanders (approved):

PR #1701, opened 2021-07-08
Adding Logging Support for Caught Exceptions

DhruvSondhi:
This PR aims to add logging support to caught exceptions.

**Description**

There are numerous exceptions that are caught but are not visible to the user. This PR aims to add support via logging for these exceptions such that it is visible when logging is turned on. Logs at `DEBUG` level 😄 

**Motivation and context**

Allows for better explanation for silent errors or exceptions that may happen. Logging them will benefit & make it more descriptive for the user to fix and see what is going on in the simulation. 

**How has this been tested?**
- [x] Testing pipeline.
- [x] Other.  Local Tests

**Type of change**

- [x] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/atom_data/base.py
+++ b/tardis/io/atom_data/base.py
@@ -164,7 +164,7 @@ class AtomData(object):
                 try:
                     dataframes[name] = store[name]
                 except KeyError:
-                    logger.debug("\n\tDataframe does not contain NAME column")
+                    logger.debug("Dataframe does not contain NAME column")
                     nonavailable.append(name)
 
             atom_data = cls(**dataframes)
@@ -173,7 +173,7 @@ class AtomData(object):
                 atom_data.uuid1 = store.root._v_attrs["uuid1"].decode("ascii")
             except KeyError:
                 logger.debug(
-                    "\n\tUUID not available for Atom Data\n\tSetting value to None"
+                    "UUID not available for Atom Data. Setting value to None"
                 )
                 atom_data.uuid1 = None
 
@@ -181,7 +181,7 @@ class AtomData(object):
                 atom_data.md5 = store.root._v_attrs["md5"].decode("ascii")
             except KeyError:
                 logger.debug(
-                    "\n\tMD5 not available for Atom Data\n\tSetting value to None"
+                    "MD5 not available for Atom Data. Setting value to None"
                 )
                 atom_data.md5 = None
 
@@ -189,18 +189,20 @@ class AtomData(object):
                 atom_data.version = store.root._v_attrs["database_version"]
             except KeyError:
                 logger.debug(
-                    "\n\tVERSION not available for Atom Data\n\tSetting value to None"
+                    "VERSION not available for Atom Data. Setting value to None"
                 )
                 atom_data.version = None
 
             # ToDo: strore data sources as attributes in carsus
 
             logger.info(
-                f"\n\tReading Atom Data with:\n\tUUID = {atom_data.uuid1}\n\tMD5  = {atom_data.md5} "
+                f"Reading Atom Data with: UUID = {atom_data.uuid1} MD5  = {atom_data.md5} "
             )
             if nonavailable:
                 logger.info(
-                    f'\n\tNon provided Atomic Data:\n\t{", ".join(nonavailable)}'
+                    "Non provided Atomic Data: {0}".format(
+                        ", ".join(nonavailable)
+                    )
                 )
 
         return atom_data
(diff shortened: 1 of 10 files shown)


tardis/analysis.py
```diff
@@ -2,12 +2,15 @@
 
 import re
 import os
+import logging
```
andrewfullard, line 5: Why are there changes in this file? They don't do anything.

tardis/analysis/opacities.py
```diff
@@ -329,7 +329,7 @@ def _calc_thomson_scattering_opacity(self):
         """Calculate the Thomson scattering opacity for each grid cell
 
         \[
-          \chi_{\mathrm{Thomson}} = n_{\mathrm{e}} \sigma_{\mathrm{T}}
+        \chi_{\mathrm{Thomson}} = n_{\mathrm{e}} \sigma_{\mathrm{T}}
```
andrewfullard, line 332: Why is this changed?

tardis/io/atom_data/base.py
```diff
@@ -164,23 +164,33 @@ def from_hdf(cls, fname=None):
...
+                    logger.debug("\n\tDataframe does not contain NAME column")
                     nonavailable.append(name)
 
             atom_data = cls(**dataframes)
 
             try:
                 atom_data.uuid1 = store.root._v_attrs["uuid1"].decode("ascii")
             except KeyError:
+                logger.debug(
+                    "\n\tUUID not available for Atom Data\n\tSetting value to NONE"
```
andrewfullard, line 176: I think NONE should be written as None to make it clear that it is the Python object None.

tardis/io/parsers/csvy.py
```diff
@@ -37,6 +40,7 @@ def load_csvy(fname):
         try:
             data = pd.read_csv(fname, skiprows=yaml_end_ind + 1)
         except pd.errors.EmptyDataError as e:
+            logger.debug("\n\tCould not Read CSV.\n\tSetting Data to NONE")
```
andrewfullard, line 43: See previous comment. Perhaps specify which "data" this is?

tardis/montecarlo/base.py
```diff
@@ -335,7 +335,7 @@ def virtual_packet_nu(self):
         try:
             return u.Quantity(self.virt_packet_nus, u.Hz)
         except AttributeError:
-            warnings.warn(
+            logger.debug(
```
andrewfullard, line 338: I think this should be a warning

tardis/montecarlo/base.py
```diff
@@ -349,7 +349,7 @@ def virtual_packet_energy(self):
         try:
             return u.Quantity(self.virt_packet_energies, u.erg)
         except AttributeError:
-            warnings.warn(
+            logger.debug(
```
andrewfullard, line 352: Same here

tardis/montecarlo/base.py
```diff
@@ -363,7 +363,7 @@ def virtual_packet_luminosity(self):
         try:
             return self.virtual_packet_energy / self.time_of_simulation
         except TypeError:
-            warnings.warn(
+            logger.debug(
```
andrewfullard, line 366: Same here

tardis/analysis/opacities.py
DhruvSondhi: Just for formatting consistency with Docstrings 😄

tardis/analysis.py
DhruvSondhi: I wanted to ask about this. Do we need to log exceptions for this part of the source code 🤔

wkerzendorf:
I think that is a nice start to things. First of all - in general this PR seems to contain a variety of different fixes and a lot of files. In general, fewer PRs targeted to specific issues is better. 
My main other question is there are `\n\t` in front of essentially all debug statements. why?

DhruvSondhi:
> My main other question is there are `\n\t` in front of essentially all debug statements. why?

Hello @wkerzendorf, This is actually due to how the formatting for the logger has been done. Continuing from #1632, I have added a new line & some tab spaces between the messages such that it is easier to read. If we remove this (`\n\t`) then the message is on the same line, without any space, making it difficult to read (what is the problem or where is the problem because everything is squashed in a single line with line wrapping, etc problems). In the end, it is just for formatting purposes. 
If you think that the formatting is unnecessary, I can revert them as they used to be (formatting wise) :smile:

andrewfullard:
Now that you have a fix for the `/n/t`, you can implement it here

DhruvSondhi:
I have implemented it in PR #1704. I think we can merge this as it is here & then I can resolve this in that PR 😄

wkerzendorf (approved):

andrewfullard (approved):

PR #1703, opened 2021-07-09
Adding Progress Bars

atharva-2001:
This PR adds progress bars to TARDIS.

**Description**

**Motivation and context**

**How has this been tested?**
- [ ] Other. 
- [X] Testing pipeline.

**Examples**

Please visit the documentation here: 
https://atharva-2001.github.io/tardis/branch/progress_bar/io/output/index.html
https://atharva-2001.github.io/tardis/branch/progress_bar/quickstart/quickstart.html

Peek 2021-08-03 23-38
Peek 2021-08-05 18-43

**Type of change**

- [ ] Bug fix. 
- [X] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [X] I have assigned and requested two reviewers for this pull request.


MarkMageeAstro (changes requested):
The progress bars look great, but I personally think this should not be a separate PR. To me, this is still part of the convergence plots because it's showing the progress of the model. I would like to see the layout of the progress bars and convergence plots together.

What is the impact on performance using these bars, if any?

Minor changes:
1. The packets bar should read "XX% of packets propagated". Perhaps even "XX% of real packets propagated".
2. There are hardcoded values for iterations and packet numbers, but these should come from the input yaml file. You can get the total packet number from (iterations - 1) * no_of_packets + last_no_of_packets

atharva-2001:
Thank you @MarkMageeAstro!
I tried 5 iterations in each case and the simulation took me around 30 seconds using progress bars and around 27 seconds without them. I will make the required changes very soon. I too believe seeing the progress bars and the convergence plots together would be better.

MarkMageeAstro:
That's great, I would consider a ~10% increase acceptable.

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -26,6 +26,23 @@
 )
 from tardis.montecarlo.montecarlo_numba import njit_dict
 from numba.typed import List
+import tqdm.notebook as tq
+from tqdm import tqdm
+
+packet_pbar = tq.tqdm(
+    total=860000,
```
marxwillia, line 33: We can't hardcode the number of packets or iterations.  These numbers need to be retrieved from the tardis config object.

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -26,6 +26,23 @@
...
+    desc="Packets",
+    bar_format="{bar}{percentage:3.0f}% packets propagated",
+)
+iteration_pbar = tq.tqdm(
+    total=20,
```
marxwillia, line 38: Retrieve number of iterations from the tardis config object.  This is a user specified parameter.

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -168,6 +186,8 @@ def montecarlo_main_loop(
     virt_packet_last_line_interaction_out_id = []
 
     for i in prange(len(output_nus)):
+        with objmode:
+            update_packet_pbar(1)
```
marxwillia, line 190: Try using `if (i % 1000 == 0)` as a condition.  You might need to also change the argument in `update_packet_pbar` to 1000.  Use the tqdm documentation here and click the links to the source code for the update function:

https://tqdm.github.io/docs/tqdm/

marxwillia (changes requested):

jaladh-singhal (commented):
The progress bar looks nice and code quality looks ok to me (considering you haven't added docstrings to existing functions that didn't have them already). Rest about code changes - since it deals with numba I'd request @andrewfullard to have a look.

Have you checked how progress bar looks on static html page (in docs)?

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -26,9 +26,107 @@
...
-def montecarlo_radial1d(model, plasma, runner):
+packet_pbar = pbar(
+    dynamic_ncols=True,
+    bar_format="{bar}{percentage:3.0f}% of packets propagated, iteration 0/?",
+)
+if type(packet_pbar).__name__ == "tqdm_notebook":
+    packet_pbar.container.close()
+
+
+def update_packet_pbar(i, current_iteration, total_iterations, total_packets):
```
andrewfullard, line 46: I'm not sure if this function should be in the base montecarlo numba file.

atharva-2001: Will adding this function to `tardis/util/base.py` be a good idea?

tardis/montecarlo/base.py
```diff
@@ -280,6 +282,7 @@ def run(
         no_of_virtual_packets : int
         nthreads : int
         last_run : bool
+        total_iterations : int
```
marxwillia, line 285: Please add a description sentence for total_iterations.

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -178,6 +196,15 @@ def montecarlo_main_loop(
     virt_packet_last_line_interaction_out_id = []
 
     for i in prange(len(output_nus)):
+        if show_progress_bar:
+            with objmode:
+                update_packet_pbar(
+                    1,
```
marxwillia, line 202: Even if this 1 should be hard coded, it would be better to define it as a variable with a descriptive name so that future developers know what it does.

tardis/util/base.py
```diff
@@ -596,3 +597,124 @@ def is_notebook():
...
+            bar.reset(total=no_of_packets)
+        if total_iterations is not None:
+            bar.reset(total=total_iterations)
+
+        # change the amount of space the prefix string of the bar takes
+        # here, either packets or iterations
+        bar.container.children[0].layout.width = "6%"
+
+        # change the length of the bar
+        bar.container.children[1].layout.width = "60%"
```
marxwillia, line 709: Have you stress tested these settings with very narrow and very wide windows for browser and terminal?

marxwillia (changes requested):
Answer questions in comments.

tardis/util/base.py
atharva-2001: For the notebook, these settings cause the bar to have a fixed length _relative_ to the length of the cell, meaning the actual bar would always be 60% the length of the cell and the string in front of the bar would always be 6% the length of the cell. These work great under normal conditions, but not ideally under _too_ narrow conditions, where the bar overlaps the string before the bar sometimes.

For the terminal, I defined the settings now [here](https://github.com/tardis-sn/tardis/pull/1703/files#diff-389972ec71fe92482e51ea9a059cc62a97a39f6cd7429575c3cbf63cbe3d94b2R618), which is makes sure that the string in the front of the bar is not cut short (and are equal) and the both the bars have only [80 characters](https://github.com/tqdm/tqdm#description-and-additional-stats), making sure they are of the same length. 

I hard coded the values for the terminal and they work pretty well. I could could code it in a way to always find out somehow how big the screen is and increase the length of the bars, but I didn't think that was necessary, as the user could always increase/decrease the size of the terminal text.

wkerzendorf:
this has preliminary approval from me (pending the others approval) - it looks very nice.

jaladh-singhal (changes requested):
Great work @atharva-2001. Everything looks good except mentioning it about in our docs is missing.

AFAIR, you were adding a "Progress bar for Simulation Run" page in "Additional outputs" section explaining options related to progress bar, different modes of display, etc. along with a note in quickstart (?) that progress bar doesn't get displayed in static page but will show up in interactive mode?

andrewfullard (approved):

jaladh-singhal (approved):
Thanks for adding the documentation

jamesgillanders (approved):
Great addition to the code!

PR #1710, opened 2021-07-13
Changed Formatting for the logger based on Log Level

DhruvSondhi:
This PR aims to add Logging Formatting based on log level. Debug Level has different formatting due to the number of messages that would be printed to the `strerr`. All other logging level follow the existing formatting (with the `\n\t` formatting).

**Description**

**Motivation and context**

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Examples**

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/atom_data/base.py
+++ b/tardis/io/atom_data/base.py
@@ -169,7 +169,7 @@ class AtomData(object):
                 try:
                     dataframes[name] = store[name]
                 except KeyError:
-                    logger.debug("Dataframe does not contain NAME column")
+                    logger.debug(f"Dataframe does not contain {name} column")
                     nonavailable.append(name)
 
             atom_data = cls(**dataframes)
--- a/tardis/io/logger/colored_logger.py
+++ b/tardis/io/logger/colored_logger.py
@@ -28,11 +28,16 @@ class ColoredFormatter(logging.Formatter):
     Custom logger class for changing levels color
     """
 
+    non_debug = "[$BOLD{name:20s}$RESET][{levelname:18s}]  \n\t{message:s} ($BOLD{filename:s}$RESET:{lineno:d})"
+    debug = "[$BOLD{name:20s}$RESET][{levelname:18s}]  {message:s} ($BOLD{filename:s}$RESET:{lineno:d})"
+
     def __init__(self, msg, use_color=True):
-        logging.Formatter.__init__(self, msg)
+        logging.Formatter.__init__(self, msg, style="{")
         self.use_color = use_color
 
     def format(self, record):
+        non_debug = formatter_message(ColoredFormatter.non_debug, True)
+        debug = formatter_message(ColoredFormatter.debug, True)
         COLOR_SEQ = "\033[1;%dm"
         RESET_SEQ = "\033[0m"
         BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE = range(8)
@@ -49,6 +54,10 @@ class ColoredFormatter(logging.Formatter):
                 COLOR_SEQ % (30 + COLORS[levelname]) + levelname + RESET_SEQ
             )
             record.levelname = levelname_color
+        if record.levelno == logging.DEBUG:
+            self._style._fmt = debug
+        else:
+            self._style._fmt = non_debug
         return logging.Formatter.format(self, record)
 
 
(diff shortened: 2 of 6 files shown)


tardis/io/logger/logger.py
```diff
@@ -7,12 +7,11 @@
 
 warnings.filterwarnings("ignore", category=pyne.utils.QAWarning)
 
-FORMAT = "[$BOLD%(name)-20s$RESET][%(levelname)-18s]  %(message)s ($BOLD%(filename)s$RESET:%(lineno)d)"
+FORMAT = "[$BOLD{name:20s}$RESET][{levelname:18s}]  \n\t{message:s} ($BOLD{filename:s}$RESET:{lineno:d})"
```
DhruvSondhi, line 10: The format of the logging message has been changed from Python 2 to Python 3 `f-string`.

tardis/io/logger/colored_logger.py
```diff
@@ -28,11 +28,16 @@ class ColoredFormatter(logging.Formatter):
     Custom logger class for changing levels color
     """
 
+    non_debug = "[$BOLD{name:20s}$RESET][{levelname:18s}]  \n\t{message:s} ($BOLD{filename:s}$RESET:{lineno:d})"
+    debug = "[$BOLD{name:20s}$RESET][{levelname:18s}]  {message:s} ($BOLD{filename:s}$RESET:{lineno:d})"
```
DhruvSondhi, line 32: Added the formatting inside the `ColoredFormatter` Class so that it can be accessed for the log messages. For both these messages, the `formatter_message` function is called for adding the relevant colours.

DhruvSondhi (commented):
Explained some of the changes done in this PR.

tardis/io/atom_data/base.py
```diff
@@ -164,7 +164,7 @@ def from_hdf(cls, fname=None):
                 try:
                     dataframes[name] = store[name]
                 except KeyError:
-                    logger.debug("Dataframe does not contain NAME column")
+                    logger.debug(f"Dataframe does not contain {name} column")
```
andrewfullard, line 167: What does this print? Is it a column header or an entire column?

tardis/io/logger/colored_logger.py
andrewfullard: Shouldn't these be `self` variables inside `__init__`?

tardis/io/logger/colored_logger.py
```diff
@@ -57,7 +66,7 @@ class ColoredLogger(logging.Logger):
     Custom logger class with multiple destinations
     """
 
-    FORMAT = "[$BOLD%(name)-20s$RESET][%(levelname)-18s]  %(message)s ($BOLD%(filename)s$RESET:%(lineno)d)"
+    FORMAT = "[$BOLD{name:20s}$RESET][{levelname:18s}]  \n\t{message:s} ($BOLD{filename:s}$RESET:{lineno:d})"
```
andrewfullard, line 69: If there is the debug/non-debug variables, why does this still exist?

tardis/io/logger/logger.py
andrewfullard: See previous comment about the debug/non_debug variables

tardis/io/atom_data/base.py
DhruvSondhi: This is what it used to print
image
And this is what it prints right now (after the change)
image

tardis/io/logger/colored_logger.py
DhruvSondhi: This is just because of the way the `__init__` method inside the `ColoredFormatter` Class works. It takes a message (this is the message) & the `use_color` variable boolean. We would need to refactor the `__init__` function for this to be removed.

tardis/io/atom_data/base.py
andrewfullard: Much better, thanks

tardis/io/logger/colored_logger.py
DhruvSondhi: Yes, this seems to be a better way to proceed. I have restructured the `colored_logger.py` to adhere to this change.

tardis/io/logger/logger.py
DhruvSondhi: Yes, I have made the required changes.

tardis/io/atom_data/base.py
DhruvSondhi, line 167: @andrewfullard @Rodot- This change needs to be merged. What should be done here? The PR will remain Draft.

PR #1712, opened 2021-07-13
Fixed Deprecated Warning with Virtual Packet Logging

DhruvSondhi:
This PR aims to fix the warning that is being presented with Virtual Packet Logging using the `NumPy.ravel()` function. Added a new `dtype=object` parameter for the `ravel` function to work properly. Fixing minor formatting changes as well.

**Description**

**Motivation and context**

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/base.py
+++ b/tardis/montecarlo/montecarlo_numba/base.py
@@ -86,28 +86,28 @@ def montecarlo_radial1d(model, plasma, runner):
 
     if montecarlo_configuration.VPACKET_LOGGING and number_of_vpackets > 0:
         runner.virt_packet_nus = np.concatenate(
-            np.array(virt_packet_nus)
+            np.array(virt_packet_nus, dtype=object)
         ).ravel()
         runner.virt_packet_energies = np.concatenate(
-            np.array(virt_packet_energies)
+            np.array(virt_packet_energies, dtype=object)
         ).ravel()
         runner.virt_packet_initial_mus = np.concatenate(
-            np.array(virt_packet_initial_mus)
+            np.array(virt_packet_initial_mus, dtype=object)
         ).ravel()
         runner.virt_packet_initial_rs = np.concatenate(
-            np.array(virt_packet_initial_rs)
+            np.array(virt_packet_initial_rs, dtype=object)
         ).ravel()
         runner.virt_packet_last_interaction_in_nu = np.concatenate(
-            np.array(virt_packet_last_interaction_in_nu)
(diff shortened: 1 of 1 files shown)


andrewfullard:
Can you add an image/info about the exact warning?

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -86,28 +86,28 @@ def montecarlo_radial1d(model, plasma, runner):
 
     if montecarlo_configuration.VPACKET_LOGGING and number_of_vpackets > 0:
         runner.virt_packet_nus = np.concatenate(
-            np.array(virt_packet_nus)
+            np.array(virt_packet_nus, dtype=object)
```
Rodot-, line 89: Why are you setting the dtype to object?  These arrays already have the correct dtype

Rodot-:
Please include the text of the warning

DhruvSondhi:
Yes Sure, @andrewfullard @Rodot-. The warning is as below:
image
```python
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  np.array(virt_packet_nus)
```

tardis/montecarlo/montecarlo_numba/base.py
DhruvSondhi: Yes but the function `numpy.concatenate` has some updates in the newer version 🤔

Rodot-:
The sequences should not be nested, the problem is that we are creating an array superfluously before concatenating.  We do not want to be treating those as objects, they should be arrays.

tardis/montecarlo/montecarlo_numba/base.py
Rodot-: Which version?

DhruvSondhi: image
Numpy Version 1.20.0

Rodot-: Is that the version specified in the tardis environment yaml?

DhruvSondhi: No, its 1.19 in the YAML

PR #1726, opened 2021-07-15
display "latest" in doc version

epassaro:
**Description**

**Motivation and context**

As @jaladh-singhal pointed out, looks more "stable". The documentation pipeline builds the `master` branch tip, so it's a good name. Also it's the same used by RTD.

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

Built locally.

**Examples**

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above. 

**Checklist**

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -129,13 +129,14 @@ if os.getenv("DISABLE_NBSPHINX") == "1":
 else:
     nbsphinx_execute = "auto"
 
+
 # -- Project information ------------------------------------------------------
 
 # This does not *have* to match the package name, but typically does
 project = setup_cfg["name"]
 author = setup_cfg["author"]
 copyright = "2013-{0}, {1}".format(
-    datetime.datetime.now().year, setup_cfg["author"])
+    datetime.datetime.now().year, author)
 
 # The version info for the project you"re documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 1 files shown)


andrewfullard (approved):

docs/conf.py
```diff
@@ -198,7 +199,7 @@
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-html_title = "{0} v{1}".format(project, release)
+html_title = "" # "{0} v{1}".format(project, release)
```
jaladh-singhal, line 202:
```suggestion
html_title = project # "{0} v{1}".format(project, release)
```
This will keep page titles (that are shown when you hover over tab) meaningful. For e.g. With this it will be "Installation - TARDIS" instead of "Installation" for https://epassaro.github.io/tardis/branch/simple-version-docs/installation.html

jaladh-singhal (changes requested):
Thanks for making quick PR. There's one thing we should take care of

DhruvSondhi (approved):

docs/conf.py
DhruvSondhi: This could be meaningful in the sense that we are keeping the `project` variable as the HTML Title

epassaro:
> Thanks for making quick PR. There's one thing we should take care of

Done!

jaladh-singhal (approved):

PR #1729, opened 2021-07-15
require time_0 to exist and be a astropy quantity

wkerzendorf:
The decay dataframe should really require a time of definition and not just assume `time_0 = 0 * u.day` if this is not given
We also make sure that time_0 is a quantity (using ducktyping). 

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [x] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/decay.py
+++ b/tardis/io/decay.py
@@ -6,10 +6,10 @@ from astropy import units as u
 class IsotopeAbundances(pd.DataFrame):
     """[summary]
 
-        Parameters
-        ----------
-        time_0 : astropy.units.Quantity
-            time where the given isotopes are defined
+    Parameters
+    ----------
+    time_0 : astropy.units.Quantity
+        time where the given isotopes are defined
     """
 
     _metadata = ["time_0"]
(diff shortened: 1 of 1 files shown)


andrewfullard:
I think existing simple configurations need to be updated so that tests pass. Also, black has failed.

tardis/io/decay.py
```diff
@@ -4,16 +4,19 @@
 
 
 class IsotopeAbundances(pd.DataFrame):
+    """[summary]
+
+        Parameters
+        ----------
+        time_0 : astropy.units.Quantity
+            time where the given isotopes are defined
```
afloers, line 12: black reformatting:
```suggestion
    Parameters
    ----------
    time_0 : astropy.units.Quantity
        time where the given isotopes are defined
```

afloers (changes requested):
I agree that the time_0 should be specified in all Tardis runs. 
Black needs to be run on the `tardis/io/decay.py` file, there were additional PEP8 violations. I fixed the ones added in this PR.
As @andrewfullard wrote tests are failing because the configurations used in the tests haven't been updated yet.

andrewfullard (changes requested):
Fix the tests :)

PR #1731, opened 2021-07-16
Moved Montecarlo Numba Logging Configuration to `debug.yaml`

### LINKED ISSUE #1732: Move Montecarlo Logging Configuration to Debug YAML
**Describe the bug**
The montecarlo numba logger configuration must be moved to `debug.yaml` to keep consistency with respect to the debugging parameters being under the same YAML file. Allows setting up the debugging parameter easy & intuitive.
Check PR #1731

DhruvSondhi:
This PR aims to move the Montecarlo Numba Configuration to `debug.yaml` to keep consistency. This moves all the debugging parameters in a single config yaml part of the TARDIS Config YAML.

Fix #1732 

**Description**

**Motivation and context**

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/schemas/debug.yml
+++ b/tardis/io/schemas/debug.yml
@@ -19,4 +19,17 @@ properties:
     type: boolean
     default: false
     description: Allows for logging on the specified logging levels
+  debug_packets:
+    type: boolean
+    default: false
+    description: Decide whether to go into debugging mode. [EXPERIMENTAL FEATURE DO NOT RELY ON IT]
+  logger_buffer:
+    type: number
+    default: 1
+    description: Provides option to not log every line.
+  single_packet_seed:
+    type:
+      - number
+    default: -1
+    description: If debug_packets is true, this is the seed for the only packet.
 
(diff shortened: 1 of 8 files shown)


andrewfullard:
Looks like you need to fix the failing tests.

DhruvSondhi:
There is a problem with this change. There is a requirement of the `debug` section to be present inside the config file (even though all the values are being accessed via the default. This is raising multiple errors where ever the `debug` section is not present. There are 2 ways to solve this problem. One of them has been implemented in this PR to fix the failing test. It is that we add `debug` section to the config files in the following manner:
```
debug:
   specific: false
```
All other values are accessed via the Default values that have been set in the debug schema. Because the debug schema is being accessed as `config.debug. __`. 
The other way of fixing is that we use the `if else` statement to access the value. If `debug` is not present then we can access the value in some other way. This could be a matter of discussion. Please discuss the same :)

andrewfullard:
I think the fix by adding the debug entry to the configs is incorrect, because if you are moving it to a specific location, that should be consistent across all configs. I would like to see your implementation of the second option you provided.

wkerzendorf:
there should not be a debug section - all of this should be under monte carlo

DhruvSondhi:
What about the configuration of the Logger? I have implemented logic wrt to a `debug` section present inside a config that can be used to select the level & the specificity for the logging messages.

PR #1738, opened 2021-07-16
TARDIS Grid

marxwillia:
**Description**
A new module for generating grids of TARDIS simulation parameters and functionality facilitating the running of large numbers of simulations.

**Motivation and context**
The grid streamlines running and saving large grids of simulation parameters.  It will make it very easy to generate grids from basic uniform abundances and then modify the models in systematic ways to add complexity.

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Examples**
A thorough example is in the docs under /io/grid/

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**


tardis/grid/__init__.py
```diff
@@ -0,0 +1 @@
+from tardis.grid.tardisgrid import *
```
andrewfullard, line 1: * imports are a bit risky, are you sure this is okay?

marxwillia: This is just so that the class can be imported using the statement `import tardis.grid`. I think most of the directories in tardis with __init__.py files do something similar.

andrewfullard:
You may want to consider adding logging messages or passthrough handling for logging through run_tardis

tardis/grid/__init__.py
marxwillia: But I can just as easily list everything out.  That's definitely safer in case the grid code grows.

marxwillia:
I still need to write some tests for this PR.

docs/io/grid/example.yml
```diff
@@ -0,0 +1,57 @@
+# Example YAML configuration for TARDIS
```
smithis7, line 1: Any specific reason we can't just use the main `tardis_example.yml`?

docs/index.rst
```diff
@@ -61,7 +61,7 @@ Mission Statement
     io/optional/index
     io/visualization/index
     io/output/index
-
+    io/grid/TardisGridTutorial
```
smithis7, line 64: Can we put this at least temporarily under `output/` instead?

smithis7 (changes requested):

tardis/grid/__init__.py
marxwillia: Going to leave as `*` import since this is the precedent in all other major modules.  We should discuss changing this practice in its own PR.

marxwillia:
> You may want to consider adding logging messages or passthrough handling for logging through run_tardis

Added kwarg pass through so that users can specify run_tardis() kwargs.

andrewfullard (approved):

tardis/grid/base.py
```diff
@@ -0,0 +1,172 @@
...
+
+class tardisGrid:
+    """
+    A class that stores a grid of TARDIS parameters and
+    facilitates running large numbers of simulations
+    easily.
+
+    Parameters
+    ----------
+    configFile : str
```
Rodot-, line 43: It looks like configFile can also be a dict, should this be documented?

tardis/grid/base.py
```diff
@@ -0,0 +1,172 @@
...
+        path to TARDIS yml file.
+    gridFrame : pandas.core.frame.DataFrame
```
Rodot-, line 45: Possibly do some type-checking here?  Could get weird if someone accidentally passes in a numpy array or a list

tardis/grid/base.py
```diff
@@ -0,0 +1,172 @@
...
+        The validated config dict read from the user provided
+        configFile. This provides the base properties for the
+        TARDIS simulation which the rows of the grid modify.
+    grid : pandas.core.frame.DataFrame
+        Dataframe where each row is a set of parameters for
+        a TARDIS simulation.
+    """
+
+    def __init__(self, configFile, gridFrame):
+        try:
```
Rodot-, line 61: Perhaps do type-checking here rather than exceptions?

Rodot- (approved):
I left a few small suggestions for clarity but functionally it looks good to me

tardis/grid/base.py
marxwillia: I don't think it's necessary since there's effectively duck typing since the grid relies on `.iloc[]` for accessing rows.

PR #1757, opened 2021-07-26
Auto-Generating Tutorials Page

smithis7:
**Description**

I have created a script that, when we build the documentation, generates an rst file that links to all of the TARDIS tutorials. I probably have not implemented the most elegant way to run the script, so if anyone has a better way of getting it to run please let me know.

While I was changing the sidebar, I moved the API section in the sidebar. I thought it would be rather silly to make that a PR on its own, and it *vaguely* had to do with this PR. ~~I made it its own commit, however, so if I need to remove that commit I will. I also added `ZENODO.rst` to `.gitignore` while I was adding `tutorials.rst` to `.gitignore` for the same reason.~~

**Motivation and context**

My goal was to create a page that lists all of the tutorials in TARDIS, which will automatically update when we add in a new tutorial (noting that all of our tutorials are under `docs/io`).

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Documentation built locally and on GitHub. 

**Examples**

See built documentation:
https://smithis7.github.io/tardis/branch/tutorials_doc/index.html

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -398,3 +398,26 @@ with open(zenodo_path, "w") as f:
     f.write(zenodo_record)
 
 print(zenodo_record)
+
+# -- Creating tutorials.rst ---------------------------------------------------
+
+notebook_list = []
+
+walk = os.walk("io/")
+
+for root, dirs, fnames in walk:
+    for fname in fnames:
+        if fname.endswith(".ipynb") and "checkpoint" not in fname:
+            notebook_list.append("* :doc:`" + root + "/" + fname[:-6] + "`")
+
+notebook_string = ""
+
+for notebook in notebook_list:
+    notebook_string += "\n" + notebook
+
+title = "Tutorials\n*********\n"
+
+description = "\nThe following pages contain the TARDIS tutorials:\n"
+
+with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
+    f.write(title + description + notebook_string)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -399,6 +399,7 @@ with open(zenodo_path, "w") as f:
 
 print(zenodo_record)
 
+
 # -- Creating tutorials.rst ---------------------------------------------------
 
 notebook_list = []
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -402,23 +402,15 @@ print(zenodo_record)
 
 # -- Creating tutorials.rst ---------------------------------------------------
 
-notebook_list = []
+notebooks = ""
 
-walk = os.walk("io/")
-
-for root, dirs, fnames in walk:
+for root, dirs, fnames in os.walk("io/"):
     for fname in fnames:
         if fname.endswith(".ipynb") and "checkpoint" not in fname:
-            notebook_list.append("* :doc:`" + root + "/" + fname[:-6] + "`")
-
-notebook_string = ""
-
-for notebook in notebook_list:
-    notebook_string += "\n" + notebook
+            notebooks += "\n* :doc:`" + root + "/" + fname[:-6] + "`"
 
 title = "Tutorials\n*********\n"
-
 description = "\nThe following pages contain the TARDIS tutorials:\n"
 
 with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
-    f.write(title + description + notebook_string)
+    f.write(title + description + notebooks)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -325,6 +325,7 @@ def to_html_ext(path):
     """Convert extension in the file path to .html"""
     return os.path.splitext(path)[0] + ".html"
 
+
 def generate_tutorials_page(app):
     notebooks = ""
 
@@ -338,6 +339,7 @@ def generate_tutorials_page(app):
 
     with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
         f.write(f"{title}\n{description}\n{notebooks}")
+
 
 def autodoc_skip_member(app, what, name, obj, skip, options):
     """Exclude specific functions/methods from the documentation"""
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -325,6 +325,19 @@ def to_html_ext(path):
     """Convert extension in the file path to .html"""
     return os.path.splitext(path)[0] + ".html"
 
+def generate_tutorials_page(app):
+    notebooks = ""
+
+    for root, dirs, fnames in os.walk("io/"):
+        for fname in fnames:
+            if fname.endswith(".ipynb") and "checkpoint" not in fname:
+                notebooks += f"\n* :doc:`{root}/{fname[:-6]}`"
+
+    title = "Tutorials\n*********\n"
+    description = "The following pages contain the TARDIS tutorials:"
+
+    with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
+        f.write(f"{title}\n{description}\n{notebooks}")
 
 def autodoc_skip_member(app, what, name, obj, skip, options):
     """Exclude specific functions/methods from the documentation"""
@@ -355,8 +368,8 @@ def create_redirect_files(app, docname):
             with open(old_html_fpath, "w") as f:
                 f.write(new_content)
 
-
 def setup(app):
+    app.connect("builder-inited", generate_tutorials_page)
     app.connect("autodoc-skip-member", autodoc_skip_member)
     app.connect("build-finished", create_redirect_files)
 
@@ -398,19 +411,3 @@ with open(zenodo_path, "w") as f:
     f.write(zenodo_record)
 
 print(zenodo_record)
-
-
-# -- Creating tutorials.rst ---------------------------------------------------
-
-notebooks = ""
-
-for root, dirs, fnames in os.walk("io/"):
-    for fname in fnames:
-        if fname.endswith(".ipynb") and "checkpoint" not in fname:
-            notebooks += f"\n* :doc:`{root}/{fname[:-6]}`"
-
-title = "Tutorials\n*********\n"
-description = "The following pages contain the TARDIS tutorials:"
-
-with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
-    f.write(f"{title}\n{description}\n{notebooks}")
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -407,10 +407,10 @@ notebooks = ""
 for root, dirs, fnames in os.walk("io/"):
     for fname in fnames:
         if fname.endswith(".ipynb") and "checkpoint" not in fname:
-            notebooks += "\n* :doc:`" + root + "/" + fname[:-6] + "`"
+            notebooks += f"\n* :doc:`{root}/{fname[:-6]}`"
 
 title = "Tutorials\n*********\n"
-description = "\nThe following pages contain the TARDIS tutorials:\n"
+description = "The following pages contain the TARDIS tutorials:"
 
 with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
-    f.write(title + description + notebooks)
+    f.write(f"{title}\n{description}\n{notebooks}")
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -50,7 +50,8 @@ Mission Statement
 
     installation
     quickstart/quickstart
-
+    tutorials
+    API <api/modules>
 
 
 .. toctree::
@@ -103,14 +104,6 @@ Mission Statement
     development/index
     CHANGELOG.md
     roadmap
-
-.. toctree::
-    :maxdepth: 2
-    :caption: API
-    :hidden:
-
-    api/modules
-
 
 
 .. toctree::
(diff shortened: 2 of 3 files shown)


epassaro:
1.  I don't see why this can't be implemented through `conf.py`, probably I'm missing something. Editing the Makefile it's not a good idea as you already pointed out.
2. Adding ZENODO.rst to `.gitignore` will break up the TARDIS citation. To keep this we should force push `ZENODO.rst` on the pipeline but I don't think that's a good idea. Probably also applies to `tutorials.rst`, I'm not a huge fan to adding too many things to `.gitignore`.

smithis7:
> 1. I don't see why this can't be implemented through `conf.py`, probably I'm missing something. Editing the Makefile it's not a good idea as you already pointed out.
> 2. Adding ZENODO.rst to `.gitignore` will break up the TARDIS citation. To keep this we should force push `ZENODO.rst` on the pipeline but I don't think that's a good idea. Probably also applies to `tutorials.rst`, I'm not a huge fan to adding too many things to `.gitignore`.

So to your first point, I could not figure out a way to do this through `conf.py`. If you have any ideas, I will try it out, but to me it seemed like the Makefile was the only way to do it.

For the second point I can revert that commit. I am purely wondering though, since `ZENODO.rst` is not in our repository, what would be messed up by adding it to `.gitignore`? I do think `tutorials.rst` should be ignored by git as we don't want it in the repository, just want it built during the docs build process.

docs/tutorials_generator.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+notebook_list = []
+
+walk = os.walk("io/")
+
+for pth, dirs, files in walk:
+    for file in files:
+        if ".ipynb" in file and "checkpoint" not in file:
```
epassaro, line 9: Try using method `endswith` for strings and `isdir` for checkpoint directories. Also we use a lot `fname` instead of `file` throughout TARDIS code.

docs/tutorials_generator.py
```diff
@@ -0,0 +1,22 @@
...
+notebook_string = ""
+
+for notebook in notebook_list:
+    notebook_string += "\n" + notebook
+
+title = "Tutorials\n*********\n"
+
+description = "\nThe following pages contain the TARDIS tutorials:\n"
+
+with open("tutorials.rst", mode="wt", encoding="utf-8") as tutorials_page:
```
epassaro, line 21: File buffers are usually called `f` or something shorter.

epassaro:
> > 1. I don't see why this can't be implemented through `conf.py`, probably I'm missing something. Editing the Makefile it's not a good idea as you already pointed out.
> > 2. Adding ZENODO.rst to `.gitignore` will break up the TARDIS citation. To keep this we should force push `ZENODO.rst` on the pipeline but I don't think that's a good idea. Probably also applies to `tutorials.rst`, I'm not a huge fan to adding too many things to `.gitignore`.
> 
> So to your first point, I could not figure out a way to do this through `conf.py`. If you have any ideas, I will try it out, but to me it seemed like the Makefile was the only way to do it.
> 
> For the second point I can revert that commit. I am purely wondering though, since `ZENODO.rst` is not in our repository, what would be messed up by adding it to `.gitignore`? I do think `tutorials.rst` should be ignored by git as we don't want it in the repository, just want it built during the docs build process.

Just copy and paste the script content at the end of `conf.py` and see what it does.

docs/index.rst
```diff
@@ -50,7 +50,8 @@ Mission Statement
 
     installation
     quickstart/quickstart
-
+    tutorials
+    API <api/modules>
```
epassaro, line 54: I like the old API subsection, but let see what others think about it!

smithis7: Ok, completely understandable. Just to put it out there, there are a few reasons I did that:
- It shortens the sidebar, which is an overall goal of mine in making the documentation easier to navigate.
- It makes the API easier to find. As we continue fixing up docstrings, I think the API should be closer to the top than it was before. I think this way puts it in a place where it makes sense.
- The single page in its own section looked sloppy IMO (I have plans for most of the other sections like that).

jaladh-singhal: I think these reasons do make sense - esp making API easier to find also for new users

docs/conf.py
```diff
@@ -398,3 +398,27 @@ def setup(app):
...
+
+for root, dirs, fnames in walk:
+    for fname in fnames:
+        if fname.endswith(".ipynb") and "checkpoint" not in fname:
+            notebook_list.append("* :doc:`" + root + "/" + fname[:-6] + "`")
+
+notebook_string = ""
+
+for notebook in notebook_list:
+    notebook_string += "\n" + notebook
```
jaladh-singhal, line 417: I don't see any reason to do it in a separate loop when you can do same thing directly above in L412

docs/conf.py
```diff
@@ -398,3 +398,27 @@ def setup(app):
...
+
+title = "Tutorials\n*********\n"
+
+description = "\nThe following pages contain the TARDIS tutorials:\n"
```
jaladh-singhal, line 421:
```suggestion
title = "Tutorials\n*********\n"
description = "\nThe following pages contain the TARDIS tutorials:\n"
```
Newlines shouldn't be used that often - they should be used when a code block needs to be logically splitted into chunks (like it's fine to keep a newline between L421 and L423)

docs/conf.py
```diff
@@ -398,3 +398,27 @@ def setup(app):
...
 print(zenodo_record)
+
+
+# -- Creating tutorials.rst ---------------------------------------------------
+
+notebook_list = []
+
+walk = os.walk("io/")
+
+for root, dirs, fnames in walk:
```
jaladh-singhal, line 409:
```suggestion
for root, dirs, fnames in os.walk("io/"):
```
No need to define it as variable unless you're reusing it

jaladh-singhal (changes requested):
This looks like a great addition - I think your code can be improved a bit (see my comments). Besides we use f-strings in most of TARDIS modules since they're more readable so please use them over strings concatenation - it'll also save you from adding \n before, after your variables (like `title`, `description`) since you can put them as `f"{title}\n{description}\n"`

Apart from this, I was wondering if we can reorder notebooks in list meaningfully or perhaps make the not-so-informative list of notebooks a bit more informative (& attractive :smile: ) by adding relevant visuals (for e.g. how einsteinpy [does it](https://gke.mybinder.org/v2/gh/einsteinpy/einsteinpy/main?filepath=index.ipynb))

epassaro:
I don't see why it's necessary to add `docs/tutorials.rst` to `.gitignore`. If there's no alternative, please add it at the bottom of the file in a new section (we are currently using the Astropy `.gitignore`).

smithis7:
> I don't see why it's necessary to add `docs/tutorials.rst` to `.gitignore`. If there's no alternative, please add it at the bottom of the file in a new section (we are currently using the Astropy `.gitignore`).

It definitely needs to be in `.gitignore`. We do not want that file in our repository. I do think it belongs in the sphinx section but I have placed it at the bottom as I understand the concern about not wanting to mess with the astropy stuff.

epassaro:
> > I don't see why it's necessary to add `docs/tutorials.rst` to `.gitignore`. If there's no alternative, please add it at the bottom of the file in a new section (we are currently using the Astropy `.gitignore`).
> 
> It definitely needs to be in `.gitignore`. We do not want that file in our repository. I do think it belongs in the sphinx section but I have placed it at the bottom as I understand the concern about not wanting to mess with the astropy stuff.

If the file is not committed then it's not on the repository. It's just on your computer marked as "untracked". There's nothing wrong with that (it's the same situation with ZENODO.rst)

smithis7:
> > > I don't see why it's necessary to add `docs/tutorials.rst` to `.gitignore`. If there's no alternative, please add it at the bottom of the file in a new section (we are currently using the Astropy `.gitignore`).
> > 
> > 
> > It definitely needs to be in `.gitignore`. We do not want that file in our repository. I do think it belongs in the sphinx section but I have placed it at the bottom as I understand the concern about not wanting to mess with the astropy stuff.
> 
> If the file is not committed then it's not on the repository. It's just on your computer marked as "untracked". There's nothing wrong with that (it's the same situation with ZENODO.rst)

Personally, I think it makes it more clear that we do not want it in the repo to have it in the `.gitignore`. I think we should defer to @wkerzendorf on this one.

smithis7:
@epassaro  I removed any edits to `.gitignore`.

docs/conf.py
```diff
@@ -398,3 +398,19 @@ def setup(app):
...
+
+
+# -- Creating tutorials.rst ---------------------------------------------------
+
+notebooks = ""
+
+for root, dirs, fnames in os.walk("io/"):
+    for fname in fnames:
+        if fname.endswith(".ipynb") and "checkpoint" not in fname:
+            notebooks += "\n* :doc:`" + root + "/" + fname[:-6] + "`"
```
jaladh-singhal, line 410:
```suggestion
            notebooks += f"\n* :doc:`{root}/{fname[:-6]}`"
```
Doesn't f-strings look neater also easier to write than string concatenation :smile:

docs/conf.py
```diff
@@ -398,3 +398,19 @@ def setup(app):
...
+
+title = "Tutorials\n*********\n"
+description = "\nThe following pages contain the TARDIS tutorials:\n"
+
+with open("tutorials.rst", mode="wt", encoding="utf-8") as f:
+    f.write(title + description + notebooks)
```
jaladh-singhal, line 416: You can do something like this (provided you remove leading & trailing `\n` from `tiltle` and `description`):
```suggestion
    f.write(f"{title}\n\n{description}\n{notebooks})
```

jaladh-singhal (changes requested):
Thanks for making changes - code looks lot neater now. Though you forgot to address my comment about using f-strings!

jaladh-singhal:
> If the file is not committed then it's not on the repository. It's just on your computer marked as "untracked". There's nothing wrong with that (it's the same situation with ZENODO.rst)

That's definitely wrong IMO @epassaro. Why? Because some new contributor built docs to add a page locally and tutorials.rst will get generated and they'll likely commit all the files together.

In general, almost auto-generated files should be gitignored. You can notice same for `docs/_build`, `tutorials.rst` is just same, difference is it got generate post-build phase.

smithis7:
> > If the file is not committed then it's not on the repository. It's just on your computer marked as "untracked". There's nothing wrong with that (it's the same situation with ZENODO.rst)
> 
> That's definitely wrong IMO @epassaro. Why? Because some new contributor built docs to add a page locally and tutorials.rst will get generated and they'll likely commit all the files together.
> 
> In general, almost auto-generated files should be gitignored. You can notice same for `docs/_build`, `tutorials.rst` is just same, difference is it got generate post-build phase.

That was my thinking as well.

smithis7:
> > If the file is not committed then it's not on the repository. It's just on your computer marked as "untracked". There's nothing wrong with that (it's the same situation with ZENODO.rst)
> 
> That's definitely wrong IMO @epassaro. Why? Because some new contributor built docs to add a page locally and tutorials.rst will get generated and they'll likely commit all the files together.
> 
> In general, almost auto-generated files should be gitignored. You can notice same for `docs/_build`, `tutorials.rst` is just same, difference is it got generate post-build phase.

I think we will discuss the `.gitignore` issue and if needed, address it in another PR.

wkerzendorf:
@smithis7 @jaladh-singhal @epassaro my main criticism of the PR is that I do not think that is how files should be generated in the Sphinx API (I might be wrong). It feels very hacky to add this to the `conf.py`

wkerzendorf:
https://stackoverflow.com/questions/7250659/how-to-use-python-to-programmatically-generate-part-of-sphinx-documentation

smithis7:
@wkerzendorf how I did it is the same way we generate the zenodo file.

I can use the link you suggested, but that involves (pretty messy IMO) changes to the `Makefile` which feels way more hacky.

smithis7:
@wkerzendorf how is this implementation? This is the more "official" way of doing it, by hooking into the sphinx build process.

jaladh-singhal (approved):
@smithis7 looks good to me except I think you must add `tutotrials.rst` to gitignore in this same PR. 

@epassaro please read this https://www.atlassian.com/git/tutorials/saving-changes/gitignore - tutorial.rst is generated at runtime and we don't want users who have built docs locally to accidentally commit it.
> - files generated at runtime, such as .log, .lock, or .tmp

jaladh-singhal:
> @smithis7 @jaladh-singhal @epassaro my main criticism of the PR is that I do not think that is how files should be generated in the Sphinx API (I might be wrong). It feels very hacky to add this to the `conf.py`

@wkerzendorf I don't think we have a better way for this not-so-common usecase. Also now that @smithis7 has hooked this function to build initiation event of sphinx, I can confirm this is right way to do it because I used same approach to create redirects in #1458:
https://github.com/tardis-sn/tardis/blob/6a0b8db607b87d91ef8312906758f7d380103b47/docs/conf.py#L361

PR #1765, opened 2021-07-30
Quantity data frame

wkerzendorf:
Add the long awaited quantity dataframe.

**Description**

**Motivation and context**

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -600,9 +600,11 @@ def is_notebook():
 class QDataFrame(pd.DataFrame):
     # normal properties
     _metadata = ["units"]
-    def __init__(self, units=[], *args, **kwargs):
+    def __init__(self, *args, **kwargs):
+        units = kwargs.pop('units', [])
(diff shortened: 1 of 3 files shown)


tardis/util/base.py
```diff
@@ -596,3 +596,15 @@ def is_notebook():
     # All other shell instances are returned False
     else:
         return False
+
+class QDataFrame(pd.DataFrame):
+    # normal properties
+    _metadata = ["units"]
+    def __init__(self, *args, **kwargs):
+        units = kwargs.pop('units', [])
```
marxwillia, line 604: Should we require the length of the units array to be equal to the number of columns? Would it be more useful to have the units attribute be a dictionary with (key,value) pairs equal to (df_column_name, unit)?

PR #1769, opened 2021-08-04
Changed the way `specific_log_level` flag works in the Logging Configuration

DhruvSondhi:
The current implementation didn't cater to a bug that I have encountered recently. This PR aims to implement the functionality such that we are able to set the `specific` flag default value in the `debug` sub section of the `tardis_config.yml`.

**Description**

**Motivation and context**

**How has this been tested?**
- [x] Testing pipeline.
- [x] Other. 

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/logger/logger.py
+++ b/tardis/io/logger/logger.py
@@ -102,15 +102,28 @@ def logging_state(log_level, tardis_config, specific_log_level):
             print(
                 "log_level is defined both in Functional Argument & YAML Configuration {debug section}"
             )
-            print(
-                f"log_level = {log_level.upper()} will be used for Log Level Determination\n"
+
+            logging_level = (
+                log_level if log_level else tardis_config["debug"]["log_level"]
             )
 
+            # Displays a message when both log_level & tardis["debug"]["log_level"] are specified
+            if log_level and tardis_config["debug"]["log_level"]:
+                print(
+                    "log_level is defined both in Functional Argument & YAML Configuration {debug section}"
+                )
+                print(
+                    f"log_level = {log_level.upper()} will be used for Log Level Determination\n"
+                )
+
     else:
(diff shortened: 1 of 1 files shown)


tardis/io/logger/logger.py
```diff
@@ -112,6 +113,9 @@ def logging_state(log_level, tardis_config, specific):
 
         if specific:
             specific = specific
```
andrewfullard, line 115: Instead of an if..else block, why not `if not specific`?

tardis/io/logger/logger.py
```diff
@@ -85,7 +86,7 @@ def logging_state(log_level, tardis_config, specific):
         Allows to set specific logging levels. Logs of the `log_level` level would be output.
     """
 
-    if "debug" in tardis_config:
+    if "debug" in tardis_config and "specific" in tardis_config["debug"]:
         specific = (
             tardis_config["debug"]["specific"] if specific is None else specific
```
andrewfullard, line 91: If `specific` is undefined, then perhaps it should be `else DEFAULT_SPECIFIC_STATE`

andrewfullard (changes requested):

tardis/io/logger/logger.py
andrewfullard: I believe this is still relevant

DhruvSondhi: Yes, I had the implementation wrong in the first place. Will change it accordingly :)

andrewfullard:
Can you write up the bug that this fixes into an issue so we know what is being fixed?

andrewfullard (approved):

Issue #1779, opened 2021-08-13
Change selected shell in the dropdown when range of shells is chosen in Custom Abundance Widget


jaladh-singhal:
In custom abundance widget when selecting radio button "range of shells", the selection in shell no. dropdown stays same - instead it should be changed to account the now selected shell range. For e.g. If I had selected 5th shell and then I selected range as 9-15, the values in abundance inputs still correspond to 5th shell. Also box selection for 5th shell in the plot still persists that should have disappeared.

It'll be better to make it like this:
- Change selected shell in shell no. dropdown to leftmost edge of selected range of shells (i.e. shell 9 as in above example). Thereby it will update abundance inputs to show abundances for 9th shell and not for 5th shell.
- Make the box selection for 5th shell disappear in the plot (otherwise there will still be two overlapping box selections)
- And most importantly: whenever "range of shells" radio button is selected, show a note that abundance values shown below are for the innermost shell in selected range.

PR #1799, opened 2021-09-18
Bug fixes and Improvements in Custom Abundance Widget

yuyizheng1112:
This PR contains several changes and bug fix to custom abundance widget.

**Description**

- Set abundances in new shell(s) to the abundances in the left (inner) shell instead of setting to 0. In the case of adding/overwriting innermost shell, set the abundances equal to the first shell.
- Added the feature required in issue #1779 .
- Fixed the bugs in `write_csv_portion` function and added config version information when saving CSVY file. Now the velocity of the output file is correct and `from_csvy` function works properly.
- FIxed the issue of "Unnormalized" warning display.

**Motivation and context**

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 

**Examples**

After adding new shell(s), the abundances in new shell(s) keep the same with the left shell:

截屏2021-09-18 下午3 51 28

When Select a range of shells, it will be highlighted in the plot and the shell No dropdown will be disabled:

截屏2021-09-18 下午3 52 21

**Type of change**

- [x] Bug fix. 
- [x] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**


jaladh-singhal (changes requested):
Very excellent work @yuyizheng1112 - this is an almost perfect PR. :clap: 

You've made it exactly as we wanted it to be. I've ran the widget locally and it works great - esp. how the range selection is now handled.

I just think since this PR makes 4 different fixes as listed by you (that should have ideally be in separate PRs but it's ok now that you've made one), you should rename PR title to something more generic (like ___ improvements in custom abundance widget) that includes all 4 points. Current tile is just apt for 1st point.

Only problem that I noticed while running it locally is the shell-overwriting warning persists even after clicking "add" button which I guess was not the case before. I'm happy to merge this once that's fixed. And I should mention unnormalized warnings work correctly now!

luharvey:
Hi Jenny, this is looking great and the csvy import and export issues are fixed. There is only one final bug I can see when trying to run the csvy file, for some reason the v_inner_boundary is slightly lower than the velocity of the first shell (about 1km/s lower so perhaps a rounding issue?). TARDIS then complains that the inner boundary velocity is lower than the lowest velocity in the model.

yuyizheng1112:
I think this pr can be merged now @jaladh-singhal @jamesgillanders

jaladh-singhal:
We'd be happy to! @yuyizheng1112 can you please quickly resolve the merge conflict by rebasing it?

jaladh-singhal (approved):
Everything works fine now! Thanks alot @yuyizheng1112

PR #1817, opened 2021-10-20
Fixing SDEC docs page

smithis7:
**Description**

This provides a permanent fix to the issue in #1811. The problem was that the SVG file format which the SDEC plots were saved as are too large (6+ MB per plot). This PR makes it so the plots on this page (not any other page) are exported as high-resolution PNG (png2x) files -- this is not as nice as SVG, but it will do for that page. The file size is much more manageable. I have also included a description of this problem and its solution in the documentation.

**Examples**

The two pages involved:
https://smithis7.github.io/tardis/branch/sdec_fix_doc/io/visualization/sdec_plot.html
https://smithis7.github.io/tardis/branch/sdec_fix_doc/development/documentation_guidelines.html (see note at the bottom of the page)

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**


smithis7:
> Thank you for fixing this! Everything looks great to me, I just had a few small questions though: Is the `InlineBackend.figure_format` statement deprecated? It says [here](https://ipython.org/ipython-doc/2/config/options/notebook.html) in this old version of docs that we must be using `InlineBackend.figure_formats` instead. However, in the [newer](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.set_matplotlib_formats) docs even that is deprecated, and it says that we should use `matplotlib_inline.backend_inline.set_matplotlib_formats()`. Maybe you might want to check out [this](https://github.com/ipython/matplotlib-inline/blob/ea6d2c1c9447cef729110ca1c9de24b69e5536a3/matplotlib_inline/backend_inline.py#L256) too.
> 
> Also, we do not have an option of `png2x` there. Is that also deprecated, or is it the same as the `retina` option?

Indeed, `figure_format` is deprecated -- great catch! I believe `%config InlineBackend.figure_formats='png2x'` is the same as putting `c.InlineBackend.figure_formats='png2x'` in the configuration file, so it is consistent with the [current documentation](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.set_matplotlib_formats).

atharva-2001 (approved):
Great work! I wondered if `%config InlineBackend.figure_formats` is deprecated, but it is not mentioned anywhere in the documentation that it is, just the function(`IPython.display.set_matplotlib_formats`) is deprecated. 
Also, both the `png2x` and the `retina` options are the [same](https://github.com/ipython/ipython/blob/76b369efde69aaa2efbb732565a2d9669f716bcf/IPython/core/pylabtools.py#L282), so I am happy with the changes.

PR #1818, opened 2021-10-22
QDataframe

wkerzendorf:
Follow-up of #1765


wkerzendorf:
@Rodot- so - there are tests as a Christmas present. There will be more - but I first want to get this started in pandas.

Rodot-:
Is this still something we want to add?  With things like Pandas 2.0 this might lead to a maintenance burden.

PR #1827, opened 2021-11-19
Adding Explanations to Spectrum Configuration

bartnikm:
**Description**
Added Explanations to the Spectrum Configuration Section of the Documentation

**Motivation and context**
The original did not have a description of how to edit method parameters, so this would be nice for newer users

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above. 

**Checklist**


docs/io/configuration/components/spectrum.rst
```diff
@@ -15,7 +15,38 @@ just an integer. Finally, the method option selects the final spectral synthesis
 * real: construct spectrum from the real packet population alone
 * virtual: use the :ref:`virtual packet scheme <virtual_packets>` for spectral synthesis
 * integrated: use the :ref:`formal integral method <formal_integral>` of Lucy 1999
- 
+
```
smithis7, line 18: So this has nothing to do with the job you've done (which is great), and I wish I had noticed this prior. I don't believe that this info is up-to-date. @andrewfullard or @wkerzendorf can confirm, but I believe that the `method` argument does nothing. I tested this out and nothing changed (including in the hdf output). Currently, TARDIS can calculate all three types (as is described in the spectrum generation section of the docs). We just have extra settings that can be specified for the virtual and integrated methods.

docs/io/configuration/components/spectrum.rst
```diff
@@ -15,7 +15,38 @@ just an integer. Finally, the method option selects the final spectral synthesis
 * real: construct spectrum from the real packet population alone
```
smithis7, line 15: I think we should link the new page on basic spectrum generation here.

tardis/io/schemas/spectrum.yml
```diff
@@ -3,10 +3,13 @@ type: object
 properties:
   start:
     type: quantity
+    description: The beginning of the spectrum to be generated
   stop:
     type: quantity
+    description: The end of the spectrum to be generated
   num:
     type: number
+    description: Number of iterations used to build spectrum
```
smithis7, line 12: This is the number of bins that we use to generate the spectrum (however you want to phrase it)

tardis/io/schemas/spectrum.yml
```diff
@@ -3,10 +3,13 @@ type: object
 properties:
   start:
     type: quantity
+    description: The beginning of the spectrum to be generated
```
smithis7, line 6: Beginning and end may want to be clarified a bit

smithis7 (changes requested):
Lines 11-13 could use some better explanation and clean-up (I know you did not write them). ``Start`` should not be capitalized and I'd probably say "value with units" as opposed to "quantity with units" as they are not expressed as astropy quantities. Also, I don't entirely like how it uses "frequency space" as I think we can explain that with less technical vocabulary.

docs/io/configuration/components/spectrum.rst
bartnikm: I did think it was weird that the method parameter was there when each method is called separately anyway. I presumed this served some function as setting a default method to be called. I will look into this

docs/io/configuration/components/spectrum.rst
bartnikm: If by new page you mean [https://tardis-sn.github.io/tardis/physics/spectrum/basic.html](url), then I think under the virtual construction we should link the page on spectrum generation with virtual packets as well ( [https://tardis-sn.github.io/tardis/physics/spectrum/virtualpackets.html](url) ), unless this page is still under development?

docs/io/configuration/components/spectrum.rst
bartnikm: For the time being, I will address this as a warning

docs/io/configuration/components/spectrum.rst
```diff
@@ -8,15 +8,42 @@ Finally, the spectrum configuration tells TARDIS information needed for spectrum
 
 .. jsonschema:: schemas/spectrum.yml
 
-``Start`` and ``end`` are given as Quantities with units. If they are given in
-frequency space they are switched around if necessary. The number of bins is
-just an integer. Finally, the method option selects the final spectral synthesis mode. Currently, there are three options:
+``start`` and ``end`` are given as values with units.  
+``num`` specifies the number of bins used to build the spectrum and must be given as an integer. 
+TARDIS produces the spectrum in three different methods. For more information on these methods, visit the
```
smithis7, line 13:
```suggestion
TARDIS produces the spectrum via three different methods. For more information on these methods, visit the
```

docs/io/configuration/components/spectrum.rst
```diff
@@ -8,15 +8,42 @@ Finally, the spectrum configuration tells TARDIS information needed for spectrum
...
+pages below:
  
-* real: construct spectrum from the real packet population alone
-* virtual: use the :ref:`virtual packet scheme <virtual_packets>` for spectral synthesis
-* integrated: use the :ref:`formal integral method <formal_integral>` of Lucy 1999
+* real: :doc:`Basic Spectrum Generation <../../../physics/spectrum/basic>`
+* virtual: :ref:`Virtual Packet Scheme <virtual_packets>`
+* integrated: :ref:`Formal Integral Method <formal_integral>`
+
+The three methods can be specified when running the simulation (see the :doc:`quickstart guide <../../../quickstart/quickstart>` for an example of this).
```
smithis7, line 20:
```suggestion
The three methods can be specified when plotting the spectrum (see the :doc:`quickstart guide <../../../quickstart/quickstart>` for an example of this).
```

smithis7 (changes requested):
Other than these quick fixes (you can just commit them directly from github) it looks good to go!

smithis7 (approved):
Great!

andrewfullard (approved):

PR #1832, opened 2021-11-26
Convert quantities to arrays when plotting in SDEC plot

atharva-2001:
This fixes the docs build that was caused due to the new matplotlib release.

**Description**

Using quantities with `matplotlib`'s `fill_between` method was causing an error. This PR changes quantities to arrays, just when plotting, everywhere in the SDEC plot so that no problems arise later as well. 

**Motivation and context**

**How has this been tested?**
- [X] Testing pipeline.
- [ ] Other. 

**Examples**

[Link](https://github.com/jaladh-singhal/tardis/runs/4301154620?check_suite_focus=true) to a failed documentation build
[working docs build](https://atharva-2001.github.io/tardis/branch/quantity_support/)
**Type of change**

- [X] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -1177,8 +1177,8 @@ class SDECPlotter:
         # Plot modeled spectrum
         if show_modeled_spectrum:
             self.ax.plot(
-                self.plot_wavelength,
-                self.modeled_spectrum_luminosity,
+                self.plot_wavelength.value,
+                self.modeled_spectrum_luminosity.value,
                 "--b",
                 label=f"{packets_mode.capitalize()} Spectrum",
                 linewidth=1,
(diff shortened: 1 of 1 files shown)


jamesgillanders (commented):
this fixes the problem, but do we want to do this, or pin the matplotlib version?

smithis7:
> this fixes the problem, but do we want to do this, or pin the matplotlib version?

We definitely want this. Pinning the matplotlib version was a backup in case this problem was not easy enough to fix.

smithis7 (commented):
My only concern here is do we allow users to use different units when plotting? And either way, we should probably have the units printed on the axes, [something like this](https://github.com/tardis-sn/tardis/blob/master/tardis/montecarlo/spectrum.py#L105).

jaladh-singhal (approved):
Thanks!

jaladh-singhal:
> My only concern here is do we allow users to use different units when plotting? And either way, we should probably have the units printed on the axes, [something like this](https://github.com/tardis-sn/tardis/blob/master/tardis/montecarlo/spectrum.py#L105).

I think that we infer unit to be shown on axes from quantity. If not, then that's case for another issue/PR @smithis7 @atharva-2001 - it's out of scope for this PR.

smithis7 (approved):
Looks good

PR #1834, opened 2021-11-30
Adding Luminosity Integral Explanation to Supernova Configuration

bartnikm:
**Description**
Added explanations for luminosity_wavelength_start and luminosity_wavelength_end parameters and made slight wording changes to the supernova configuration section of the documentation

**Motivation and context**
The original did not have an explanation for the luminosity integral limits being set to 0 and inf

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

https://bartnikm.github.io/tardis-bartnikm/branch/supernova_configuration/io/configuration/components/supernova.html
**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [x] None of the above. 

**Checklist**

### DIFF
--- a/docs/io/configuration/components/supernova.rst
+++ b/docs/io/configuration/components/supernova.rst
@@ -4,11 +4,11 @@
 Supernova Configuration
 ***********************
 
-The supernova component of the configuration file contains some key information about the supernova being modeled, namelt the time since the supernova and the luminosity the user wishes TARDIS should output:
+The supernova component of the configuration file contains some key information about the supernova being modeled, namely the time since the supernova and the luminosity the user wishes TARDIS should output:
 
 .. jsonschema:: schemas/supernova.yml
 
-As luminosity (in units of energy/s) is computed by integrating over the spectral luminosity (in units of energy/s/wavelength), TARDIS sums over all discrete energy packets to compute luminosity when ran, hence why we set `luminosity_wavelength_start` to 0 and `luminosity_wavelength_end` to infinity.
+As luminosity (in units of energy/s) is computed by integrating over the spectral luminosity (in units of energy/s/wavelength), TARDIS sums over all discrete energy packets to compute luminosity when ran, hence why `luminosity_wavelength_start` is set to 0 and `luminosity_wavelength_end` to infinity. It is recommended to not change these parameters unless desired.
 
 During a run of TARDIS, we attempt to converge the output spectrum to match the requested luminosity
 (see :ref:`est_and_conv`).
--- a/docs/io/configuration/components/supernova.rst
+++ b/docs/io/configuration/components/supernova.rst
@@ -4,13 +4,13 @@
 Supernova Configuration
 ***********************
 
-The supernova component of the configuration file contains some key information about the supernova being modeled, namely the time since the supernova and the luminosity the user wishes TARDIS should output:
+The supernova component of the configuration file contains some key information about the supernova being modeled, namely the time since the supernova (which is used throughout TARDIS calculations) and the luminosity the user wishes TARDIS should output:
 
 .. jsonschema:: schemas/supernova.yml
 
-As luminosity (in units of energy/s) is computed by integrating over the spectral luminosity (in units of energy/s/wavelength), TARDIS sums over all discrete energy packets to compute luminosity when ran, attempting to converge the output spectrum to match the luminosity requested (see :ref:`est_and_conv`). The range over which TARDIS sums these energy packets is set from 0 to infinity via the `luminosity_wavelength_start` and `luminosity_wavelength_end`, respectively, to generate a spectrum whose luminosity is known across the entire spectrum by default. However, if in the event only the luminosity within a certain range of wavelengths is known, then `luminosity_wavelength_start` and `luminosity_wavelength_end` can be changed as necessary to reflect this, allowing TARDIS to attempt to create a spectrum whose luminosity within the set range will converge to the value defined in `luminosity_requested`.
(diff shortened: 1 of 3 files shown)


smithis7 (commented):
Removing distance from the schema is causing tests to fail. We can try and fix those tests, or we can just remove that change. Probably the latter is better.

docs/io/configuration/components/supernova.rst
```diff
@@ -4,9 +4,11 @@
 Supernova Configuration
 ***********************
 
-The supernova component of the configuration file contains some key information about the supernova being modeled:
+The supernova component of the configuration file contains some key information about the supernova being modeled, namely the time since the supernova and the luminosity the user wishes TARDIS should output:
 
 .. jsonschema:: schemas/supernova.yml
 
+As luminosity (in units of energy/s) is computed by integrating over the spectral luminosity (in units of energy/s/wavelength), TARDIS sums over all discrete energy packets to compute luminosity when ran, hence why `luminosity_wavelength_start` is set to 0 and `luminosity_wavelength_end` to infinity. It is recommended to not change these parameters unless desired.
```
andrewfullard, line 11: I would say "required" rather than "desired" at the end

bartnikm: Good call!

bartnikm:
> Removing distance from the schema is causing tests to fail. We can try and fix those tests, or we can just remove that change. Probably the latter is better.

I think for now we can remove the change, but we should either look for uses for it or ways to fix the tests to remove it down the line.

smithis7 (commented):
What you wrote sounds good. Just one general thing that I think may be useful here. While it is true that we add up the luminosities between `luminosity_wavelength_start` and `luminosity_wavelength_end`, perhaps a better way to think about it in terms of configuring TARDIS is that the `luminosity_requested` is the luminosity of the supernova being modeled between `luminosity_wavelength_start` and `luminosity_wavelength_end`. If we use 0 and infinity, it is just the luminosity of the entire thing, but if the researcher only has, for example, the luminosity between 400 and 700 nm, they could insert that here.

bartnikm:
> What you wrote sounds good. Just one general thing that I think may be useful here. While it is true that we add up the luminosities between `luminosity_wavelength_start` and `luminosity_wavelength_end`, perhaps a better way to think about it in terms of configuring TARDIS is that the `luminosity_requested` is the luminosity of the supernova being modeled between `luminosity_wavelength_start` and `luminosity_wavelength_end`. If we use 0 and infinity, it is just the luminosity of the entire thing, but if the researcher only has, for example, the luminosity between 400 and 700 nm, they could insert that here.

Ah I was under the impression that the integration limits were constants that shouldn't be changed. I'll make a note of that, thanks!

docs/io/configuration/components/supernova.rst
```diff
@@ -4,9 +4,24 @@
 Supernova Configuration
 ***********************
 
-The supernova component of the configuration file contains some key information about the supernova being modeled:
+The supernova component of the configuration file contains some key information about the supernova being modeled, namely the time since the supernova and the luminosity the user wishes TARDIS should output:
```
smithis7, line 7: namely the time since the supernova explosion (which is used throughout the TARDIS calculations) and the luminosity....

docs/io/configuration/components/supernova.rst
```diff
@@ -4,9 +4,24 @@
...
 
 .. jsonschema:: schemas/supernova.yml
 
-During a run of TARDIS, we attempt to converge the output spectrum to match the requested luminosity
-(see :ref:`est_and_conv`).
\ No newline at end of file
+As luminosity (in units of energy/s) is computed by integrating over the spectral luminosity (in units of energy/s/wavelength), TARDIS sums over all discrete energy packets to compute luminosity when ran, attempting to converge the output spectrum to match the luminosity requested (see :ref:`est_and_conv`). The range over which TARDIS sums these energy packets is set from 0 to infinity via the `luminosity_wavelength_start` and `luminosity_wavelength_end`, respectively, to generate a spectrum whose luminosity is known across the entire spectrum by default. However, if in the event only the luminosity within a certain range of wavelengths is known, then `luminosity_wavelength_start` and `luminosity_wavelength_end` can be changed as necessary to reflect this, allowing TARDIS to attempt to create a spectrum whose luminosity within the set range will converge to the value defined in `luminosity_requested`.
```
smithis7, line 11: maybe also mention that luminosity can be given in standard units (such as erg/s or J/s) or logarithmic units such as log_lsun

docs/io/configuration/components/supernova.rst
```diff
@@ -4,9 +4,24 @@
...
+As an example, here is a sample code which will generate a specturm where only the luminosity of the visible light portion of the spectrum is given.
+
+.. code-block:: yaml
+        
+    supernova:
+      luminosity_requested: 9.44 log_lsun
+      time_explosion: 13 day
+      luminosity_wavelength_start: 400 nm
+      luminosity_wavelength_end: 700 nm
+
```
smithis7, line 22: In this example, the output spectrum will have a luminosity of approximately 10^{9.44}L_{sun} within the visible range.

docs/io/configuration/components/supernova.rst
```diff
@@ -4,9 +4,24 @@
...
+.. warning::
```
smithis7, line 23: I think this note is for the other PR (the spectrum one). I don't think we accept those inputs as frequency unless you tested it out

smithis7 (approved):
Awesome!

andrewfullard (approved):

Issue #1829, opened 2021-11-24
Kilonova-model crash when no recombination data available


gleck97:
**Describe the bug**
The bug appears when trying to run a (kilonova) model where _only_ elements with atomic number Z > 28 are abundant and when using "nebular" and "dilute-lte" as ionization and excitation for the plasma state. In this case, TARDIS does not seem to find any recombination data (which so far seems to be provided up to Z = 28, see "knox_long_recombination_zeta.dat") and the model crashes when trying to compute the plasma state.

**To Reproduce**
As mentioned above, run a model using
1. only elements above Z = 28 (atomic data is also needed for them)
2. ionization: nebular
3. excitation: dilute-lte

in the config file.

**Additional context**
I created a branch for myself to work on this and found a way to solve this issue (see https://github.com/gleck97/tardis/blob/master/tardis/plasma/properties/ion_population.py in lines 166-183), but I would be happy to receive feedback or comments to improve on it and bring it up to the standard of the code as much as possible.

chvogl:
Fixed with merge of #2150.

PR #1836, opened 2021-12-02
Fix for Issue "Kilonova-model crash when no recombination data available"

gleck97:
Fix for issue #1829 

**Description**
New version: In the case if there is no zeta_data, a data frame with the required dimension is created and filled with dummy values for the zeta data. 

**Motivation and context**
If no such data was provided, TARDIS crashed.

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/plasma/properties/ion_population.py
+++ b/tardis/plasma/properties/ion_population.py
@@ -164,23 +164,22 @@ class PhiSahaNebular(ProcessingPlasmaProperty):
     @staticmethod
     def get_zeta_values(zeta_data, ion_index, t_rad):
         """
-        OLD VERSION:
-        zeta_t_rad = zeta_data.columns.values.astype(np.float64)
-        zeta_values = zeta_data.loc[ion_index].values.astype(np.float64)
+        Attributes
+        ----------
+
         """
-        # SUGGESTION:
-        ion_index_in_zeta = ion_index.isin(zeta_data.index) # returns a boolean array, False means an index has not been found in the zeta data file
-        if False not in ion_index_in_zeta: 
+        ion_index_in_zeta = (
+            ion_index.isin(zeta_data.index)
+            if hasattr(ion_index, "isin")
+            else ion_index in zeta_data.index
+        )
+        if False not in ion_index_in_zeta:
             zeta_values = zeta_data.loc[ion_index].values.astype(np.float64)
             zeta_t_rad = zeta_data.columns.values.astype(np.float64)
-        else: # now zeta data found in the file -> zeta_data is empty
-            # for now: hard-coded dummy value, may better be added as a config parameter (-> how?)
-            dummy_zeta_val = 0.3
-            # for now, the following temp_values have to be included as zeta_data is empty -> is there a better solution?
-            temp_values = [2000.0,4000.0,6000.0,8000.0,10000.0,12000.0,14000.0,16000.0,18000.0,20000.0] 
-            zeta_data_copy = pd.DataFrame(dummy_zeta_val,columns=temp_values,index=ion_index)
-            zeta_values = zeta_data_copy.values.astype(np.float64)
-            zeta_t_rad = zeta_data_copy.columns.values.astype(np.float64)
+        else:
+            dummy_zeta_val = 1.0
+            zeta_t_rad = np.linspace(2000, 40000, 10, endpoint=True)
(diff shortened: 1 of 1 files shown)


tardis/plasma/properties/ion_population.py
```diff
@@ -163,8 +163,24 @@ def calculate(
 
     @staticmethod
     def get_zeta_values(zeta_data, ion_index, t_rad):
+        """
```
Rodot-, line 166: The docstring should adhere to the tardis documentation standards.  We use PEP 8 as the standard for our docstrings.

tardis/plasma/properties/ion_population.py
```diff
@@ -163,8 +163,24 @@ def calculate(
...
+        OLD VERSION:
         zeta_t_rad = zeta_data.columns.values.astype(np.float64)
         zeta_values = zeta_data.loc[ion_index].values.astype(np.float64)
+        """
+        # SUGGESTION:
+        ion_index_in_zeta = ion_index.isin(zeta_data.index) # returns a boolean array, False means an index has not been found in the zeta data file
```
Rodot-, line 172: This is currently throwing an error in the tests, please check it out.

tardis/plasma/properties/ion_population.py
Rodot-: It should be more clear what the data types of the arguments are since this will help resolve the below comment.

tardis/plasma/properties/ion_population.py
Rodot-, line 172:
```suggestion
        ion_index_in_zeta = ion_index.isin(zeta_data.index) if hasattr(ion_index, 'isin') else ion_index in zeta_data.index # returns a boolean array, False means an index has not been found in the zeta data file
```

tardis/plasma/properties/ion_population.py
afloers: Have a look at https://tardis-sn.github.io/tardis/development/code_quality.html#docstrings for the docstrings. You can also use https://github.com/psf/black to automatically format your code.

tardis/plasma/properties/ion_population.py
```diff
@@ -163,8 +163,24 @@ def calculate(
 
     @staticmethod
     def get_zeta_values(zeta_data, ion_index, t_rad):
+        """
+        OLD VERSION:
         zeta_t_rad = zeta_data.columns.values.astype(np.float64)
         zeta_values = zeta_data.loc[ion_index].values.astype(np.float64)
+        """
+        # SUGGESTION:
```
afloers, line 171:
```suggestion
```
this can be removed

tardis/plasma/properties/ion_population.py
```diff
@@ -163,8 +163,24 @@ def calculate(
...
+            zeta_values = zeta_data.loc[ion_index].values.astype(np.float64)
+            zeta_t_rad = zeta_data.columns.values.astype(np.float64)
+        else: # now zeta data found in the file -> zeta_data is empty
+            # for now: hard-coded dummy value, may better be added as a config parameter (-> how?)
+            dummy_zeta_val = 0.3
+            # for now, the following temp_values have to be included as zeta_data is empty -> is there a better solution?
+            temp_values = [2000.0,4000.0,6000.0,8000.0,10000.0,12000.0,14000.0,16000.0,18000.0,20000.0] 
+            zeta_data_copy = pd.DataFrame(dummy_zeta_val,columns=temp_values,index=ion_index)
+            zeta_values = zeta_data_copy.values.astype(np.float64)
+            zeta_t_rad = zeta_data_copy.columns.values.astype(np.float64)
```
afloers, line 183:
```suggestion
            zeta_t_rad = np.linspace(2000,20000, 10, endpoint=True)
            zeta_values = np.full(zeta_t_rad.shape, dummy_zeta_val).astype(np.float64)
```
would this work? I don't think the `zeta_data_copy` dataframe is used for anything else, so we can skip this step and use numpy methods instead.

gleck97:
Hi! Thank you both for your commands! At long last I've just integrated and committed them to my local master. Feel free to have a look and comment if you notice anything else that I should change! Should also be in PEP8 for which I used black in VSCode.

Issue #1819, opened 2021-10-23
test_logging really slow


wkerzendorf:

DhruvSondhi:
Hello @wkerzendorf, can you please elaborate a little more on what the issue is?

PR #1850, opened 2022-01-11
Speed Up Logging Tests

atharva-2001:
**Description**

**Motivation and context**

This might help address issue https://github.com/tardis-sn/tardis/issues/1819 

**How has this been tested?**
- [X] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [X] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/logger/tests/test_logging.py
+++ b/tardis/io/logger/tests/test_logging.py
@@ -7,17 +7,34 @@ from tardis.io.logger.logger import LOGGING_LEVELS
 from tardis import run_tardis
 
 
-def test_logging_simulation(atomic_data_fname, caplog):
+@pytest.fixture(scope="module")
+def config_verysimple(config_verysimple, atomic_data_fname):
+    """
+    Further simplify the `config_verysimple` fixture for testing logging.
+    """
+    config_verysimple.montecarlo.iterations = 2
+    config_verysimple.montecarlo.no_of_packets = 400
+    config_verysimple.montecarlo.last_no_of_packets = -1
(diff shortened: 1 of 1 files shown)


tardis/io/logger/tests/test_logging.py
```diff
@@ -7,17 +7,34 @@
 from tardis import run_tardis
 
 
-def test_logging_simulation(atomic_data_fname, caplog):
+@pytest.fixture(scope="session")
```
wkerzendorf, line 10: so I wonder if the fixtures can live more centrally?

atharva-2001: Yes, we do have fixtures in multiple `confest.py` files, which I assume are resued. (This fixture does too, I just wanted to modify it to make it even simpler)

tardis/io/logger/tests/test_logging.py
```diff
@@ -7,17 +7,34 @@
...
 
-    simulation = Simulation.from_config(config)
 
+def test_logging_simulation(caplog, config_verysimple, atomic_data_fname):
+    """
+    Testing the logs for simulations runs
+    """
+    config_verysimple["atom_data"] = atomic_data_fname
+    simulation = Simulation.from_config(config_verysimple)
     simulation.run()
```
atharva-2001, line 38: Just to make sure this is done, this code can be replaced by the `simulation_verysimple` fixture in the main `conftest.py` file.

PR #1847, opened 2021-12-23
Add a ModelState class

atharva-2001:
**Description**

**Motivation and context**

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

- [ ] My change requires a change to the documentation.
    - [ ] I have updated the documentation accordingly.
    - [ ] (optional) I have built the documentation on my fork following [the instructions](https://tardis-sn.github.io/tardis/development/documentation_preview.html).
- [ ] I have assigned and requested two reviewers for this pull request.

PR #1870, opened 2022-01-24
Add Abundance to ModelState class

atharva-2001:
In continuation of PR https://github.com/tardis-sn/tardis/pull/1847 

**Description**

**Motivation and context**

**How has this been tested?**
- [ ] Testing pipeline.
- [ ] Other. 

**Examples**

**Type of change**

- [ ] Bug fix. 
- [X] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -385,6 +385,9 @@ class Radial1DModel(HDFWriterMixin):
         return self.model_state.density
 
     def calculate_abundance(self):
+        """
+        Calculate model abundance.
+        """
         if not self.raw_isotope_abundance.empty:
             self._abundance = self.raw_isotope_abundance.decay(
                 self.time_explosion
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -93,6 +93,11 @@ class ModelState:
         """Outer radius of model shells."""
         return self.geometry.r_outer.values * self.geometry_units["r_outer"]
 
+    @property
+    def elemental_abundance(self):
+        """Elemental Abundance"""
+        return self.isotope_abundance.groupby(level=0).sum()
+
 
 class Radial1DModel(HDFWriterMixin):
     """
@@ -190,16 +195,13 @@ class Radial1DModel(HDFWriterMixin):
             )[self.v_boundary_inner_index + 1 : self.v_boundary_outer_index + 1]
         )
 
-        isotope_abundance = abundance
-        if not self.raw_isotope_abundance.empty:
-            isotope_abundance = self.raw_isotope_abundance.decay(
-                self.time_explosion
-            ).merge(self.raw_abundance)
+        raw_abundance = self.raw_abundance.copy()
+        raw_abundance["mass_number"] = -1
+        raw_abundance = raw_abundance.set_index(
+            [raw_abundance.index, "mass_number"]
+        )
+        isotope_abundance = raw_abundance.append(self.raw_isotope_abundance)
 
-        isotope_abundance = isotope_abundance.loc[
-            :, self.v_boundary_inner_index : self.v_boundary_outer_index - 1
-        ]
-        isotope_abundance.columns = range(len(isotope_abundance.columns))
         self.model_state = ModelState(
             v_inner=v_inner,
             v_outer=v_outer,
@@ -207,7 +209,7 @@ class Radial1DModel(HDFWriterMixin):
             r_outer=self.time_explosion * v_outer,
             time_explosion=self.time_explosion,
             density=density,
-            isotope_abundance=isotope_abundance
+            isotope_abundance=isotope_abundance,
         )
 
         if t_inner is None:
@@ -704,10 +706,10 @@ class Radial1DModel(HDFWriterMixin):
             ), "CSVY field descriptions exist without corresponding csv data"
             if unsupported_columns != set():
                 logger.warning(
-                "The following columns are "
-                "specified in the csvy model file,"
-                 f" but are IGNORED by TARDIS: {str(unsupported_columns)}"
-            )
+                    "The following columns are "
+                    "specified in the csvy model file,"
+                    f" but are IGNORED by TARDIS: {str(unsupported_columns)}"
+                )
 
         time_explosion = config.supernova.time_explosion.cgs
 
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -186,7 +186,7 @@ class Radial1DModel(HDFWriterMixin):
         self.time_explosion = time_explosion
         self._electron_densities = electron_densities
         self.raw_abundance = self._abundance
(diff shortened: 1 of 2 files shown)


tardis/model/base.py
```diff
@@ -285,38 +381,23 @@ def velocity(self):
...
-        if not self.raw_isotope_abundance.empty:
-            self._abundance = self.raw_isotope_abundance.decay(
-                self.time_explosion
-            ).merge(self.raw_abundance)
-        abundance = self._abundance.loc[
-            :, self.v_boundary_inner_index : self.v_boundary_outer_index - 1
-        ]
-        abundance.columns = range(len(abundance.columns))
-        return abundance
+        return self.model_state.abundance
```
atharva-2001, line 402: https://github.com/tardis-sn/tardis/blob/bf9b70833cfd260c651a1d52e83c37125f4ca52b/tardis/model/tests/test_base.py#L286-L288

This test failed in the last commit because `abundance` depends upon `raw_isotope_abundance`, and changing the latter does not change the former. 
One way to do this is creating another function for calculating abundance, and assigning the result to `model_state.abundance`. This is done in https://github.com/tardis-sn/tardis/pull/1870/commits/1254a1e7b6e6c4bf349a041d604fb30d72dfd3e5 , is it a good way to do this?

wkerzendorf: so - how about you get isotope abundances and then internally we calculate elemental abundances

atharva-2001: Well, we would also need to pass in `raw_abundance`, `v_boundary_inner_index` and `v_boundary_outer_index` too, because the abundance calculation depends on these things. Would that be alright?

wkerzendorf: no - we need to cut the isotope abundance beforehand.

wkerzendorf: we should never use `v_boundary_inner_index` in any of the model state

wkerzendorf: a good way to think about it - the modelstate essentially only stores. There are no cutting operations within the classs.

tardis/model/base.py
```diff
@@ -171,23 +185,30 @@ def __init__(
...
         density = (
             self.homologous_density.calculate_density_at_time_of_simulation(
                 self.time_explosion
             )[self.v_boundary_inner_index + 1 : self.v_boundary_outer_index + 1]
         )
+
+        raw_abundance = self.raw_abundance.copy()
+        raw_abundance["mass_number"] = -1
+        raw_abundance = raw_abundance.set_index([raw_abundance.index, "mass_number"])
+        isotope_abundance = raw_abundance.append(self.raw_isotope_abundance)
```
atharva-2001, line 201: This should not be decayed, should it?

```suggestion
        isotope_abundance = raw_abundance.append(self.raw_isotope_abundance.decay(500))
```

tardis/model/tests/test_base.py
```diff
@@ -377,6 +378,28 @@ def test_density(self):
...
+        raw_abundance = raw_abundance.set_index(
+            [raw_abundance.index, "mass_number"]
+        )
+        expected_isotope_abundance = raw_abundance.append(
+            self.model.raw_isotope_abundance.decay(self.model.time_explosion)
+        )
+
+        pd.testing.assert_frame_equal(
+            self.model.model_state.isotope_abundance, expected_isotope_abundance
+        )
```
atharva-2001, line 401: Should we be testing this against saved data?

tardis/model/base.py
```diff
@@ -376,16 +393,43 @@ def density(self):
...
+
+    @property
+    def raw_isotope_abundance(self):
+        return self._raw_isotope_abundance
+
+    @raw_isotope_abundance.setter
+    def raw_isotope_abundance(self, value):
+        self._raw_isotope_abundance = value
+        self.model_state.isotope_abundance = self.isotope_abundance
+
```
atharva-2001, line 429: In [this](https://github.com/tardis-sn/tardis/blob/c70594674b4e0998f3ea8a349911c1096e877331/tardis/model/tests/test_base.py#L283) test, `raw_isotope_abundance` is redefined. The test used to fail because the change was not reflected in `model_state`'s `isotope_abundance` and correspondingly in the `abundance` property. 

I thought defining a setter method would be a good idea to make sure `model_state`'s `isotope_abundance` is updated as well.

tardis/model/base.py
```diff
@@ -376,16 +393,43 @@ def density(self):
...
+        raw_abundance.columns = pd.RangeIndex(start=0, stop=len(raw_abundance.columns), step=1)
+
+        if not self._raw_isotope_abundance.empty:
+            self._abundance = raw_abundance.append(
+                self._raw_isotope_abundance.decay(self.time_explosion)
+            )
+            norm_factor = self._abundance.sum(axis=0)
+            self._abundance /= norm_factor
+        else:
+            self._abundance = raw_abundance
```
atharva-2001, line 412: This is a small replacement for the merge method [here](https://github.com/tardis-sn/tardis/blob/c70594674b4e0998f3ea8a349911c1096e877331/tardis/io/decay.py#L112).
I could clean it up if this method is alright.

tardis/model/base.py
```diff
@@ -376,16 +393,43 @@ def density(self):
...
+        return isotope_abundance
+
+    @property
+    def raw_isotope_abundance(self):
+        return self._raw_isotope_abundance
+
+    @raw_isotope_abundance.setter
+    def raw_isotope_abundance(self, value):
+        self._raw_isotope_abundance = value
+        self.model_state.isotope_abundance = self.isotope_abundance
```
atharva-2001, line 428: In [this](https://github.com/tardis-sn/tardis/blob/c70594674b4e0998f3ea8a349911c1096e877331/tardis/model/tests/test_base.py#L283) test, `raw_isotope_abundance` is redefined. The test used to fail because the change was not reflected in `model_state`'s `isotope_abundance` and correspondingly in the `abundance` property.

I thought defining a setter method would be a good idea to make sure that `model_state`'s `isotope_abundance` is updated as well.

tardis/model/base.py
```diff
@@ -84,6 +93,11 @@ def r_outer(self):
         """Outer radius of model shells."""
         return self.geometry.r_outer.values * self.geometry_units["r_outer"]
 
+    @property
+    def elemental_abundance(self):
+        """Elemental Abundance"""
+        return self.isotope_abundance.groupby(level=0).sum()
```
wkerzendorf, line 99:
```suggestion
        # the dataframe has both mass and proton number and 
        # we are grouping by proton number to get the elemental abundances
        return self.isotope_abundance.groupby(level=0).sum()
```

wkerzendorf:
Closing this in favor of @code-explorer

PR #1874, opened 2022-01-25
move transport related parts to r_packet_transport

wkerzendorf:
from r_packet.py (which is only for definitions)


tardis/montecarlo/montecarlo_numba/tests/test_packet.py
```diff
@@ -6,6 +6,7 @@
 import tardis.montecarlo.montecarlo_numba.calculate_distances as calculate_distances
 import tardis.montecarlo.montecarlo_numba.frame_transformations as frame_transformations
 import tardis.montecarlo.montecarlo_numba.opacities as opacities
+import tardis.montecarlo.montecarlo_numba.r_packet_transport
```
andrewfullard, line 9: I would change this to add `as r_packet_transport`

tardis/montecarlo/tests/test_montecarlo.py
```diff
@@ -50,6 +50,7 @@
 import pandas as pd
 import tardis.montecarlo.montecarlo_numba.formal_integral as formal_integral
 import tardis.montecarlo.montecarlo_numba.r_packet as r_packet
+import tardis.montecarlo.montecarlo_numba.r_packet_transport
```
andrewfullard, line 53: Same as above

Rodot- (approved):

Issue #1871, opened 2022-01-24
Changelog workflow is broken


andrewfullard:
**Describe the bug**

Changelog workflow is refusing to push, claiming that it hasn't got a branch checked out.

**To Reproduce**

Run changelog workflow

**Screenshots**

![image](https://user-images.githubusercontent.com/20341051/150811378-71cbe504-4baa-40b2-9ec6-790d5d2ecf7f.png)

madoyi:
I'd like to help out with this issue, is there a specific implementation you had in mind ?

rohithvarma3000:
Hey, I have been exploring this issue for a day. https://github.com/actions/checkout/issues/124#issuecomment-584264556 This comment explains the reason why this issue arises. I think a push to a PR related branch is also considered as a trigger. So we need to do some study regarding when this workflow should be called.

The simplest way to fix the issue will be through a `git checkout` and then performing the push.

Also I had a suggestion. I think we need to setup a conditional statement to prevent the workflow from running in forks of the repository. This is because the fork will not have the `BOT_TOKEN` secret and the workflow may give an error.

I would love to work on this issue as well.

rohithvarma3000:
I have a made a mistake in my previous comment. The workflow is being triggered correctly and the PR is not causing the error. The detached head occurs due to the way git handles tags. When we make a commit over the tag, the commit is not attached to any branch. The tag still points to the previous commit. So we need to modify the tag accordingly to handle this change.

epassaro:
Replaced by the pre-release workflow. 

Closing.

PR #1878, opened 2022-01-26
Fix changelog action workflow

rohithvarma3000:
Modify the git flow in the changelog action workflow

**Description**
This PR aims to solve the detached head issue raised by github actions when the tag triggers the workflow. The action first deletes the tag in the runner. Then the tag is recreated to point to the new commit. The change is then force pushed to origin.

**Motivation and context**
Fix for issue #1871 

**How has this been tested?**
- [ ] Testing pipeline.
- [x] Other. 
Tested by pushing tags to my fork

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -6,7 +6,7 @@ name: changelog
 on:
   push:
     tags:
-    - 'release-*'
+      - 'release-*'
 
   workflow_dispatch:  # Manual trigger
 
@@ -35,5 +35,9 @@ jobs:
       - name: Push changes
         run: |
           git add CHANGELOG.md
-          git commit -m "Post-release updates [auto]"
-          git push
+          git commit --amend --no-edit
+          GIT_TAG_MESSAGE=$(git tag -l --format='%(contents)' ${{ github.ref_name }})
+          git tag -d ${{ github.ref_name }}
+          git tag -a ${{ github.ref_name }} -m "$GIT_TAG_MESSAGE"
+          git checkout master
+          git push --tags --force


.github/workflows/update-changelog.yml
```diff
@@ -35,5 +35,9 @@ jobs:
...
         run: |
           git add CHANGELOG.md
-          git commit -m "Post-release updates [auto]"
-          git push
+          git commit --amend --no-edit
+          GIT_TAG_MESSAGE=$(git tag -l --format='%(contents)' ${{ github.ref_name }})
+          git tag -d ${{ github.ref_name }}
+          git tag -a ${{ github.ref_name }} -m "$GIT_TAG_MESSAGE"
+          git checkout master
+          git push --tags --force
```
andrewfullard, line 43: Why is a force push needed?

rohithvarma3000: The tag that we intend to rewrite will exist in the remote. The lack of a force push will give the error `already exists`. So to overwrite the tag we need a force push

epassaro:
This issue is related to the @tardis-bot token permissions and the fact our `master` branch is write protected.

rohithvarma3000:
@epassaro I am not sure about the permissions that the tardis bot has. But I don't think that it is alone responsible. I have made a testing branch in my fork to test this workflow. There I gave the default `GITHUB_TOKEN` to the workflow. Even in that case the workflow gives the `detached head` issue when it is triggered automatically by the tag condition. I have also observed in the `tardis-sn/tardis` repository that the changelog workflow always gave this error except when it was manually trigger or scheduled.

Upon enquire into the issue, I figured that the issue occurs due to the nature of how the checkout action works. When we consider a tag based trigger, the checkout action will perform the `git checkout <tag-name>` command and it does not checkout the branch as usual. So when we try to add a commit, it does not get appended to the branch. Instead it enters the `detached head` state. [This](https://www.cloudbees.com/blog/git-detached-head) is a good article that explains the problem.

So to deal with this, I have used this approach to modify and force push the tag to the repository. If you think we need to append to a branch as usual, then we need to figure out a different way to automatically trigger the workflow.

rohithvarma3000:
Also I have updates the PR to use better git tag commands. I had used an unnecessary complicated approach before to deal with the tags.

epassaro:
Hi @rohithvarma3000, thanks for your PR.

We're going to change the way we push changes from CI/CD to the repository and soon this action would not be needed anymore.

PR #1883, opened 2022-01-31
Fix line list sorting

chvogl:
This PR sorts the line list by frequency instead of wavelength. This guarantees that the line list is in the correct order even if there are small inconsistencies in the atomic data.

**Motivation and context**
The atomic data from the Boyle 2017 paper crashes the code because there are differences of up to 4e-6 between the stored frequencies and the frequencies calculated from the wavelengths. As a result, the frequency can also increase along the line list (when sorting by wavelength) resulting in a nu_diff < 0 error.

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/io/atom_data/base.py
+++ b/tardis/io/atom_data/base.py
@@ -347,7 +347,7 @@ class AtomData(object):
             )
         ]
 
-        self.lines.sort_values(by="wavelength", inplace=True)
+        self.lines.sort_values(by="nu", ascending=False, inplace=True)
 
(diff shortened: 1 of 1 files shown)


chvogl:
The tests fail because sorting by frequency does not necessarily lead to the same order as sorting by wavelength for lines that have the same frequency.

PR #1886, opened 2022-02-01
Continuum branch [attempted merge with master]

wkerzendorf:


Rodot-:
What's up with all the C-code still in here?  Is that just left over from when we started the project?

wkerzendorf:
fails - some numba issue
[tmp_pytest.txt](https://github.com/tardis-sn/tardis/files/8128099/tmp_pytest.txt)

PR #1901, opened 2022-02-21
Fix PyTest fixtures and updated mainloop to properly compile

Rodot-:
added in a special fixture for the montecarlo_1e5 model tests as to a void mixing config fixtures.  Set the montecarlo_main_loop to recompile itself on the first iteration ot set proper config options

Test for numba 1e5 model was failing when run on it's own but not when run as part of the test suite.  This was caused by the single packet seed setting incorrectly setting every packet to the same seed, which was removed.  This indicated a larger problem where the montecarlo main loop was not properly being updated with the config changes after subsequent runs, so a change was added to recompile the main loop on the first iteration of each tardis run. 

**Motivation and context**
Tests failing inconsistently

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/base.py
+++ b/tardis/montecarlo/montecarlo_numba/base.py
@@ -63,7 +63,7 @@ def montecarlo_radial1d(
     packet_seeds = montecarlo_configuration.packet_seeds
 
     number_of_vpackets = montecarlo_configuration.number_of_vpackets
-
+    if iteration == 0: montecarlo_main_loop.recompile() # Make sure we update
     (
         v_packets_energy_hist,
         last_interaction_type,
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -138,7 +138,7 @@ def tardis_ref_data(tardis_ref_path, generate_reference):
         yield store
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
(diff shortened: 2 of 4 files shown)


andrewfullard (approved):

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -63,7 +63,7 @@ def montecarlo_radial1d(
     packet_seeds = montecarlo_configuration.packet_seeds
 
     number_of_vpackets = montecarlo_configuration.number_of_vpackets
-
+    if iteration == 0: montecarlo_main_loop.recompile() # Make sure we update
```
wkerzendorf, line 66: this is dangerous and only needed for testing. Is that something we should have a config option on?

Rodot-: I don't think this is dangerous, I think this is necessary for tardis to work properly if config options change between runs in the same kernel

Rodot-: This is really just an initialization step

PR #1942, opened 2022-04-01
STELLA to TARDIS parser

wkerzendorf:
See title


andrewfullard:
Solved more recently with other changes.

PR #1949, opened 2022-04-12
First Objective for "Monte Carlo 2D Packet Visualization"

jayantbhakar:
Here I have completed the first objective for "Monte Carlo 2D Packet Visualization". I have added the jupyter notebook file as a commit and added the figure obtained for single packet tracking in the description.

**Description**
Graph of mu vs r for a single packet:
Screenshot from 2022-04-12 07-55-01


Rodot-:
Can you update this to set the number of iterations to 1?  Since you are only running a single packet, running more than one iteration won't really help you solve the plasma state.  Perhaps you should add comments as well explaining why you set the config options the way you did.  I also see virtual packets being run in there and I'm not sure why that is.

It would also be useful to show the shells on the plot so we can see when the packet crosses the boundaries.

jayantbhakar:
I will update it as soon as possible. Thanks for your review.

jayantbhakar:
Also, I have a doubt, what should I set the number of virtual packets to? Since I have read in the documentation that virtual packets are used to reduce noise in the spectrum generation. And virtual packets are generated whenever a real packet is initially emitted from the photosphere and second whenever it interacts with the plasma. So I thought having virtual packets more than 1 would be fine. And thats why i thought of choosing a number greater than 1 and less than 5. Am I getting it right, I am not sure about it. Can you tell me about how many virtual packets should I use?

jayantbhakar:
I have updated the notebook with the following changes:

- set iterations to 1
- set virtual packets to 0
- added comments to the code wherever description was required
- plotted the shell boundaries in the graph

The final plot looks like this-

Screenshot from 2022-04-14 21-48-34

andrewfullard:
For this particular project, virtual packets are not relevant because they follow simple paths that are not useful to visualize.

Rodot-:
> Also, I have a doubt, what should I set the number of virtual packets to? Since I have read in the documentation that virtual packets are used to reduce noise in the spectrum generation. And virtual packets are generated whenever a real packet is initially emitted from the photosphere and second whenever it interacts with the plasma. So I thought having virtual packets more than 1 would be fine. And thats why i thought of choosing a number greater than 1 and less than 5. Am I getting it right, I am not sure about it. Can you tell me about how many virtual packets should I use?

So, you are right that the virtual packets help decrease the montecarlo noise, but each packet is essentially just a single montecarlo sample.  For this project we're just tracking the trajectory of an individual packet, so it doesn't matter what the bulk behavior is or the level of noise.  Since you're just tracking a single rpacket you only need to run a single rpacket and nothing else.  Good question though!

jayantbhakar:
Got that. And I have set the number of virtual packets to 0 in the updated file. So that's fine I guess.

PR #1992, opened 2022-04-29
Fix low temperature problem of Van Regemorter approximation

chvogl:
The exponential in the calculation of the van Regemorter approximation can become infinite, e.g. for 100 K and transition energies bigger than roughly 6.5 eV:
`np.exp(((6.5 * units.eV) / (100 * units.K * const.k_B)).to(''))`
The product of the exponential integral E1 and the exponential, however, stays finite. I have added an implementation of 
the product that does not calculate the exponential for large values but instead uses the Laurent series expansion.

**How has this been tested?**
- [x] Testing pipeline.
- [ ] Other. 

**Type of change**

- [x] Bug fix. 
- [ ] New feature. 
- [ ] Breaking change. 
- [ ] None of the above. 

**Checklist**

### DIFF
--- a/tardis/plasma/properties/atomic.py
+++ b/tardis/plasma/properties/atomic.py
@@ -3,7 +3,7 @@ import logging
 import numpy as np
 import pandas as pd
 from numba import njit
-from scipy.special import expn
+from scipy.special import exp1
 from scipy.interpolate import PchipInterpolator
 from collections import Counter as counter
 import radioactivedecay as rd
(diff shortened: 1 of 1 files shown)


epassaro (approved):

Rodot- (changes requested):
Please provide new tests for these class methods to verify and guarantee their behavior.  The test should verify that 500 is a good upper limit for the scipy/numpy calculation.

Rodot- (approved):

PR #2013, opened 2022-05-07
Replaced the pandas .ix method with loc/iloc

jayantbhakar:
In the analysis.py file, I have replaced the depreciated pandas .ix function with its new .loc and .iloc functions.


epassaro:
Please follow the TARDIS pull request template.

jayantbhakar:
I have added the mailmap, I think it was causing some issues earlier.

PR #2020, opened 2022-05-10
Fixes for CITATION.cff

epassaro:
**Description**

- Wait for the Zenodo webhook, otherwise `CITATION.cff` is not modified.
- `doi2cff update` does not update the `title` field and author order, use `doi2cff init` instead

**Motivation and context**

**How has this been tested?**

- [ ] Testing pipeline
- [ ] Other

**Examples**

**Type of change**

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] None of the above 

**Checklist**

- [ ] I have updated the documentation according to my changes
- [ ] I have built the documentation by applying the `build_docs` label to this pull request (if you don't have enough privileges a reviewer will do it for you)
- [x] I have requested two reviewers for this pull request

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -87,9 +87,14 @@ jobs:
           title: Post-release ${{ env.DATE }}
           body: |
             *\*beep\* \*bop\**
-            
+
             Hi, human.
-            
+
             These are the changes required after the latest TARDIS release.
+
+            > :warning: **Warning:** 
+            >
+            > The `post_release` workflow waits for a webhook and then updates `CITATION.cff`. \
+            If the file is not changed by this pull request, check the latest release on [Zenodo's website](https://zenodo.org/record/592480), close this pull request, and run the [run the workflow manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) again.
           labels: documentation,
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yml
```diff
@@ -36,6 +36,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Wait for Zenodo webhook
+        run: sleep 120
```
andrewfullard, line 40: Hardcoded sleep seems like an issue- is there a way to link the action as a wait instead?

epassaro: Nope. That's out of our control, depends on how much time requires Zenodo to make a new release on their server. 2 mins seems enough.

andrewfullard: What happens if this wait time is exceeded?

epassaro: A pull request is opened just with changes in CHANGELOG.rst, CITATION.cff will not updated because the release was the same at that moment the job started.

epassaro: This is done, the warning looks like this: https://github.com/tardis-sn/tardis/pull/2018

andrewfullard (approved):

PR #2029, opened 2022-05-18
Create Dockerfile and docker workflow

epassaro:
**Description**

- Adds a `Dockerfile` with a fully functioning TARDIS.
- Adds a workflow to build and push Docker images of TARDIS repo with the labels: `master` (latest changes, gets overwritten on every push to master), `<tag-name>` (on release), and `latest` (points to latest tag, on release).

**Motivation and context**

Run TARDIS everywhere.

**How has this been tested?**

- [ ] Testing pipeline
- [x] Other

Locally build and run the Docker image:

```
docker build . -t tardis-rt/tardis:latest
docker run -p 8888:8888 tardis-rt/tardis:latest
```

To re-start an exited container, first check the `NAMES` field of your container with `docker ps -a` and then

```
docker start -ia <container_name>
```

**Examples**

**Type of change**

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] None of the above 

**Checklist**

- [ ] I have updated the documentation according to my changes
- [ ] I have built the documentation by applying the `build_docs` label to this pull request (if you don't have enough privileges a reviewer will do it for you)
- [x] I have requested two reviewers for this pull request


.github/workflows/docker.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Extract metadata
+        uses: docker/metadata-action@v4
+        with:
+          images: tardis-rt/tardis
```
epassaro, line 35: I assumed this new account is going to be named `tardis-rt`.

epassaro:
> I don't know very much about Docker, I'll try this out when I get the time

Great, tell me if you have any doubt.

Do you know someone in TARDIS that knows about Docker and should be a reviewer of this PR?

DhruvSondhi (approved):

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN conda run -n tardis pip install /tmp/repo \
+    && echo "conda activate tardis" >> ~/.bashrc
+
+RUN mkdir /workdir
+COPY docs/quickstart.ipynb /workdir
+WORKDIR /workdir
+
+RUN conda clean --all \
+    && apt-get autoremove --purge -y \
+    && rm -rf /var/lib/apt/lists/* \
```
jaladh-singhal, line 18: Just curious, why are we cleaning apt lists?

jaladh-singhal (dismissed):
This is a very great addition @epassaro :tada: 

I tried running it locally - it works perfectly except I had to install curl otherwise quickstart notebook wasn't able to download tardis_example.yml. Perhaps, we can add following step in dockerfile:
```Dockerfile
RUN apt-get update && apt-get install -y \
curl
```

But I understand if we don't since quickstart notebook isn't the only thing container users are gonna use. As you know better, it's your call whether we should add this or not, hence I'm approving it!

epassaro:
> This is a very great addition @epassaro tada
> 
> I tried running it locally - it works perfectly except I had to install curl otherwise quickstart notebook wasn't able to download tardis_example.yml. Perhaps, we can add following step in dockerfile:
> 
> ```dockerfile
> RUN apt-get update && apt-get install -y \
> curl
> ```
> 
> But I understand if we don't since quickstart notebook isn't the only thing container users are gonna use. As you know better, it's your call whether we should add this or not, hence I'm approving it!

That's why I'm changing the `curl` command to `wget` in #2042. `wget` is always present in minimal Docker images and VMs while `curl` is usually not. Installing `curl` will bloat the image with ~500MB, and since the Quickstart is there just to show TARDIS is fully functional I don't think it's worth it.

Dockerfile
epassaro: From [Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/):

> In addition, when you clean up the apt cache by removing /var/lib/apt/lists it reduces the image size, since the apt cache is not stored in a layer. Since the RUN statement starts with apt-get update, the package cache is always refreshed prior to apt-get install.

DhruvSondhi (changes requested):

jaladh-singhal:
> > This is a very great addition @epassaro tada
> > I tried running it locally - it works perfectly except I had to install curl otherwise quickstart notebook wasn't able to download tardis_example.yml. Perhaps, we can add following step in dockerfile:
> > ```dockerfile
> > RUN apt-get update && apt-get install -y \
> > curl
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But I understand if we don't since quickstart notebook isn't the only thing container users are gonna use. As you know better, it's your call whether we should add this or not, hence I'm approving it!
> 
> That's why I'm changing the `curl` command to `wget` in #2042. `wget` is always present in minimal Docker images and VMs while `curl` is usually not. Installing `curl` will bloat the image with ~500MB, and since the Quickstart is there just to show TARDIS is fully functional I don't think it's worth it.

Sounds good @epassaro!

Knights-Templars:
Closed in favor of PR #2569.

PR #2052, opened 2022-06-13
Created a smart HDF Reader which can re-produce simulation object

### LINKED ISSUE #1251: Create a smart HDF Reader which can re-produce simulation object
Currently simulation objects are usually saved to disk as hdf files. But current implementation of HDF file writer (`to_hdf`) don't save units and only selected properties are saved. And whenever that hdf is read, the interfaces using it have to create a separate hdf classmethod to extract information from it and then convert to quantities (to get data in same format as a simulation object gives). 

It will be great if we can have a HDF Reader which handles all of this hassle, by creating a simulation object (or something very close) from the hdf file. This means we also need to enhance capabilities of current HDF writer so that the new HDF reader can get all information from HDF it wants.

This requires an in-depth analysis of how simulation object is created by a tardis run and how it is saved as hdf file, so as to convert the saved hdf file back to simulation object.

satwik-kambham:
### :pencil: Description

**Type:** :rocket: `feature`

Feature to allow storing and loading simulation objects from HDF files.
- Adds interfaces for storing and loading simulation, runner, plasma and model objects.

closes #1251 

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/io/model_reader.py
+++ b/tardis/io/model_reader.py
@@ -1,5 +1,9 @@
 # reading different model files
 
+from tardis.io.config_reader import ConfigurationNameSpace
+from tardis.montecarlo.base import MontecarloRunner
+from tardis.util.base import parse_quantity, is_valid_nuclide_or_elem
+
 import warnings
 import numpy as np
 from numpy import recfromtxt, genfromtxt
@@ -8,15 +12,11 @@ from astropy import units as u
 from radioactivedecay import Nuclide
 from radioactivedecay.utils import Z_DICT, elem_to_Z
 import h5py
-from tardis.io.config_reader import ConfigurationNameSpace
-from tardis.montecarlo.base import MontecarloRunner
 
 import logging
 
 # Adding logging support
 logger = logging.getLogger(__name__)
-
-from tardis.util.base import parse_quantity, is_valid_nuclide_or_elem
 
 
 class ConfigurationError(Exception):


Rodot-:
Just as a reminder from our meeting today and so others see it.  In addition to making the model to and from hdf functions, there should be a set of tests that make sure these functions work.  The test should create a simulation, run for a single iteration, export the runner and model (two separate tests) to hdf then delete them from the simulation.  Load up the model and runner from the hdfs, add them back to the simulation, and run for another iteration.  Verify the results are the same as running for two iterations (This might be tricky since there is some information about the random number generator state that needs to be preserved).

Additionally, make sure to add your name and email to the mailmap file!

tardis/io/model_reader.py
```diff
@@ -533,3 +536,394 @@ def parse_csv_abundances(csvy_data):
...
+        "v_boundary_inner_cgs": model.v_boundary_inner,
+        "v_boundary_outer_cgs": model.v_boundary_outer,
+        "w": model.w,
+        "t_rad_cgs": model.t_rad,
+        "r_inner_cgs": model.r_inner,
+        "density_cgs": model.density,
+    }
+
+    for key, value in model_dict.items():
+        if key.endswith("_cgs"):
```
Rodot-, line 825: Maybe instead check if `value` has a `unit` attribute

andrewfullard:
Very good work! codecov thinks you need to add more tests, but I'm not sure if that will change much.

tardis/io/model_reader.py
```diff
@@ -7,6 +7,9 @@
 from astropy import units as u
 from radioactivedecay import Nuclide
 from radioactivedecay.utils import Z_DICT, elem_to_Z
+import h5py
```
AlexHls, line 10: Import statements could be reordered to match [PEP 8 ordering](https://peps.python.org/pep-0008/#imports), but this won't affect functionallity

AlexHls (approved):
Looks great!

tardis/io/model_reader.py
```diff
@@ -533,3 +536,394 @@ def parse_csv_abundances(csvy_data):
...
+        "volume_cgs": runner.volume,
+    }
+
+    try:
+        runner_dict["single_packet_seed"] = runner.single_packet_seed
+    except AttributeError:
+        runner_dict["single_packet_seed"] = None
+
+    for key, value in runner_dict.items():
+        if key.endswith("_cgs"):
```
Rodot-, line 608: I'm a little confused here, I was under the impression that variable names ending in `_cgs` were unitless arrays and the `_cgs` implied that the units of the array was cgs.  Are there any attributes that trigger this conditional?

Additionally, it might be useful to just ignore all units and just save everything as unitless arrays in cgs units.  The reader will have to know which units correspond to the arrays when read in but this can be hard coded with something like a dictionary mapping.  I think avoiding saving the units as well would make the directory structure a bit simpler and more in line with the way that the simulation hdf is currently output.

tardis/io/model_reader.py
```diff
@@ -533,3 +536,394 @@ def parse_csv_abundances(csvy_data):
...
+            runner_dict[key] = [value.cgs.value, value.unit.to_string()]
+
+    integrator_settings = runner.integrator_settings.items()
```
Rodot-, line 611: just keep in mind that `.items()` is an iterator.  Might be good to turn these into a list or dictionary directly.

Rodot- (approved):

PR #2053, opened 2022-06-13
Change developer installation command

epassaro:
### :pencil: Description

Change developer mode installation command in docs. The `pip install -e .` is equivalent to `python setup.py develop` but solves and installs packages defined in package configuration files. This could lead to installing packages via `pip` to the local user directory (for example `astropy` in `~.local`) if the `tardis` environment is not activated.

There's nothing wrong with `pip install -e .` but could cause issues to new users.

- [ ] Testing pipeline
- [ ] Other method (describe)


rohithvarma3000 (approved):
Looks good to me.

PR #2082, opened 2022-07-01
Added documetation on how to compare environment before update

jaladh-singhal:
Added a note in "Environment update" section in installation docs on how to compare environment against env file. This is useful for checking if it needs update.


epassaro:
This PR contained erroneous information @wkerzendorf. 

Sorry for not giving a review earlier.

PR #2092, opened 2022-07-14
Arepo 3d grid 

AlexHls:
### :pencil: Description

**Type:** :rocket: `feature` | :roller_coaster: `infrastructure`

Prototype for creating Cartesian 3D models from Arepo snapshots. Proposes a CSVY file format containing a list of grid (see attached file)
[3d_csvy_proposal.csvy.txt](https://github.com/tardis-sn/tardis/files/9181062/3d_csvy_proposal.csvy.txt)

A few notes on the 3D CSVY file:

- The cell coordinates refer to the corner pointing into the negative direction in each axis
- The header contains additional information compared to the 1D CSVY files:

1. `model_photosphere_radius` : The radius of the photospheric sphere. Only cells outside of this radius should be considered belonging to the photosphere
2. `model_cellsize` : Width of the (cubic) cells. _This information could also be inferred from the cell coordinates, so this could technically be removed_

- [ ] Testing pipeline
- [ ] Other method (describe)
- [x] My changes can't be tested (automated testing would introduce proprietary arepo-snap-util package as dependency which is not possible. The relevant function from said package is written in C and rewriting it in Python is out of scope.) 


tardis/io/parsers/arepo.py
```diff
@@ -744,6 +741,219 @@ def create_profile(
...
+        numthreads : int
+            Number of threads used in mapping procedure.
+
+        Returns
+        -------
+        profile : 3DProfile object
+
+        """
+
+        print("Mapping velocities to 3D grid...")
```
andrewfullard, line 789: Do we expect this to take long enough that print statements are necessary to track progress?

docs/io/configuration/components/models/converters/arepo_to_tardis.ipynb
```diff
@@ -390,6 +390,46 @@
     "\n",
```
andrewfullard, line 390: <p>...although <strong>they</strong> <strong>cannot currently</strong> be used... ...<strong>functionality</strong>...</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/2092/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

tardis/io/parsers/arepo.py
AlexHls: Yes, depending on the chosen resolution, the mapping procedure can take >10 minutes (I guess if you try to map a snapshot with 55 species to a 200 grid over an hour is still realistic)

PR #2099, opened 2022-07-18
Arbitrary isotope decay handling and positronium

andrewfullard:
### :pencil: Description

**Type:** :rocket: `feature`

These changes allow for an arbitrary selection of radioactive isotopes to be added to an energy deposition calculation. The isotopes must be part of a `nuclear` database, which this code reads in. Also included some refactoring of the packet initialization process.

Positronium continuum handling is also included, supporting 2 and 3 photon decay processes with a controlling fraction.

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/energy_input/energy_source.py
```diff
@@ -350,3 +253,11 @@ def read_artis_lines(isotope, path_to_data):
         sep="  ",
         index_col=False,
     )
+
+
+def get_nuclear_lines_database(
+    path="~/Downloads/tardisnuclear/decay_radiation.h5",
```
afloers, line 259: does the decay radiation data base always live in `~/Downloads/tardisnuclear`? this seems dangerous

tardis/energy_input/indivisible_packets.py
```diff
@@ -295,7 +361,7 @@ def main_gamma_ray_loop(
     photoabsorption_opacity="tardis",
     pair_creation_opacity="tardis",
     seed=1,
-    path_to_artis_lines="",
+    path_to_decay_data="",
```
afloers, line 364: is `path_to_decay_data` the file that is hardcoded in `tardis/energy_input/energy_source.py` line 259?

tardis/energy_input/energy_source.py
andrewfullard: It's a kwarg because it's the standard location from `nuclear`, but the user can override

tardis/energy_input/indivisible_packets.py
andrewfullard: Yes, the user can override the path with this kwarg

tardis/energy_input/indivisible_packets.py
```diff
@@ -157,123 +261,85 @@ def initialize_packets(
...
-                GXPacketStatus.IN_PROCESS,
-                k,
-                decay_time,
-            )
+            if isotope_name in parents:
+                tau_end = taus[parents[isotope_name]]
+            else:
+                tau_end = 0
+
+            for c in range(count):
```
wkerzendorf, line 301: rename `count` to something more descriptive

tardis/energy_input/energy_source.py
Rodot-, line 259: Is this general for everyone?  Does tardisnuclear always install here?

tardis/energy_input/indivisible_packets.py
```diff
@@ -75,53 +79,153 @@ def sample_energy(energy, intensity):
...
     """Samples the decay time from the mean half-life
     of the isotopes (needs restructuring for more isotopes)
 
     Parameters
     ----------
-    isotope : string
-        Isotope as e.g. Ni56
-    taus : dict
-        Mean half-life for each isotope
+    start_tau : float64
```
Rodot-, line 90: If these are half-lifes perhaps it better to call them something like t_12 or t_halflife since tau is sometimes used for the decay e-folding time

tardis/energy_input/energy_source.py
andrewfullard: Yes, but I have moved the path value to the main function as a kwarg so it is clearer for the user

tardis/energy_input/indivisible_packets.py
andrewfullard: They are actually the e-folding time i.e. the mean lifetime, so I have clarified.

PR #2103, opened 2022-07-22
New geometry class

code-explorer:
### :pencil: Description

**Type:** ::rocket: `restructure` 

New geometry class for model state restructure.

### :pushpin: Resources

Model State Plan


tardis/model/base.py
```diff
@@ -84,6 +84,26 @@ def r_outer(self):
         return self.geometry.r_outer.values * self.geometry_units["r_outer"]
 
 
+class Geometry:
+    """
+    Holds information about model geometry
+
+    Parameters
+    ----------
+    radius : astropy.units.quantity.Quantity
```
Rodot-, line 93: Give this the properties `r_inner`, `r_outer`, `v_inner`, and `v_outer` as numpy arrays containing units

tardis/model/base.py
```diff
@@ -84,6 +84,26 @@ def r_outer(self):
         return self.geometry.r_outer.values * self.geometry_units["r_outer"]
 
 
+class Geometry:
```
Rodot-, line 87: For now, just make a class called `Radial1DGeometry` which will eventually inherit from a Geometry parent class.  The `Radial1DGeometry` should contain the attributes needed to run tardis with a radial 1D geometry

tardis/model/base.py
Rodot-: Make sure it has methods to compute volumes in each shell

Rodot- (changes requested):

tardis/model/base.py
```diff
@@ -84,6 +84,47 @@ def r_outer(self):
         return self.geometry.r_outer.values * self.geometry_units["r_outer"]
 
 
+class Radial1DGeometry:
+    """
+    Holds information about model geometry
```
andrewfullard, line 89: Be specific about the geometry type

tardis/model/base.py
```diff
@@ -84,6 +84,47 @@ def r_outer(self):
...
+
+    Parameters
+    ----------
+    radius : astropy.units.quantity.Quantity
+    velocity : astropy.units.quantity.Quantity
+    """
+
+    def __init__(self, radius, velocity):
+        self.geometry_df = pd.DataFrame(
+            {"radius": radius, "velocity": velocity}
```
andrewfullard, line 99: Dataframes can't read astropy quantities, so they need to be explicitly set to CGS units with the `.to_cgs` method and then converted to a pure numpy array with the `.value` method.

andrewfullard (changes requested):

tardis/model/base.py
```diff
@@ -84,6 +84,47 @@ def r_outer(self):
...
+    Holds information about model geometry
+
+    Parameters
+    ----------
+    radius : astropy.units.quantity.Quantity
+    velocity : astropy.units.quantity.Quantity
+    """
+
+    def __init__(self, radius, velocity):
+        self.geometry_df = pd.DataFrame(
```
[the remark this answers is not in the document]
andrewfullard: To be clear, we have decided to cut the dataframe step. Then the class can just have `self.radius` and `self.velocity`. The properties can remain, they are just created without the dataframe. `self.radius` and `self.velocity` can keep their astropy quantities as a result, you can just use `to(u.cm)` for example.

tardis/model/base.py
```diff
@@ -84,6 +84,41 @@ def r_outer(self):
...
+    """
+    Holds information about model geometry for radial 1D models.
+
+    Parameters
+    ----------
+    radius : astropy.units.quantity.Quantity
+    velocity : astropy.units.quantity.Quantity
+    """
+
+    def __init__(self, radius, velocity):
```
wkerzendorf, line 97: the call signature should be `def __init__(self, r_inner, r_outer, v_inner, v_outer):`

andrewfullard (approved):

tardis/model/base.py
```diff
@@ -84,6 +84,27 @@ def r_outer(self):
         return self.geometry.r_outer.values * self.geometry_units["r_outer"]
 
 
+class Radial1DGeometry:
+    """
+    Holds information about model geometry for radial 1D models.
+
+    Parameters
+    ----------
+    radius : astropy.units.quantity.Quantity
```
Rodot-, line 93: Make sure the description of the doc-string matches the parameters of the __init__

AlexHls (dismissed):

andrewfullard (approved):

Rodot- (dismissed):

wkerzendorf (approved):

andrewfullard (approved):

AlexHls (approved):

Rodot- (approved):

PR #2105, opened 2022-07-22
Documentation for Features In-Progress

smithis7:
### :pencil: Description

**Type:** :memo: `documentation`

Currently, the development roadmap section of the TARDIS documentation is highly outdated. In addition to updating it, this PR makes a place in the documentation for placing documentation for features not yet implemented into TARDIS, as this has been demonstrated to be necessary given all the new developments going on in the collaboration.

This is in accordance with discussions at TARDIS-con.

Note: I think I may be missing something about Hydrogen NLTE. Please let me know if this is the case, and if so what to put for it.

### :pushpin: Resources

See the new page: https://tardis-sn.github.io/tardis/pull/2105/contributing/in_progress/index.html

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [x] Other method (describe)
- [ ] My changes can't be tested (explain why)

Docs built on GitHub.

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [x] I updated the documentation according to my changes
- [x] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #2107, opened 2022-07-27
Added positronium notes

andrewfullard:
### :pencil: Description

Added some documentation for positronium discussion. Probably won't stay under the "energy input" header, but it's the best place to keep a record.

### :pushpin: Resources

See https://en.wikipedia.org/wiki/Positronium 
Positronium spectrum https://ui.adsabs.harvard.edu/abs/1949PhRv...75.1696O 
SN Ia commentary (section 2.2.2) https://ui.adsabs.harvard.edu/abs/2004ApJ...613.1101M/abstract

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/docs/physics/energy_input/positronium.rst
+++ b/docs/physics/energy_input/positronium.rst
@@ -1,0 +1,23 @@
+**************************
+Positronium
+**************************
+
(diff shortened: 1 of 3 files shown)


docs/physics/energy_input/positronium.rst
```diff
@@ -0,0 +1,23 @@
+**************************
```
smithis7, line 1: line of *** is too long, should be ``***********`` on top and bottom (one * for each character in the title, otherwise sphinx complains).

isaacgsmith (changes requested):
Can we hold off on this until #2105 is merged? That will give us a good location for both the gammaray deposition and the positronium docs.

PR #2114, opened 2022-08-10
Experimental model state

code-explorer:
### :pencil: Description

**Type:** :rocket: `restructure`

Restructuring model state with new composition and geometry classes.

### :pushpin: Resources

Restructure plan

### :vertical_traffic_light: Testing

How did you test these changes?

- [x] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

PR #2118, opened 2022-08-17
Integrating experimental model state

code-explorer:
### :pencil: Description

**Type:** : :rocket: `restructure`

Replacing old model state with restructured model state from #2114  .

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/model/base.py
```diff
@@ -21,6 +24,127 @@
 
 logger = logging.getLogger(__name__)
 
+STABLE_ISOTOPE_MASS_NUMBER = {
```
wkerzendorf, line 27: don't mix code and data.

tardis/model/base.py
```diff
@@ -21,6 +24,127 @@
...
+    "Ds": 281,
+    "Rg": 272,
+    "Cn": 285,
+    "Nh": 286,
+    "Fl": 289,
+    "Mc": 289,
+    "Lv": 293,
+    "Ts": 294,
+    "Og": 294,
+}
```
andrewfullard, line 146: Is this hardcoded information necessary? Can you recover it from the radioactivedecay module?

tardis/model/tests/test_base.py
```diff
@@ -425,20 +387,43 @@ def test_radial_1D_geometry_volume(simulation_verysimple, index, expected):
...
+    [
+        ((8, 0), 5.4470099e53),
+        ((16, 10), 5.0367073e52),
+        ((20, 19), 2.0231745e51),
+    ],
+)
+def test_model_state_number(simple_model_state, index, expected):
+    model_state = simple_model_state
+
+    assert_almost_equal((model_state.number).loc[index], expected, decimal=-47)
```
andrewfullard, line 424: Why are the decimal parameters negative? I am probably just forgetting the syntax :)

tardis/model/base.py
```diff
@@ -145,6 +269,40 @@ def elemental_number_density(self):
         )
 
 
+class ModelState_Experimental:
```
andrewfullard, line 272: If it is implemented into the standard Radial1DModel, perhaps this is no longer experimental? @wkerzendorf

tardis/model/tests/test_base.py
code-explorer: The comparisons are being made between numbers in the range 1e50, a position decimal parameter signifies how many digits after the decimal point we need to look at whereas a negative parameter signifies how many digits before the decimal we need to ignore.

andrewfullard: Okay. I thought that e.g. 1.0124e50 should be compared with decimal=4?

tardis/model/base.py
```diff
@@ -119,13 +118,15 @@ class Composition:
...
     atomic_mass_unit: astropy.units.Unit
 
     Attributes
     ----------
-    number_density : pd.DataFrame
-        Number density of each isotope in each shell.
+    atomic_mass : pd.DataFrame
+        Atomic mass of elements calculated for each shell.
+    elemental_number_density : pd.DataFrame
+        Number density of each element in each shell.
```
AlexHls, line 129: Is it really elements we're storing here or individual isotopes?

AlexHls (approved):

andrewfullard (approved):

tardis/model/base.py
code-explorer: The mass of the individual isotopes are being added up based on their abundances(or mass fractions) and so we are storing the properties for the elements here.

PR #2120, opened 2022-08-22
Update arepo docs notebook to be pre-run

AlexHls:
### :pencil: Description

**Type:** :beetle: `bugfix` | :memo: `documentation`

Updates the arepo-parser documentation to use a pre-run jupyter notebook to avoid downloading reference files on each build. Additionally the documentation now uses the actual snapshot now since notebook is pre-run anyway, i.e. arepo-snap-utils is required to pre-run the notebook.

- [x] Testing pipeline
- [ ] Other method (describe)


andrewfullard (approved):

smithis7 (changes requested):
I am not a huge fan of the pre-running idea, especially since now it cannot be interactively run as it uses the actual snapshot. If we do go with this though, the few things I would add to this PR:
- In `.gitignore` make sure all generated files are included, and make sure any files no longer generated are removed.
- Make sure all files needed to run the notebook (assuming you have the correct packages installed) like the snapshot are included in the repo so the notebook can be updated if needed.

Edit: I also didn't know that the notebook was switched to being pre-run -- I think I missed that PR. As mentioned before I think it's best to clear output.

AlexHls:
> I am not a huge fan of the pre-running idea, especially since now it cannot be interactively run as it uses the actual snapshot. If we do go with this though, the few things I would add to this PR:
> 
>     * In `.gitignore` make sure all generated files are included, and make sure any files no longer generated are removed.
> 
>     * Make sure all files needed to run the notebook (assuming you have the correct packages installed) like the snapshot are included in the repo so the notebook can be updated when needed.
> 
> 
> Edit: I also didn't know that the notebook was switched to being pre-run -- I think I missed that PR. As mentioned before I think it's best to clear output.

As discussed in the slack, the whole reason to set this notebook to being pre-run is to avoid having to download the huge snapshot files from the LFS every time the docs have to be built. If this is no longer desired, this PR could be closed.
I don't see any other solution to the problem other than having the notebook being pre-run in order to avoid the file download from the LFS (it probably is even less desired to include a several 100MB file in the main repo just to build the docs), but if you have a better solution @smithis7 , let me know.
Also I don't really see people running this notebook interactively unless they are already familiar with Arepo, in which case they most certainly have their own snapshot. The whole JSON thing was a hack to begin with to avoid the arepo-snap-util dependency, and is an edge case at the very best. To generate the JSON file, you will most likely have a valid snapshot anyway.

@andrewfullard what do you think?

smithis7:
> > I am not a huge fan of the pre-running idea, especially since now it cannot be interactively run as it uses the actual snapshot. If we do go with this though, the few things I would add to this PR:
> > ```
> > * In `.gitignore` make sure all generated files are included, and make sure any files no longer generated are removed.
> > 
> > * Make sure all files needed to run the notebook (assuming you have the correct packages installed) like the snapshot are included in the repo so the notebook can be updated when needed.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Edit: I also didn't know that the notebook was switched to being pre-run -- I think I missed that PR. As mentioned before I think it's best to clear output.
> 
> As discussed in the slack, the whole reason to set this notebook to being pre-run is to avoid having to download the huge snapshot files from the LFS every time the docs have to be built. If this is no longer desired, this PR could be closed. I don't see any other solution to the problem other than having the notebook being pre-run in order to avoid the file download from the LFS (it probably is even less desired to include a several 100MB file in the main repo just to build the docs), but if you have a better solution @smithis7 , let me know. Also I don't really see people running this notebook interactively unless they are already familiar with Arepo, in which case they most certainly have their own snapshot. The whole JSON thing was a hack to begin with to avoid the arepo-snap-util dependency, and is an edge case at the very best. To generate the JSON file, you will most likely have a valid snapshot anyway.
> 
> @andrewfullard what do you think?

If that's the issue then I understand, then I would just still remove unnecessary files from the `.gitignore`. Perhaps also adding a note that the notebook can't be run interactively without the proper files.

smithis7 (approved):

andrewfullard (approved):

andrewfullard:
It's fine by me

wkerzendorf (approved):

PR #2132, opened 2022-10-03
Adding rate matrix index

sonachitchyan:
### :pencil: Description

**Type:**  `infrastructure`

This PR addds the NLTEIndexHelper class which has the output `rate_matrix_index`. This will be used for constructing the rate equation for solving NLTE ionization and NLTE excitation. Index helper returns MultiIndex with atomic number, ion number and whether the species is treated in lte or nlte_ionization. If none of those are the case, it keeps the level number as the third index for further usage in NLTE excitation treatment.

- [x] Testing pipeline
- [ ] Other method (describe)


wkerzendorf:
@sonachitchyan can you update the ddescription of the PR

wkerzendorf:
@sonachitchyan also adding tests that just load the infrastructure (not even solving it successfully would be great)

tardis/plasma/properties/rate_matrix_index.py
```diff
@@ -0,0 +1,42 @@
...
+                    self.nlte_ionization_species,
+                    nlte_excitation_species,
+                )
+            ),
+            names=levels.names,
+        ).drop_duplicates()
+        return rate_matrix_index
+
+    def calculate_rate_matrix_index(
+        self, levels, nlte_ionization_species, nlte_excitation_species
```
andrewfullard, line 31: If you already have `self` you should be able to use `self.nlte_ionization_species` in this method

tardis/plasma/standard_plasmas.py
```diff
@@ -28,6 +29,7 @@
     adiabatic_cooling_properties,
     two_photon_properties,
     isotope_properties,
+    nlte_properties_new,
```
andrewfullard, line 32: Why the `_new` in the name?

sonachitchyan: it was just a temporary solution by me and Christian. We just need a way of differentiating between this and numerical nlte that already exists in the code.

tardis/plasma/properties/rate_matrix_index.py
sonachitchyan: done

tardis/plasma/standard_plasmas.py
wkerzendorf: this is confusing. We should find a good workaround

wkerzendorf:
Very nice PR. I think the only thing left is the `_new` thing.

tardis/plasma/standard_plasmas.py
sonachitchyan: I can name it `nlte_solver_properties`

andrewfullard (approved):

andrewfullard (approved):

PR #2134, opened 2022-10-11
Replacing NumbaModel with NumbaRadial1DGeometry

code-explorer:
### :pencil: Description

**Type:**  :rocket: `restructure`

Replacing NumbaModel with NumbaRadial1DGeometry.

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/numba_interface.py
+++ b/tardis/montecarlo/montecarlo_numba/numba_interface.py
@@ -13,6 +13,32 @@ from tardis.montecarlo import (
 
 C_SPEED_OF_LIGHT = const.c.to("cm/s").value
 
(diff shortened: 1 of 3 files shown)


tardis/montecarlo/montecarlo_numba/numba_interface.py
```diff
@@ -13,6 +13,32 @@
...
 
+numba_geometry_spec = [
+    ("r_inner", float64[:]),
+    ("r_outer", float64[:]),
+    ("v_inner", float64[:]),
+    ("v_outer", float64[:]),
+]
+
+
+@jitclass(numba_geometry_spec)
```
wkerzendorf, line 24: just use `jitclass` on the Geometry class in model one.

PR #2143, opened 2022-10-31
Refactoring mass fraction

satwik-kambham:
### :pencil: Description

**Type:** :rocket: `restructure`

The initial mass fraction that will be read in will just be in `IsotopeMassFraction` dataframe which contains 
`A, Z, mass_fraction`
for “stable” elements A = -1


andrewfullard:
@wkerzendorf we still want this I think

PR #2144, opened 2022-11-03
Moving mass fraction and composition

satwik-kambham:
### :pencil: Description

**Type:**  :rocket: `restructure`

Moving composition and IsotopeAbundance to new composition sub-package and refactoring IsotopeAbundance into IsotopeMassFraction


andrewfullard:
@wkerzendorf do we want to use this?

PR #2150, opened 2022-11-09
Kilonova missing zeta

gleck97:
### :pencil: Description

This PR improves upon the suggestions from PR 1875 which I have closed.

**Type:** :beetle: `bugfix` 

The original problem from PR 1875 was: In case one runs a simulation with elements of atomic number > 28 only (as for kilonovae e.g.), in NLTE mode TARDIS would crash as it does not find any recombination (zeta) data. If one included at least one element with Z <= 28, the simulation would not crash.

Proposed changes here to fix the issue: In tardis/plasma/properties/atomic.py, just add "and keys" in the if condition in line 635. Creating an additional filter for zeta data turned out not to be necessary.
This if statement fills up missing zeta data with dummy values in the "else" condition if elements > 28 are included. However, this "else" condition was not reached if ONLY Z > 28 elements are provided as the list called "keys" is empty in that case. To force TARDIS to enter the "else" condition anyway, "and keys" checks if "keys" is not empty. If it is empty, the "else" condition is entered and executed, filling up the zeta dataframe with dummy values for all required ions.

What could be done in addition is to move all zeta data massaging which is done in ion_population.py, namely in get_zeta_values from the PhiSahaNebular class, to atomic.py. This would make the code even more consistent at this particular point. Andreas suggested I could do this as well as I have just checked out the code at this point.

### :pushpin: Resources

Running a simulation is enough, printing / checking with debugger the resulting dataframe for zeta data showed that the changes work fine.

- [ ] Testing pipeline
- [x ] Other method (describe)

- [ x] I requested two reviewers for this pull request

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -70,6 +70,9 @@ Frederik Beaujean <beaujean@mpp.mpg.de> Frederik Beaujean <schoenerhans16@gmx.de
 Gaurav Gautam <gautam1168@gmail.com>
 Gaurav Gautam <gautam1168@gmail.com> gautam1168 <gautam1168@gmail.com>
 
+Gerrit Leck <gerritleck@web.de>
+Gerrit Leck <gerritleck@web.de> Gerrit Leck <g.leck@gsi.de>
+
 Isaac Smith <smithis7@msu.edu> 
 Isaac Smith <smithis7@msu.edu> Isaac Smith <71480393+smithis7@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


afloers (changes requested):
@gleck97 please add your name and email to the .mailmap file in the main folder of TARDIS.

afloers (approved):

andrewfullard (approved):

PR #2159, opened 2022-11-18
Fix relativistic packet initialization

chvogl:
### :pencil: Description

**Type:** :beetle: `bugfix`

TARDIS currently does not initialize packets correctly for highly relativistic flows. The reason (found by @ssim) is that the inner boundary, on which we initialize the packets, is not comoving with the material. As a result, we need to adjust the distribution of comoving angles and energies.
(I will add a more detailed description later.)

With the changes in this PR, the flux from the TARDIS inner boundary matches the theoretical expectation for the emission of a highly relativistic sharp photosphere (beta = 0.2) as derived by @ssim:
relativistic_flux_theory_tardis_comparison
Also, TARDIS now produces W=0.5 directly at the inner boundary even for highly relativistic velocities as expected for zero limb darkening in the comoving frame.

- [ ] Testing pipeline
- [X] Other method (check against Stuart's prediction, w=0.5)

### DIFF
--- a/tardis/montecarlo/base.py
+++ b/tardis/montecarlo/base.py
@@ -106,7 +106,12 @@ class MontecarloRunner(HDFWriterMixin):
 
         self.seed = seed
         if packet_source is None:
-            self.packet_source = source.BlackBodySimpleSource(seed)
+            if not enable_full_relativity:
+                self.packet_source = source.BlackBodySimpleSource(seed)
+            else:
+                self.packet_source = source.BlackBodySimpleSourceRelativistic(
+                    seed
+                )
         else:
             self.packet_source = packet_source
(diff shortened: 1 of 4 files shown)


ssim (approved):
Well this looks like an excellent result to me - thanks for trying this out, Christian - and I'm actually very pleased / relieved to see that this makes the W=0.5 make sense again!?

chvogl:
@andrewfullard @wkerzendorf could one of you have a quick glance at the PR? I don't think this has to be a detailed review. For now, some high-level feedback on what is missing (e.g. more detailed explanations) would be sufficient.

andrewfullard (approved):
Looks good to me!

PR #2194, opened 2023-01-23
Cache LFS objects in the tests workflow

atharva-2001:
### :pencil: Description

Write a complete description of your changes, including the necessary context or any piece of information required to understand your work.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,6 +68,7 @@ jobs:
       - name: Git LFS Pull
         run: git lfs pull
         working-directory: tardis-refdata
+        if: steps.lfs-cache.outputs.cache-hit != 'true'
 
       - name: Setup environment
         uses: conda-incubator/setup-miniconda@v2
@@ -91,7 +92,7 @@ jobs:
         run: pip install -e .
 
       - name: Run tests
-        run: pytest tardis -x ${{ env.PYTEST_FLAGS }}
+        run: pytest tardis ${{ env.PYTEST_FLAGS }}
 
       - name: Upload to Codecov
         run: bash <(curl -s https://codecov.io/bash)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,6 +52,7 @@ jobs:
         with:
           repository: tardis-sn/tardis-refdata
           path: tardis-refdata
+          lfs: false
 
       - name: Create LFS file list
         run: git lfs ls-files -l | cut -d' ' -f1 | sort > .lfs-assets-id
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,9 +47,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # - name: Download reference data
-      #   run: bash .ci-helpers/download_reference_data.sh
-
       - name: Clone tardis-sn/tardis-refdata
         uses: actions/checkout@v2
         with:
@@ -70,11 +67,6 @@ jobs:
       - name: Git LFS Pull
         run: git lfs pull
         working-directory: tardis-refdata
-
-      - name: see file size
-        run: ls -lrtsh -a
-        working-directory: tardis-refdata
-    
 
       - name: Setup environment
         uses: conda-incubator/setup-miniconda@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,6 +69,11 @@ jobs:
         run: git lfs pull
         working-directory: tardis-refdata
         if: steps.lfs-cache.outputs.cache-hit != 'true'
+      
+      - name: Git LFS Checkout
+        run: git lfs checkout
+        working-directory: tardis-refdata
+        if: steps.lfs-cache.outputs.cache-hit == 'true'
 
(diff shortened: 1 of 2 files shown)


sonachitchyan:
@atharva-2001 be sure to add your info into the `.mailmap`

atharva-2001:
The tests here are not failing, but the tests [here](https://github.com/atharva-2001/tardis/actions/runs/3995806498) are. I wonder why that is happening. However the files are being cached!

epassaro:
I don't think this is going to work as expected. 

You need to use a conditional step, like we do in this example:

```yaml
      - uses: actions/cache@v2
        with:
          path: ${{ matrix.prefix }}
          key: conda-${{ matrix.label }}-${{ hashFiles('conda-${{ matrix.label }}.lock') }}-${{ env.CACHE_NUMBER }}
        id: cache-conda

      - name: Update environment
        run: mamba update -n tardis --file conda-${{ matrix.label }}.lock
        if: steps.cache-conda.outputs.cache-hit != 'true'
```

In the example, the second step only runs if the cache exists. In your code, all the steps are running on every commit.

atharva-2001:
> I think this could be achieved by doing:
> 1. Run the cache step.
> 2. Clone the repo without the LFS objects (there's an option for that)
> 3. Get the LFS assets (I'm assuming this step is going to work independently of pulling the LFS objects, but I don't know)
> 4. Make a conditional step that pulls the LFS objects if the cache does not exist.

But this is what I'm doing here right? Here is what I think the workflow does: 
1. Clone the repo without LFS Objects
2. Generate a file(`.lfs-assets-id`) with all the hashes of the file names. This file is then used to generate the cache key. If any of these files change, then the key would be invalid.
3. If the cache key matches, the items would be restored and `git lfs pull` wouldn't do anything. If it fails, there would be nothing in the lfs folder and the items would be downloaded(that way we won't need a conditional statement.)

epassaro:
I think you are right @atharva-2001, but I think we should keep the `lfs: false` flag and the conditional statement explicit if the result is the same, just to be more clear in case we need to revisit this in the future.

andrewfullard:
> I think you are right @atharva-2001, but I think we should keep the `lfs: false` flag and the conditional statement explicit if the result is the same, just to be more clear in case we need to revisit this in the future.

Why lfs: false? Confusing to me.

atharva-2001:
> Why lfs: false? Confusing to me.

The LFS objects would be saved as cache and we wouldn't need to get them every time the workflow runs. We would only need to get LFS objects if the reference data repository is updated(because that would invalidate the cache key) and we are doing that using `git lfs pull`.

.github/workflows/tests.yml
```diff
@@ -72,7 +91,7 @@ jobs:
         run: pip install -e .
 
       - name: Run tests
-        run: pytest tardis ${{ env.PYTEST_FLAGS }}
+        run: pytest tardis -x ${{ env.PYTEST_FLAGS }}
```
chvogl, line 94: I don't think we should exit instantly on the first error or failed test. What do other people think?

.github/workflows/tests.yml
```diff
@@ -47,8 +47,27 @@ jobs:
...
+      - name: Restore LFS cache
+        uses: actions/cache@v2
+        id: lfs-cache
+        with:
+          path: tardis-refdata/.git/lfs
+          key: ${{ runner.os }}-lfs-${{ hashFiles('tardis-refdata/.lfs-assets-id') }}-v1
+
+      - name: Git LFS Pull
+        run: git lfs pull
+        working-directory: tardis-refdata
```
chvogl, line 70: I agree with Ezequiel that it would be more instructive to make this step conditional on the `cache-hit` even if it doesn't make a difference in practice.

.github/workflows/tests.yml
andrewfullard: Agreed and Atharva has solved that

andrewfullard (approved):

chvogl (approved):
Looks good to me!

chvogl:
@atharva-2001 Any idea why the tests are failing?

atharva-2001:
I'm afraid I'm yet to look at it. I'd soon fix this and let you know!

.github/workflows/tests.yml
```diff
@@ -47,8 +47,28 @@ jobs:
...
+        if: steps.lfs-cache.outputs.cache-hit != 'true'
```
chvogl, line 71: Apparently, the tests fail if we do not `git lfs pull`. Is it possible that we need to run `git lfs checkout` even if we recover the lfs stuff from the cache?

andrewfullard (approved):

.github/workflows/tests.yml
chvogl: Ok. This was the problem.

atharva-2001: Yes! Thank you for suggesting this. I did `git lfs checkout` in commit: https://github.com/tardis-sn/tardis/pull/2194/commits/8e68779e206ced452693cef26d5c22e64bcda896

chvogl (approved):
Looks great!

Issue #2191, opened 2023-01-19
The Tests Pipeline Is Failing on my fork


atharva-2001:
**Describe the bug**

The test pipeline is failing. 
To check:
- [Link](https://github.com/atharva-2001/tardis/actions/runs/3958861195) to the failed workflow.
- [Link](https://github.com/tardis-sn/tardis/compare/master...atharva-2001:test_branch?expand=1) to the branch.

**To Reproduce**

Run the tests workflow.

**Screenshots**

**System**
 - OS:
    - [X] GNU/Linux
    - [ ] macOS

 - Environment (`conda list`):

**Additional context**

epassaro:
The test pipeline is failing on your fork. That's because your master branch is 179 commits behind upstream.

atharva-2001:
But the branch on which the workflow failed is [synced](https://github.com/tardis-sn/tardis/compare/master...atharva-2001:test_branch?expand=1) with the master branch on the TARDIS repository, right?

epassaro:
I don't know, the link is not correct I guess. I assume you can see your branch but I don't.

epassaro:
Are working with a new atomic file?

MD5SUM and UUID1 signatures are no longer encoded in new Carsus atomic files, I think that's the problem.

epassaro:
The pipeline passed on the master branch.

chvogl:
I guess this is no longer relevant @atharva-2001? Closing for now.

PR #2195, opened 2023-01-24
Reading nlte_excitation_species from config

sonachitchyan:
### :pencil: Description

This PR makes sure that `nlte_excitation_species` is read from the config file and passed to the class that constructs the `rate_matrix_index` which will then be used for NLTE excitation treatment.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/io/tests/test_config_reader.py
+++ b/tardis/io/tests/test_config_reader.py
@@ -190,7 +190,7 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
 
     Parameter
     ---------
-        `tardis_config_verysimple` : YAML File
+        `tardis_config_verysimple_nlte` : YAML File
 
     Result
     ------
@@ -230,7 +230,7 @@ def test_plasma_nlte_exc_section_config(tardis_config_verysimple_nlte):
 
     Parameter
     ---------
-        `tardis_config_verysimple` : YAML File
+        `tardis_config_verysimple_nlte` : YAML File
 
     Result
     ------
--- a/tardis/io/tests/test_config_reader.py
+++ b/tardis/io/tests/test_config_reader.py
@@ -220,6 +220,44 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
                 raise ValueError("Nlte ionization species not in continuum.")
     assert ve.type is ValueError
 
+def test_plasma_nlte_exc_section_config(tardis_config_verysimple_nlte):
+    """
+    Configuration Validation Test for Plasma Section of the Tardis Config YAML File.
+
+    Validates:
+        nlte_excitation_species: should be included in continuum_interaction
+
+    Parameter
+    ---------
+        `tardis_config_verysimple` : YAML File
+
+    Result
+    ------
+        Assertion based on validation for specified values
+    """
+    conf = Configuration.from_config_dict(
+        tardis_config_verysimple_nlte, validate=True, config_dirname="test"
+    )
+    tardis_config_verysimple_nlte["plasma"]["continuum_interaction"][
+        "species"
+    ] = [
+        "He I",
+    ]
+    tardis_config_verysimple_nlte["plasma"]["nlte_excitation_species"] = ["H I"]
+    with pytest.raises(ValueError) as ve:
+        nlte_excitation_species = tardis_config_verysimple_nlte["plasma"][
+            "nlte_excitation_species"
+        ]
+
+        for species in nlte_excitation_species:
+            if not (
+                species
+                in tardis_config_verysimple_nlte["plasma"][
+                    "continuum_interaction"
+                ]["species"]
+            ):
+                raise ValueError("Nlte excitation species not in continuum.")
+    assert ve.type is ValueError
 
 def test_spectrum_section_config(tardis_config_verysimple):
     """
--- a/tardis/io/tests/test_config_reader.py
+++ b/tardis/io/tests/test_config_reader.py
@@ -210,11 +210,9 @@ def test_plasma_nlte_section_config(
         "He I",
     ]
     tardis_config_verysimple_nlte["plasma"]["nlte_ionization_species"] = ["H I"]
+    config = Configuration.from_config_dict(tardis_config_verysimple_nlte)
     with pytest.raises(PlasmaConfigError) as ve:
-        plasma = assemble_plasma(
-            tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data
-        )
-    assert ve.type is PlasmaConfigError
+        assemble_plasma(config, nlte_raw_model, nlte_atom_data)
 
 
 def test_plasma_nlte_exc_section_config(
@@ -240,11 +238,9 @@ def test_plasma_nlte_exc_section_config(
         "He I",
     ]
     tardis_config_verysimple_nlte["plasma"]["nlte_excitation_species"] = ["H I"]
-    with pytest.raises(PlasmaConfigError) as ve:
-        plasma = assemble_plasma(
-            tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data
-        )
-    assert ve.type is PlasmaConfigError
+    config = Configuration.from_config_dict(tardis_config_verysimple_nlte)
+    with pytest.raises(PlasmaConfigError):
+        plasma = assemble_plasma(config, nlte_raw_model, nlte_atom_data)
 
 
 def test_spectrum_section_config(tardis_config_verysimple):
--- a/tardis/io/tests/test_config_reader.py
+++ b/tardis/io/tests/test_config_reader.py
@@ -8,6 +8,8 @@ from jsonschema.exceptions import ValidationError
 
 from tardis.io import config_reader
 from tardis.io.config_reader import Configuration
+from tardis.plasma.exceptions import PlasmaConfigError
+from tardis.plasma.standard_plasmas import assemble_plasma
 
 
 def data_path(filename):
@@ -181,7 +183,7 @@ def test_plasma_section_config(tardis_config_verysimple):
     assert ve.type is ValueError
 
 
-def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
+def test_plasma_nlte_section_config(tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data,):
     """
     Configuration Validation Test for Plasma Section of the Tardis Config YAML File.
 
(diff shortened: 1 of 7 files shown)


tardis/io/tests/test_config_reader.py
```diff
@@ -221,6 +221,46 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
+def test_plasma_nlte_exc_section_config(tardis_config_verysimple_nlte):
+    """
+    Configuration Validation Test for Plasma Section of the Tardis Config YAML File.
+
+    Validates:
+        nlte_excitation_species: should be included in continuum_interaction
+
+    Parameter
+    ---------
+        `tardis_config_verysimple` : YAML File
```
andrewfullard, line 233: _nlte missing

sonachitchyan: fixed, thanks for noticing!

andrewfullard (dismissed):
Looks good, tests pass, I'm happy

tardis/io/tests/test_config_reader.py
```diff
@@ -221,6 +221,46 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
+
+        for species in nlte_excitation_species:
+            if not (
+                species
+                in tardis_config_verysimple_nlte["plasma"][
+                    "continuum_interaction"
+                ]["species"]
+            ):
+                raise ValueError("Nlte excitation species not in continuum.")
+    assert ve.type is ValueError
```
chvogl, line 261: I am very confused about what this test is supposed to do. You set values for the `continuum_interaction_species` and the `nlte_excitation_species` in the config by hand. Then you check that the `nlte_excitation_species` are not in the `continuum_interaction_species`  with a piece of code that only exists in this test. This check fails because the values were set correctly in the config and you raise a `ValueError` by hand. The test, which expects a `ValueError`,  then obviously passes. As far as I can tell, the only thing that is being tested (in a very convoluted way) is that the config values exist and can be set to the desired values. I know that the test if derived from already existing tests but they also do not seem to make much more sense. Maybe I am just missing something?

tardis/plasma/standard_plasmas.py
```diff
@@ -175,15 +180,33 @@ def assemble_plasma(config, model, atom_data=None):
...
-                if not (species in config.plasma.continuum_interaction.species):
-                    raise PlasmaConfigError(
-                        f"NLTE ionization species {species} not in continuum species."
-                    )
+        if (
+            config.plasma.nlte_ionization_species
+            or config.plasma.nlte_excitation_species
+        ):
+            nlte_ionization_species = []
+            nlte_excitation_species = []
```
chvogl, line 188: Why is this needed?

tardis/plasma/standard_plasmas.py
```diff
@@ -175,15 +180,33 @@ def assemble_plasma(config, model, atom_data=None):
...
-            nlte_ionization_species = config.plasma.nlte_ionization_species
-            for species in nlte_ionization_species:
-                if not (species in config.plasma.continuum_interaction.species):
-                    raise PlasmaConfigError(
-                        f"NLTE ionization species {species} not in continuum species."
-                    )
+        if (
+            config.plasma.nlte_ionization_species
+            or config.plasma.nlte_excitation_species
+        ):
```
chvogl, line 186: Why is this if clause needed? You later do individual checks `if config.plasma.nlte_ionization_species` and `if config.plasma.nlte_excitation_species` anyways.

tardis/io/tests/test_config_reader.py
sonachitchyan: I'm not sure why this is the way to do it. I mimicked it from the existing tests. Maybe @andrewfullard would have a good answer?

tardis/plasma/standard_plasmas.py
sonachitchyan: so that if one of them is empty, it passes an empty list.

tardis/plasma/standard_plasmas.py
sonachitchyan: As I am saving them in the same plasma property, I added this part to make sure that that property is added no matter if one of them is empty or not.

tardis/plasma/standard_plasmas.py
chvogl: I see. I think you could also achieve this by moving the assignment `nlte_ionization_species = config.plasma.nlte_ionization_species` and `nlte_excitation_species = config.plasma.nlte_excitation_species` here since the default value for both is an empty list (as defined in the yaml schema).

tardis/plasma/standard_plasmas.py
chvogl: Thanks for clarifying. That makes sense. I guess you could still get rid of the if clause if you do the the assignments `nlte_ionization_species = config.plasma.nlte_ionization_species` and `nlte_excitation_species = config.plasma.nlte_excitation_species` regardless of whether they are actually needed. However, I don't think this a much better solution.

sonachitchyan: so, I got rid of those 2 lines, and the code still seems to be working.

sonachitchyan: I got rid of the reassignment for `nlte_ionization_species` and `nlte_excitation_species`. Is there anything else I should change?

chvogl: I think you need to move the assignments `nlte_ionization_species = config.plasma.nlte_ionization_species` and `nlte_excitation_species = config.plasma.nlte_excitation_species` out of the inner two if clauses. Otherwise, these variables are not defined if they the values from the config are empty lists. Alternatively, you can also use `config.plasma.nlte_excitation_species` and `config.plasma.nlte_ionization_species` in line 206/207.

tardis/io/tests/test_config_reader.py
wkerzendorf: To do this properly we should follow this

def test_fails_but_bad_style():
    try:
        x = 1 / 1
        assert False
    except ZeroDivisionError:
        assert True

wkerzendorf (dismissed):

tardis/io/tests/test_config_reader.py
sonachitchyan: I updated the tests, now they get skipped, and I'm not sure if that's a good or a bad thing. I pushed it here, so that you guys can see them too.

tardis/io/tests/test_config_reader.py
```diff
@@ -190,35 +196,55 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
+        "species"
+    ] = [
+        "He I",
+    ]
+    tardis_config_verysimple_nlte["plasma"]["nlte_excitation_species"] = ["H I"]
+    with pytest.raises(PlasmaConfigError) as ve:
+        plasma = assemble_plasma(
+            tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data
+        )
+    assert ve.type is PlasmaConfigError
```
chvogl, line 247:
```suggestion
```
The type of the Exception is already checked by the `pytest.raises`. I don't think checking the type after the with clause adds anything but I might be wrong.

tardis/io/tests/test_config_reader.py
```diff
@@ -190,35 +196,55 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
-                    "continuum_interaction"
-                ]["species"]
-            ):
-                raise ValueError("Nlte ionization species not in continuum.")
-    assert ve.type is ValueError
+    with pytest.raises(PlasmaConfigError) as ve:
+        plasma = assemble_plasma(
+            tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data
+        )
+    assert ve.type is PlasmaConfigError
```
chvogl, line 217:
```suggestion
```
The type of the Exception is already checked by the pytest.raises. I don't think checking the type after the with clause adds anything but I might be wrong.

tardis/io/tests/test_config_reader.py
```diff
@@ -190,35 +196,55 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
-                in tardis_config_verysimple_nlte["plasma"][
-                    "continuum_interaction"
-                ]["species"]
-            ):
-                raise ValueError("Nlte ionization species not in continuum.")
-    assert ve.type is ValueError
+    with pytest.raises(PlasmaConfigError) as ve:
+        plasma = assemble_plasma(
+            tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data
+        )
```
chvogl, line 216: The new tests make much more sense than the old ones but aren't they strictly speaking plasma tests?
What do other people think?

tardis/io/tests/test_config_reader.py
```diff
@@ -190,35 +196,55 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
-                species
-                in tardis_config_verysimple_nlte["plasma"][
-                    "continuum_interaction"
-                ]["species"]
-            ):
-                raise ValueError("Nlte ionization species not in continuum.")
-    assert ve.type is ValueError
+    with pytest.raises(PlasmaConfigError) as ve:
+        plasma = assemble_plasma(
+            tardis_config_verysimple_nlte, nlte_raw_model, nlte_atom_data
```
chvogl, line 215: I think you need to pass a Configuration object not a config_dict?

sonachitchyan: you were right! I'll push the fixed version once the tests pass on my local machine.

andrewfullard (approved):

tardis/io/tests/test_config_reader.py
```diff
@@ -190,35 +196,55 @@ def test_plasma_nlte_section_config(tardis_config_verysimple_nlte):
...
+):
+    """
+    Configuration Validation Test for Plasma Section of the Tardis Config YAML File.
+
+    Validates:
+        nlte_excitation_species: should be included in continuum_interaction
+
+    Parameter
+    ---------
+        `tardis_config_verysimple_nlte` : YAML File
```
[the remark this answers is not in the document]
chvogl:
```suggestion
        `tardis_config_verysimple_nlte` : YAML File
        `nlte_raw_model` : A simple model
        `nlte_atom_data` : An example atomic dataset
```

tardis/io/tests/test_config_reader.py
chvogl: Nobody seems to care. Let's move on.

chvogl (approved):

sonachitchyan:
So the tests failed out of the blue. I will look into it.

chvogl:
> So the tests failed out of the blue. I will look into it.

This looks extremely similar to #2191.

sonachitchyan:
> > So the tests failed out of the blue. I will look into it.
> 
> This looks extremely similar to #2191.

yes, the error happens at the exact same spot. @atharva-2001 do you know why this might be happening?

chvogl:
This might not be particularly useful but one difference between the failed and successful test runs is that for the failed runs the cached conda environment is not found. This then triggers the "Update environment" step.

sonachitchyan:
> This might not be particularly useful but one difference between the failed and successful test runs is that for the failed runs the cached conda environment is not found. This then triggers the "Update environment" step.

I made a separate PR, to try to see if I can fix this. If that works, it should all be fine afterwards.

andrewfullard (approved):

PR #2212, opened 2023-02-19
Simulation Object Restructure (First Phase)

Soumya624:
### :pencil: Description

**Type:** :rocket: `feature` 

Created a new class ConvergencePlot under the file name tardis/simulation/convergence_plot.py and move damped_converge and _get_convergence_status methods in that. Changes are there in tardis/simulation/base.py and tardis/simulation/convergence_plot.py

### :pushpin: Resources

- [ ] Tested Locally


Soumya624:
@Rodot- Please have a look at this PR now. Tried to maintain a fixed commit history. Three files have been modified, including the mailmap

tardis/simulation/convergence_plot.py
```diff
@@ -0,0 +1,59 @@
+import logging
+import numpy as np
+
+logger = logging.getLogger(__name__)
+
+
+class ConvergencePlot:
+    @staticmethod
```
sonachitchyan, line 8: if it's a static method, you don't need `self` as an argument.

tardis/simulation/base.py
```diff
@@ -88,32 +85,6 @@ def reshape_plasma_state_store(self, executed_iterations):
 
 
 class Simulation(PlasmaStateStorerMixin, HDFWriterMixin):
-    """A composite object containing all the required information for a
```
Rodot-, line 91: What happened to the doc string?

Soumya624: Added that part. Please have a look!

tardis/simulation/convergence_plot.py
Soumya624: @sonachitchyan removed that in the latest PR. Please check

Rodot-: Did you remove the self from the calls in the simulation as well?  What was the error?

Soumya624: @Rodot- I think now it's resolved. Please have a look!

tardis/simulation/convergence_plot.py
```diff
@@ -0,0 +1,59 @@
...
+    def damped_converge(value, estimated_value, damping_factor):
+        return value + damping_factor * (estimated_value - value)
+
+    def _get_convergence_status(
```
Rodot-, line 12: Is this meant to be a static method?

tardis/simulation/convergence_plot.py
```diff
@@ -0,0 +1,59 @@
+import logging
+import numpy as np
+
+logger = logging.getLogger(__name__)
+
+
+class ConvergencePlot:
```
Rodot-, line 7: This class seems to be able to do everything with static methods and there's no __init__.  Does this actually need to be a class or could the whole thing just be a module?  If it is a class, what attributes does it have and keep track of?

Rodot- (commented):
I think you should include the plotting part of advance_state in the ConvergencePlot class as well.  Ideally, the class should entirely handle all of the convergence information and calculations and simply give back the results to be incorporated into the simulation

tardis/simulation/convergence_plot.py
Soumya624: @Rodot- 
Yeah we can refactor this class as a module with standalone functions. But, even though the current implementation of ConvergencePlot does not have any instance variables, it may be useful to keep it as a class if there is a possibility that instance variables may be added in the future to track and store the convergence data across multiple function calls.

Please let me know if there's any changes required!

PR #2226, opened 2023-02-27
[GSoC] Grotrian Diagram Visualisation: First Objective

AyushiDaksh:
### :pencil: Description

**Type:** 🌞 : `GSoC`

This is my solution (as a Jupyter notebook) to the first objective of the GSoC'23 project: Grotrian Diagram Visualisation.

### :pushpin: Resources
[GSoC'23 Page](https://summerofcode.withgoogle.com/)
[GSoC'23 Ideas Page](https://tardis-sn.github.io/gsoc_2023/ideas/)

- [ ] Testing pipeline
- [ ] Other method (describe)
- [x] My changes can't be tested (explain why)


aoifeboyle:
Nice start! Maybe you could plot it with steps instead of a continuous line like in Figs. 2 and 3 of https://arxiv.org/pdf/1611.05938.pdf ? And maybe make the y-axis log scale to show the neutral component more clearly?

AyushiDaksh:
@aoifeboyle Done! Please have a look again when you get time.

MarkMageeAstro:
Good job Ayushi! I would have a look at some of the other plots in TARDIS, like the convergence plots and custom abundance widget, to see how we normally plot things like velocities and fractional abundances

jamesgillanders:
Hi @AyushiDaksh,

The step format of your density plots is good! However, I suggest you have a think about where the steps should break, since the current format does not represent the current density structure of the model (i.e. think about whether each density should "step" at the inner, outer or middle of the shells).

Also, the x axis is labelled as km/s, but the values have not been scaled down by a factor of 1000 to reflect the conversion from m/s.

AyushiDaksh:
@jamesgillanders 
Sure, I'll have a look at the step breaks and the unit issue.

> Good job Ayushi! I would have a look at some of the other plots in TARDIS, like the convergence plots and custom abundance widget, to see how we normally plot things like velocities and fractional abundances

@MarkMageeAstro Sure I'll have a look at the other existing plots!

AyushiDaksh:
@jamesgillanders I corrected the x-axis scale (correctly to km/s) and also shifted the steps correctly (now the plot step drops at the outer walls of each shell).
Please review and let me know if the plots look good now.

MarkMageeAstro:
Nice that you've now got it as a fraction of total abundance, but the plotly version doesn't agree with the matplotlib version.

AyushiDaksh:
@MarkMageeAstro Thanks for the review.

I think both the plots **do** agree? Can you please elaborate on the dissimilarity you found? (Keep in mind that the way the units are displayed on plotly is different i.e. 1p = 10^-12 and so on)

MarkMageeAstro:
> @MarkMageeAstro Thanks for the clarification! I have updated the plotly axes formatting accordingly. Please have a look at your convenience!

Looks good.


> Also, I'm beginning my GSoC application for this project now. Is there anything specific to this project that you want me to keep in mind while preparing my application?

We want to hear your ideas for how you think the project objectives should be implemented, so make sure you include good detail on that.

AyushiDaksh:
@MarkMageeAstro The middle velocities are for the square markers. I have used the option "mid" in the step plot so the markers are placed in the middle of each step, while the actual steps happen at the mid point of the line joining the consecutive markers.

More clearly,
Say the first marker is at v_mid[0] and the second marker is at v_mid[1], then the first step is happening at (v_mid[0]+v_mid[1])/2 due to the "mid" kwarg in the step plot.

Hence the steps are happening at the correct location, even when I'm only extracting the mid velocities from the model.

MarkMageeAstro:
> @jamesgillanders The middle velocities are for the square markers. I have used the option "mid" in the step plot so the markers are placed in the middle of each step, while the actual drops happen at the mid point of the line joining the consecutive markers.
> 
> More clearly, Say the 0th square marker is at v_mid[0] and the 1st marker is at v_mid[1], then the first drop is happening at (v_mid[0]+v_mid[1])/2 due to the "mid" kwarg in the step plot, which is the position of the outer velocity of 0th shell.
> 
> Hence the steps are happening at the correct location, even when I'm only extracting the mid velocities from the model.

I think part of the issue with this is the first and last cells. If your line starts at v_mid[0], then the plot doesn't show v_inner[0]. Likewise v_mid[-1] doesn't show v_outer[-1]. You can see this from the plot -- the first and last cells both look like they're smaller than the others, but I believe they should be the same size.

AyushiDaksh:
@MarkMageeAstro  Yes you're right, I see your point. I have updated the plots accordingly (used outer velocities and also added the photosphere velocity for complete end on left sides). Please review.

jamesgillanders:
@AyushiDaksh Yes! This is exactly how I thought they should look -- good work! I would suggest focussing on writing the project proposal now

AyushiDaksh:
@MarkMageeAstro @jamesgillanders 

I have made my code more modular and general:
- Now the functionality is implemented in `tardis.visualization.widgets.ion_density.IonDensityWidget` (similar to other widgets). It has a @classmethod `from_simulation` which takes in a simulation and returns a widget object. Calling the `display` function on this objects displays it.
- The widget allows the user to select an element from a dropdown and display the corresponding fractional ion densities for that element. It automatically filters out ions which have zero ion densities in all shells.
- Also added a detailed custom hover text.

Please see the preview below.

https://user-images.githubusercontent.com/37770155/229315568-06d24931-b7a8-4ff0-bcc1-3a9ee8ef0fc9.mov

PR #2227, opened 2023-02-28
[GSoC] Simulation Object Restructure: First Objective

AyushiDaksh:
### :pencil: Description

**Type:** 🌞 `GSoC`

This is my solution to the first objective of the GSoC'23 project: Simulation Object Restructure.

Here, I have created a separate class called ConvergenceStrategy which is called internally by the Simulation class.

### :pushpin: Resources
[GSoC'23 Page](https://summerofcode.withgoogle.com/)
[GSoC'23 Ideas Page](https://tardis-sn.github.io/gsoc_2023/ideas/)

- [x] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/simulation/base.py
+++ b/tardis/simulation/base.py
@@ -25,6 +25,150 @@ from tardis.montecarlo.montecarlo_numba.r_packet import (
 logger = logging.getLogger(__name__)
 
 
+class ConvergenceStrategy(object):
+    def __init__(self, config, model):
+        """
+        Create a new ConvergenceStrategy instance from a Configuration object and Model object.
+
+        Parameters
+        ----------
+        config : tardis.io.config_reader.Configuration
+        model: tardis.model.Radial1DModel
+        """
+        self.config = config.montecarlo.convergence_strategy
+        self.model = model
+        self._converged = (
+            False  # Whether the convergence has truly converged or not
+        )
+        self._consecutive_converges_count = (
+            0  # Counts the number of consecutive converges
+        )
+
+    @property
+    def converged(self):
+        return self._converged
+
+    def convergence_step(
+        self,
+        estimated_t_rad,
+        estimated_w,
+        estimated_t_inner,
+        iterations_executed,
+    ):
+        """
+        Performs a single convergence step
+
+        Parameters
+        ----------
+        estimated_t_rad : astropy.units.Quantity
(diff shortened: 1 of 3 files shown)


atharva-2001 (commented):
Thanks for creating this pull request! Did tests pass with your changes(assuming you provided a path to the reference data)?

tardis/simulation/base.py
```diff
@@ -216,74 +200,10 @@ def estimate_t_inner(
...
-
     def advance_state(self):
         """
         Advances the state of the model and the plasma for the next
-        iteration of the simulation. Returns True if the convergence criteria
-        are met, else False.
-
-        Returns
-        -------
-            converged : bool
```
sonachitchyan, line 286: you should still have what the function returns(or what it updates in this case)

sonachitchyan (changes requested):
Thank you for the PR! There is one small thing that I commented about one of the docstrings. Everything else looks great. As a next step, can you add a simple unit test and be sure that the documentation shows explicitly shows how the convergence plot is treated?

tardis/simulation/base.py
AyushiDaksh: Fixed this!

tardis/simulation/base.py
```diff
@@ -441,11 +340,11 @@ def run(self):
...
             )
             self.iterate(self.no_of_packets)
-            self.converged = self.advance_state()
+            self.advance_state()
             if hasattr(self, "convergence_plots"):
                 self.convergence_plots.update()
             self._call_back()
-            if self.converged:
+            if self.convergence_strategy.converged:
                 if self.convergence_strategy.stop_if_converged:
```
Rodot-, line 348: Perhaps this should be a single conditional and the convergence_strategy should have a separate single attribute that tells it to stop or not based on whether or not stop_if_converged is true.

AyushiDaksh: I see. I'll add a new @property "should_stop = self.converged and self.stop_if_converged" in the ConvergenceStrategy class. We'll only check this property in the simulation class above.

If you can think of a better name than "should_stop", please let me know

AyushiDaksh: @Rodot- I've handled this, please review

I also wanted to ask, what test cases can I add for this PR?

PR #2229, opened 2023-03-01
Velocity Packet Tracker Visualisation (First Phase)

Soumya624:
### :pencil: Description

Created a file named 'vpt_visualisation.ipynb' inside the 'docs' folder to generate a SDEC plot. And Made a second plot showing the abundance of each element as a function of velocity. 

Apart from that, created a plot to show the fractional abundance of different elements as a function of inner velocity, with the density of the shell overlaid as a black dashed line on the right axis.

And made a 'Bubble Chart', where each element is represented by a circle with a size proportional to its fractional abundance at the outermost shell and a color indicating its atomic symbol. The size and position of the outermost shell circle is fixed, while the position of the element circles varies with velocity. This plot is useful for comparing the densities of different elements at the outermost shell across a range of velocities.

**Type:** :rocket: `feature` 

 `closes` the first objective of Velocity Packet Tracker Visualisation

### :pushpin: Resources

Examples, notebooks and links to useful references.

- [ ] Locally Tested


atharva-2001:
Would mind sharing an nbveiwer link and running the notebook? I wasn't able to see any plots on ReviewNB.

jamesgillanders:
Hello @Soumya624, 

I have a few comments regarding this notebook.

1. Visualising Abundance plot
 - we typically visualise the velocity of the eject in km/s -- can you update the units of the x-axis to reflect this.
 - Also, we typically use the units of g/cm^3 for density -- can you update this too?
 - You have used a step function to represent the shell densities -- good! However, you need to think about where this step function breaks, since it currently doesn't represent the true step-like behaviour of the density of the model (i.e. have a think about where the density "steps" -- is it at the inner or outer boundary of the shells, or in the middle?).
 - Check that you are using the correct densities from the model -- it might be worth looking again at some of the documentation to see if you can find the correct densities for use here.


2. Density vs Velocity for Different Elements
 - where did the step-like functionality go for the density profile -- that was good? This continuous curve is not reflective of the true density profile in the model!
 - Your figure plots 2 legends, with many duplicated entries -- address this please.
 - See above comments for axis units.

3. Bubble Chart plot
 - I cannot understand the purpose of this figure. Can you please provide a better explanation for its purpose? And maybe explain why you think it is useful?

Other general comments:
 - Always tag 2 reviewers for your PR. If you don't, it will likely be missed by reviewers, since they will filter for the ones their input has been requested on first!

Soumya624:
Hi @atharva-2001 , @jamesgillanders ,

Thanks for your review. On the way to update your suggestions. 

Regarding the bubble chart, in the case of this simulation, the bubble chart can be used to show the density of each element at the outermost shell as a function of velocity. The size of each bubble represents the fractional abundance of the corresponding element at the outermost shell, and the color represents the atomic symbol of the element. This allows us to compare the densities of different elements at the outermost shell across a range of velocities, and to see how the density of each element changes as the velocity increases.

The bubble chart can also be useful for identifying relationships and trends that might not be immediately apparent from other types of plots. For example, we might notice that certain elements have larger or smaller bubbles than we would expect based on their fractional abundances, or that the bubbles for certain elements tend to cluster in certain regions of the plot. This can help us identify areas of the simulation that might be worth investigating in more detail.

Please let me know if this can work!

jamesgillanders:
Hi @Soumya624, are there any updates on this? We haven't seen any updated code or plots

Soumya624:
Hi @jamesgillanders ,

Updated the units as you mentioned. Please have a look. Making PR's to update the remaining points!

Thanks

MarkMageeAstro:
Hi @Soumya624, I still can't see the plots in the ReviewNB

Soumya624:
Hi @MarkMageeAstro ,

Thanks for pointing out the ReviewNB. Please check this link once https://app.reviewnb.com/tardis-sn/tardis/commit/b1e4e05fda84c28974de98454419b3325f6ed53e/

Thanks

MarkMageeAstro:
> Hi @MarkMageeAstro ,
> 
> Thanks for pointing out the ReviewNB. Please check this link once https://app.reviewnb.com/tardis-sn/tardis/commit/b1e4e05fda84c28974de98454419b3325f6ed53e/
> 
> Thanks

Thanks, I can see the plots now. 

Have another look at the units. Compare your velocities to the convergence plots.

Have a look at some of ther other comments @jamesgillanders left as well, particularly with regards to the steps. 

I'm also a bit unsure of why the bubble chart is better than the simple line plot, but I'm happy for you to try new things. If you want to show a more developed version with other shells included then we can look at that, but for now I would suggest the objectives and application take priority.

MarkMageeAstro:
Hi @Soumya624 any update on your proposal? The deadline is fast approaching!

Soumya624:
> Hi @Soumya624 any update on your proposal? The deadline is fast approaching!

Hi @MarkMageeAstro,

I have shared the drive link of the proposal with you in Gitter. Please have a look. And please lemme know your if there are any changes required from my end. I would be happy to do so.

Thanks!

Issue #2216, opened 2023-02-21
Montecarlo Packet Progress bar does not go to 100%


Rodot-:
**Describe the bug**
When running the montecarlo main loop, the progress bar updates with an approximation for the number of packets completed when run in multithreaded mode.  At the end of the loop, the progress bar is no longer updated so the display never shows it reaching 100%.  

To solve this problem, at the end of the loop another update call to the progress bar should be made to fill out the remaining progress.  

**To Reproduce**
Run a tardis model with >=2 threads

**System**
 - OS:
    - [x] GNU/Linux
    - [x] macOS

 - Environment (`conda list`):

**Additional context**

AyushiDaksh:
@Rodot- Are you still working on this? If not, can I pick it up?

Rodot-:
I'm not working on this myself so it is definitely up for grabs

shreyas3156:
@Rodot- There seems to be a delay of a few seconds (even after the particular cell in the notebook has finished running), after which the packet progress bar does update to 100%. I could reproduce this for **1** <= no. of threads <= **4**.

shreyas3156:
Hi @Rodot- I have made a PR that aims to solve this issue. It would be great if you could review it.

atharva-2001:
@Rodot- Does PR https://github.com/tardis-sn/tardis/pull/2237 fix this issue?

MarkMageeAstro:
Seems to be fixed now

PR #2237, opened 2023-03-09
MonteCarlo packet progress bar completes to 100%

shreyas3156:
### :pencil: Description

**Type:** :beetle: `bugfix` 

The Monte-Carlo packet progress bar used to update during every iteration but did not complete to 100% once it was over. An update call needed to be included at the end of each iteration.

`resolves` #2216 

### :pushpin: Resources

[tqdm Docs
](https://tqdm.github.io/docs/tqdm/)

- [ ] Testing pipeline
- [x] Other method (describe)

Before:
Before

After:
After

### DIFF
--- a/tardis/montecarlo/montecarlo_numba/base.py
+++ b/tardis/montecarlo/montecarlo_numba/base.py
@@ -29,7 +29,7 @@ from tardis.montecarlo.montecarlo_numba.single_packet_loop import (
 )
 from tardis.montecarlo.montecarlo_numba import njit_dict
 from numba.typed import List
-from tardis.util.base import update_iterations_pbar, update_packet_pbar
+from tardis.util.base import update_iterations_pbar, update_packet_pbar, refresh_packet_pbar
 
 
 def montecarlo_radial1d(
@@ -128,6 +128,7 @@ def montecarlo_radial1d(
             virt_packet_last_line_interaction_out_id
         ).ravel()
     update_iterations_pbar(1)
+    refresh_packet_pbar()
     # Condition for Checking if RPacket Tracking is enabled
     if montecarlo_configuration.RPACKET_TRACKING:
         runner.rpacket_tracker = rpacket_trackers
--- a/tardis/util/base.py
+++ b/tardis/util/base.py
@@ -699,6 +699,14 @@ def update_packet_pbar(i, current_iteration, no_of_packets, total_iterations):
     packet_pbar.update(i)
 
 
+def refresh_packet_pbar():
+    """
+    Refresh packet progress bar after each iteration.
+
+    """
+    packet_pbar.refresh()
+
+
 def update_iterations_pbar(i):
     """
     Update progress bar for each iteration.


sonachitchyan:
Thanks for the PR! Please make sure to run`black` on the files that you have modified.

shreyas3156:
> Thanks for the PR! Please make sure to run`black` on the files that you have modified.

@sonachitchyan Thank you for the quick review! I have made the changes after running it.

sonachitchyan (approved):
Looks good to me!

tardis/util/base.py
```diff
@@ -699,6 +699,14 @@ def update_packet_pbar(i, current_iteration, no_of_packets, total_iterations):
...
 
 
+def refresh_packet_pbar():
+    """
+    Refresh packet progress bar after each iteration.
+
+    """
+    packet_pbar.refresh()
+
+
```
DhruvSondhi, line 709: Moving this inside the above function may do the trick. Please let me know if that works. Also add a single line comment for the same instead of having a function.

DhruvSondhi (changes requested):
This is maybe a better approach. I am not sure what you are doing (which is like in the end, when all the iterations have finished, you are refreshing the progress bar) and if this is a good idea. Maybe try this, as it refreshes the progress bar after each packet.

tardis/montecarlo/montecarlo_numba/base.py
```diff
@@ -128,6 +132,7 @@ def montecarlo_radial1d(
             virt_packet_last_line_interaction_out_id
         ).ravel()
     update_iterations_pbar(1)
+    refresh_packet_pbar()
```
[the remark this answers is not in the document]
shreyas3156: I agree that refreshing the progress bar after every packet makes the most logical sense. 

However, the bar is indeed updated automatically after every packet is sent. The issue arises at the end of any given iteration when `packet_pbar.update()` is called to update the packet but the change is not reflected. This can be solved by refreshing the bar at the end of every iteration (not just the final one).

tardis/util/base.py
shreyas3156: Since the update/refresh does not occur only at the end of each iteration, I decided to call it when the iteration ends. Adding the `packet_pbar.refresh()` at every packet would increase the simulation time severely (around 1 min/iteration) in contrast to refreshing only at the end (70s for 20 iterations on my system).

This is why I could not find a suitable place to add the refresh anywhere in the function above  (in `update_packet_pbar()`). Also because one cannot conclude when the final packet is sent (probably because it is numba jitted in parallel mode).

Thus, the only remaining solution was to either add the `packet_bar.refresh()` within the `update_iterations_pbar()` function as it is called at the end of every iteration, or have a function of its own.

DhruvSondhi: > Thus, the only remaining solution was to either add the `packet_bar.refresh()` within the `update_iterations_pbar()` function as it is called at the end of every iteration, or have a function of its own.

I suppose it is better to have just this line `packet_bar.refresh()` after the `update_iterations_pbar(1)` as having a function just for this is not good, in my opinion, and is just overhead with a call within a functional call.

shreyas3156: > I suppose it is better to have just this line `packet_bar.refresh()` after the `update_iterations_pbar(1)` as having a function just for this is not good, in my opinion, and is just overhead with a call within a functional call.

The packet_pbar and iterations_pbar instances of the progress bars are initialized and processed in `tardis/utility/base.py`. So, `packet_bar.refresh()` after `update_iterations_pbar(1)` would require instantiating the progress bars in `montecarlo_numba/base.py` as well, which may not be the best approach.

I agree that it is an avoidable overhead, in which case I can just add it in the `update_iterations_pbar()` method after `iterations_pbar.update(i)`. It would look something like:

```
def update_iterations_pbar(i):
    iterations_pbar.update(i)
    packet_bar.refresh()
```

DhruvSondhi: This would be inefficient as the current implementation updates the progress bar by 1 per cent after a specific increase in the progress as the packets are propagated. What this may do is that after every 1 per cent, the progress bar will get refreshed and updated. For this, I suppose you can look into making something like this: (pseudo code here)
```
If the progress is above let's say, 95% (would be good if this is as close to 100 as possible or maybe even 100):
  refresh the bar for each iteration until it reaches the end,
else:
  keep doing what it initially was doing,
```
What happens with the `numba jitted` functions is that the information is present with the compiler (machine level compiler) but is not actually passed onto into Python runtime (Python interpreter), which causes loss of information as hence it doesn't understand how to add all the multithreaded information together from different cores. 

Another possible solution could be that you update the progress bar by returning the instance of the progress bar from the `update_iterations_pbar()` function and then update it when all the progress is finished. What I mean by this is that even when we are running TARDIS in multithreaded mode, the progress bar takes into consideration all the progress made on all the cores. And in the end, please correct me, it is still compiled on a single core. So a single call for refreshing the progress bar at the end after the completion of all the propagated packets may do the trick. It works for all other iterations, expect the last one. This would be more efficient.

shreyas3156: > This would be inefficient as the current implementation updates the progress bar by 1 per cent after a specific increase in the progress as the packets are propagated. What this may do is that after every 1 per cent, the progress bar will get refreshed and updated. 

The `update_iterations_pbar()` is called to update the _iterations_ progress bar (`iterations_pbar`) at the end of each iteration, only after the _packet_ progress bar (`packet_pbar`) has finished updating for that iteration, i.e., after all the packets for that iteration have been sent. So I think this refresh would not interfere with the `packet_pbar` while the packets are being sent and only update it after sending 100% of the packets.

> What happens with the numba jitted functions is that the information is present with the compiler (machine level compiler) but is not actually passed onto into Python runtime (Python interpreter), which causes loss of information as hence it doesn't understand how to add all the multithreaded information together from different cores.

Thank you for the clear description. This may be why there is no possible way to fetch the completion percentage as the number of packets sent through different cores is not trackable. 

> So a single call for refreshing the progress bar at the end after the completion of all the propagated packets may do the trick.

My intention with the code block mentioned in the previous reply was precisely this. Since  `update_iterations_pbar()` is called at the end of every iteration, adding the refresh part within it would also be subject to a call at the end of each iteration when all the packets have propagated. Please correct me if I've misunderstood.

atharva-2001 (approved):

Issue #2102, opened 2022-07-19
relative paths for atomic data and configuration are not intuitive


wkerzendorf:
I set the path to the config file as `w7/config.yml` and in the `config.yml` there is an entry `atom_data: kurucz_cd23_chianti_H_He.h5`. It did complain that `w7/kurucz.h5` was not found and it also does not exist in the standard repository. I wonder if it is cleaner to specify either a filename or an atomic dataset.

shreyas3156:
Hi @wkerzendorf, I plan to work on resolving this issue and I came across the following code in `standard_plasmas.py`
```
 if atom_data is None:
        if "atom_data" in config:
            if os.path.isabs(config.atom_data):
                atom_data_fname = config.atom_data
            else:
                atom_data_fname = os.path.join(
                    config.config_dirname, config.atom_data
                )
```
The last case where `config_dirname` is used to define `atom_data_fname` looks a bit counter-intuitive because the `atom_data` directory and `config` directory may be quite unrelated (please correct me if I'm wrong).

One possible solution is replacing `config.config_dirname` with `get_data_dir()` as it may make more sense to join `atom_data` to `data_dir`

PR #2240, opened 2023-03-15
Fix relative atom data path to eliminate use of configuration directory

shreyas3156:
### :pencil: Description

**Type:** :beetle: `bugfix` 

`resolves` #2102 

This PR aims to solve the Atom Data relative path issue that arises when the configuration file is not in the same directory as the notebook/file calling `run_tardis()`. The issue arises because we currently use `config.config_dirname` in `standard_plasmas.py` to find the path to the atom data file which seems counter-intuitive because config and atom data are quite unrelated.

The solution proposed in this PR:

1. Moves all logic that handles `atom_data_fname` from `standard_plasmas.py` to `tardis/io/atom_data/util.py`, with the rationale that resolution of the file path is more appropriate for the `resolve_atom_data_fname()` method in the latter.

2. Eliminates the usage of `config.config_dirname` altogether and instead relies on `data_dir` to find the atom data file.

### :pushpin: Resources

- [x] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -169,6 +169,9 @@ Sampark Sharma <samparksharma2000@gmail.com>
 Shilpi Prasad <shilpiprd@gmail.com>
 Shilpi Prasad <shilpiprd@gmail.com> Shilpi <72646134+shilpiprd@users.noreply.github.com>
 
+Shreyas Singh <shreyashsingh31@gmail.com>
+Shreyas Singh <shreyashsingh31@gmail.com> shreyas3156 <shreyashsingh31@gmail.com>
+
 Satwik Kambham <satwik.kambham@gmail.com>
 Satwik Kambham <satwik.kambham@gmail.com> code-explorer <satwik.kambham@gmail.com>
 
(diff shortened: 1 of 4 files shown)


andrewfullard (changes requested):

andrewfullard (approved):

andrewfullard (dismissed):
Looks good. Could add some docstrings to the tests to make the coverage check happy.

shreyas3156:
> Looks good. Could add some docstrings to the tests to make the coverage check happy.

Sure, I'll add the docstrings.

andrewfullard:
Looks like the test is failing

shreyas3156:
> Looks like the test is failing

Yes, the test passed on my local when I specified the `tardis-refdata` path to the `DEFAULT_DATA_DIR` while running pytest. I'm not sure what the `get_data_dir()` call in the tests in github actions receives so I've decided to instead mock the `get_data_dir()` function and I'll amend the recent commit with the change.

shreyas3156:
Sorry, I mistakenly removed the other reviewers while requesting a review! :(

andrewfullard (approved):

andrewfullard:
@wkerzendorf I think this should be revisited with the restructure

andrewfullard:
Duplicated by other fixes more recently

Issue #1976, opened 2022-04-25
Widgets code shows up during the commandline 


wkerzendorf:
It should be disabled when running on the commandline. 
```
VBox(children=(FigureWidget({█▏                                                            119918/500000 [00:12<00:38, 9986
    'data': [{'type': 'scatter', 'uid': 'a781058b-c248-4e8b-a41f-c9232fc3fe15', 'xaxis': 'x', 'yaxis': 'y'},
             {'type': 'scatter', 'uid': '1bed6058-4acd-4c65-b2e0-cd0062fc8e29', 'xaxis': 'x2', 'yaxis': 'y2'}],
    'layout': {'height': 450,
               'legend': {'title': {'text': 'Iterations'}, 'traceorder': 'reversed'},
               'margin': {'b': 25, 'l': 10, 'pad': 0, 'r': 135, 't': 25},
               'template': '...',
               'xaxis': {'anchor': 'y',
                         'domain': [0.0, 0.45],
                         'tickformat': 'g',
                         'title': {'text': '$\\text{Velocity}~[\\text{km}~\\text{s}^{-1}]$'}},
               'xaxis2': {'anchor': 'y2',
                          'domain': [0.55, 1.0],
                          'matches': 'x',
                          'tickformat': 'g',
                          'title': {'text': '$\\text{Velocity}~[\\text{km}~\\text{s}^{-1}]$'}},
               'yaxis': {'anchor': 'x',
                         'domain': [0.0, 1.0],
                         'nticks': 15,
                         'tickformat': 'g',
                         'title': {'text': '$T_{\\text{rad}}\\ [\\text{K}]$'}},
               'yaxis2': {'anchor': 'x2',
                          'domain': [0.0, 1.0],
                          'nticks': 15,
                          'tickformat': 'g',
                          'title': {'text': '$W$'}}}
```

shreyas3156:
Hi @wkerzendorf, if the display() method is called from the command-line, would it be better to print a message like "Please run the widget inside a notebook" or raise an error instead?

mohakmalviya:
hello sir! i would like to work on this issue can you please elaborate it??

andrewfullard:
This specific issue is fixed (but there are similar commandline display issues)

PR #2253, opened 2023-03-22
Fix for widgets displaying code in commandline

shreyas3156:
### :pencil: Description

**Type:** :beetle: `bugfix`

`resolves` #1976 

The widgets displayed code when called from the command-line as shown in the issue above. This PR adds a check so that the `display()` methods in the widgets run only for notebooks. When called from command-line, a message "Please use a notebook to display the widget" is printed.

For the tests, I have used a pytest.Monkeypatch context manager to use a session-scoped fixture and mock the `is_notebook()` response.

### :pushpin: Resources

[Monkeypatch.context](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.context)

- [x] Testing pipeline
- [x] Other method - Ran the widgets in the command-line

image

@wkerzendorf @atharva-2001

### DIFF
--- a/tardis/visualization/widgets/tests/test_custom_abundance.py
+++ b/tardis/visualization/widgets/tests/test_custom_abundance.py
@@ -5,6 +5,7 @@ import tardis
 import numpy as np
 import numpy.testing as npt
 
+from tardis.tests.test_util import monkeysession
 from tardis.visualization.widgets.custom_abundance import (
     CustomAbundanceWidgetData,
     CustomYAML,
@@ -30,7 +31,7 @@ def yml_data(example_configuration_dir: Path, atomic_dataset):
 
 
 @pytest.fixture(scope="module")
-def caw(yml_data):
+def caw(yml_data, monkeysession):
     """Fixture to contain a CustomAbundanceWidget
     instance generated from a YAML file tardis_configv1_verysimple.yml.
 
@@ -40,6 +41,10 @@ def caw(yml_data):
         CustomAbundanceWidget generated from a YAML
     """
     caw = CustomAbundanceWidget(yml_data)
+    monkeysession.setattr(
+        "tardis.visualization.widgets.custom_abundance.is_notebook",
+        lambda: True,
+    )
     caw.display()
     return caw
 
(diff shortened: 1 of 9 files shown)


tardis/visualization/widgets/tests/test_custom_abundance.py
```diff
@@ -88,6 +89,10 @@ def caw(yml_data):
         CustomAbundanceWidget generated from a YAML
     """
     caw = CustomAbundanceWidget(yml_data)
+    monkeysession.setattr(
+        "tardis.visualization.widgets.custom_abundance.is_notebook",
+        lambda: True,
+    )
```
atharva-2001, line 95: Was the issue for Convergence plots or for the custom abundance widget?
Also, I think we should raise an error but I also would like to hear what others say about this. If we do so, maybe then we would want a test that mimics the command line and see if it raises an error?

shreyas3156: > Was the issue for Convergence plots or for the custom abundance widget?

You're right. I just noticed that the issue was created for the convergence plots. But the code was also visible for the three Widgets, so I made the changes. 

I will add the checks for command-line in the Convergence plots.

> Also, I think we should raise an error but I also would like to hear what others say about this. If we do so, maybe then we would want a test that mimics the command line and see if it raises an error?

Raising an error would definitely be a more prudent way to handle it.

shreyas3156: @atharva-2001 I have added a check where RuntimeError is raised if `run_tardis` is called from the command line and if `show_convergence_plots` is set to True (which was the only case where the code was being printed). 

The default value of `show_convergence_plots` was set to True in `simulation/base.py` but False in `tardis/base.py`. I changed the former to False as well, otherwise the tests which simply call `Simulation.from_config` using the default as True would've failed.

jamesgillanders:
@atharva-2001 Can you take a look at this please and let us know what you think?

jamesgillanders:
@atharva-2001 Can you take a look at this please and let us know what you think?

atharva-2001 (approved):

andrewfullard (approved):

MarkMageeAstro (approved):

PR #2255, opened 2023-03-25
GSoC'23 Grotrian Diagrams: first objective

MariamH78:
### :pencil: Description

**Type:** :rocket: `feature` 

This is the PR to view and review the files and the notebooks I've created for GSoC's first objective.

- [ ] Testing pipeline
- [x] Other method (describe)
        - The notebook tests the full functionality of the class.


MarkMageeAstro:
Hi Mariam,

Nice work on the PR! The code looks nicely laid out. 

It would be nice to have ions of the same element on the same plot, rather than a specific plot for each one. I think labels with actual element names would also be more useful. I suggest looking at how this is done with ions in the SDEC plot and then try to implement a similar style here.

MariamH78:
Hey @MarkMageeAstro!

I've updated the repository to try and match what you're talking about. Not sure if I understood your comment completely: I only modified my own code to allow for atom names, but I didn't take into account using roman numerals to denote charge.

Also I found a good compromise for using subplots/not using them: Users can mix and match however they like.

Does that work?

jamesgillanders:
Hi @MariamH78 -- good start with the PR. I have a few comments.

1. Is it possible to rescale the y-axis so that we are looking at the ratio of ion density, which is much more informative for our purposes? i.e. plot (Si+ / Si), (Si2+ / Si).

2. Can you change the plot style to include some step functionality, instead of a simple line connecting the points? The reason being that the simulation does not have a continuous density structure -- in fact, the density in each shell is constant (and so the density of the Si ions will be too) and this should be reflected by the plot.

3. Can you make the different ions have different coloured points/lines? I find it difficult to differentiate between the light/medium/dark blues.

MariamH78:
Hi @jamesgillanders,

Before I start working, I need to make sure I properly understand what you meant.

1. So every point (or step) would be the ratio between  let's say, singly/doubly/whatever ionized ion OVER neutral silicon, not ionized OVER total silicon (neutral + ionized)?
2. Got that one, will implement it right away.
3. Aaand I was thinking the same thing, the blues are pretty and I love them but they may not be easy to distinguish for everyone, so yeah, we're dropping the blues 👍👍

MarkMageeAstro:
> Hi @jamesgillanders,
> 
> Before I start working, I need to make sure I properly understand what you meant.
> 
>     1. So every point (or step) would be the ratio between  let's say, singly/doubly/whatever ionized ion OVER neutral silicon, not ionized OVER total silicon (neutral + ionized)?
> 
>     2. Got that one, will implement it right away.
> 
>     3. Aaand I was thinking the same thing, the blues are pretty and I love them but they may not be easy to distinguish for everyone, so yeah, we're dropping the blues 👍👍

1. Every step should be the ratio of singly/doubly/whatever to total silicon. We usually think of these things as "how much of all my silicon is in this state?" so it makes sense to use the total silicon.
3. The colours are still a bit difficult for me to see. Can you try a range of colours like what we have in some of the other plots?

For the multiple plots, am I understanding this right? You can pass a list of how many plots you want and each one can contain however many ionisation states? Nice work!

Any update on your proposal? The deadline is very soon!

Issue #1653, opened 2021-06-17
Add units to BlackBodySimpleSource


isaacgsmith:
We would like to make the `BlackBodySimpleSource.create_packets()` method (in `tardis/montecarlo/packet_source.py`) unit-sensitive. That is, we would like to take in quantities with units and return quantities with units.

**Additional context**
A potential thing to think about here is how this method is used in the code, and having that be unit-sensitive as well. Remembering that this class can be replaced by a user's own packet source, do we want to require that the user's function includes units? That may be something for @wkerzendorf to answer.

DivyaPamecha:
Hello @smithis7 @wkerzendorf . I am new to open source and I found this issue interesting. I tried researching about this but couldn't find any resources. Could you please help me with this issue?

rishav-karanjit:
Hey, I really wanna work on this issue. Can this issue be assigned to me?

jayantbhakar:
Hi, is this issue still open? I want to know exactly how the unit will be taken as input?

AyushiDaksh:
@wkerzendorf In case of Custom packet source, do we really want to force the user to take in quantities with units and return quantities with units?

laksh-krishna-sharma:
@wkerzendorf  
Hey, I really want to work on this issue. 
As I am very curious about modern physics and also having a pretty good understanding about physics and Python as well.
So please can this issue be assigned to me please?

andrewfullard:
Since we use a packet collection and that has to be numba compatible, units aren't possible.

PR #2259, opened 2023-03-27
Add units to black body simple source

AyushiDaksh:
### :pencil: Description

This change makes the `BlackBodySimpleSource.create_packets()` method (in `tardis/montecarlo/packet_source.py`) unit-sensitive. That is, it now takes in quantities with units and return quantities with units (instead of float).

The documentation corresponding to Custom Packet Source has also been updated accordingly (to force the user to write a unit-sensitive custom function).

The unit sensitive quantities that the function takes are:
- Temperature
- Initial packet radius
- Time elapsed since explosion

The unit sensitive quantities that the function returns are:
- Packet radii (cm)
- Packet frequencies (Hz)
- Packet energies (erg)

I observed that the `radius` input and `radii` output were already coded in a unit-sensitive manner. The changes broke things for other quantities (especially the `nus` and `energies`) but I handled them accordingly in MonteCarloRunner (from where the function is being called) and test cases.

However, the `MonteCarloRunner` object is still **not** unit sensitive (it stores these quantities as float, which further get stored as float in `PacketCollection`) and the same gets propagated to other parts of the code (like `RPacket`, which affects `Numba`). I tried to make the whole workflow unit-sensitive, but it started breaking a lot of things.

Hence, for now, I only made `create_packets` unit sensitive, and contained the ripple effect in MonteCarloRunner (which remains unit-insensitive itself). Do we want to make the whole workflow unit-sensitive?

`resolves` [Issue 1653](https://github.com/tardis-sn/tardis/issues/1653)

- [✅] Testing pipeline: Without and with ref data

- [✅] I requested two reviewers for this pull request
- [✅] I updated the documentation according to my changes

@smithis7 @jamesgillanders @atharva-2001


tardis/montecarlo/base.py
```diff
@@ -228,10 +228,10 @@ def _initialize_packets(
             )
 
         mc_config_module.packet_seeds = seeds
-        self.input_r = radii
-        self.input_nu = nus
+        self.input_r = radii.to(u.cm)
```
smithis7, line 231: Should this have a `.value`?

smithis7 (changes requested):
This looks great overall. Make sure you clear output on documentation notebooks before you commit, and also the packet initialization documentation will need to be fixed. If you build the docs locally you will see any notebooks that need to be updated.

tardis/montecarlo/base.py
AyushiDaksh: @smithis7 
No, radii was already coded in a unit sensitive manner, so no changes were needed here. I just decided to add that "to" function as a sanity check (refactoring), because otherwise the codebase assumes that quantities would be given in CGS. (The test cases pass without that "to" function too)

Also, "mus" is a dimensionless quantity, so nothing to be done there.

AyushiDaksh: @smithis7 I have also cleared Jupyter cell outputs and ran the documentation build. Just a minor change was needed in 1 other notebook, fixed it. The documentation build was successful.

Please review

smithis7:
There is an issue with the docs build on our end, when I get the chance I can check out your PR and build docs locally.

AyushiDaksh:
Sure, thanks @smithis7 !

sonachitchyan:
Good work! I feel like making it unit sensitive is a good idea, but make sure when it gets passed to the montecarlo part, you're only using the values, so that the code doesn't slow down. Also, please rebase this so that we can see if the tests pass.

AyushiDaksh:
@sonachitchyan Thanks for the review! :)

> I feel like making it unit sensitive is a good idea
Do you mean making the whole MonteCarloRunner unit sensitive (This PR only makes packet source unit sensitive, and handles the initialization in MonteCarloRunner accordingly)? Keep in mind that PacketCollection, RPacket would still need to stay unit **insensitive** to avoid breaking Numba (unless we want to make everything unit sensitive).

The docs are building locally except the following error:

`/Users/ayushi/Documents/tardis/tardis/montecarlo/packet_source.py:docstring of tardis.montecarlo.packet_source.BasePacketSource.create_blackbody_packet_nus:9: ERROR: Unexpected indentation.`

I don't think my changes are causing this though, as I get the same error when building docs on the updated master branch. Should I ignore this?

Also, the GitHub build-docs workflow failed due to a `Connection reset by peer` error. Should we run this again?

sonachitchyan (dismissed):

AyushiDaksh:
@Rodot- Can you please review this? Thanks!

AyushiDaksh:
Some HDF5 error is failing the docs test on Github which doesn't seem related to my changes

tardis/montecarlo/packet_source.py
```diff
@@ -129,24 +132,26 @@ class BlackBodySimpleSource(BasePacketSource):
...
+                no_of_packets : int
+                    Number of packets
+        <<<<<<< HEAD
+                radius : float64
+        =======
+                rng : numpy random number generator
+                radius : astropy.Quantity
+        >>>>>>> a0ca0e139 (Add units to create_packets)
+                    Initial packet radius
+
```
atharva-2001, line 148: This was probably a merge conflict being resolved but was staged.

tardis/montecarlo/packet_source.py
```diff
@@ -164,25 +169,21 @@ def create_packets(
...
-            Packet frequencies
-        array
-            Packet directions
-        array
-            Packet energies
+        astropy.Quantity : Packet radii
+        astropy.Quantity : Packet frequencies
+        astropy.Quantity : Packet directions
+        astropy.Quantity : Packet energies
         """
```
atharva-2001, line 187: I wonder if these are according to our docstring standards- it's the name of the variable and then the variable's type, right? The next line would have more description of the variable.

tardis/montecarlo/base.py
```diff
@@ -340,7 +340,7 @@ def run(
         self._initialize_geometry_arrays(model)
 
         self._initialize_packets(
```
atharva-2001, line 342: This function doesn't have a docstring. Would you mind adding that?

docs/io/optional/custom_source.ipynb
atharva-2001, line 1: Docstrings also need to be fixed here

atharva-2001 (changes requested):
Hi @AyushiDaksh, I made a few small comments. Could you please fix them and re-request my review?

tardis/montecarlo/packet_source.py
AyushiDaksh: @atharva-2001 For parameters of the function, the docstring convention is `variable_name (type): description`
However, for return values (like the above lines), the convention is `type: description` (because return values aren't strictly associated with a variable_name)

docs/io/optional/custom_source.ipynb
AyushiDaksh: https://github.com/tardis-sn/tardis/pull/2259#discussion_r1248459956

tardis/montecarlo/packet_source.py
AyushiDaksh: Resolved this!

tardis/montecarlo/base.py
AyushiDaksh: Added!

AyushiDaksh:
@atharva-2001 Thanks for the review! I had made the changes you mentioned and re-requested a review

andrewfullard:
This will be handled as part of a larger packet source effort

PR #2262, opened 2023-03-28
Added units to BlackBodySimpleSource

Jae124:
### :pencil: Description

**Type:** :beetle: `bugfix`

As a first pull request, I changed the data types of temperature and radius of create_packets() from float64 to astropy.units.quantity.Quantity. 
Also added unit conversion in create_blackbody_packet_nus() to u.K, create_uniform_packets() to u.erg. Since the base.py of montecarlo calls create_packets(), modified parameters of base.py also. Also modified tests to the correct format, temperature input from float64 to u.K Quantity object.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/montecarlo/packet_source.py
```diff
@@ -96,7 +97,7 @@ def create_blackbody_packet_nus(
         xis_prod = np.prod(xis[1:], 0)
         x = ne.evaluate("-log(xis_prod)/l")
 
-        return x * (const.k_B.cgs.value * temperature) / const.h.cgs.value
+        return (x * (const.k_B.cgs.value * temperature.to(u.K)) / const.h.cgs.value).to(u.Hz)
```
smithis7, line 100: This is not exactly how to do it. Nothing in a function with units should have `.value`, and I wouldn't force it to be in Hz, only do that where it is used in the code.

tardis/montecarlo/packet_source.py
```diff
@@ -48,9 +49,9 @@ def create_uniform_packet_energies(no_of_packets, rng):
 
         Returns
         -------
-            energies for packets : numpy.ndarray
+            energies for packets : numpy.ndarray in Quantity units
         """
-        return np.ones(no_of_packets) / no_of_packets
+        return (np.ones(no_of_packets) / no_of_packets).to(u.erg)
```
smithis7, line 54: I would do `* u.erg` as the `.to` method is primarily used for unit conversions.

tardis/montecarlo/base.py
```diff
@@ -357,10 +357,10 @@ def run(
         self._initialize_geometry_arrays(model)
 
         self._initialize_packets(
-            model.t_inner.value,
+            model.t_inner,
             no_of_packets,
             iteration,
-            model.r_inner[0],
```
smithis7, line 363: Why are we getting rid of `[0]`?

smithis7 (commented):
This is a good start. The rest of the TARDIS code (specifically where the blackbody functions are called) needs to be updated, as does the packet initialization documentation.

tardis/montecarlo/base.py
smithis7: The `[0]` selects the first shell, so we are using the inner boundary of the first shell.

PR #2270, opened 2023-04-02
[GSOC] First objective: STARDIS Parallelization and GPU

nevasini:


smithis7:
Thank you for the submission! The goal of the objective here is to compare the speeds of both functions for each bin size, not comparing the output.

nevasini:
> Thank you for the submission! The goal of the objective here is to compare the speeds of both functions for each bin size, not comparing the output.

Agreed understood the pitfalls, I have tried to compare against the bin size as you wanted to. Please review it once more

docs/physics/montecarlo/initialization.ipynb
```diff
@@ -71,10 +71,39 @@
   },
```
smithis7, line 71: <p>Try this with just numba, and you won't need/want to change anything within the function from how it was before, only the decorator</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/2270/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

PR #2280, opened 2023-04-19
Prevent tests from running on forks

atharva-2001:
### :pencil: Description

This is to prevent consumption of LFS resources. 

### :pushpin: Resources

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
https://github.com/atharva-2001/tardis/actions/runs/4742893534

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ defaults:
 
 jobs:
   build:
-    if: github.repository_owner==tardis-sn
+    if: github.repository_owner=='tardis-sn'
     strategy:
       matrix:
         include:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ defaults:
 
 jobs:
   build:
-
+    if: github.repository_owner==tardis-sn
     strategy:
       matrix:
         include:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ defaults:
 
 jobs:
   build:
-    if: github.repository_owner=='tardis-sn'
+    if: github.repository_owner == 'tardis-sn'
     strategy:
       matrix:
         include:


andrewfullard:
I take it forks cannot use caches in the same way? And even if they can, they will still have to generate a cache to begin with?

andrewfullard (dismissed):

.github/workflows/tests.yml
```diff
@@ -29,7 +29,7 @@ defaults:
 
 jobs:
   build:
-
+    if: github.repository_owner=='tardis-sn'
```
epassaro, line 32: Leave spaces before and after ==

atharva-2001:
https://github.com/tardis-sn/tardis/blob/b3bc761aa9c7d9a185b2d76eb504048c1e71e1de/.github/workflows/tests.yml#L22-L24

I'm afraid I don't know if this works completely as intended.
The workflow should run only once for a commit, right?
`${{ github.ref }}` is different for forks and for upstream repo.
https://github.com/atharva-2001/tardis/actions/runs/4798029406/jobs/8535766612
https://github.com/tardis-sn/tardis/actions/runs/4798029889/jobs/8535760860

Here is the workflow file.
https://github.com/tardis-sn/tardis/actions/runs/4798029889/workflow

epassaro:
> https://github.com/tardis-sn/tardis/blob/b3bc761aa9c7d9a185b2d76eb504048c1e71e1de/.github/workflows/tests.yml#L22-L24
> 
> I'm afraid I don't know if this works completely as intended. The workflow should run only once for a commit, right? `${{ github.ref }}` is different for forks and for upstream repo. https://github.com/atharva-2001/tardis/actions/runs/4798029406/jobs/8535766612 https://github.com/tardis-sn/tardis/actions/runs/4798029889/jobs/8535760860
> 
> Here is the workflow file. https://github.com/tardis-sn/tardis/actions/runs/4798029889/workflow

As far as I remember, that creates a group depending if the workflow runs on a branch of the repository or runs on a pull request. 

The part ${{ github.head_ref || github.ref }} ` means, if `github.head_ref` does not exist fallback to `github.ref`. One of these variables exist if the workflow runs in a branch of the repo, the other exist in pull requests (or both scenarios?, I don't remember).

It wasn't designed with forks in mind, I don't know what happens there.

atharva-2001:
I agree- this is working correctly. My apologies. I also found this- might be of interest:
https://stackoverflow.com/a/75403978/11974464

.github/workflows/tests.yml
atharva-2001: Done via https://github.com/tardis-sn/tardis/pull/2280/commits/8351415447c05920aa85e0d657e4ba811dc20528

andrewfullard (approved):

ymatsumu (approved):

PR #2305, opened 2023-06-02
Fix to helium indexing in NLTE

aoifeboyle:
### :pencil: Description

**Type:** :beetle: `bugfix`

Using g.loc[2,1,0], for example, in the helium NLTE file results in the error 'cannot join with no overlapping index names'. This fixes it (by replacing with float(g.loc[2,1,0])). 

- [ ] Testing pipeline
- [x] Other method (describe)

Other method: It runs now and gives the correct output. 


tardis/plasma/properties/helium_nlte.py
```diff
@@ -48,7 +48,7 @@ def calculate(
         helium_population.loc[0, 0] = 0.0
         # He II excited states
         he_two_population = level_boltzmann_factor.loc[2, 1].mul(
-            (g.loc[2, 1, 0] ** (-1.0))
+            (float(g.loc[2, 1, 0]) ** (-1.0))
```
chvogl, line 51: Do you know why casting `g.loc[2, 1, 0]` to a float solves your indexing problem? g should be a pandas Series of ints and the loc operation should return an int. Casting the int to a float should not make a difference in this case. What is the type of `g.loc[2, 1, 0]` for you? Is it a pandas Series? What is your pandas version?

sonachitchyan: what if instead we do `g.loc[[2,1,0]]` ? Won't this just keep the type of the element?

aoifeboyle: @chvogl @sonachitchyan

So g.loc[2,1,0] gives me:

ipdb> g.loc[2,1,0]
[py.warnings         ][WARNING]  
        <stdin>:1: PerformanceWarning: indexing past lexsort depth may impact performance.
 (warnings.py:109)
atomic_number  ion_number  level_number
2              1           0               2
Name: g, dtype: int64

g is indeed a pandas Series and my pandas version is 1.0.5 but I started this PR because someone told me they were having this error.

g.loc[2,1,0].values[0] returns the right value (the zero because it returns an array). Is that preferable to the float() casting?

chvogl: @aoifeboyle Thanks a lot for the reply. I am quite puzzled. g.loc[2,1,0] returns an int for me with the same pandas version. I suppose g is a DataFrame in your case (while it is a Series for me)?
I would not switch to `g.loc[2,1,0].values[0]` since this would not work if `g.loc[2,1,0]` returns an int. The float casting at least works in both scenarios.

aoifeboyle: @chvogl Very strange! I just pulled the current Tardis master version and it still breaks at g.loc[2,1,0] and when I use the debugger to do the following I still get a Series:

ipdb> type(g)
<class 'pandas.core.series.Series'>

But g.loc[2,1,0] still returns an object including the indices. 

Should we merge the float() version or investigate further? If you like I can send you the input files I'm using and you can see if it triggers the same error in your version of Tardis. But if the float() should work for everyone than I guess that's also ok.

chvogl: I am Ok with merging the float version, but I am happy to have a look if you send me the input files. My curiosity has been piqued.

chvogl (approved):

PR #2365, opened 2023-07-20
integrate radiation field

wkerzendorf:
New integration of radiation field


tardis/radiation_field/base.py
```diff
@@ -0,0 +1,70 @@
...
+        self.dilution_factor = dilution_factor
+        self.t_rad = self.t_radiative
+        self.w = self.dilution_factor
+        self.opacities = opacities
+        self.source_function = packet_source
+
+
+def convert_config_to_radiationfield_state(
+    config, no_of_shells, temperature=None
+):  ### make it a to_method in the end
```
andrewfullard, line 62: Good because otherwise this is really clunky!

tardis/radiation_field/base.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+from astropy import units as u
+
+from tardis.montecarlo.packet_source import BasePacketSource
+from tardis.montecarlo.montecarlo_numba.numba_interface import OpacityState
+
+
+class RadiationField:
```
andrewfullard, line 8: Do we want to be clear about this being a radiationfield*state*?

tardis/radiation_field/base.py
```diff
@@ -0,0 +1,62 @@
...
+        opacities: OpacityState,
+        packet_source: BasePacketSource,
+    ):
+        self.t_radiative = t_radiative
+        self.dilution_factor = dilution_factor
+        self.opacities = opacities
+        self.packet_source = packet_source
+
+
+class MonteCarloRadiationFieldState:
```
andrewfullard, line 36: Should this inherit from the above class to reduce code duplication? Is it even necessary? There is duplication of t_rad.

tardis/radiation_field/base.py
wkerzendorf: sure

tardis/radiation_field/base.py
wkerzendorf: I just copied from the previous implementation. But ultimately, we could also only have that and then add a superclass later.

wkerzendorf:
@andrewfullard I've included your suggestions.

andrewfullard (approved):

andrewfullard:
Looks like `black` and tests are needed

PR #2386, opened 2023-08-09
Integrate density in state object

wkerzendorf:
another stab at the restructure


andrewfullard:
Seems generally sensible. Would it be possible to fix at least the density tests? I think it should be possible to solve the abundance widget issue as well.

tardis/model/tests/test_density.py
```diff
@@ -10,24 +10,22 @@
...
     if hasattr(actual, "cgs"):
         actual = actual.cgs.value
-    path = os.path.join("homologous_density", "density_0")
+    path = os.path.join("model", "density")
     expected = pd.read_hdf(hdf_file_path, path)
     assert_almost_equal(actual, expected.values)
 
 
 def test_hdf_time_0(hdf_file_path, simulation_verysimple):
-    actual = simulation_verysimple.model.homologous_density.time_0
```
andrewfullard, line 28: Are we sure that `time_0` and `time_explosion` are equivalent?

tardis/io/model/density.py
```diff
@@ -0,0 +1,187 @@
...
+    else:
+        raise ValueError(
+            f"Unrecognized density type " f"'{density_configuration.type}'"
+        )
+    return density_0, time_0
+
+
+def parse_config_v1_density(config: Configuration) -> u.Quantity:
+    """
+    Create a new HomologousDensity instance from a Configuration object.
```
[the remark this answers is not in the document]
wkerzendorf: can we ignore this for the next PR. I'm shifting the files around but do not want to change things twice.

tardis/model/tests/test_density.py
wkerzendorf: in this case they are because we shift everything to time_explosion. the `model` object only has things at the same time.

andrewfullard (dismissed):

andrewfullard (approved):
If all of the refdata tests are going to pass, I'm happy.

PR #2409, opened 2023-09-10
Add units to BlackBodySimpleSource

### LINKED ISSUE #1653: Add units to BlackBodySimpleSource

laksh-krishna-sharma:
Description:

This pull request introduces unit-sensitivity to the "BlackBodySimpleSource.create_packets()". The primary goal is to ensure that this method accepts input quantities with units and returns quantities with units, enhancing its usability and consistency.

Context:

The BlackBodySimpleSource class is a packet source generates Blackbody packets, and this modification aims to make it unit sensitive. By enforcing units, I improve the clarity of how the method should be used and allow users to work with units seamlessly.

Changes: 

Fixes #1653 
I have verified this pull request:

* Ensuring radius has units in meter by adding .to(u.meter). line no. 122
* Ensuring temperature has units of Kelvin by adding .to(u.Kelvin). line no. 123
* Adding *u.meter in line no. 152
* Adding * u.dimensionless_unscaled in line no. 234

### DIFF
--- a/tardis/montecarlo/packet_source.py
+++ b/tardis/montecarlo/packet_source.py
@@ -105,9 +105,9 @@ class BlackBodySimpleSource(BasePacketSource):
     Parameters
     ----------
     radius : float64
-        Initial packet radius
+        Initial packet radius with units (e.g., '3e10 m')
     temperature : float
-        Absolute Temperature.
+        Absolute Temperature with units (e.g., '5600 K')
     base_seed : int
         Base Seed for random number generator
     legacy_secondary_seed : int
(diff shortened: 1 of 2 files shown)


tardis/montecarlo/packet_source.py
```diff
@@ -119,8 +119,8 @@ def from_model(cls, model, *args, **kwargs):
         return cls(model.r_inner[0], model.t_inner.value, *args, **kwargs)
 
     def __init__(self, radius=None, temperature=None, **kwargs):
-        self.radius = radius
-        self.temperature = temperature
+        self.radius = radius.to(u.meter)    # Ensure radius has units in meter
```
atharva-2001, line 122: Do we know if this is an astropy quantity?

PR #2415, opened 2023-09-20
In gamma ray code : Changed times to np.geomspace

Knights-Templars:


tardis/energy_input/gamma_ray_transport.py
```diff
@@ -356,15 +356,7 @@ def main_gamma_ray_loop(
     ), "Error, simulation start time greater than end time!"
 
     if time_space == "log":
-        times = np.zeros(time_steps + 1)
-
-        # log time steps
-        for i in range(time_steps + 1):
```
wkerzendorf, line 362: this looks like a good choice (example comment)

andrewfullard:
I think this is a great change, but I would like to see a before/after comparison of the time steps to make sure this does the same thing as my old code.

Knights-Templars:
> I think this is a great change, but I would like to see a before/after comparison of the time steps to make sure this does the same thing as my old code.

Thanks @andrewfullard. I have tested with `np.isclose` and is numericaally the same.

tardis/energy_input/gamma_ray_transport.py
```diff
@@ -356,15 +356,7 @@ def main_gamma_ray_loop(
...
-            times[i] = (
-                np.log(time_start)
-                + (np.log(time_end) - np.log(time_start)) / time_steps * i
-            )
-            times[i] = np.exp(times[i])
+        times = np.geomspace(time_start, time_end, time_steps + 1)
```
wkerzendorf, line 359:
```suggestion
        times = np.logspace(time_start, time_end, time_steps + 1, base=np.e)
```

wkerzendorf: I think that's better (example comment)

PR #2434, opened 2023-10-09
Added a function to calculate shell masses

Knights-Templars:
Function to calculate shell masses

PR #2435, opened 2023-10-09
Added a function to calculate isotope mass

Knights-Templars:


andrewfullard:
Much like #2434 this ought to be a method of the model IMO.

PR #2436, opened 2023-10-09
Make inventories

Knights-Templars:
Create a list of inventories.

### DIFF
--- a/tardis/energy_input/gamma_ray_transport.py
+++ b/tardis/energy_input/gamma_ray_transport.py
@@ -260,6 +260,48 @@ def initialize_packets(
     )
 
 
+def calculate_shell_masses(model):
+    """Function to calculate shell masses
+    Parameters
+    ----------
+    model : tardis.Radial1DModel
+        The tardis model to calculate gamma ray propagation through
+    Returns
+    -------
+    numpy.ndarray
+        shell masses in units of g
+
+    """
+    outer_velocities = model.v_outer.to("cm/s").value
+    inner_velocities = model.v_inner.to("cm/s").value
+    ejecta_density = model.density.to("g/cm^3").value
+    ejecta_volume = model.volume.to("cm^3").value
+    shell_masses = ejecta_volume * ejecta_density
+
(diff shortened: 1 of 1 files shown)


tardis/energy_input/gamma_ray_transport.py
```diff
@@ -260,6 +260,85 @@ def initialize_packets(
...
+    )
+
+    raw_isotope_mass = raw_isotope_abundance.apply(
+        lambda x: x * calculate_shell_masses(model), axis=1
+    )  # calculate masses of isotope
+
+    return raw_isotope_mass
+
+
+def create_inventories(model, time_end=80.0):
```
andrewfullard, line 305: I think time_end should be set up to require a quantity input, not just float

tardis/energy_input/gamma_ray_transport.py
```diff
@@ -260,6 +260,85 @@ def initialize_packets(
...
+    inventories = (
+        raw_isotope_mass.to_inventories()
+    )  # inventories are in disintegrations for x g of isotope
+    shell_masses = calculate_shell_masses(model)
+
+    inventory_list = []
+    for shell, inv in enumerate(inventories):
+
+        decay_per_gram = (
+            inv / shell_masses[shell]
```
andrewfullard, line 331: `inv` is not enough to get the total decays. You need to use the method `cumulative_decays`. They are also already in per gram so you do not need to divide by shell masses again

andrewfullard: You can't divide an inventory by mass

tardis/energy_input/gamma_ray_transport.py
```diff
@@ -260,6 +260,85 @@ def initialize_packets(
...
+        )  # decays are in disintegrations for 1 g of isotope
+        total_decays = decay_per_gram.cumulative_decays(
```
andrewfullard, line 333: This would work if you just replace `decay_per_gram` with `inv`.

Issue #2442, opened 2023-10-11
Duplicate level index when using continuum interaction


AlexHls:
**Describe the bug**

When enabling continuum interactions duplicate level indices appear (see screenshot), causing TARDIS to crash during the re-indexing process ``tardis/plasma/properties/transition_probabilities.py:362``

Currently #2443 implements a bandaid fix that throws a warning. Once Carsus implemented a fix upstream, this warning should be removed and an exception should be raised instead.

**To Reproduce**

**Screenshots**

![2023-10-11T15:21:24,292949360-04:00](https://github.com/tardis-sn/tardis/assets/70367168/c3d2bfe7-6423-436d-8a4d-5e114f1ce731)

**System**
 - OS:
    - [x] GNU/Linux
    - [ ] macOS

 - Environment (`conda list`):

**Additional context**

chvogl:
Should be dealt with in Carsus (see carsus#384).

chvogl:
Is this not what #2444 is for?

PR #2443, opened 2023-10-12
Fix duplicate entries in continuum line list

AlexHls:
### :pencil: Description

**Type:** :beetle: `bugfix` 

Attemts to fixe #2442 

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/plasma/properties/atomic.py
+++ b/tardis/plasma/properties/atomic.py
@@ -464,6 +464,7 @@ class LevelIdxs2LineIdx(HiddenPlasmaProperty):
             ],
             names=["source_level_idx", "destination_level_idx"],
         )
+
         level_idxs2line_idx = pd.Series(
             np.arange(len(index)), index=index, name="lines_idx"
         )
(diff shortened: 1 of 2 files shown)


tardis/plasma/properties/transition_probabilities.py
```diff
@@ -330,11 +330,20 @@ def calculate(
...
-            0,
-            "transition_type",
-            level_idxs2transition_idx.at[("k", "bf"), "transition_type"],
-        )
+        # Since 'at' seems to be faster than 'loc' we try it first
+        # However, if the index is not unique, 'at' will raise a ValueError
+        try:
+            fb_cooling_probs.insert(
+                0,
+                "transition_type",
```
wkerzendorf, line 338: but these are different - right? one inserts `value[0]` the other inserts the value. Why are there sometimes multiple? are they duplicate? Should we drop them?

tardis/plasma/properties/atomic.py
```diff
@@ -467,6 +467,14 @@ def calculate(self, atomic_data):
         level_idxs2line_idx = pd.Series(
             np.arange(len(index)), index=index, name="lines_idx"
         )
+
+        # Check for duplicate indices
```
AlexHls, line 471: We already get duplicate indices here, completely unrelated to the actual continuum lines. For now I added a ``ValueError`` to catch this early, but I'm not sure if we can simply drop the duplicate indices.

tardis/plasma/properties/transition_probabilities.py
AlexHls: This bit is just a workaround to avoid an error from this. This should (probably) be removed before merging. I suspect that this is only a symptom of the issue. The duplicates are already present in ``atomic.py``.

wkerzendorf (approved):

sonachitchyan (approved):

wkerzendorf (changes requested):
Please follow the TARDIS bot instructions

sonachitchyan (approved):

Issue #2021, opened 2022-05-10
TARDIS May produce different results when multithreading.


Rodot-:
**Describe the bug**
Updating the estimators when in parallel during the montecarlo main loop is not an atomic operation, resulting in invalid values when run with multiple threads.

Should be fixed by adding multithreaded tests.

**To Reproduce**
Hard to compare, see continuum branch

**Screenshots**

**System**
 - OS:
    - [x] GNU/Linux
    - [x] macOS

 - Environment (`conda list`):

**Additional context**

jayantbhakar:
I think the same issue is with progress bars. I am trying to figure out a way for it. I have gone through its code and the multithread loop is causing the problem. I didn't find a way to control that loop for single execution of the progress bars as well.

Rodot-:
Should be noted while this was partially fixed by #2022 there may still be inconsistencies in the v_packet energy histograms as they are not properly handled in parallel (though the likelihood they run into such problems is lower, but non-zero)

PR #2447, opened 2023-10-16
Fix numba parallel issues

AlexHls:
### :pencil: Description

**Type:** :beetle: `bugfix`

Fixes the numba parallel issues mentioned in #2021 (at least for normal packets)

### :pushpin: Resources

None.

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/docs/io/configuration/index.rst
+++ b/docs/io/configuration/index.rst
@@ -9,11 +9,6 @@ The necessary parameters for a TARDIS simulation are provided via a `YAML <https
 configuration file. The following sections respectively describes what can or must be included in the
 configuration file, shows an example configuration file, describes how TARDIS checks to make sure a configuration
 file is valid, and demonstrates how a YAML configuration file is read in.
-
-.. caution:: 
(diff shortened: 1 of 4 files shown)


andrewfullard:
This is great to see fixed. Do you have examples showing the improvement to reproducibility?

AlexHls:
Yes, here is a plot showing the comparison between a serial and a parallel run (10 threads) of the ``tardis_example.yml``:
test_parallel
You can see that the difference is at the level of machine precision.

Before the fix this would have been the result:
test_parallel

I didn't do more extensive testing, but I think this is such a trivial fix/ clear case that more extensive testing isn't necessary

andrewfullard:
Great, thank you! Next up we need to actually make the monte carlo loop parallel with numba, but that is a LOT more work.

wkerzendorf:
@Rodot- signed off.

wkerzendorf (approved):

Rodot- (commented):
Looks good to me!

PR #2453, opened 2023-10-25
Fix bug in relativistic packet source

chvogl:
### :pencil: Description

**Type:** :beetle: `bugfix`

#2366 broke the BlackBodySimpleSourceRelativistic by changing `mus` to `nus`. The packets source thus assigns frequencies between 0 and 1. This PR fixes this.

- [ ] Testing pipeline
- [x] Other method (describe)

### DIFF
--- a/tardis/montecarlo/packet_source.py
+++ b/tardis/montecarlo/packet_source.py
@@ -313,7 +313,7 @@ class BlackBodySimpleSourceRelativistic(BlackBodySimpleSource):
         self.beta = ((self.radius / self.time_explosion) / const.c).to("")
         return super().create_packets(no_of_packets)
 
-    def create_packet_nus(self, no_of_packets):
+    def create_packet_mus(self, no_of_packets):
         """
         Create zero-limb-darkening packet :math:`\mu^\prime` distributed
         according to :math:`\\mu^\\prime=2 \\frac{\\mu^\\prime + \\beta}{2 \\beta + 1}`.
@@ -327,7 +327,7 @@ class BlackBodySimpleSourceRelativistic(BlackBodySimpleSource):
 
         Returns
         -------
-        array of frequencies
+        Directions for packets
             numpy.ndarray
         """
         z = self.rng.random(no_of_packets)


chvogl:
I still need to add tests for the relativistic packet source (to make sure something similar doesn't happen again).
I will remove the draft status afterwards.

wkerzendorf:
and my apologies - that was my bad

tardis/montecarlo/tests/test_packet_source.py
```diff
@@ -43,6 +47,22 @@ def blackbodysimplesource(self, request):
         yield cls.bb
         montecarlo_configuration.LEGACY_MODE_ENABLED = False
 
+    @pytest.fixture(scope="class")
+    def blackbodysimplesourcerelativistic(self, request):
```
wkerzendorf, line 51:
```suggestion
    def blackbody_simplesource_relativistic(self, request):
```

tardis/montecarlo/tests/test_packet_source.py
```diff
@@ -75,3 +93,47 @@ def test_bb_packet_sampling(
         assert np.all(np.isclose(nus, ref_df["nus"]))
         assert np.all(np.isclose(mus, ref_df["mus"]))
         assert np.all(np.isclose(unif_energies, ref_df["energies"]))
+
+    def test_bb_packet_sampling_relativistic(
+        self,
+        tardis_ref_data,
+        blackbodysimplesourcerelativistic,
```
wkerzendorf, line 100:
```suggestion
        blackbody_simplesource_relativistic,
```

wkerzendorf:
is this done?

chvogl:
I think this should be good enough for now. The tests are not great but I think fixing the bug has priority.

tardis/montecarlo/tests/test_packet_source.py
```diff
@@ -75,3 +93,47 @@ def test_bb_packet_sampling(
...
+        blackbody_simplesource_relativistic.beta = 0.25
+
+        nus = blackbody_simplesource_relativistic.create_packet_nus(100)
+        unif_energies = (
+            blackbody_simplesource_relativistic.create_packet_energies(100)
+        )
+        blackbody_simplesource_relativistic._reseed(2508)
+        mus = blackbody_simplesource_relativistic.create_packet_mus(10)
+
+        gamma = np.sqrt(1 - 0.25**2) ** -1
```
jamesgillanders, line 118: Should the 0.25 be specified instead as blackbody_simplesource_relativistic.beta? Is there a reason for it to be hard-coded?

tardis/montecarlo/tests/test_packet_source.py
```diff
@@ -75,3 +93,47 @@ def test_bb_packet_sampling(
...
+        ref_df = tardis_ref_data["/packet_unittest/blackbody"]
+        expected_nus = ref_df["nus"]
+        expected_unif_energies = ref_df["energies"] * 1.6 / gamma
```
jamesgillanders, line 121: Can I get a reminder of what the 1.6 represents? I did look through all this a few months ago, but I have since forgotten many of the details, so a quick reminder or explanation would be nice!

jamesgillanders (commented):
The code looks good Christian! Some minor comments but probably related to my poor understanding as opposed to the PR...

tardis/montecarlo/tests/test_packet_source.py
chvogl: No specific reason. I changed it now.

tardis/montecarlo/tests/test_packet_source.py
chvogl: 1.6 is the "static_inner_boundary2cmf_factor" , which is calculated as (2 * beta + 1) / (1 - beta**2). The factor appears because the inner boundary is not comoving with the material (contrary to what one would intuitively assume). If you think it is better, I can calculate the factor here instead of hardcoding it. The idea behind hardcoding it was to avoid code duplication between the test and the method being tested.

jamesgillanders: I think if it is explicit somewhere else then it's probably fine. You're the expert -- if it makes sense to do it this way, then that's good enough for me!

jamesgillanders (approved):
I'm happy with this

andrewfullard (approved):

andrewfullard:
Merging this despite fails- OSX fails are HDF breaking, docstr-cov is not important.

PR #2454, opened 2023-10-25
Fix zero density

AlexHls:
### :pencil: Description

**Type:** :beetle: `bugfix`
Removes line that sets innermost shell density to zero when reading a ``csvy`` model file. At least as far as I can tell there is no apparent reason for this to be done. In the worst case, this can lead to a ``PlasmaIonizationError: n_electron just turned "nan" - aborting`` error since ``number_density`` becomes infinite.

- [x] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -594,7 +594,7 @@ class SimulationState(HDFWriterMixin):
                 csvy_model_config.datatype.fields[density_field_index]["unit"]
             )
             density_0 = csvy_model_data["density"].values * density_unit
-            density_0 = density_0.to("g/cm^3")[1:]
+            density_0 = density_0.to("g/cm^3")
             density = calculate_density_after_time(
                 density_0, time_0, time_explosion
             )
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -595,7 +595,6 @@ class SimulationState(HDFWriterMixin):
             )
             density_0 = csvy_model_data["density"].values * density_unit
             density_0 = density_0.to("g/cm^3")[1:]
-            density_0 = density_0.insert(0, 0)
             density = calculate_density_after_time(
                 density_0, time_0, time_explosion
             )


andrewfullard (dismissed):
Happy since the tests are passing

wkerzendorf:
@andrewfullard are we worried that this will break things?

andrewfullard:
> @andrewfullard are we worried that this will break things?

It's possible. Looking at the CSVY tests, they don't really cover this at all as far as I can tell.

AlexHls:
@marxwillia do you have any idea what the purpose of this line is?

tardis/model/base.py
```diff
@@ -595,7 +595,6 @@ def from_csvy(cls, config, atom_data=None):
             )
             density_0 = csvy_model_data["density"].values * density_unit
             density_0 = density_0.to("g/cm^3")[1:]
```
DeerWhale, line 597: Probably should get rid of the `[1:]` at the end line 597, just to make sure the dimension is consistent.

marxwillia:
I don't think this should break anything.  The first row of the CSVY data should be ignored when setting up the model since the physical quantities correspond to the value of those parameters within the shell.  I believe we set up the v_inner_boundary to throw an error if it was set to a value lower than the first velocity, ie we force the photosphere to be defined above the velocity of the first line in the CSVY file.  I think the reason we replaced the first density value with 0.0 was to emphasize that it's not used.

AlexHls:
I think this value is used, if not physically than at least in some calculations down the line. I.e. I get a 0 number density for the populations in zeroth shell (which is calculated using this zero density value). It could be possible that my csvy model is miss-configured, but than we should rather throw an Exception, warning the user about a miss-configured csvy model, rather than silently modifying the input data.

AlexHls:
What the code does so far in ``tardis/model/base.py:515`` :

- It loads the geometry from the csvy, which essentially loads the velocity.
- When no ``v_inner_boundary`` or ``v_outer_boundary`` is specified, it takes the first or last element, respectively to set those values. Essentially that means that for a csvy file with 125 cells/ rows, we get ``v_inner.shape = (124,)`` and ``v_outer.shape = (124,)``. To my understanding, this is the indented outcome.
- The code then proceeds to load the density (``tardis/model/base.py:586``). This is where things get weird:
  - If the csvy does not have a ``density`` attribute, it calls the ``parse_csvy_density function``. This function exists both (identically) in ``tardis/io/model/parse_density_input.py`` and tardis/io/model/readers/density.py``, although only the former is used. Not sure if this is intended, but this seems highly redundant to me
  - This function reads in the velocity again , this time adding a 0 velocity in the zeroth element to be able to calculate a ``v_middle`` of the shape n+1 where n is the number of 'density shells'. It then proceeds to parse the density section of the configuration file (i.e. not the csvy file). I'm not sure how sensible this mixed approach is, but in of itself, this part seems of (at least apart from the 0 velocity bit, but this is at least half sensible in the context of the reconstruction we're doing there)
  - If this ``density`` attribute is not present, the code proceeds to read the density from the csvy file. Here not much should be needed, since the data is already loaded in the ``csvy_model_data`` dataframe, loaded through the ``load_csvy`` function. Interestingly enough the ``load_csvy`` function already skips the first line! This means we never load the first line! I feel this should be documented very visibly! Apart from this, the only thing this section does is adding some units and setting the zeroth density to 0 for no apparent reason. This has nothing to do with the velocity shape etc. or the calculation of the density at the requested time.

In my opinion, setting the density to 0 can be safely omitted. There should maybe be a few places where some warning should be raised, but this is a different issue.

PR #2456, opened 2023-10-27
Rename calls to `get_thread_id()` to `_get_thread_id()` 

smokestacklightnin:
### :pencil: Description

**Type:** :beetle: `bugfix` 

Rename calls to `get_thread_id()` to `_get_thread_id()`  because the former doesn't exist in current versions of `numba`.

This was accomplished using the command

```
find . -type f -name "*.py" -exec sed -i 's/get_thread_id/_get_thread_id/g' {} +
```

### :pushpin: Resources

The reference for the above command can be found [here](https://victoria.dev/blog/how-to-replace-a-string-with-sed-in-current-and-recursive-subdirectories/).

- [X] Testing pipeline


andrewfullard:
This is not correct for the current tardis env

PR #2458, opened 2023-11-01
Enable nlte ionization as plasma component

AlexHls:
### :pencil: Description

**Type:**  :rocket: `feature` 

Enables the use of NLTE ionization (and implicitly excitation) treatment during the plasma construction. In the current implementation, it is not compatible with the ``helium_treatment`` options ``recomb-nlte`` and ``numerical-nlte``. This is due to the fact that both ``recomb-nlte`` and the NLTE rate equation solver provide ``ion_number_density`` and ``electron_densities``, potentially overwriting each other. As this could lead to unexpected behavior, I decided to simply disallow the simultaneous use of these options to prevent unintended results. (I could not find an obvious incompatibility with ``numerical-nlte`` since this does not seem to provide the above mentioned quantities.. However, since it is in the same if else block, having this option enabled would block the addition of the NLTE ionizaition/excitation properties.) 

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/plasma/properties/transition_probabilities.py
+++ b/tardis/plasma/properties/transition_probabilities.py
@@ -84,7 +84,9 @@ class SpMatrixSeriesConverterMixin(object):
             idx2reduced_idx.loc[q_indices[1]].values,
         )
         max_idx = idx2reduced_idx.max() + 1
-        matrix = sp.coo_matrix((series, q_indices), shape=(max_idx, max_idx))
+        matrix = sp.coo_matrix(
+            (series.astype(np.float64), q_indices), shape=(max_idx, max_idx)
+        )
         return matrix
 
     @staticmethod
--- a/tardis/plasma/properties/transition_probabilities.py
+++ b/tardis/plasma/properties/transition_probabilities.py
@@ -48,7 +48,9 @@ def normalize_trans_probs(p):
     index = p.index.get_level_values("source_level_idx")
     p_norm = p / p_summed.loc[index].values
     p_norm = p_norm.fillna(0.0)
-    breakpoint()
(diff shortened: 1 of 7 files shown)


AlexHls:
The typing fixes are a bit of a monkey patch, but it avoids possible issues in several places where the NLTE tests failed. They probably don't fix the underlying issues, i.e. that some variables are created with the wrong datatype. I couldn't find where this happens, so this fix should do the trick until another solution (e.g. enforcing static typing) is found.

This should be ready for review now.

tardis/plasma/properties/transition_probabilities.py
```diff
@@ -41,10 +41,16 @@ def normalize_trans_probs(p):
...
-    p_summed = p.groupby(level=0).sum()
+    # Dtype conversion is needed for pandas to return nan instead of
+    # a ZeroDivisionError in cases where the sum is zero.
+    p = p.astype(np.float64)
+    p_summed = p.groupby(level=0).sum().astype(np.float64)
     index = p.index.get_level_values("source_level_idx")
     p_norm = p / p_summed.loc[index].values
     p_norm = p_norm.fillna(0.0)
+    # Convert back to original dtypes to avoid typing problems later on
+    # in the number code.
```
andrewfullard, line 52: numba, not number.

andrewfullard (dismissed):

andrewfullard (approved):
This continues to remind me that we need to interrogate assemble_plasma to better deal with configurations.

DeerWhale (approved):

AlexHls:
@andrewfullard or @wkerzendorf , can you merge this? I don't think well get all checks to pass in this PR, the issues seem unrelated.

PR #2470, opened 2023-11-10
Add astropy import guard

AlexHls:
### :pencil: Description

**Type:** :beetle: `bugfix` 

Fixes the import issue described in #2081 or at least throws a warning that TARDIS should be imported first.

### DIFF
--- a/tardis/__init__.py
+++ b/tardis/__init__.py
@@ -12,13 +12,20 @@ __all__ = []
 # ----------------------------------------------------------------------------
 
 import sys
+import warnings
 
 # ----------------------------------------------------------------------------
 
-from astropy import physical_constants, astronomical_constants
+if ("astropy.units" in sys.modules) or ("astropy.constants" in sys.modules):
+    warnings.warn(
+        "Astropy is already imported externally. Astropy should be imported"
+        " after TARDIS."
+    )
+else:
+    from astropy import physical_constants, astronomical_constants
 
-physical_constants.set("codata2014")
-astronomical_constants.set("iau2012")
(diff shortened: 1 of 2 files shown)


andrewfullard (approved):

tardis/__init__.py
```diff
@@ -12,13 +12,20 @@
...
 import sys
+import warnings
 
 # ----------------------------------------------------------------------------
 
-from astropy import physical_constants, astronomical_constants
+if ("astropy.units" in sys.modules) or ("astropy.constants" in sys.modules):
+    warnings.warn(
+        "Astropy is already imported externally. Astropy should be imported"
+        " after TARDIS."
```
Knights-Templars, line 22: Sounds like a quick fix for me now.

andrewfullard (approved):

tardis/__init__.py
AlexHls: I don't think that there is any real 'fix' for this as the Exception comes from Astropy itself. This is just a guard such that TARDIS continues running and doesn't leave the user with a potentially cryptic Exception. At the worst, one uses a different version of the Astropy constants. This will probably mess with e.g. test results, but I think a warning is enough and the code doesn't need to raise an Exception at this point. 

On a side note, ``ruff`` formatting and checking sorts the Astropy import before the TARDIS import, so this is bound to occur at some point.

Knights-Templars:
This PR may be a temporary fix for now, and this can be merged. The corresponding issue is #2081

Knights-Templars (approved):

sonachitchyan (approved):

PR #2445, opened 2023-10-12
Restructure/abundance

wkerzendorf:
And the next one - useful for @jvshields @Knights-Templars 

Still in draft mode but good to look at.

PR #2475, opened 2023-11-21
Add assert statement to catch negative abundances

AlexHls:
Adds some assert statements to catch negative abundances in the input, preventing the simulation from running with nonsensical inputs.

Covers at least the csvy input where something like this is probably the most likely place to occur.

### DIFF
--- a/tardis/io/model/readers/csvy.py
+++ b/tardis/io/model/readers/csvy.py
@@ -150,8 +150,6 @@ def parse_csv_abundances(csvy_data):
     )
     assert (
         isotope_abundance < 0.0
-    ).sum().sum() == 0, (
-        "Isotope abundances must be positive. Negative abundances found csvy."
-    )
+    ).sum().sum() == 0, "Isotope abundances must be positive. Negative abundances found in csvy."
 
     return abundance.index, abundance, isotope_abundance
--- a/tardis/io/model/readers/csvy.py
+++ b/tardis/io/model/readers/csvy.py
@@ -143,4 +143,15 @@ def parse_csv_abundances(csvy_data):
                 element_symbol_string
             ].tolist()
 
+    assert (
+        abundance < 0.0
+    ).sum().sum() == 0, (
+        "Abundances must be positive. Negative abundances found in csvy."
+    )
+    assert (
+        isotope_abundance < 0.0
+    ).sum().sum() == 0, (
+        "Isotope abundances must be positive. Negative abundances found csvy."
+    )
+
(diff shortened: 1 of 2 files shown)


tardis/io/model/readers/csvy.py
```diff
@@ -143,4 +143,15 @@ def parse_csv_abundances(csvy_data):
...
 
+    assert (
+        abundance < 0.0
+    ).sum().sum() == 0, (
+        "Abundances must be positive. Negative abundances found in csvy."
+    )
+    assert (
+        isotope_abundance < 0.0
+    ).sum().sum() == 0, (
+        "Isotope abundances must be positive. Negative abundances found csvy."
```
andrewfullard, line 154: found in

tardis/model/parse_input.py
```diff
@@ -152,4 +152,15 @@ def parse_abundance_section(config, atom_data, geometry):
...
 
+    assert (
+        abundance < 0.0
+    ).sum().sum() == 0, (
+        "Abundances must be positive. Negative abundances found in csvy."
+    )
+    assert (
+        isotope_abundance < 0.0
+    ).sum().sum() == 0, (
+        "Isotope abundances must be positive. Negative abundances found csvy."
```
andrewfullard, line 163: found in

Also, is this guaranteed to be CSVY here?

andrewfullard (approved):

tardis/model/parse_input.py
```diff
@@ -152,4 +152,15 @@ def parse_abundance_section(config, atom_data, geometry):
     if atom_data is not None:
         elemental_mass = atom_data.atom_data.mass
 
+    assert (
+        abundance < 0.0
+    ).sum().sum() == 0, (
```
jvshields, line 157: I'm a bit confused about what this .sum().sum() == 0 is doing. Can you describe that operation?

wkerzendorf:
@AlexHls can we just add this to making the composition object in #2445

tardis/model/parse_input.py
AlexHls: It essentially sums up all ``True`` (=1) and ``False`` (=0) values. Since all values in the expression before that should be ``False``, the sum should evaluate to 0. A single ``True`` value would mean the sum is larger than 0. The double sum is just to cover both axes of the DataFrame. As far as I know there is now built-in method to do this in pandas. A simple ``assert (abundance >=0).all()`` doesn't seem to work with DataFrames

PR #2478, opened 2023-11-27
Add ruff rules and docs

AlexHls:
Adds ruff rules to the project config in preparation for an eventual ruff pre-commit hook. Adopts the rules used by astropy. Also adds some rudimentary documentation on its usage.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,0 +1,230 @@
+# Ruff configuration file adapted from astropy
+extend = "pyproject.toml"
+lint.ignore = [
+    # NOTE: to find a good code to fix, run:
+    # ruff --select="ALL" --statistics astropy/<subpackage>
+
+    # flake8-annotations (ANN) : static typing
+    "ANN001",  # Function argument without type annotation
+    "ANN003",  # `**kwargs` without type annotation
+    "ANN201",  # Public function without return type annotation
+    "ANN202",  # Private function without return type annotation
+    "ANN401",  # Use of `Any` type
+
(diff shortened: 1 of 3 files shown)


wkerzendorf (dismissed):

pyproject.toml
```diff
@@ -48,3 +48,83 @@ omit-covered-files = false
...
+"docs/*.py" = [
+    "INP001",  # implicit-namespace-package. The examples are not a package.
+]
+
+[tool.ruff.lint.flake8-annotations]
+ignore-fully-untyped = true
+mypy-init-return = true
+
+[tool.ruff.lint.isort]
+known-first-party = ["astropy", "extension_helpers"]
```
wkerzendorf, line 127:
```suggestion
known-first-party = ["tardis", "carsus", "stardis"]
```

.ruff.toml
```diff
@@ -0,0 +1,230 @@
+# Ruff configuration file adapted from astropy
+extend = "pyproject.toml"
+lint.ignore = [
```
andrewfullard, line 3: Are these global ignores?

andrewfullard (dismissed):
Would like to see the output of it run on TARDIS

.ruff.toml
AlexHls: Yes, these ignores should apply globally.

AlexHls:
@wkerzendorf I think you need to force merge it since the docs workflow is still failing

wkerzendorf (approved):

PR #2479, opened 2023-11-27
Check if atom data download exists

wkerzendorf:
and do not redownload unless forced.


andrewfullard (approved):
Hopefully this helps the docs issue

andrewfullard (approved):

PR #2490, opened 2023-12-04
Fix viz sdec

AlexHls:
Fixes the SDEC plot

- Updates the variable name ``model`` to ``simulation_state``
- Allows to plot a SDEC in cases where no visual or real packets are available
- Makes the blackbody photosphere optional (not really helpful in my cases)
- Some fixes to ensure ruff compliance (not really required, but while I'm at it, could be removed if not wanted)


andrewfullard (approved):
Good change.

AlexHls:
If there are no virtual packets, you can't generate the plotter object. However, with my fix, you can generate a plotter object and the 'real' spectrum works perfectly fine. The plot of the virtual spectrum obviously fails, but this is tu be expected. I assume it will work the other way around, but I haven't tested it. One could probably make the error handling a bit more elaborate, I.e. avoid raising an exception, but imo this is beyond the scope of this pr and a simple fix has to suffice for now.

In short: my fix makes the sdec work if your hdf doesn't contain virtual packets (for the real spectrum)

wkerzendorf:
the tests fail?

AlexHls:
> the tests fail?

No idea why. Tests also fail locally, but there different ones fail... All of them seem completely unrelated to the visualization, i.e. all visualization tests pass...

AlexHls:
@wkerzendorf Now the tests pass for some reason, but Codecov is complaining all of a sudden... Not sure if this is an actual issue or just some artifact...

jvshields (approved):

PR #2500, opened 2023-12-31
Restructure/estimators

wkerzendorf:
An attempt at restructuring the estimator code. 

requires #2492 to be merged.

PR #2501, opened 2024-01-05
Fix/nlte ionization matrix

wkerzendorf:
The negative value problem seems to have been caused by a greater sign error in the negative value fixer module

REQUIRES THE merge of #2500


andrewfullard:
87 changed files?

wkerzendorf:
> 87 changed files?

yes - this depends on #2500. It's otherwise difficult to work with.

PR #2504, opened 2024-01-18
Update number of shells in SimulationState 

DeerWhale:
### :pencil: Description

**Type:** :beetle: `bugfix` | :vertical_traffic_light: `testing` 

Updated the no_of_shells in SimulationState to be the number of active shells, which affect the dimensionality of csvy models when the inner boundary velocity is set to be a value in between the first and last shell.

From there, found some other similar bugs that would cause issue for csvy models that has v_inner_bounday in the middle, including: 

- The geometric dilution factor: the [r_boundary_inner in this equation](https://tardis-sn.github.io/tardis/physics/setup/model.html#:~:text=taken%20up%20by%20the%20sphere%20is) should be the radius of the first shell has v <= v_inner_boundary (Note: if directly change to v_inner_active[0] will cause error on the first w value, since v_middle[0] can be smaller than v_inner_active[0])
- The packet_source radius should start at r_inner_active[0] not the raw r_inner[0], which caused the issue that there is no packet escaping if the csvy model has deep inactive shells 

Another modification is in DiluteBlackBodyRadiationFieldState" -- include the geometry input, so only check the assertion on dilution factor >0 for active shells (for cases that v_inner is in the middle, the raw dilution factors contains nan values.)
 

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -135,7 +135,9 @@ class SimulationState(HDFWriterMixin):
     @dilution_factor.setter
     def dilution_factor(self, new_dilution_factor):
         if len(new_dilution_factor) == self.no_of_shells:
-            self.radiation_field_state.dilution_factor = new_dilution_factor
+            self.radiation_field_state.dilution_factor[
+                self.geometry.v_inner_boundary_index : self.geometry.v_outer_boundary_index
+            ] = new_dilution_factor
         else:
             raise ValueError(
                 "Trying to set dilution_factor for unmatching number"
@@ -151,7 +153,9 @@ class SimulationState(HDFWriterMixin):
     @t_radiative.setter
     def t_radiative(self, new_t_radiative):
         if len(new_t_radiative) == self.no_of_shells:
-            self.radiation_field_state.t_radiative = new_t_radiative
+            self.radiation_field_state.t_radiative[
+                self.geometry.v_inner_boundary_index : self.geometry.v_outer_boundary_index
+            ] = new_t_radiative
         else:
             raise ValueError(
                 "Trying to set t_radiative for different number of shells."
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -224,7 +224,7 @@ class SimulationState(HDFWriterMixin):
 
     @property
     def no_of_shells(self):
-        return self.geometry.no_of_shells
+        return self.geometry.no_of_shells_active
 
     @property
     def no_of_raw_shells(self):
--- a/tardis/model/base.py
+++ b/tardis/model/base.py
@@ -132,7 +132,7 @@ class SimulationState(HDFWriterMixin):
 
     @dilution_factor.setter
     def dilution_factor(self, new_dilution_factor):
-        if len(new_dilution_factor) == self.no_of_shells_active:
+        if len(new_dilution_factor) == self.no_of_shells:
             self.radiation_field_state.dilution_factor = new_dilution_factor
         else:
             raise ValueError(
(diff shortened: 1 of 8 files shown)


tardis/model/base.py
```diff
@@ -260,15 +260,6 @@ def from_config(cls, config, atom_data):
...
 
-        nuclide_mass_fraction = parse_abundance_config(
-            config, geometry, time_explosion
-        )
-
-        # using atom_data.mass.copy() to ensure that the original atom_data is not modified
-        composition = Composition(
-            density, nuclide_mass_fraction, atom_data.atom_data.mass.copy()
-        )
-
```
DeerWhale, line 271: This modification is jus to delete the duplicated codes

tardis/model/base.py
```diff
@@ -133,7 +133,9 @@ def dilution_factor(self):
     @dilution_factor.setter
     def dilution_factor(self, new_dilution_factor):
         if len(new_dilution_factor) == self.no_of_shells:
-            self.radiation_field_state.dilution_factor = new_dilution_factor
+            self.radiation_field_state.dilution_factor[
+                self.geometry.v_inner_boundary_index : self.geometry.v_outer_boundary_index
+            ] = new_dilution_factor
```
DeerWhale, line 138: Probably should have a test for the case that the v_inner_boundary is in between shells

jvshields (approved):

andrewfullard:
Unfortunately this will need a small rebase

tardis/model/parse_input.py
DeerWhale, line 1: These modifications made the radiation_field_state only carry the t_radiative and dilution_factor for the active shells.

DeerWhale:
> Unfortunately this will need a small rebase

Rebased to upstream/master now.

tardis/model/parse_input.py
```diff
@@ -720,6 +718,12 @@ def calculate_geometric_dilution_factor(geometry):
     return 0.5 * (
         1
         - np.sqrt(
-            1 - (geometry.r_inner[0] ** 2 / geometry.r_middle**2).to(1).value
+            1
+            - (
+                geometry.r_inner[geometry.v_inner_boundary_index] ** 2
```
andrewfullard, line 723: Could this be `geometry.r_inner_active[0]`?

tardis/model/parse_input.py
```diff
@@ -720,6 +718,12 @@ def calculate_geometric_dilution_factor(geometry):
...
+                / geometry.r_middle**2
```
andrewfullard, line 724: Double check that `r_middle` is correct here (matching the active radii)

tardis/model/parse_input.py
DeerWhale: Tried that but since r_inner_active[0] is overwrite by the v_inner_boundary value (which can be values in between the discrete shells), sometimes it can be a bigger value than the first element in r_middle, which cause the first w to be nan, and fails the assertion in the diluteBBradiationField.

tardis/model/parse_input.py
DeerWhale: r_middle is the raw shells, but if we would like to keep the raw shape of the shells in radiationfield, then we should use r_middle I think.

andrewfullard (approved):

tardis/model/radiation_field_state.py
```diff
@@ -17,18 +17,37 @@ class DiluteBlackBodyRadiationFieldState:
...
         # ensuring that the radiation_field has both
         # dilution_factor and t_radiative equal length
         assert len(t_radiative) == len(dilution_factor)
-        assert np.all(t_radiative > 0 * u.K)
-        assert np.all(dilution_factor > 0)
+        if (
+            geometry is not None
+        ):  # check the active shells only (this is used when setting up the radiation_field_state)
+            assert np.all(
+                t_radiative[
```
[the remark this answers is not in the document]
DeerWhale: That is actually already being taken of, since the estimators only use the active shells, the simulation_state slice the active shells from the "raw" radiation_field_state.

jvshields (approved):

PR #2509, opened 2024-02-02
Precommit

ntchen:
### :pencil: Description

.pre-commit-config.yaml is added. It integrates isort (new), ruff and black. ruff and black will check the local rules, e.g., .ruff.toml. I did not do `pre-commit run --all-files`, since it influences too many files. For new commits, it only scans the modified files.

Text need to be added to the documents:
#### Setting Up Pre-Commit Hooks

To ensure code quality and consistency, we use `pre-commit` hooks in this project. Please follow these steps to set up `pre-commit` on your local machine:

Install `pre-commit` by running:
`pip install pre-commit`

Set up the pre-commit hooks with:
`pre-commit install`

This needs to be done only once per repository. The pre-commit hooks will now automatically run on each commit to ensure your changes meet our code quality standards.

N/A

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)
- [x] it is an independent file

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -78,7 +78,7 @@ Gaurav Gautam <gautam1168@gmail.com> gautam1168 <gautam1168@gmail.com>
 Gerrit Leck <gerritleck@web.de>
 Gerrit Leck <gerritleck@web.de> Gerrit Leck <g.leck@gsi.de>
 
-Isaac Smith <smithis7@msu.edu> 
+Isaac Smith <smithis7@msu.edu>
 Isaac Smith <smithis7@msu.edu> Isaac Smith <71480393+smithis7@users.noreply.github.com>
 Isaac Smith <smithis7@msu.edu> smithis7 <71480393+smithis7@users.noreply.github.com>
 Isaac Smith <smithis7@msu.edu> smithis7 <smithis7@msu.edu>
@@ -163,6 +163,8 @@ Nathan Kowalski <56104113+kowal180@users.noreply.github.com>
 Nilesh Patra <37436956+nileshpatra@users.noreply.github.com>
 
 Nolan Brown <nolanbrown01@gmail.com>
+
+Nutan Chen <nutan.chen@gmail.com>
 
 Pratik Patel <pratikpatel15133@gmail.com>
 Pratik Patel <pratikpatel15133@gmail.com> Pratik151 <pratikpatel15133@gmail.com>
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,19 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.14
+    hooks:
+      - id: ruff
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+  - repo: https://github.com/psf/black
+    rev: 23.12.1
+    hooks:
+      - id: black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,3 @@ repos:
     rev: 5.13.2
     hooks:
       - id: isort
-  - repo: https://github.com/psf/black
-    rev: 23.12.1
-    hooks:
-      - id: black
--- a/docs/contributing/development/code_quality.rst
+++ b/docs/contributing/development/code_quality.rst
@@ -19,7 +19,7 @@ TARDIS follows the `PEP 8 <https://www.python.org/dev/peps/pep-0008/>`_ style gu
 Black
 -----
 `Black <https://black.readthedocs.io/en/stable/index.html>`_ is a PEP 8 compliant opinionated code formatter. At TARDIS. we use Black to automatically conform to PEP 8. It is already installed in the TARDIS conda environment, so all you have to do is to run Black before commiting your changes: ::
-    
+
     black {source_file_or_directory}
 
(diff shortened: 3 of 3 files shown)


ntchen:
I did not do `pre-commit run --all-files`, since it influences too many files.

ntchen:
if necessary, we can replace ruff (linter and formatter) with flak8 (linter only), since there is black (formatter). 

```
  - repo: https://github.com/pycqa/flake8
    rev: 7.0.0
    hooks:
      - id: flake8
 ```

**Comment by ntchen** [comment]
I just noticed another pull request for pre-commit configuration after creating this,  so this is not really necessary, but there is still some difference. #2508

**Comment by Rodot-** [comment]
> I just noticed another pull request for pre-commit configuration after creating this, so this is not really necessary, but there is still some difference. #2508

PR #2487

**Comment by Rodot-** [comment]
@AlexHls @andrewfullard How does this compare to the other pre-commit PR?

**Comment by AlexHls** [comment]
This adds different hooks and could be merged independently (I think @andrewfullard was against action actionlint anyways, so that PR can potentially be closed anyway)

File: .pre-commit-config.yaml [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+    hooks:
+      - id: ruff
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+  - repo: https://github.com/psf/black
+    rev: 23.12.1
+    hooks:
+      - id: black
```
AlexHls, line 19: I would not run black and ruff at the same time. The ruff config mimicks blak formatting, however there are some subtle differences.

Ultimately both formating and linting should be taken care of by ruff. Having both black and ruff doing the same thing will only lead to conflicts.

Since this PR should only affect new files the choice of formatter should not matter and as such I would vote for ruff being the sole formatting tool. Besides, the ``pyproject.toml`` already contains the formatting configuration for ruff.

.mailmap
```diff
@@ -78,7 +78,7 @@ Gaurav Gautam <gautam1168@gmail.com> gautam1168 <gautam1168@gmail.com>
 Gerrit Leck <gerritleck@web.de>
 Gerrit Leck <gerritleck@web.de> Gerrit Leck <g.leck@gsi.de>
 
-Isaac Smith <smithis7@msu.edu> 
+Isaac Smith <smithis7@msu.edu>
```
AlexHls, line 81: Why does this line change? Should be cleaned up.

AlexHls (changes requested):

andrewfullard:
My main concern with all precommit hooks is any potential barrier to development. Will this require a change to the TARDIS environment to use these tools, for example? How seamless is it?

ntchen:
> My main concern with all precommit hooks is any potential barrier to development. Will this require a change to the TARDIS environment to use these tools, for example? How seamless is it?

It is seamless and does not require a change. It checks the local rules, e.g., .ruff.toml, as written in the description. That's why I followed the previous setup to use ruff and black; otherwise, it can be ruff only or black + flake8 (this also answers the above @AlexHls ).

Users have the option to not use it. When utilized, it only scans the committed files.

.mailmap
ntchen: there was a space in the end and fixed by the pre-commit. 200 files have similar problems, if use `pre-commit run --all-files`, but not necessary to scan all now.

.pre-commit-config.yaml
andrewfullard: Does `ruff` perform the same formatting as `black`? Do they follow the same rules?

andrewfullard:
> > My main concern with all precommit hooks is any potential barrier to development. Will this require a change to the TARDIS environment to use these tools, for example? How seamless is it?
> 
> It is seamless and does not require a change. It checks the local rules, e.g., .ruff.toml, as written in the description. That's why I followed the previous setup to use ruff and black; otherwise, it can be ruff only or black + flake8 (this also answers the above @AlexHls ).
> 
> Users have the option to not use it. When utilized, it only scans the committed files.

Excellent. Please add the instructions for setup (the ones you have in the PR description are fine) to the developer's guide part of the documentation.

.pre-commit-config.yaml
AlexHls: Not necessarily. In principle it should be a ``black`` drop in replacement (see [docs](https://docs.astral.sh/ruff/formatter/#black-compatibility)). However, in a very limited number of cases it will fail the ``black`` checks. In an ideal word there would be a 'all in one' switch where the formatting provider is changed to ruff.

For reference, my IDE automatically applies ruff formatting and I never had issues. But I recall other people having issues.

In my opinion, ruff should be adopted as the default formatter moving forward and black should only be used case-by-case where the pipeline rejects the formatting.

andrewfullard:
@ntchen could you add the instructions for setup? I am also concerned about the black+ruff combo as Alex mentioned

.pre-commit-config.yaml
ntchen: removed black

ntchen:
> @ntchen could you add the instructions for setup? I am also concerned about the black+ruff combo as Alex mentioned

black has been removed and the instructions are in the document. can you take a look?

AlexHls:
@andrewfullard what do you think of adding a formatting commit (either in this pr or a separate one), enforcing ruff formatting and adding a ruff pre-commit hook to ensure consistent formatting. If we're concerned about the git blame history, a ``.git-blame-ignore-revs`` entry could be an option. 
I'm just concerned that otherwise is may take quite a while until ruff formatting takes over 'naturally'. (Not sure if this would require changes to the cicd.)

andrewfullard:
> @andrewfullard what do you think of adding a formatting commit (either in this pr or a separate one), enforcing ruff formatting and adding a ruff pre-commit hook to ensure consistent formatting. If we're concerned about the git blame history, a `.git-blame-ignore-revs` entry could be an option. I'm just concerned that otherwise is may take quite a while until ruff formatting takes over 'naturally'. (Not sure if this would require changes to the cicd.)

Well, given that `ruff` and `black` are currently interchangeable aside from line length, which we can set in ruff.toml, I don't think it's an issue unless you want to apply *all* of ruff's rules such as import sorting from the start.

andrewfullard (approved):

AlexHls (approved):

PR #2522, opened 2024-02-20
prevent float overflow when calculating He I population

DeerWhale:
### :pencil: Description

When the T_radiative in the shell is below 404 K, the highlighted part in the screenshot below will return Inf due to float overflow, which cause the nan values in the calculation of electron density. 

This PR initially aims to prevent the overflow by capping the value to be the max float value allowed in the program. However a better way to do this maybe raise an error when the input model T_radiative is below certain threshold (1000K is chosen in this PR), so the user can double check the input model. 

Fun fact, the low T_rad also cause kernel crash when using the macroatom setting. 

image

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/model/parse_input.py
+++ b/tardis/model/parse_input.py
@@ -678,6 +678,11 @@ def parse_csvy_radiation_field_state(
             geometry, packet_source
         )
 
+    if np.any(t_radiative < 1000 * u.K):
+        raise ValueError(
+            f"Radiative temperature is too low in shell {np.argmin(t_radiative)} (T_rad = {t_radiative[np.argmin(t_radiative)]})"
+        )
+
     if hasattr(csvy_model_data, "columns") and (
         "dilution_factor" in csvy_model_data.columns
     ):
--- a/tardis/model/parse_input.py
+++ b/tardis/model/parse_input.py
@@ -679,10 +679,10 @@ def parse_csvy_radiation_field_state(
         )
 
     if np.any(t_radiative < 1000 * u.K):
-        logging.warn(
+        logging.critical(
             "Radiative temperature is too low in some of the shells, temperatures below 1000K "
             f"(e.g., T_rad = {t_radiative[np.argmin(t_radiative)]} in shell {np.argmin(t_radiative)} in your model) "
-            "are not accurately handled by Tardis.",
+            "are not accurately handled by TARDIS.",
         )
 
     if hasattr(csvy_model_data, "columns") and (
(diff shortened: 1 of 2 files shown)


andrewfullard:
Good find. My issue with this solution is that nothing in a TARDIS simulation should ever be that cold!

DeerWhale:
> Good find. My issue with this solution is that nothing in a TARDIS simulation should ever be that cold!

The issue only occurs in the most outer layer in some of the cases (in some of the models there are shells >40k km/s with low T_rad ). 

Alternative is printing out a warning state why the overflow is encountered there, so the user can modify their models accordingly with that Info?

andrewfullard:
I would suggest a combination of both. A warning that the n_electron calculation has overflowed, telling the user that their temperatures are too low (with a printout of the problem temperature value and possibly the shells that are affected)

andrewfullard (dismissed):

tardis/model/parse_input.py
```diff
@@ -678,6 +678,11 @@ def parse_csvy_radiation_field_state(
             geometry, packet_source
         )
 
+    if np.any(t_radiative < 1000 * u.K):
+        raise ValueError(
+            f"Radiative temperature is too low in shell {np.argmin(t_radiative)} (T_rad = {t_radiative[np.argmin(t_radiative)]})"
```
jvshields, line 683: Maybe add a short description of why this is a problem. Like say, "Temperature is too low in shell, temperatures below 1000K (your chosen value) are not accurately handled by Tardis."

jvshields: Something that suggests what the threshold is, and what you need to do to adapt your model so as to avoid a crash.

jvshields (approved):

andrewfullard (approved):

andrewfullard (dismissed):
A warning is perhaps optimistic but we shall see how it goes

tardis/model/parse_input.py
```diff
@@ -678,6 +679,13 @@ def parse_csvy_radiation_field_state(
             geometry, packet_source
         )
 
+    if np.any(t_radiative < 1000 * u.K):
+        warnings.warn(
```
jvshields, line 683: please use logging.warn (and remove the warnings import)

DeerWhale: Done.

jvshields (approved):

wkerzendorf (approved):

Issue #2519, opened 2024-02-13
Convergence plot correction 


jamesgillanders:
**Describe the bug**
In the convergence plot widget, both T and W are plotted against shell velocity. These plots should present their data as a step function, instead of a scatter line, since the T and W values are constant throughout each individual shell.

**To Reproduce**
Run the convergence plot.

Vivek-Joshi17:
Sir, I'm willing to work on this issue with your permission.

jamesgillanders:
Go for it!

On 25 Feb 2024, at 17:59, Vivek Joshi ***@***.***> wrote:



Sir, I'm willing to work on this issue with your permission.

sarthak-dv:
@jamesgillanders Since this issue was pending since weeks, I have taken the liberty to fix it.
[PR containing fix](https://github.com/tardis-sn/tardis/pull/2528).
Please review, thanks in advance!

PR #2528, opened 2024-03-04
Convergence plot correction: Use step plots

sarthak-dv:
### :pencil: Description

**Type:** :beetle: `bugfix`

In the convergence plot widget, both T and W are plotted against shell velocity. These plots should present their data as a step function, instead of a scatter line, since the T and W values are constant throughout each individual shell.

convergence

closes [Issue#2519](https://github.com/tardis-sn/tardis/issues/2519)

@jamesgillanders @MarkMageeAstro Please review. Also can you please apply the `build_docs` label since the plots in the docs will change?

- [ ] Testing pipeline
- [x] Other method: Ran convergence plot to see the plot line shape change


sarthak-dv:
@jamesgillanders @MarkMageeAstro Can you please review this fix when you have time?
Also can you please apply the `build_docs` label since the plots in the docs will change?

Thanks in advance!

atharva-2001:
hi, can you please put up a screenshot showing the changes you made in the pr description?

sarthak-dv:
@atharva-2001 Done. Also, can you please put the `build_docs` label on this PR?

MarkMageeAstro:
Maybe someone else knows, but is 'hv' or 'vh' correct? 

Each shell is defined by an inner and outer velocity, so does the temperature, etc. correspond to the inner or outer boundary? I guess that'll determine which one to use or maybe I'm misunderstanding something

andrewfullard:
> Maybe someone else knows, but is 'hv' or 'vh' correct?
> 
> Each shell is defined by an inner and outer velocity, so does the temperature, etc. correspond to the inner or outer boundary? I guess that'll determine which one to use or maybe I'm misunderstanding something

This is actually a rather interesting question about how TARDIS works. Temperature and density are actually defined at the *centre* of the cell.

jamesgillanders:
Temperature etc. are constant throughout each shell. The point to consider and to address is how that looks on a step plot that is generated by using e.g., v_inner values. In that case the step profile should not be "mid"

sarthak-dv:
@jamesgillanders @andrewfullard @MarkMageeAstro 

I believe "hv" is correct given that we're considering inner velocities. (Reference for different modes [here](https://chart-studio.plotly.com/~cjtorresj/30.embed). "hv" changes value at the current x-coordinate (inner_boundary_i) and stays put until the next x-value (inner_boundary_{i+1} = outer_boundary_i)

If I chose outer velocities, then "vh" would have been correct.

"hvh" would be correct if I chose mid velocities for the shells.

sarthak-dv:
@MarkMageeAstro

The plot itself steps down at the correct locations, as seen in the attached image:

convergence_step_example

The last step down is at the inner boundary of the last shell. However, it isn't extended in a horizontal line because the plot doesn't know where to "end" the line segment i.e. outer boundary of last shell).
To fix this, I'll have to additionally pass the outer boundary of the last shell with the same t_rad i.e. (x=v_outer[-1], y=t_rad[-1])

Let me know your thoughts here.

MarkMageeAstro:
Okay, I see where the confusion has come now. In your first plot, I assumed the step down was at the highest velocity (i.e. v_outer of the last shell), but it's clear from your latest version that's not the case. It's also more obvious which velocities correspond to the temperatures when you print out the complete boundaries for each cell:

Screenshot 2024-03-12 at 12 30 19

Yes, I think it would be good to make sure the line starts at the inner boundary and extends to the outer boundary of the full model.

sarthak-dv:
@MarkMageeAstro I have updated my branch with the changes. Now the plot looks like this:

convergence_plot_corrected


It's weird that I force-pushed the recent changes to this branch after rebasing master, but they don't show up on the pull request.
However, when I see the [diff between my branch and tardis master,](https://github.com/tardis-sn/tardis/compare/master...sarthak-dv:tardis:convergence-plot-correction) I can see these changes.

sarthak-dv:
Can you also please rerun the checks, since the PR was behaving weirdly in between

jamesgillanders (dismissed):
Approved!

sarthak-dv:
Thanks James!

@MarkMageeAstro This required another approval before merge. Please let me know if this is good to go.

andrewfullard:
It looks like the convergence plot tests are failing, and `black` needs to be run on the code.

sarthak-dv:
@andrewfullard I've fixed the convergence plot tests and ran Black formatter on the file.

sarthak-dv:
@MarkMageeAstro Can you please review this? @jamesgillanders has already approved and it needs two approvals to merge

MarkMageeAstro (approved):

PR #2529, opened 2024-03-04
PR to separate main gamma ray loop

Knights-Templars:
This is a draft PR. In this PR, we have functions calculating the distribution of packets to each isotope, fractional energy per isotope per decay chain in each shell. We also separate the main_gamma_ray_loop from the gamma_ray_transport.


tardis/energy_input/gamma_ray_transport.py
```diff
@@ -505,6 +513,36 @@ def decay_chain_energies(
...
+                    else 0.0
+                )
+                for isotopes in decay_energy[shell][parent_isotope]
+            }
+            for parent_isotope in decay_energy[shell]
+        }
+        for shell in decay_energy
+    }
+
+    return fractional_decay_energy
```
wkerzendorf, line 543: that looks complex - would a dataframe be better here?

tardis/energy_input/main_gamma_ray_loop.py
```diff
@@ -0,0 +1,285 @@
...
+    print(packets_per_isotope_df)
+    total_energy = energy_df.sum().sum()
+    print("Total gamma-ray energy:")
+    print(total_energy * u.keV.to("erg"))
+
+    iron_group_fraction = iron_group_fraction_per_shell(model)
+    number_of_packets = packets_per_isotope_df.sum().sum()
+    print("Total number of packets:", number_of_packets)
+    individual_packet_energy = total_energy / number_of_packets
+    print("Energy per packet:", individual_packet_energy)
```
wkerzendorf, line 151: please use log statements.

tardis/energy_input/GXPacket.py
```diff
@@ -32,6 +33,8 @@ class GXPacketStatus(IntEnum):
     ("shell", int64),
     ("time_current", float64),
     ("tau", float64),
+    ("Z", int64),
```
wkerzendorf, line 36: I get the feeling that using the same infrastructure as the tracker is better than attaching source to each packet @andrewfullard

andrewfullard: I agree.

tardis/energy_input/gamma_ray_transport.py
Knights-Templars: Noted. Thanks!

tardis/energy_input/main_gamma_ray_loop.py
Knights-Templars: Noted.

PR #2531, opened 2024-03-07
[GSOC] Created a memory usage profile of the TARDIS example simulations

officialasishkumar:
### :pencil: Description

This PR is completes the second part of the objective of GSoC'24 idea "TARDIS Benchmarking and Performance Improvement". I have used memray and scalene to generate the memory usage profile. I have used the example simulation mentioned in [quickstart](https://tardis-sn.github.io/tardis/quickstart.html). 

Memray report:
image

Scalene report:
image

### :pushpin: Resources
https://bloomberg.github.io/memray
https://github.com/plasma-umass/scalene

- [ ] Testing pipeline
- [ ] Other method (describe)
- [x] My changes can't be tested (explain why)

cc: @atharva-2001 


wkerzendorf:
Could you write a bit about what you found? What you might want to optimize?

officialasishkumar:
After going through the memory profiler, here are the functions/lines which are taking most memory and the possible optimizations that we can use: 

1. [montecarlo_main_loop](https://github.com/tardis-sn/tardis/blob/39c128afa1be54ed2dac3fdf11e098a87b432b37/tardis/montecarlo/montecarlo_numba/base.py#L31): This function itself takes more than 2GB of memory (after checking on scalene). Possible optimizations which we can adopt to make it take less memory are: 
 - we can create reusable objects (r_packet, vpacket_collection, rpacket_tracker) outside the loop and update their properties inside the loop instead of creating new objects for each packet. 
 - analysing where copy is not necessary since a lot of np.copy() is happening in this function. 
 - maybe doing in place operations in some places 
 - Where appropriate, replacing loops with list comprehensions or use map/filter functions, which can be more memory-efficient.

2. [tardis/montecarlo/montecarlo_numba/numba_interface.py](https://github.com/tardis-sn/tardis/blob/39c128afa1be54ed2dac3fdf11e098a87b432b37/tardis/montecarlo/montecarlo_numba/numba_interface.py#L169): This line takes around 600MB because of the copy operation. We can avoid it since the value doesn't change while shifting it to contigous block in memory. 

3. Being more specific with the import statements and only import packages that are necessary

> Could you write a bit about what you found? What you might want to optimize?

officialasishkumar:
Updated: Added screenshots, html and replaced jupyter notebook with python code.

andrewfullard:
Great! Can you describe what scalene does differently to memray?

PR #2535, opened 2024-03-09
Path.replace bug

Sumit112192:
### :pencil: In file tardis/io/atom_data/util.py 
`fname` sometimes becomes a path object causing the bug, and on line 38, we seem to apply the `replace` method of `str` on the path object, which has its own `replace` method.

In the file `tardis/simulation/base.py` on line 665, atom_data_fname becomes a path object, causing this error. A better way would be to make atom_data_fname str here. Could anyone suggest a better wat to fix this bug? 
https://github.com/tardis-sn/tardis/blob/39c128afa1be54ed2dac3fdf11e098a87b432b37/tardis/simulation/base.py#L660-L667

**Type:** :beetle: `bugfix`

### :pushpin: Resources

image


Sumit112192:
@wkerzendorf @atharva-2001 Can you please review this request? Thanks.

sarthak-dv:
That's a great catch @Sumit112192 !

I think another way is that instead of manually replacing ".h5" with "" (which seems a slightly dirty thing to do), we stick to Path object, as use `with_suffix` function, as below:

```
In [12]: path
Out[12]: PosixPath('/Users/sarthak/Downloads/example.jpeg')

In [13]: path.with_suffix("")
Out[13]: PosixPath('/Users/archil/Downloads/example')
```

Of course we'll have to make sure that this change is backward compatible, i.e. the function `resolve_atom_data_fname` continues to accept path and strings too.

andrewfullard:
Could you instead ensure that `fname` is always a `Path` object instead of casting to a string?

Sumit112192:
In the below code snippet, 
inside the body of the `if` statement, `fname` is initialized as a `Str` object ( Line 663 ), and  
inside the body of the `else` statement, `fname` is initialized as a `Path` object ( Line 666 ).
https://github.com/tardis-sn/tardis/blob/f515d16a20c7e70c593c18a74e62f456e804875d/tardis/simulation/base.py#L661-L667

In the `resolve_atom_data_frame` function
https://github.com/tardis-sn/tardis/blob/f515d16a20c7e70c593c18a74e62f456e804875d/tardis/io/atom_data/util.py#L13-L36

It's supposed to have str as input and str as output.

Should I instead convert the Path object on line 666 to a `Str` object as that seems to resolve the issue and carry on the before objective of `resolve_atom_data_fname`

@andrewfullard @sarthak-dv

andrewfullard:
I would prefer if you modify `resolve_atom_data_fname` to use `Path` objects instead

Sumit112192:
@andrewfullard Okay, I will do that.

andrewfullard:
@Sumit112192 do you have an expected timeline for this PR?

Sumit112192:
I am really sorry for the delay. I had some college work to do this week. I will start working on it from Friday night. @andrewfullard

andrewfullard:
Please run `black` on your changes.

andrewfullard (approved):
Thanks for this fix!

DeerWhale:
Thanks for spotting the bug and the fix! I think this fix the case when the atom data file in the same directory as the config file, however I still get error when I only specific the name of the atom_data file (not the absolute path) that exist in the tardis_data folder, which should still work as one of the check in the linked part of code below:
https://github.com/tardis-sn/tardis/blob/ab869ea8586302b6f67cf95f4ed6a11235c9cf76/tardis/io/atom_data/util.py#L31-L36

Could you attempt to fix those cases as well? Thanks!

tardis/io/atom_data/util.py
```diff
@@ -35,9 +35,9 @@ def resolve_atom_data_fname(fname):
         )
         return fpath
 
-    atom_data_name = fname.replace(".h5", "")
+    atom_data_name = fname.with_suffix("")
```
DeerWhale, line 38: Since fname is an absolute path coming from "atom_data_fname" in simulation/base.py in stead of the file name only, should we use `fname.stem` here as suggested by https://github.com/tardis-sn/tardis/pull/2535#issuecomment-1987317427?

Sumit112192:
> Thanks for spotting the bug and the fix! I think this fix the case when the atom data file in the same directory as the config file, however I still get error when I only specific the name of the atom_data file (not the absolute path) that exist in the tardis_data folder, which should still work as one of the check in the linked part of code below:
> 
> https://github.com/tardis-sn/tardis/blob/ab869ea8586302b6f67cf95f4ed6a11235c9cf76/tardis/io/atom_data/util.py#L31-L36
> 
> Could you attempt to fix those cases as well? Thanks!

Sure, why not !!!

Sumit112192:
@DeerWhale Are the changes good?

andrewfullard (approved):

DeerWhale (approved):

DeerWhale:
> @DeerWhale Are the changes good?

Look great! Thanks a lot for the fix!

PR #2549, opened 2024-03-15
Path replace bug

rudrakshnalbalwar:
Type : Bug Fix

Description: Addressing Inconsistent File Path Handling in tardis
Description:

This pull request fixes a bug in tardis related to how file paths are handled in tardis/io/atom_data/util.py and potentially other parts of the codebase.

Problem:

The variable fname in resolve_atom_data_fname can be either a string representing a filename or a path object. When fname is a path object, attempting to use the replace method on line 38 leads to an error because path objects have their own replace method. This inconsistency in data types causes unexpected behavior.

Solution:

This pull request implements the following fix:

Explicit Type Conversion (Recommended): The code is modified to ensure fname is always a string before applying the replace method. This is achieved by checking if fname is an instance of os.PathLike (a common base class for path-like objects) and converting it to a string using str(fname).
Benefits:

Fixes the bug causing errors when fname is a path object.
Improves code clarity and maintainability by ensuring consistent data types.
Testing:

Unit tests (if available) should be updated to verify the fix.
Manual testing can be performed to ensure the functionality of resolve_atom_data_fname and related functions is not affected.
Additional Notes:

Similar issues in other parts of the codebase (e.g., tardis/simulation/base.py) might need to be addressed.
Consider using type hints (PEP 484) to explicitly define expected data types for function arguments and return values.

Please review this pull request and let me know if you have any questions.


andrewfullard:
Thanks for the submission. We went with #2535 instead, but you can still refer to this PR in your GSoC application as a reference for your code.

PR #2560, opened 2024-03-23
Bug t radiative

Sumit112192:
### :pencil: Description

**Type:** :beetle: `bugfix`| memo: `documentation`
image
ref. https://tardis-sn.github.io/tardis/physics/setup/model.html#Temperatures

https://github.com/tardis-sn/tardis/blob/1718002b4db7fd1a93e475844582ed9b41aa7cdf/tardis/model/parse_input.py#L726-L731
The code doesn't match with the documentation. So, I have changed it accordingly.
I have directly used `packet_source.temperature` rather than first computing `lambda_wien_inner`

Slight documentation typos: 
The documentation string of `parse_structure_config` was repeated twice, so I removed the extra one.
Also, there were some instances where `v_inner, v_outer` were labeled as `r_inner, r_outer`. So, I have changed them as well.

Also, If we pass `packet_source` as one of the function arguments, why define it again in the body?
https://github.com/tardis-sn/tardis/blob/1718002b4db7fd1a93e475844582ed9b41aa7cdf/tardis/model/parse_input.py#L589-L618
### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

@andrewfullard @jamesgillanders Please review this pull request.


tardis/model/parse_input.py
```diff
@@ -723,10 +687,8 @@ def calculate_t_radiative_from_t_inner(geometry, packet_source):
     Quantity
         The calculated radiative temperature.
     """
-    lambda_wien_inner = const.b_wien / packet_source.temperature
-    t_radiative = const.b_wien / (
-        lambda_wien_inner
-        * (1 + (geometry.v_middle - geometry.v_inner_boundary) / const.c)
+    t_radiative = packet_source.temperature / (
+        (1 - (geometry.v_middle - geometry.v_inner_boundary) / const.c)
```
andrewfullard, line 691: The documentation is incorrect here: the `1 - ` should be `1 + ` because temperature should *decrease* further out. The other change (packet_source.temperature) is fine.

andrewfullard: You can modify the documentation to fix the sign of the function

DeerWhale:
Thanks for spotting the duplication of the packet initialization, I have opened an issue #2585 and our team members will look into this further. Feel free to dive deeper if you are interested as well!

tardis/model/parse_input.py
Sumit112192: Yes !!! That makes sense. I will change the documentation and will redo this change to the original.

Sumit112192:
@andrewfullard I have made the necessary changes.

The changes that I have made till now.
1. Updated the documentation so that doppler factor is $1+\frac{\Delta v}{v}$ instead of $1-\frac{\Delta v}{v}$
2. Fixed some common typos in the documentation strings of the functions
3. Change suggested by Black in the file `tardis/model/parse_input.py`

Sumit112192:
I don't know why Black's test is failing on `tardis/model/parse_input.py`, but I have made the change that Black suggested.

image

Sumit112192:
It seems that the suggestions by black installed through `sudo apt install black` on Ubuntu are different from those of black installed through `pip install black`.

Sumit112192:
Sorry for all those commits. I am still trying to figure out how branching works in Git, and hopefully, I will not repeat the same mistake.

andrewfullard:
> It seems that the suggestions by black installed through `sudo apt install black` on Ubuntu are different from those of black installed through `pip install black`.

Make sure you use the version of black that is installed with the TARDIS environment

andrewfullard (approved):

tardis/model/parse_input.py
```diff
@@ -33,42 +33,6 @@ def parse_structure_config(config, time_explosion, enable_homology=True):
...
-    -----
-    This function parses the structure configuration data and returns the parsed electron
-    densities, temperature, geometry, and density. The structure configuration can be of
-    type 'specific' or 'file'. If it is of type 'specific', the velocity and density are
-    parsed from the configuration. If it is of type 'file', the velocity and density are
-    read from a file. The parsed data is used to create a homologous radial 1D geometry object.
-    """
-    """
-    Parse the structure configuration data.
-
```
DeerWhale, line 71: Is this deletion of the doc strings intended?

Sumit112192: They were repeated twice.

DeerWhale: Sorry my bad didn't notice it, thanks for spotting the duplicates!

Knights-Templars (approved):

PR #2119, opened 2022-08-21
adding RPacketPlotter to TARDIS

jayantbhakar:
### :pencil: Description

**Type:** :rocket: `feature`

This PR is based on the Montecarlo 2D packet visualization project. It aims to write a module for the RPacketPlotter tool which creates a 2D animated graph of the RPackets or the real packets used in the Montecarlo simulation.

Example code to plot is shown below:

`rpacket_plotter = RPacketPlotter.from_simulation(sim, no_of_packets=10)`
`rpacket_plotter.generate_plot()`

Here's how the plot looks like:

Light Theme:

Screencast-from-02-09-22-12-47-5

Dark theme:

darktheme_PR

Also, link issues affected by this pull request by using the keywords: `close`, `closes`, `closed`, `fix`, `fixes`, `fixed`, `resolve`, `resolves` or `resolved`. 

### :pushpin: Resources

Examples, notebooks, and links to useful references.

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing locally: Here is the log of pytests that I have run locally.
  
Screenshot from 2022-10-10 10-05-38

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #2562, opened 2024-03-26
Reference Data Generation Fixes

atharva-2001:
### :pencil: Description

**Type:** :beetle: `bugfix`  :vertical_traffic_light: `testing` |

- [X] change hdf properties for simulation.transport

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ on:
 
 env:
   CACHE_NUMBER: 0  # increase to reset cache manually
-  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data
+  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data -vvv
                 --cov=tardis --cov-report=xml --cov-report=html 
   CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
@@ -172,6 +172,12 @@ jobs:
       - name: Run tests
         run: pytest tardis ${{ env.PYTEST_FLAGS }}
         working-directory: ${{ env.TARDIS_PIP_PATH }}
+        if: always()
 
       - name: Upload to Codecov
         run: bash <(curl -s https://codecov.io/bash)
+      
+      - name: Refdata Generation tests
+        run: pytest tardis ${{ env.PYTEST_FLAGS }} --generate-reference
+        working-directory: ${{ env.TARDIS_PIP_PATH }}
+        if: always()
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ on:
 
 env:
   CACHE_NUMBER: 0  # increase to reset cache manually
-  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data -vvv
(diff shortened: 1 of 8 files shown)


tardis/tests/test_tardis_full.py
```diff
@@ -74,24 +71,24 @@ def get_ref_data(key):
 
         return get_ref_data
 
-    def test_j_blue_estimators(self, transport, refdata):
-        j_blue_estimator = refdata("j_blue_estimator").values
+    # def test_j_blue_estimators(self, transport, refdata):
```
andrewfullard, line 74: I assume these comments are temporary

atharva-2001: yes, should be fixed once I make the RadiationFieldMCEstimators a subclass of `HDFWriterMixin` to save it as an HDF.

andrewfullard:
Can you explain the test failures?

atharva-2001:
It's because the reference data needs to be updated. The update refdata workflow will not work on this PR because the YAML file on the master does not have steps which add the regression tests settings. This PR fixes that. Once this PR is in, we can either run the workflow or I can update the repo and the tests should work then.

andrewfullard:
This is simply a change in *where* the reference data are located in the HDF file, and not a change in the numbers, right?

.github/workflows/tests.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 env:
   CACHE_NUMBER: 0  # increase to reset cache manually
-  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data
+  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data -vvv
```
andrewfullard, line 18: Should `-vvv` be removed now?

.github/workflows/tests.yml
```diff
@@ -172,6 +172,12 @@ jobs:
       - name: Run tests
         run: pytest tardis ${{ env.PYTEST_FLAGS }}
         working-directory: ${{ env.TARDIS_PIP_PATH }}
+        if: always()
```
andrewfullard, line 175: Should this be removed before merge?

.github/workflows/tests.yml
atharva-2001: I was thinking to keep it, it helps tell us exactly what tests are skipped right?

.github/workflows/tests.yml
atharva-2001: I can, but I didn't want any other step to influence(or stop) this step. Is that the right thing to do?

.github/workflows/tests.yml
andrewfullard: I agree, but it also makes the logs very long. I think it should be removed unless we are doing specific debugging.

.github/workflows/tests.yml
andrewfullard: I see, so even if other installation steps fail, it will always attempt to run the tests?

atharva-2001: The same with the refdata generation step, it could happen that these tests fail and those pass, so I didn't want above steps to influence this, if that is alright

atharva-2001: yes

atharva-2001:
> This is simply a change in where the reference data are located in the HDF file, and not a change in the numbers, right?

No, I don't think there is a change in numbers.

.github/workflows/tests.yml
atharva-2001: Done in 9dd7c854ffa842e3e40f6317bc7533269f33a97e

andrewfullard:
I would like to get this merged ASAP, but it would be best after #2119 because that changes quite a few files.

andrewfullard:
After rebasing #2119 it actually changes only a few files (good!) so this should be mergeable now?

andrewfullard (approved):

PR #2525, opened 2024-02-28
Montecarlo configuration refactor

andrewfullard:
Absolutely horrible refactor that converts the Montecarlo configuration globals into a jitclass

PR #2568, opened 2024-04-01
Montecarlo configuration refactor, Numba developer proposed version

andrewfullard:
### :pencil: Description

**Type:** :beetle: `bugfix`

This is the version of the refactor that uses the methods proposed here: https://hackmd.io/Wmy-e2RsQCSqjJsP0VAmbg?both which allow us to send config objects to different functions with smart compilation that accounts for changes in the config while retaining the performance benefits of skipping paths (this effectively compiles JIT functions to each config).

- [x] Testing pipeline
- [ ] Other method (describe)


andrewfullard:
Total test duration (Moria): 683s
Total test duration #2525 (Moria): 622s

This is likely because the obj2strdict is being called in the single packet loop.

andrewfullard:
need to rebase

andrewfullard:
Note to self: rebase

andrewfullard:
Not relevant any more

PR #2573, opened 2024-04-04
Bugfix: Added a conditional statement and changed the error type that was being handled in `tardis.io.model_reader.read_uniform_abundances`.

KC210404:
### :pencil: Description

**Type:** :beetle: `bugfix`

While working on the first objective plots for GSOC'24, I found a bug in the function: tardis.io.model_reader.reader.read_uniform_abundances (model_reader.py file). 

`Bug Description:` According to the schema of the abundances sub-section of the model section, after validation, the key model_isotope_time_0 gets added to the validated abundances section. However, when we pass this validated abundances section to the function read_uniform_abundances, there is no condition to handle model_isotope_time_0. Neither is it a "type" nor is it an elemental or isotopic species. Consequently, it reaches the last else statement (nested-else), where actually the isotopic species should go to, which are passed to the Nuclide of the radioactivedecay library of Python to get the multi-level index of the species in the isotope_abundance DataFrame. But since model_isotope_time_0 is not an isotopic species and is a compulsory inclusion at the validation stage, it must be handled by adding a string comparison like with the "type"-key in the abundances section. We can't loop until the last but one element (model_isotope_time_0 is the last key), because in cases where the config-dict is not validated then this key wouldn't be added, and we would lose the last species abundance.

Secondly, we need to handle such strings which are not either elemental or isotopic species for which the try-except block with a runtime error getting excepted is already added. However, when such strings are passed to the Nuclide function, radioactivedecay.utils.NuclideStrError gets generated, not a runtime error. So these two changes were done to the code.

Example: validated-abundances section of tardis_example.yml file: {'type': 'uniform', 'O': 0.19, 'Mg': 0.03, 'Si': 0.52, 'S': 0.19, 'Ar': 0.04, 'Ca': 0.03, 'model_isotope_time_0': <Quantity 0. s>}

`Note:` Key inclusions such as the "model_isotope_time_0" key occur at the validation stage for different sections of the YAML file configuration. So, when a specific section is parsed from the validated configuration dictionary (which is done by the functions in model_reader.py), it is important that such key inclusions be handled and debugged.

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/io/model_reader.py
```diff
@@ -169,7 +169,7 @@ def read_uniform_abundances(abundances_section, no_of_shells):
     )
 
     for element_symbol_string in abundances_section:
-        if element_symbol_string == "type":
+        if element_symbol_string == "type" or element_symbol_string == "model_isotope_time_0":
```
DeerWhale, line 172:
```suggestion
        if (
            element_symbol_string == "type"
            or element_symbol_string == "model_isotope_time_0"
        ):
```

DeerWhale (approved):

jvshields (approved):

PR #2575, opened 2024-04-05
Added time stamp in mass fraction dataframe

Knights-Templars:
### :pencil: Description

**Type:** :rocket: `feature` 

added time stamp for the mass fraction dataframe as shown in the notebook.

### :vertical_traffic_light: Testing feature

- [X] Testing needs to be implemented


andrewfullard:
The shell index seems to have gone missing

tardis/energy_input/gamma_ray_channel.py
```diff
@@ -3,13 +3,64 @@
...
+        isotope time coming from HESMA model (typically 100 s) in days.
+
+    Returns
+    -------
+    isotope_mass_fraction : pd.DataFrame
+        isotope mass fractions at different times including the initial time.
+    """
+
+    # linear time spacing
+    time_delta = np.linspace(time_start, time_end, time_step)
```
andrewfullard, line 44: Instead of taking in the start/end/step, please use whatever time array is passed in (so that we can support log-space time for example)

Knights-Templars: Changed!

Knights-Templars:
> The shell index seems to have gone missing

Shell indices are column names.

tardis/energy_input/gamma_ray_channel.py
```diff
@@ -3,13 +3,57 @@
...
+    model_isotope_time,
+):
+    """
+    Function to evolve the isotope abundance over time.
+
+    Parameters
+    ----------
+    raw_isotope_mass_fraction : pd.DataFrame
+        mass fractions of isotopes.
+    time_delta : np.array
```
andrewfullard, line 26: Is this actually a time delta? We have to be careful to pick the correct time to measure the decay at. I would suggest the "effective time" array, which is the central time of each time bin.

Knights-Templars:
This PR will be part of connecting the data frame with the packet source.

PR #2578, opened 2024-04-11
Updating embedded links in docs with new file names "tutorial_" and "how_to_"

erinvisser:
### :pencil: Description

Renaming files, in [PR #2563](https://github.com/tardis-sn/tardis/pull/2563), based on their type of documentation (following Diataxis framework) caused issues in the documentation, because these files are embedded throughout the documentation.
I went through and found circumstances of these embedded links and updated the file names. I also went through the index pages for each section and updated the file names accordingly.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)


DeerWhale (approved):

jvshields:
This looks like a positive change to me, but I do have one comment maybe worth discussing on a larger level. I'm not sure calling things "tutorial" will signify a difference compared to a "how to" to somebody trying approaching our documentation to learn. We've talked a good bit now about the difference, and the goals of one versus another, but before we talked about those things I'm not sure I would have seen a difference. 

Trying to put myself in the shoes of somebody approaching our documentation, I think if I saw either "how to" or "tutorial" I'd expect to be learning about how to use the software in either way (though people can disagree if they think otherwise). Maybe instead of saying "tutorial" we should aim for a more descriptive title for the tutorials to make it clear that those tutorials aim to teach somebody how the thing works. Like instead of "tutorial_convergence_plot" it would be called something like "Constructing_and_customizing_convergence_plots." 

This sounds much harder and like a lot more work, but I'm mostly not convinced that the word "tutorial" will mean something specific to a user the way it means something to us, having talked about diataxis.

jvshields (approved):

PR #2587, opened 2024-04-16
Trigger docs only when something inside the docs folder changed

KasukabeDefenceForce:
### :pencil: Description

I am trying to implement a conditional logic in the docs workflow so that the docs trigger only if there is a change in any of the files in the docs directory.

### :pushpin: Resources

- [x] Testing pipeline
- [ ] Other method (describe)


andrewfullard:
Why are the tests failing in this PR?

atharva-2001:
Can you please include example workflows- conditions when the workflow ran successfully?

atharva-2001 (changes requested):
Thanks, Abhinav, I believe the workflow will now run if there are changes in the docs folder. But someone will still be able to run it manually too using the build-docs label(without skipping), as a test?

atharva-2001:
> Why are the tests failing in this PR?
The regression cache was corrupt(size was in bytes). Deleted the cache and now it works.

.github/workflows/build-docs.yml
```diff
@@ -37,47 +36,63 @@ defaults:
...
         if: github.event_name == 'pull_request_target'
 
+      - name: Check for changes in documentation
+        run: |
+          if git diff origin/master..."$(git rev-parse --abbrev-ref HEAD)" --name-only | cat | grep 'docs' | grep -q .; then
+              num_files=$(git diff --name-only origin/master...HEAD | grep '^docs/' | wc -l)
+              echo "Changes found in documentation files: $num_files"
+              exit 0
+          else
+              echo "No changes found in documentation files"
```
wkerzendorf, line 63:
```suggestion
              echo "No changes found in documentation files - will stop running the pipeline."
```

KasukabeDefenceForce:
> Thanks, Abhinav, I believe the workflow will now run if there are changes in the docs folder. But someone will still be able to run it manually too using the build-docs label(without skipping), as a test?

What should be its behavior when we manually run it? If there are no changes in the docs then it should not do anything or it should run all the steps?

andrewfullard:
With `build-docs` label: should always run the full docs build on PR commits
With force run in actions list: run the full docs build
With docs change: automatically run the full docs build

.github/workflows/build-docs.yml
```diff
@@ -37,47 +36,63 @@ defaults:
...
+          activate-environment: tardis
+          use-mamba: true
+        if: steps.docs_check.outcome == 'success'
 
       - uses: actions/cache@v2
         with:
           path: /usr/share/miniconda3/envs/tardis
           key: conda-linux-64-${{ hashFiles('conda-linux-64.lock') }}-${{ env.CACHE_NUMBER }}
         id: cache-conda
+        if: steps.docs_check.outcome == 'success'
```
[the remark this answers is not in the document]
KasukabeDefenceForce: Yes, the redundant step is removed and now it initially checks for changes and exits if there are no changes, showing a green tick.

KasukabeDefenceForce:
> Thanks, Abhinav, I believe the workflow will now run if there are changes in the docs folder. But someone will still be able to run it manually too using the build-docs label(without skipping), as a test?

Yes, we can run it manually, by either manually running the workflow or by adding the build docs label to the PR.

.github/workflows/build-docs.yml
KasukabeDefenceForce: i have updated the message.

atharva-2001:
There is a merge conflict. Also, can you provide example workflow links?

.github/workflows/build-docs.yml
```diff
@@ -36,42 +35,73 @@ defaults:
...
+              exit 1
+          fi
+        id: docs_check
+        if: steps.trigger_check.outcome != 'success'
+        continue-on-error: true
 
+  build-docs:
     runs-on: ubuntu-latest
+    needs: Check-for-changes
+    if: needs.Check-for-changes.outputs.trigger-check-outcome == 'success' || needs.Check-for-changes.outputs.docs-check-outcome == 'success'
```
atharva-2001, line 79:
```suggestion
    if: needs.check-for-changes.outputs.trigger-check-outcome == 'success' || needs.check-for-changes.outputs.docs-check-outcome == 'success'
```

.github/workflows/build-docs.yml
```diff
@@ -36,42 +35,73 @@ defaults:
     shell: bash -l {0}
 
 jobs:
-  build-docs:
+  Check-for-changes:
```
atharva-2001, line 38:
```suggestion
  check-for-changes:
```

.github/workflows/build-docs.yml
KasukabeDefenceForce: I have renamed the job using lower case.

.github/workflows/build-docs.yml
KasukabeDefenceForce: I have renamed the job using lower case.

andrewfullard (approved):

atharva-2001 (approved):

PR #2603, opened 2024-04-30
NumberDensity as Input to plasma

wkerzendorf:
first start of carving a position for the new plasma


tardis/model/base.py
```diff
@@ -161,6 +161,45 @@ def t_radiative(self, new_t_radiative):
...
+            :,
+            self.geometry.v_inner_boundary_index : self.geometry.v_outer_boundary_index,
+        ]
+        elemental_number_density.columns = range(
+            len(elemental_number_density.columns)
+        )
+        return elemental_number_density
+
+    @property
+    def isotopic_number_density(self):
```
andrewfullard, line 184: I believe the composition.isotopic_number_density produces the correct values because it uses isotope masses rather than the effective element masses, so it should be possible to use that instead.

andrewfullard (dismissed):
It would be good to address my comment to reduce code duplication, but it's not a blocker.

tardis/model/base.py
Knights-Templars: Please make an issue!

andrewfullard (approved):

andrewfullard:
This can be merged if docs build locally

atharva-2001 (approved):

PR #2651, opened 2024-06-10
Last Interaction radius Plot

sarthak-dv:
### :pencil: Description

**Type:** :rocket: `feature`

This PR is part of the Velocity Packet Tracker Visualisation tool. It introduces a new Python script interaction_radius_plot.py containing the InteractionRadiusPlotter class and a Jupyter notebook interaction_radius_plot.ipynb to demonstrate accessing the class and visualizing the plot.
The plots are generated using both Matplotlib and Plotly.

This PR aims to discuss and finalize the look and feel of the Last Interaction Radius plot.

Matplotlib:
mpl

Plotly:
plotly

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -14,18 +14,13 @@ from tardis.io.configuration import config_reader
 from tardis.io.configuration.config_reader import Configuration
 from tardis.io.util import YAMLLoader, yaml_load_file
 from tardis.model import SimulationState
-from tardis.transport.montecarlo import RPacket
-from tardis.transport.montecarlo.numba_interface import (
-    NumbaModel,
-    opacity_state_initialize,
-)
-from tardis.transport.montecarlo.packet_collections import (
-    VPacketCollection,
-)
 from tardis.simulation import Simulation
 from tardis.tests.fixtures.atom_data import DEFAULT_ATOM_DATA_UUID
 from tardis.tests.fixtures.regression_data import RegressionData
-from tardis.transport.montecarlo import RPacket, opacity_state_initialize
+from tardis.transport.montecarlo import RPacket
+from tardis.transport.montecarlo.numba_interface import (
+    opacity_state_initialize,
+)
 from tardis.transport.montecarlo.packet_collections import (
     VPacketCollection,
 )
@@ -68,7 +63,11 @@ class BenchmarkBase:
     def tardis_ref_path(self):
         # TODO: This route is fixed but needs to get from the arguments given in the command line.
         #       /app/tardis-refdata
-        return "/app/tardis-refdata"
+        ref_data_path = Path(
+            Path(__file__).parent.parent,
+            "tardis-refdata",
+        ).resolve()
+        return ref_data_path
 
     @property
     def atomic_dataset(self) -> AtomData:
@@ -258,6 +257,36 @@ class BenchmarkBase:
         return opacity_state_initialize(
             self.nb_simulation_verysimple.plasma,
             line_interaction_type="macroatom",
+            disable_line_scattering=self.nb_simulation_verysimple.transport.montecarlo_configuration.DISABLE_LINE_SCATTERING,
+            continuum_processes_enabled=self.nb_simulation_verysimple.transport.montecarlo_configuration.CONTINUUM_PROCESSES_ENABLED,
+        )
+
+    @property
+    def verysimple_enable_full_relativity(self):
+        return self.nb_simulation_verysimple.transport.enable_full_relativity
+
+    @property
+    def verysimple_disable_line_scattering(self):
+        return (
+            self.nb_simulation_verysimple.transport.montecarlo_configuration.DISABLE_LINE_SCATTERING
+        )
+
+    @property
+    def verysimple_continuum_processes_enabled(self):
+        return (
+            self.nb_simulation_verysimple.transport.montecarlo_configuration.CONTINUUM_PROCESSES_ENABLED
+        )
+
+    @property
+    def verysimple_tau_russian(self):
+        return (
+            self.nb_simulation_verysimple.transport.montecarlo_configuration.VPACKET_TAU_RUSSIAN
+        )
+
+    @property
+    def verysimple_survival_probability(self):
+        return (
+            self.nb_simulation_verysimple.transport.montecarlo_configuration.SURVIVAL_PROBABILITY
         )
 
     @property
--- a/benchmarks/transport_montecarlo_interaction.py
+++ b/benchmarks/transport_montecarlo_interaction.py
@@ -24,8 +24,11 @@ class BenchmarkMontecarloMontecarloNumbaInteraction(BenchmarkBase):
         init_nu = packet.nu
         init_energy = packet.energy
         time_explosion = self.verysimple_time_explosion
+        enable_full_relativity = self.verysimple_enable_full_relativity
 
-        interaction.thomson_scatter(packet, time_explosion)
+        interaction.thomson_scatter(
+            packet, time_explosion, enable_full_relativity
+        )
 
         assert np.abs(packet.mu - init_mu) > 1e-7
         assert np.abs(packet.nu - init_nu) > 1e-7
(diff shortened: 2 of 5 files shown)


MarkMageeAstro:
Nice work Sarthak, the plot seems to be coming along nicely. 

I think we should have a consistent style across both matplotlib and plotly -- they should both look as close as possible to each other. I prefer the style of the current plotly version so I would try and update the matplotlib version to match it. Although I would also use a step plot for both rather than filled histograms. I prefer steps to filled because in the filled histograms you can get overlapping, translucent colours and that means they might not exactly match what's shown in the legend.

sarthak-dv:
Thanks, @MarkMageeAstro for the feedback!
I'll implement the changes to ensure that Matplotlib matches the current Plotly style, and I'll switch both to use step plots instead of filled histograms.

MarkMageeAstro:
Rather than specifying bins as an integer in your hist function, you can give it a list of values. I'm not sure exactly what the right parameter name will be, but there will be some property like self.velocity or self.simulation_state.velocity, something like that. It should give you an array of the inner and outer boundaries of the cells in the model that you can pass to hist. This way you can be sure the plot lines up exactly with the cells in the model for all elements

sarthak-dv:
@MarkMageeAstro @jamesgillanders  I have generated step plots in both Plotly and Matplotlib, adjusted the bins based on the velocity array. Please review them and confirm if they meet your expectations.

Matplotlib:

matplotlib(step)


Plotly:

plotly(step)

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+        -------
+        Plotter
+        """
+        hdfstore = pd.HDFStore(hdf_fpath)
+        time_explosion = hdfstore['/simulation/plasma/scalars']['time_explosion'] * u.s
+        return cls(
+            dict(
+                virtual=sdec.SDECData.from_hdf(hdf_fpath, "virtual"),
+                real=sdec.SDECData.from_hdf(hdf_fpath, "real"),
+            ),
```
jaladh-singhal, line 92: time_explosion, velocity needs to be passed in this instantiation too?

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+    element_symbol2atomic_number,
+    species_string_to_tuple,
+    roman_to_int,
+    int_to_roman,
+)
+
+import matplotlib.pyplot as plt
+import matplotlib.cm as cm
+import matplotlib.colors as clr
+import plotly.graph_objects as go
```
jaladh-singhal, line 18: [nitpick] a good practice is to group imports in the follow following pattern:

imports from other packages

imports from local package (tardis)

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+        hdfstore = pd.HDFStore(hdf_fpath)
+        time_explosion = hdfstore['/simulation/plasma/scalars']['time_explosion'] * u.s
+        return cls(
+            dict(
+                virtual=sdec.SDECData.from_hdf(hdf_fpath, "virtual"),
+                real=sdec.SDECData.from_hdf(hdf_fpath, "real"),
+            ),
+        )
+
+    def _parse_species_list(self, species_list):
```
jaladh-singhal, line 95: we already have similar/same method in SDEC plot. why not make that public and generalize it if needed and share methods in these two classes

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+                    for species_id in temp_list
+                ]
+
+                self._species_list = requested_species_ids
+                self._keep_colour = keep_colour
+        else:
+            self._species_list = None
+        return
+
+    def _make_colorbar_labels(self):
```
jaladh-singhal, line 189: same here

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+                    species_name.append(label)
+                elif atomic_number not in self._keep_colour:
+                    # otherwise add the ion to the label list
+                    label = f"{atomic_symbol} {ion_numeral}"
+                    species_name.append(label)
+
+        self._species_name = species_name
+        return
+
+    def _make_colorbar_colors(self):
```
jaladh-singhal, line 227: same here

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+        packets_mode : str, optional
+            Packet mode, either 'virtual' or 'real'. Default is 'virtual'.
+        ax : matplotlib.axes.Axes, optional
+            Axes object to plot on. If None, creates a new figure.
+        figsize : tuple, optional
+            Size of the figure. Default is (11, 6).
+        cmapname : str, optional
+            Colormap name. Default is 'jet'. A specific colormap can be chosen, such as "jet", "viridis", "plasma", etc.
+        species_list : list of str
+            List of species to plot. Default is None.
```
jaladh-singhal, line 307: There's no benefit of setting default here because you are raising ValueError for None, it should be a required parameter instead.

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+        species_list=None,
+    ):
+        """
+        Generate the last interaction radius distribution plot using plotly.
+
+        Parameters
+        ----------
+        packets_mode : str, optional
+            Packet mode, either 'virtual' or 'real'. Default is 'virtual'.
+        species_list : list of str
```
jaladh-singhal, line 390: same here

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+        plot_data = []
+
+        for species_counter, identifier in enumerate(self.species):
+            g_df = groups.get_group(identifier)
+            r_last_interaction = g_df["last_interaction_in_r"].values * u.cm
+            v_last_interaction = (r_last_interaction / self.time_explosion).to(
+                "km/s"
+            )
+            plot_data.append(v_last_interaction)
+            color = f"rgba({int(255*self._color_list[species_counter][0])}, {int(255*self._color_list[species_counter][1])}, {int(255*self._color_list[species_counter][2])}, 1)"
```
jaladh-singhal, line 429: there's `to_rgb255_string` in SDECPlotter, you can use that here and maybe move that staticmethod to plot_util and use in both classes.

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+                    name=name,
+                    opacity=0.75,
+                )
+            )
+        fig.update_layout(
+            xaxis_title="Last Interaction Velocity (km/s)",
+            yaxis_title="Packet Count",
+            template="plotly_white",
+            font=dict(size=14),
+            yaxis=dict(tickformat=".1e"),
```
jaladh-singhal, line 459: can you adjust tickformat for x-axis too to make it similar to mpl (or exponent if not possible) - we never want "1k, etc." in scientific plots

tardis/visualization/tools/interaction_radius_plot.py
```diff
@@ -0,0 +1,461 @@
...
+                        shape='hv',
+                    ),
+                    name=name,
+                    opacity=0.75,
+                )
+            )
+        fig.update_layout(
+            xaxis_title="Last Interaction Velocity (km/s)",
+            yaxis_title="Packet Count",
+            template="plotly_white",
```
jaladh-singhal, line 457: changing template isn't necessary - we keep it default and allow user to change it if needed - see convergence plots for example

jaladh-singhal (changes requested):
Excellent work @sarthak-dv - this is coming along really nicely. I'm glad to see that you have already modularised the code along with docstrings and code comments!

I have few more suggestions to make it better - to minimize redundancy between this and SDECPlotter. Don't hesitate in exposing SDECPlotter's method if needed here. Just make sure to test that you didn't break it in doing so :)

Additional comments:
- you can move your demo notebook to docs/ with other visualisation demos
- good next step will be to start writing tests for your module

Thanks for keeping up the code quality - the variable names are also great!

PR #2640, opened 2024-06-02
Refactor and add more benchmarks for montecarlo

officialasishkumar:
### :pencil: Description

**Type:** :rocket: `feature`

Refactor montecarlo Bencmarks and adding new benchmarks that cover most of the functions inside montecarlo.

### :pushpin: Resources

Examples, notebooks, and links to useful references.

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #2656, opened 2024-06-13
fix benchmark action

officialasishkumar:
### :pencil: Description

**Type:** :beetle: 🚀 

- Fix github actions for benchmarks
-  Add a webpage for every PR that leads to a benchmark commit graph (eg: https://tardis-sn.github.io/tardis-benchmarks/#benchmark_run_tardis.Benchmarkruntardis.time_run_tardis)

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -95,7 +95,7 @@ jobs:
         if: github.event_name != 'pull_request_target'
         uses: peaceiris/actions-gh-pages@v4
         with:
-          personal_token: ${{ secrets.PERSONAL_TOKEN }}
+          personal_token: ${{ secrets.BOT_TOKEN }}
           publish_branch: ${{ env.DEPLOY_BRANCH }}
           publish_dir: .asv/html
           keep_files: true
@@ -192,7 +192,7 @@ jobs:
         if: github.event_name == 'pull_request_target'
         uses: peaceiris/actions-gh-pages@v4
         with:
-          personal_token: ${{ secrets.PERSONAL_TOKEN }}
+          personal_token: ${{ secrets.BOT_TOKEN }}
           publish_branch: ${{ env.DEPLOY_BRANCH }}
           publish_dir: .asv/html
           destination_dir: ${{ env.DEST_DIR }}
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -98,6 +98,7 @@ jobs:
           personal_token: ${{ secrets.PERSONAL_TOKEN }}
           publish_branch: ${{ env.DEPLOY_BRANCH }}
           publish_dir: .asv/html
+          keep_files: true
           external_repository: tardis-sn/tardis-benchmarks
           user_name: "TARDIS Bot"
           user_email: "tardis.sn.bot@gmail.com"
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -85,8 +85,10 @@ jobs:
       - name: Install asv
         run: pip install asv
 
-      - run: git remote add upstream https://github.com/tardis-sn/tardis
-      - run: git fetch upstream master
+      - name: Fetch from upstream master
+        run: |
+          git remote add upstream https://github.com/tardis-sn/tardis
+          git fetch upstream master
 
       - name: Accept all asv questions
         run: asv machine --yes
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -65,11 +65,6 @@ jobs:
       - name: Install asv
         run: pip install asv
 
-      - name: Fetch from upstream master
-        run: |
-          git remote add upstream https://github.com/tardis-sn/tardis
-          git fetch upstream master
-
       - name: Accept all asv questions
         run: asv machine --yes
 
@@ -108,10 +103,10 @@ jobs:
         run: asv run --config asv.conf.json
 
       - name: Compare Master and PR head
-        run: asv compare upstream/master HEAD --config asv.conf.json | tee asv-compare-output.log
+        run: asv compare origin/master HEAD --config asv.conf.json | tee asv-compare-output.log
 
       - name: Compare Master and PR head but only show changed results
-        run: asv compare upstream/master HEAD --only-changed --config asv.conf.json | tee asv-compare-changed-output.log
+        run: asv compare origin/master HEAD --only-changed --config asv.conf.json | tee asv-compare-changed-output.log
 
       - name: Benchmarks compare output
         id: asv_pr_vs_master
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -66,6 +66,11 @@ jobs:
           path: benchmarks/data/kurucz_cd23_chianti_H_He.h5
           key: atom-data
 
+      - name: Copy atom data to refdata
+        run: |
+          mkdir -p tardis-refdata/atom_data
+          cp benchmarks/data/kurucz_cd23_chianti_H_He.h5 tardis-refdata/atom_data/
+
       - name: Setup Mamba
         uses: mamba-org/setup-micromamba@v1
         with:
@@ -79,6 +84,9 @@ jobs:
 
       - name: Install asv
         run: pip install asv
+
+      - run: git remote add upstream https://github.com/tardis-sn/tardis.git
+      - run: git fetch upstream master
 
       - name: Accept all asv questions
         run: asv machine --yes
@@ -115,15 +123,15 @@ jobs:
           target-branch: main
           target-directory: .asv
 
-      - name: Compare HEAD with master if PR
+      - name: Run benchmarks
         if: github.event_name == 'pull_request_target'
-        continue-on-error: true # TODO: step failed sporadically while testing
-        run: |
-          asv continuous ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | tee asv-cont-output.log  
-          if grep -q failed asv-cont-output.log; then 
-            echo "Some benchmarks have errors!"
-            exit 1
-          fi
+        run: asv run --config asv.conf.json
+      
+      - name: Fetch the latest changes
+        if: github.event_name == 'pull_request_target'
+        run: | 
+          git remote add upstream https://github.com/tardis-sn/tardis.git
+          git fetch upstream master
 
       - name: Compare Master and PR head
         run: asv compare upstream/master HEAD --config asv.conf.json | tee asv-compare-output.log
@@ -143,21 +151,25 @@ jobs:
         with:
           path: asv-compare-changed-output.log
 
+      - name: Remove .asv
+        run: rm -rf .asv
+
       - name: Run benchmarks for base to head commits of PR
         if: github.event_name == 'pull_request_target'
         run: |
+          asv run master..${{ github.event.pull_request.head.ref }} | tee asv-output-PR.log  
+            if grep -q failed asv-output-PR.log; then 
+              echo "Some benchmarks have failed!"
+              exit 1
+            fi
           mkdir -p PR-${{ github.event.number }}
-          cd PR-${{ github.event.number }}
-          asv run ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} | tee asv-output-PR.log  
-          if grep -q failed asv-output-PR.log; then 
-            echo "Some benchmarks have failed!"
-            exit 1
-          fi
-
+          cp -r .asv PR-${{ github.event.number }}
+      
       - name: Generate Graphs and HTML of PR
         if: github.event_name == 'pull_request_target'
-        working-directory: PR-${{ github.event.number }}
-        run: asv publish
+        run: |
+          cd PR-${{ github.event.number }}
+          asv publish
 
       - name: Delete env files of PR
         if: github.event_name == 'pull_request_target'
@@ -176,6 +188,7 @@ jobs:
           user-email: tardis.sn.bot@gmail.com
           target-branch: main
           target-directory: PR
+
       
       - name: Find Comment
         if: always() && github.event_name == 'pull_request_target'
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -86,3 +86,6 @@ benchmarks/data/*.h5
 .asv/
 pkgs/
 release_hashes.txt
+
+#PR directory (contains benchmarks)
+PR-*
```
atharva-2001, line 91: Why add this?

officialasishkumar: I misunderstood something. Removed it.

andrewfullard:
Please test by making a version on your fork

officialasishkumar:
> Please test by making a version on your fork and add link to test results here

https://github.com/AkashKumar7902/tardis/actions/runs/9705432873/job/26787490302?pr=6

the error in the later half is due to the generation of file which is relying upon tardis repo. The compare is working which means the bot will display the comparision output. 

cc: @andrewfullard

officialasishkumar:
> Please test by making a version on your fork and add link to test results here

https://github.com/AkashKumar7902/tardis/actions/runs/9705432873/job/26787490302?pr=6

the error in the later half is due to the generation of file which is relying upon tardis repo. The compare is working which means the bot will display the comparision output.

andrewfullard:
> > Please test by making a version on your fork and add link to test results here
> 
> https://github.com/AkashKumar7902/tardis/actions/runs/9705432873/job/26787490302?pr=6
> 
> the error in the later half is due to the generation of file which is relying upon tardis repo. The compare is working which means the bot will display the comparision output.

image
 
Is there an excess step here?

officialasishkumar:
Yes

officialasishkumar:
Yes, that's an extra step. I have removed it. 

here is the new action run: https://github.com/AkashKumar7902/tardis/actions/runs/9722403407/job/26836253099?pr=9

.github/workflows/benchmarks.yml
```diff
@@ -80,6 +85,9 @@ jobs:
       - name: Install asv
         run: pip install asv
 
+      - run: git remote add upstream https://github.com/tardis-sn/tardis
```
DeerWhale, line 88: please add name for these two actions.

jvshields:
This may be helpful - https://stackoverflow.com/questions/61393981/getting-base-branch-sha-on-pull-request-in-github-action-workflow

officialasishkumar:
https://github.com/sympy/sympy/blob/master/.github/workflows/runtests.yml

cc: @andrewfullard @atharva-2001

atharva-2001:
> Yes, that's an extra step. I have removed it.
> 
> here is the new action run: [officialasishkumar/tardis/actions/runs/9733397076/job/26860263830?pr=12](https://github.com/officialasishkumar/tardis/actions/runs/9733397076/job/26860263830?pr=12)


This workflow used this workflow file-
https://github.com/officialasishkumar/tardis/actions/runs/9733397076/workflow?pr=12
which does not have your commits. Is the workflow being correctly tested? 
Note that these are pull request target workflows, so they will only pick the workflow file from the target(master) branch, not from the pull request branch.

officialasishkumar:
> Note that these are pull request target workflows, so they will only pick the workflow file from the target(master) branch, not from the pull request branch.

Here is the new workflow run: https://github.com/officialasishkumar/tardis/actions/runs/9746049892/job/26895430197?pr=13
Here is the master branch benchmarks.yml that was updated yesterday: https://github.com/officialasishkumar/tardis/blob/master/.github/workflows/benchmarks.yml

atharva-2001:
The workflow runs on this commit-
https://github.com/officialasishkumar/tardis/tree/bff68859d97a0764e38e5e598c48940e748e45a9

Not the master branch as far as I understand.

Click on the workflow file on the left side of this page-
https://github.com/officialasishkumar/tardis/actions/runs/9746049892/job/26895430197?pr=13

then click on commit hash, then browse files, and you will see this
https://github.com/officialasishkumar/tardis/tree/bff68859d97a0764e38e5e598c48940e748e45a9

officialasishkumar:
@atharva-2001, the workflow file might be because my branch is not really updated with master for a while (approx 2 months). I think the workflow is running with the current master branch as you can see in the build that the jobs that are runned are aligned with the current changes and not the old one.

.github/workflows/benchmarks.yml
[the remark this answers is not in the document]
officialasishkumar: Added

.github/workflows/benchmarks.yml
```diff
@@ -115,21 +123,15 @@ jobs:
...
-          asv continuous ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }}  | tee asv-cont-output.log  
-          if grep -q failed asv-cont-output.log; then 
-            echo "Some benchmarks have errors!"
-            exit 1
-          fi
+        run: asv run --config asv.conf.json
 
       - name: Compare Master and PR head
-        run: asv compare ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | tee asv-compare-output.log
+        run: asv compare upstream/master HEAD --config asv.conf.json | tee asv-compare-output.log
```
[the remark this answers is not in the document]
officialasishkumar: image
Unfortunately, the method of fetching the hash without hardcoding the branch hasn't been successful. Given that the master branch name is unlikely to change, would it be acceptable to use a hardcoded approach in this case?

.github/workflows/benchmarks.yml
```diff
@@ -66,6 +66,11 @@ jobs:
           path: benchmarks/data/kurucz_cd23_chianti_H_He.h5
           key: atom-data
 
+      - name: Copy atom data to refdata
+        run: |
+          mkdir -p tardis-refdata/atom_data
+          cp benchmarks/data/kurucz_cd23_chianti_H_He.h5 tardis-refdata/atom_data/
```
atharva-2001, line 72: I will suggest using the setup lfs action-
https://github.com/tardis-sn/tardis/blob/e88a31a809eb041068fffeeb16e1585b8e51f3a5/.github/workflows/tests.yml#L57-L58


Right now, there is a step in this workflow which uses Azure to download the atomic file. We're moving away from Azure. Since we already have caches, I'd suggest reusing the cache here and copying the file over to the benchmarks folder. That way, we won't have extra cache(we're near the limit).

.github/workflows/benchmarks.yml
```diff
@@ -115,21 +125,15 @@ jobs:
           target-branch: main
           target-directory: .asv
 
-      - name: Compare HEAD with master if PR
+      - name: Run benchmarks
         if: github.event_name == 'pull_request_target'
-        continue-on-error: true # TODO: step failed sporadically while testing
-        run: |
-          asv continuous ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }}  | tee asv-cont-output.log
```
atharva-2001, line 122: What is the difference between asv continuous and asv compare?

atharva-2001, line 122: What is the difference between asv continuous and asv compare?

.github/workflows/benchmarks.yml
```diff
@@ -143,6 +147,46 @@ jobs:
...
+        if: github.event_name == 'pull_request_target'
+        run: |
+          echo $(git rev-parse HEAD) > commit_hashes.txt
+          echo $(git merge-base HEAD upstream/master) >> commit_hashes.txt
+
+          asv run HASHFILE:commit_hashes.txt | tee asv-output-PR.log  
+            if grep -q failed asv-output-PR.log; then 
+              echo "Some benchmarks have failed!"
+              exit 1
+            fi
```
atharva-2001, line 160: So if a pull request has like 15 commits, it will run on all of them?

.github/workflows/benchmarks.yml
```diff
@@ -143,6 +147,46 @@ jobs:
...
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.BOT_TOKEN }}
+        with:
+          source-directory: PR-${{ github.event.number }}
+          destination-github-username: tardis-sn
+          destination-repository-name: tardis-benchmarks
+          user-email: tardis.sn.bot@gmail.com
+          target-branch: main
+          target-directory: PR
+
```
atharva-2001, line 188: Have you tested this? This will rewrite a new folder called PR each time the workflow runs on any pull request. If I am not wrong, I doubt if this is the good way to publish a website. Suppose if there are two pull requests running the thing at the same time, then how would we know what pr the published website is referring to?

asv.conf.json
```diff
@@ -9,7 +9,7 @@
         "python setup.py build",
         "PIP_NO_BUILD_ISOLATION=false python -mpip wheel --no-deps --no-index -w {build_cache_dir} {build_dir}"
     ],
-    "branches": ["master"],
+    "branches": ["upstream/master", "HEAD"],
```
atharva-2001, line 12: What does HEAD mean here exactly?

atharva-2001 (changes requested):

.github/workflows/benchmarks.yml
officialasishkumar: No, only 2

atharva-2001: I see

.github/workflows/benchmarks.yml
officialasishkumar: the doc says 
> Run a side-by-side comparison of two commits for continuous integration.

Not sure what continuous integration here means. That's why I have used asv compare.

asv.conf.json
officialasishkumar: last commit in the current branch

.github/workflows/benchmarks.yml
officialasishkumar: I have added new changes to solve this.

andrewfullard (dismissed):
I think we should try and merge this before we worry about solving the crashing individual benchmarks in #2640

.github/workflows/benchmarks.yml
```diff
@@ -80,6 +65,11 @@ jobs:
       - name: Install asv
         run: pip install asv
 
+      - name: Fetch from upstream master
+        run: |
+          git remote add upstream https://github.com/tardis-sn/tardis
+          git fetch upstream master
```
atharva-2001, line 71: this is what the checkout does right? Perhaps you could just do git remote add upstream URL?

.github/workflows/benchmarks.yml
```diff
@@ -143,6 +125,83 @@ jobs:
...
+        env:
+          EVENT: ${{ github.event_name }}
+
+      - name: Deploy ${{ env.DEST_DIR }}
+        if: github.event_name == 'pull_request_target'
+        uses: peaceiris/actions-gh-pages@v4
+        with:
+          personal_token: ${{ secrets.BOT_TOKEN }}
+          publish_branch: ${{ env.DEPLOY_BRANCH }}
+          publish_dir: .asv/html
```
atharva-2001, line 197: Perhaps we can remove this? Since .asv/html folder no longer exists?

.github/workflows/benchmarks.yml
officialasishkumar: what do you mean? git remote add upstream url is already present.

.github/workflows/benchmarks.yml
```diff
@@ -101,35 +91,27 @@ jobs:
...
-          destination-github-username: tardis-sn
-          destination-repository-name: tardis-benchmarks
-          user-email: tardis.sn.bot@gmail.com
-          target-branch: main
-          target-directory: .asv
-
-      - name: Compare HEAD with master if PR
+          personal_token: ${{ secrets.BOT_TOKEN }}
+          publish_branch: ${{ env.DEPLOY_BRANCH }}
+          publish_dir: .asv/html
```
atharva-2001, line 100: Same here?

.github/workflows/benchmarks.yml
```diff
@@ -143,6 +125,83 @@ jobs:
...
+          EVENT: ${{ github.event_name }}
+
+      - name: Deploy ${{ env.DEST_DIR }}
+        if: github.event_name == 'pull_request_target'
+        uses: peaceiris/actions-gh-pages@v4
+        with:
+          personal_token: ${{ secrets.BOT_TOKEN }}
+          publish_branch: ${{ env.DEPLOY_BRANCH }}
+          publish_dir: .asv/html
+          destination_dir: ${{ env.DEST_DIR }}
```
atharva-2001, line 198: 👍

.github/workflows/benchmarks.yml
atharva-2001: Or this is option mandatory?

.github/workflows/benchmarks.yml
atharva-2001: The first step just does this, perhaps the remote is origin there. If yes then we can perhaps reduce this step here.

atharva-2001 (changes requested):

.github/workflows/benchmarks.yml
officialasishkumar: Yes, for deploying

.github/workflows/benchmarks.yml
officialasishkumar: doesn't work. 

https://github.com/officialasishkumar/tardis/actions/runs/9779319538/job/26998276265?pr=46

.github/workflows/benchmarks.yml
atharva-2001: I see

.github/workflows/benchmarks.yml
atharva-2001: By default I think it is origin, so it does not find upstream, just rename origin to upstream if you don;t want origin

officialasishkumar: updated

atharva-2001 (approved):
Great work!

.github/workflows/benchmarks.yml
atharva-2001: Thanks

andrewfullard (approved):
If Atharva is happy, I am also happy. Cross your fingers!

PR #2672, opened 2024-06-21
R packet tracker fix

### LINKED ISSUE #2671: Bug in how RPacketTracker tracks the rpacket.
**Describe the bug**
Can someone please verify if the current implementation of RPacketTracker has a bug?
In the single_packet_loop function, the trace_packet function returns the distance, interaction_type, and delta_shell that the packet went through
https://github.com/tardis-sn/tardis/blob/26b60059dfd0cad0c2651d52137e45b609c696e8/tardis/transport/montecarlo/single_packet_loop.py#L147-L158
interaction_type could be BOUNDARY, LINE, ESCATTERING, etc.
https://github.com/tardis-sn/tardis/blob/26b60059dfd0cad0c2651d52137e45b609c696e8/tardis/transport/montecarlo/single_packet_loop.py#L162-L163
https://github.com/tardis-sn/tardis/blob/26b60059dfd0cad0c2651d52137e45b609c696e8/tardis/transport/montecarlo/single_packet_loop.py#L203-L204
https://github.com/tardis-sn/tardis/blob/26b60059dfd0cad0c2651d52137e45b609c696e8/tardis/transport/montecarlo/single_packet_loop.py#L230-L231
Here's the issue
Say the packet went through a LINE interaction in shell_id=1; we track that in rpacket_tracker correctly, but now, if it went through a BOUNDARY, we are now using the previous value of r_packet to say that we have got a LINE interaction with the updated values of r and shell_id. Even though we are not going through a LINE interaction.Please correct me if I am wrong here.A proposed solution would be to update the rpacket_tracker only when it's not BOUNDARY Interaction.

**Screenshots**
image
As we can see, the last interaction type keeps getting copied over when the packet moves to the next boundary.

**System**
 - OS:
    - [x] GNU/Linux
    - [ ] macOS

Sumit112192:
### :pencil: Description

**Type:** :beetle: `bugfix`
Fixes #2671 
This PR fixes the way RPacketTracker tracks the packet. Also, since no tests were written for rpacket_tracker. I wrote some of them.

`pytest -vvs tardis/transport/montecarlo/tests/test_rpacket_tracker.py --tardis-refdata=/home/sam/TARDIS/tardis-refdata`

@wkerzendorf @andrewfullard


andrewfullard (approved):

tardis/transport/montecarlo/single_packet_loop.py
```diff
@@ -264,8 +264,9 @@ def single_packet_loop(
             )
         else:
             pass
-        if montecarlo_configuration.ENABLE_RPACKET_TRACKING:
-            rpacket_tracker.track(r_packet)
+        if interaction_type != InteractionType.BOUNDARY:
```
Knights-Templars, line 267: Please add `and`

PR #2697, opened 2024-07-10
refactor to compile time constants

wkerzendorf:
@Sumit112192 might be interesting


Sumit112192:
Should I base my PR on this PR and then reimplement my PRs? Or should I use the logic here on my PR because it seems incomplete?

PR #2714, opened 2024-07-17
restructure of rpacket_tracker

Sumit112192:
### :pencil: Description

**Type:**  :roller_coaster: `infrastructure` and `restructure`

This PR restructures the rpacket_tracker code to initialize only when the track_rpacket flag is set. Apart from that, I will also add tests for the rpacket_tracker and change how it registers an interaction.


Sumit112192:
@wkerzendorf @andrewfullard Please review.

Sumit112192:
Profiling:
Master Branch:
image
FixRPacketTracker Branch:
image
Memory usage has gone done.

tardis/conftest.py
```diff
@@ -243,3 +252,32 @@ def simulation_verysimple_vpacket_tracking(config_verysimple, atomic_dataset):
...
+    Parameters
+    ----------
+    config_verysimple_rpacket_tracking : tardis.io.configuration.config_reader.Configuration
+    atomic_dataset : AtomData
+
+    Returns
+    -------
+    simulation object with track_rpacket enabled
+    """
+    config_rpacket_tracking.montecarlo.iterations = 3
```
andrewfullard, line 271: Do we need 3 iterations for the test to be useful?

andrewfullard (approved):

tardis/conftest.py
Sumit112192: I don't think so. But it shouldn't take too much time.

Sumit112192:
Completet as PR #2716 is merged

PR #2758, opened 2024-07-29
Opacity state slice

Rodot-:
### :pencil: Adds a __geitem__ operation to the opacity state class which allows slicing and indexing by shell

**Type:**  :rocket: `feature`
The opacity state should be able to be sliced in order to run the Montecarlo on sub-slices of the geometry.  Let me know if there are other features or considerations to think about.

- [x] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/opacities/opacity_state.py
+++ b/tardis/opacities/opacity_state.py
@@ -106,7 +106,7 @@ class OpacityState:
         self.photo_ion_activation_idx = photo_ion_activation_idx
         self.k_packet_idx = k_packet_idx
 
-    def __getitem__(self, i):
+    def __getitem__(self, i: slice):
         """Get a shell or slice of shells of the attributes of the opacity state
 
         Args:
--- a/tardis/opacities/opacity_state.py
+++ b/tardis/opacities/opacity_state.py
@@ -110,7 +110,7 @@ class OpacityState:
         """Get a shell or slice of shells of the attributes of the opacity state
 
         Args:
-            i (int, slice): shell index or slice
+            i (slice): shell slice.  Will fail if slice is int since class only supports array types
 
         Returns:
             OpacityState : a shallow copy of the current instance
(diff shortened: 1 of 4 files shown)


andrewfullard (changes requested):
I hate that this is in 2 places but oh well. Please remove the print statements from the test.

andrewfullard (approved):

tardis/opacities/opacity_state.py
```diff
@@ -106,6 +106,41 @@ def __init__(
...
+
+        Args:
+            i (int, slice): shell index or slice
+
+        Returns:
+            OpacityState : a shallow copy of the current instance
+        """
+        # NOTE: This currently will not work with continuum processes since it does not slice those arrays
+        return OpacityState(
+            self.electron_density[i],
```
wkerzendorf, line 120: check if `i` is an `int` and then raise an error

Rodot-: This wouldn't allow one to slice the opacity state then

tardis/transport/montecarlo/numba_interface.py
```diff
@@ -112,6 +112,56 @@ def __init__(
         self.photo_ion_activation_idx = photo_ion_activation_idx
         self.k_packet_idx = k_packet_idx
 
+    def __getitem__(self, i):
```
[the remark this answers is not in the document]
Rodot-: I was mistaken

tardis/opacities/opacity_state.py
andrewfullard: To be clear, Wolfgang is concerned about what happens if somebody requests `OpacityState[5]` and passes that to something else.

wkerzendorf: yep. OpacityState[5] should return either a length 1 OPacity state or fail

Rodot-: Which do you prefer?

PR #2770, opened 2024-07-31
Restructure/detailed balance foundation

wkerzendorf:


andrewfullard:
Tests rely on https://github.com/tardis-sn/tardis-regression-data/pull/24 merging first.

andrewfullard (approved):
I'm happy with this, but I have also done a lot of work on it so it needs other reviews.

jvshields (dismissed):
As far as I can tell it looks fine, but it's always hard to tell with the restructure PRs which parts are renames or functional changes worth paying attention to, or are just files and functions being moved around to places that make more sense.

andrewfullard:
> As far as I can tell it looks fine, but it's always hard to tell with the restructure PRs which parts are renames or functional changes worth paying attention to, or are just files and functions being moved around to places that make more sense.

In this case @jvshields I think the tests are most illustrative because we compare this new structure with the previous way(s) of doing things.

andrewfullard:
Tests are now failing for a different but related reason. `electron_densities` seems to stick around in memory for too long. Unclear why.

andrewfullard (dismissed):

PR #2773, opened 2024-08-01
Opacity state restructure

Rodot-:
### :pencil: Adds and opacity solver and breaks up the opacity state into constituent parts.  Current designed to be compatible with the legacy plasma

**Type:**  :rocket: `feature` 

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/opacities/continuum/continuum_state.py
```diff
@@ -0,0 +1,162 @@
...
+        photo_ion_cross_sections,
+        chi_bf,
+        ff_cooling_factor,
+        fb_emission_cdf,
+        photo_ion_idx,
+        k_packet_idx,
+    ):
+        """Current State of the Continuum Required for Opacity Computation
+
+        Args:
```
andrewfullard, line 19: Wrong docstring style

Rodot-: I think I used the wrong one for most of them, I'll go back and edit them

tardis/opacities/opacity_state.py
```diff
@@ -4,6 +4,79 @@
 
 from tardis.opacities.tau_sobolev import calculate_sobolev_line_opacity
 from tardis.transport.montecarlo.configuration import montecarlo_globals
+from tardis.opacities.continuum.continuum_state import ContinuumState
+from tardis.opacities.macro_atom.macroatom_state import MacroAtomState
+
+
+class OpacityStatePython:
```
wkerzendorf, line 11: rename this to `OpacityStatePython` to `OpacityState` and rename the other one to numba

tardis/transport/montecarlo/base.py
```diff
@@ -94,6 +99,12 @@ def __init__(
         mc_tracker.DEBUG_MODE = debug_packets
         mc_tracker.BUFFER = logger_buffer
 
+        opacity_solver_config = {
+            "line_interaction_type": self.line_interaction_type,
+            "disable_line_scattering": self.montecarlo_configuration.DISABLE_LINE_SCATTERING,
+        }
+        self.opacity_solver = OpacitySolver(opacity_solver_config)
```
wkerzendorf, line 106: make them keyword arguments.

tardis/transport/montecarlo/base.py
```diff
@@ -94,6 +99,12 @@ def __init__(
         mc_tracker.DEBUG_MODE = debug_packets
         mc_tracker.BUFFER = logger_buffer
 
+        opacity_solver_config = {
+            "line_interaction_type": self.line_interaction_type,
+            "disable_line_scattering": self.montecarlo_configuration.DISABLE_LINE_SCATTERING,
```
wkerzendorf, line 104: this should be a transportsolver option. please make an issue to change that.

tardis/opacities/opacity_solver.py
```diff
@@ -0,0 +1,50 @@
+from tardis.opacities.tau_sobolev import calculate_sobolev_line_opacity
+from tardis.opacities.opacity_state import (
+    OpacityStatePython,
+)
+import numpy as np
+
+
+class OpacitySolver(object):
+    def __init__(self, opacities_config):
```
wkerzendorf, line 9: add them class attributes

tardis/opacities/opacity_solver.py
```diff
@@ -0,0 +1,55 @@
...
+        OpacityState
+        """
+        tau_sobolev = calculate_sobolev_line_opacity(
+            legacy_plasma.atomic_data.lines,
+            legacy_plasma.level_number_density,
+            legacy_plasma.time_explosion,
+            legacy_plasma.stimulated_emission_factor,
+        )
+
+        if self.disable_line_scattering:
```
jvshields, line 48: It's quite odd to calculate this and then set it to 0 if it's disabled, and doesn't save on any computation. Can we just initialize this as 0s of the right shape instead?

tardis/opacities/tests/test_opacity_state_numba.py
```diff
@@ -0,0 +1,83 @@
...
+from tardis.opacities.opacity_state import opacity_state_to_numba
+from tardis.opacities.opacity_solver import OpacitySolver
+import numpy.testing as npt
+import numpy as np
+
+
+@pytest.mark.parametrize(
+    "line_interaction_type,sliced",
+    [
+        ("scatter", False),
```
DeerWhale, line 11: Add a parameter that scatter = True to test the `macroatom = None` case in opacity_state.py line 42.

Rodot-: Currently, because the initialization is dependent on the legacy plasma and is built from it in the opacity_state.from_legacy_plasma method, the line_interaction_type is not actually used by the solver in the current version.  This test will be added once the new plasma is in.

tardis/opacities/tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from copy import deepcopy
+from tardis.simulation import Simulation
+
+
+@pytest.fixture(scope="package")
```
andrewfullard, line 6: I find it is better to try and move fixtures to a higher conftest level if they are duplicates of other fixtures. Because this adds another simulation setup and run (with only 10 packets, I suppose) which has the same name. I am not sure if `package` scope is correct. Perhaps it should be `module`?

andrewfullard (dismissed):

tardis/opacities/tests/conftest.py
Rodot-: I'll switch the scope to module, I was having some trouble finding the right fixtures so I opted to simply duplicate the old opacity_state tests here

andrewfullard (approved):

jvshields (approved):

PR #2781, opened 2024-08-06
Update docstrings of Tracker tests

Sumit112192:
### :pencil: Description

**Type:** :roller_coaster: `documentation`

Earlier the docstring meant that the `shell_id` is of the last line interaction. 
For eg, if interaction_type is [Line, Escatter, Escatter, Line, Escatter] and the shell_id is [0, 1, 2, 2, 3, 4]. So the docstring sounded like it should return 3 since the last line interaction happened in the 3rd shell, but that was not the case.

The fixtures return the shell_id when the last interaction is strictly line
For eg, if interaction_type is [Line, Escatter, Escatter, Line, Escatter] and the shell_id is [0, 1, 2, 2, 3, 4]. In this case, since the last interaction is Escatter, it should not return anything.

### DIFF
--- a/tardis/transport/montecarlo/tests/test_rpacket_last_interaction_tracker.py
+++ b/tardis/transport/montecarlo/tests/test_rpacket_last_interaction_tracker.py
@@ -12,7 +12,7 @@ from tardis.transport.montecarlo.packet_trackers import (
 def interaction_type_in_use(
     nb_simulation_verysimple,
 ):
-    """Last interaction types of rpacket from LastInteractionTracker"""
+    """Last interaction types of rpacket from LastInteractionTracker class"""
     transport_state = nb_simulation_verysimple.transport.transport_state
     interaction_type = transport_state.last_interaction_type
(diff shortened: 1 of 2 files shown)


Sumit112192:
I have updated the docstring so that if someone works on the tracker classes, they should be able to modify it.

tardis/transport/montecarlo/tests/test_rpacket_last_interaction_tracker.py
```diff
@@ -23,7 +23,9 @@ def shell_id_in_use(
     nb_simulation_verysimple,
     interaction_type_in_use,
 ):
-    """Last interaction types of rpacket from LastInteractionTracker"""
+    """
+    shell_id when last interaction is line from LastInteractionTracker class
```
atharva-2001, line 27:
```suggestion
    `shell_id` when last interaction is line from `LastInteractionTracker` class
```

atharva-2001 (approved):

PR #2783, opened 2024-08-06
Simple Atom data refactor

andrewfullard:
### :pencil: Description

This uses  dataclasses to begin simplifying the atom data.

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/io/atom_data/collision_data.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+import pandas as pd
+
+
+@dataclass
+class CollisionData:
```
wkerzendorf, line 7: I think that is a special CMFGEN collision class. We need to look into this. There are two different formats

tardis/io/atom_data/simple_atom_data.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+import pandas as pd
+
+
+@dataclass
+class SimpleAtomData:
+    data: pd.DataFrame
```
wkerzendorf, line 8: what is that?

tardis/io/atom_data/simple_atom_data.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+import pandas as pd
+
+
+@dataclass
+class SimpleAtomData:
```
Rodot-, line 7: Is it worthwhile to make this a data class with a single attribute?

tardis/io/atom_data/util.py
```diff
@@ -51,3 +51,32 @@ def resolve_atom_data_fname(fname):
         f"Atom Data {fname} is not found in current path or in TARDIS data repo. {atom_data_name} "
         "is also not a standard known TARDIS atom dataset."
     )
+
+
+def set_atom_data_attributes(atom_data, store, attribute):
```
Rodot-, line 56: This seems a little broad, maybe some sort of schema for the AtomData is worthwhile?

tardis/io/atom_data/macro_atom_data.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+
+import pandas as pd
+
+
+@dataclass
+class MacroAtomData:
```
Rodot-, line 7: Would be good to have docstrings that talk more about what the format of these dataframes should be

tardis/io/atom_data/base.py
```diff
@@ -321,41 +290,57 @@ def __init__(
...
+            collision_data, collision_data_temperatures, yg_data
+        )
 
-        self.collision_data = collision_data
-        self.collision_data_temperatures = collision_data_temperatures
+        self.collision_data = collected_collision_data.data
+        self.collision_data_temperatures = collected_collision_data.temperatures
 
-        self.synpp_refs = synpp_refs
+        self.synpp_refs = SimpleAtomData(synpp_refs).data
```
Rodot-, line 319: I see a lot of `se.fx = SimpleAtomData(x).data` which from what I can see isn't doing anything except assigning `self.x=x`, why have this wrapper and the extra step here?

tardis/io/atom_data/base.py
```diff
@@ -223,35 +225,9 @@ def from_hdf(cls, fname=None):
...
-                atom_data.md5 = None
-
-            try:
-                atom_data.version = store.root._v_attrs["database_version"]
-            except KeyError:
-                logger.debug(
-                    "VERSION not available for Atom Data. Setting value to None"
-                )
-                atom_data.version = None
+            set_atom_data_attributes(atom_data, store, "uuid1")
```
Rodot-, line 228: Why not make this a method or an operator overload?

Rodot- (commented):
Could you expand a bit on the broader design goals of this PR?

tardis/io/atom_data/simple_atom_data.py
wkerzendorf: yep - I'm also not sure as they are all dataframes. I think getting this PR merged without such a dataclasses and then adding PRs for specific classes that do more might be the right way forward.

andrewfullard: That's fair enough! I can easily remove them. The intent was to provide a baseline structure of classes so that they can be expanded with methods, and this one being likely the base class.

tardis/io/atom_data/util.py
andrewfullard: This was some refactoring to clean up a lot of try-except blocks from the HDF reader classmethod. This would certainly be replaced if a schema was implemented.

tardis/io/atom_data/base.py
andrewfullard: See previous comments- this wrapper should theoretically provide a uniform interface for pre- and post- sliced data. But we can do that with pure dataframes if we prefer (though there are a lot of little weird things done to the dataframes in the current atomdata class)

tardis/io/atom_data/base.py
andrewfullard: I was simply refactoring the try-except blocks out. If there's a more pythonic way, I can modify this.

wkerzendorf: I think the question was why we don't have this as a method as part of the class.

tardis/io/atom_data/collision_data.py
wkerzendorf: I think for now, I would have a Chianti1CollisionData and a CMFGENCollisionData

tardis/io/atom_data/simple_atom_data.py
wkerzendorf: I think I would remove them for now until we have a better idea about how we want to proceed.

tardis/io/atom_data/base.py
```diff
@@ -322,24 +292,40 @@ def __init__(
...
         self.zeta_data = zeta_data
 
-        self.collision_data = collision_data
-        self.collision_data_temperatures = collision_data_temperatures
+        chianti_collision_data = ChiantiCollisionData(
+            collision_data, collision_data_temperatures
+        )
+
+        cmfgen_collision_data = CMFGENCollisionData(
+            yg_data, collision_data_temperatures
```
wkerzendorf, line 318: I don't think that is quite right. I'm pretty sure the yg_data stores it as part of the columns. Or ChiantiCollisionData. have a quick look in the plasma

andrewfullard: Looking at the plasma `YgData` class, it uses both `yg_data` and `collision_data_temperatures`. See https://github.com/andrewfullard/tardis/blob/atom-data-refactor-v2/tardis/plasma/properties/atomic.py#L676 and https://github.com/andrewfullard/tardis/blob/atom-data-refactor-v2/tardis/plasma/properties/atomic.py#L687. For the Chianti-source collision_data, it is used in NLTEData, see https://github.com/andrewfullard/tardis/blob/atom-data-refactor-v2/tardis/io/atom_data/nlte_data.py#L18-L19 and https://github.com/andrewfullard/tardis/blob/atom-data-refactor-v2/tardis/io/atom_data/nlte_data.py#L86. 

Oddly in the `from_hdf` method in this class, `collision_data` is set to "dummy value" if the data is sourced from CMFGen, see https://github.com/andrewfullard/tardis/blob/atom-data-refactor-v2/tardis/io/atom_data/base.py#L199. That is why I did not include it in the CMFGen dataclass.

Rodot- (approved):

PR #2791, opened 2024-08-12
Adds a radiation field convergence solver class

andrewfullard:
Has basically the same interface as ConvergenceSolver

### :pencil: Description

**Type:**  :rocket: `feature` 

Radiative temperature and dilution factor have been unified in many parts of the code as a radiation field object. This PR creates a convergence class to handle them as one item, with the option to check convergence for them both simultaneously (currently defaults to off and is not settable).

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/simulation/radiation_field_convergence.py
```diff
@@ -0,0 +1,89 @@
+from tardis.plasma.radiation_field.planck_rad_field import (
+    DilutePlanckianRadiationField,
+)
+from tardis.simulation.convergence import ConvergenceSolver
+
+
+class RadiationFieldConvergenceSolver:
+    converge_separately = True
```
wkerzendorf, line 8: I'd like to expand this a bit on what the rate of convergence is for temperature, what rate of convergence for dilution, factor. Maybe even have individual classes for the damped and undamped.

tardis/simulation/radiation_field_convergence.py
```diff
@@ -0,0 +1,89 @@
...
+                no_of_shells,
+            )
+        )
+
+        if self.converge_separately:
+            return temperature_converged, dilution_factor_converged
+
+        return temperature_converged and dilution_factor_converged
+
+    def converge(self, radiation_field, estimated_radiation_field):
```
wkerzendorf, line 64: should we call this solve?

tardis/simulation/radiation_field_convergence.py
andrewfullard: You would rather that be explicitly listed at the top, then modifying them with the input config, rather than just using the input config as it is now?

wkerzendorf: I think so - similar to the plasma_solver_factory. But we can talk about this offline.

tardis/simulation/base.py
```diff
@@ -272,6 +275,11 @@ def advance_state(self, emitted_luminosity):
         -------
             converged : bool
         """
+        current_radiation_field_state = DilutePlanckianRadiationField(
+            self.simulation_state.t_radiative,
```
Rodot-, line 279: Please use the `simulation_state.radiation_field_state.*` properties.  These are passed to the plasma which acts on the entire base geometry-sized arrays. Adding a slice or active method to the DilutePlanckianRadiationField would be preferable if a user is trying to only access the exposed information

Rodot- (changes requested):

tardis/simulation/base.py
andrewfullard: Okay, I couldn't find the radiation_field_state on the simulation_state when I looked (not very carefully) so I can easily use that instead.

Rodot-: It is passed as an argument to the SimulationState constructor and the t_radiative and dilution_factor properties grab it from it's radiation_field_state attribute.

Rodot- (approved):

tardis/simulation/tests/test_radiation_field_convergence.py
```diff
@@ -0,0 +1,58 @@
...
+    dilution_factor = np.array([0.1, 0.2, 0.3])
+    radiation_field = DilutePlanckianRadiationField(
+        temperature, dilution_factor
+    )
+    return radiation_field
+
+
+@pytest.fixture(scope="function")
+def partially_converged_radiation_field():
+    temperature = [1000, 2000, 3000] * u.K
```
jvshields, line 25: Add a little documentation to explain what's going on here.

andrewfullard, line 25: Add comment noting why this fixture is different

tardis/simulation/tests/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.fixture(scope="module")
+def config(example_configuration_dir: Path):
+    return Configuration.from_yaml(
+        example_configuration_dir / "tardis_configv1_verysimple.yml"
+    )
+
+
+@pytest.fixture(scope="function")
```
jvshields, line 15: Investigate why this is here and if any tests modify the fixture

andrewfullard:
On reflection, not necessary

PR #2796, opened 2024-08-14
Add status to r packet last interaction tracker

Sumit112192:
### :pencil: Description

**Type:** :rocket: `feature` 

This PR adds a `status` property to the RPacketLastInteractionTracker as it can be used to determine whether the packet was reabsorbed or emitted.


andrewfullard:
What exactly do you mean by "emitted"? Escaped the outer boundary of the simulation?

Sumit112192:
> What exactly do you mean by "emitted"? Escaped the outer boundary of the simulation?

```
class PacketStatus(IntEnum):
    IN_PROCESS = 0
    EMITTED = 1
    REABSORBED = 2
    ADIABATIC_COOLING = 4
```

Yes.

andrewfullard (dismissed):

DeerWhale (approved):

Sumit112192:
@DeerWhale There is some issue with the current logic, so I have to change the current code in the PR. But it will need some discussion. The current code will always show the packet status to be `IN_PROCESS` since the track function is invoked only when it is Line Interaction or Escatter interaction but not boundary interaction, so it will miss the last boundary interaction.

PR #2797, opened 2024-08-14
V inner boundary solver workflow

Rodot-:
### :pencil: 

Adds a workflow to solve for the v_inner_boundary of the simulation

Based on PR #2730 Which should be merged first

**Type:** :rocket: `feature` 

This adds a complete workflow as a subclass of the SimpleSimulation workflow from PR #2730.  This workflow adds convergence criteria and handling the the inner-boundary velocity value.  This PR also makes small edits to the configuration schema and parser to allow accessing properties of the v_inner_boundary convergence. 

### :pushpin: Resources
A notebook is provided in tardis/docs/workflows

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #2800, opened 2024-08-15
V inner formal integral

Rodot-:
### :pencil: 

Fixes the formal integral to handle changes in geometry

**Type:** :beetle: `bugfix`

When the formal integral is run with the v_inner solver, if the final active geometry state is different than that of the initial geometry mismatches in plasma and opacity quantities occur.  This PR rectifies this by slicing these arrays to the correct size

Depends on PR #2797 

- [x] Testing pipeline
- [ ] Other method (describe)

CUDA version of the formal integral was tested locally


andrewfullard:
@Rodot- this needs to be rebased now

andrewfullard (dismissed):

andrewfullard (approved):

tardis/workflows/util.py
```diff
@@ -0,0 +1,87 @@
...
+    delta_nu = bins_high - bins_low
+    n_bins = len(delta_nu)
+
+    taus = taus[1 : n_bins * bin_size + 1]
+    freqs = freqs[1 : n_bins * bin_size + 1]
+
+    ct = simulation_state.time_explosion.cgs.value * const.c.cgs.value
+    t_rad = simulation_state.radiation_field_state.temperature.cgs.value
+
+    h = const.h.cgs.value
```
jvshields, line 49: Not a fan of these three values being saved like this in this function.

tardis/workflows/v_inner_solver.py
```diff
@@ -0,0 +1,353 @@
...
+
+        estimated_v_inner = self.estimate_v_inner()
+
+        estimates[0].update(
+            {"v_inner_boundary": estimated_v_inner, "mask": self.property_mask}
+        )
+
+        return estimates
+
+    def reproject(self, a1, m1, a2, m2):
```
jvshields, line 115: Not sure if this is would be used elsewhere, but it seems like quite a generic function to live in this class. Maybe this should live in util.py? 

Also, a1, m1, a2, and m2 could be a little more explicit. Just named "array1" and "mask1".

docs/workflows/v_inner_solver_workflow.ipynb
jvshields, line 1: This notebook is okay, but if it serves as the documentation for this workflow, it'd be nice to have a bit more explanation of what the workflow is, why you'd want to use it, and how you'd want to use it.

tardis/workflows/util.py
```diff
@@ -0,0 +1,87 @@
...
+            Planck Mean Optical Depth
+    """
+    index = plasma.atomic_data.lines.nu.index
+    taus = plasma.tau_sobolevs.loc[index]
+    freqs = plasma.atomic_data.lines.nu.values
+    order = np.argsort(freqs)
+    freqs = freqs[order]
+    taus = plasma.tau_sobolevs.values[order]
+
+    extra = bin_size - len(freqs) % bin_size
```
jvshields, line 32: I can't tell what the point of this bit of code is. Could there be more informative names, or at least a comment on why you need to handle the extra frequency bins like this?

tardis/workflows/util.py
```diff
@@ -0,0 +1,87 @@
...
+    freqs = freqs[1 : n_bins * bin_size + 1]
+
+    ct = simulation_state.time_explosion.cgs.value * const.c.cgs.value
+    t_rad = simulation_state.radiation_field_state.temperature.cgs.value
+
+    h = const.h.cgs.value
+    c = const.c.cgs.value
+    kb = const.k_B.cgs.value
+
+    def B(nu, T):
```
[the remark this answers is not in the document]
andrewfullard: I agree with the comments on this function, but it is something that is out of scope of this PR to improve

tardis/workflows/v_inner_solver.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def run(self):
+        """Run the TARDIS simulation until convergence is reached"""
+        converged = False
+        while self.completed_iterations < self.total_iterations - 1:
+            logger.info(
+                f"\n\tStarting iteration {(self.completed_iterations + 1):d} of {self.total_iterations:d}"
+            )
+
+            opacity_states = self.solve_opacity()
```
DeerWhale, line 316: the opacity_states should be a property of the class as well so it is updated every iteration

tardis/workflows/v_inner_solver.py
```diff
@@ -0,0 +1,357 @@
...
+            get_tau_integ(
+                self.plasma_solver,
+                self.opacity_states["opacity_state"],
+                self.simulation_state,
+            )[self.mean_optical_depth]
+        )
+
+        interpolator = interp1d(
+            tau_integ,
+            self.simulation_state.geometry.v_inner,  # Only use the active values as we only need a numerical estimate, not an index
```
DeerWhale, line 64: I agree with Jack's previous comments that we should only use the active shells when interpolating the tau_integ since the non-active shells carry the old values from previous iterations, so add the slicing `[self.simulation_state.geometry.v_inner_boundary_index:]` here

tardis/workflows/v_inner_solver.py
```diff
@@ -0,0 +1,360 @@
...
+        """
+        Compute the Rosseland Mean Optical Depth,
+        Estimate location where v_inner makes t=2/3 (or target)
+        Extrapolate with exponential fits
+
+        Need some way to return and inspect the optical depths for later logging
+        """
+        tau_integ = np.log(
+            get_tau_integ(
+                self.plasma_solver,
```
DeerWhale, line 58:
```suggestion
                self.plasma_solver[self.simulation_state.geometry.v_inner_boundary_index:],
```

DeerWhale (dismissed):

tardis/workflows/util.py
```diff
@@ -0,0 +1,86 @@
...
+    kappa_tot = kappa_thom + kappa_exp
+    kappa_rosseland = (
+        (udnu * kappa_tot**-1).sum(axis=0) / (udnu.sum(axis=0))
+    ) ** -1
+
+    dr = (
+        simulation_state.geometry.r_outer - simulation_state.geometry.r_inner
+    ).cgs.value
+    dtau = kappa_planck * dr
+    planck_integ_tau = np.cumsum(dtau[::-1])[::-1]
```
jvshields, line 83: Why are these two lines being reversing the arrays  twice?

tardis/workflows/util.py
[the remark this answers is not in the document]
jvshields: Okay, but can we at least name it blackbody or something something that is more pep8 compliant than just a single capitalized letter? (Along with U below)

tardis/workflows/util.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+from astropy import constants as const
+
+
+def get_tau_integ(plasma, opacity_state, simulation_state, bin_size=10):
```
jvshields, line 5: Being that this function isn't using numba or anything, and we're free using astropy units throughout this, can we just not do all of these .values throughout the function? This also solves the comment below about saving things like const.h.cgs.value below. Rather, I think we should just use const.h in the function and apply this throughout to let astropy units work as intended. Tau should come out as unitless in the end anyway, but the code will be much cleaner and we'll know more explictly that the units agree.

tardis/workflows/v_inner_solver.py
```diff
@@ -0,0 +1,360 @@
...
+
+# TODO:
+# Option to estimate initial v_inner from electron opacity
+# Add option for desired optical depth
+# Think about csvy vs other formats for specifying grid
+# Handle non-explicit formats when going out of the simulation
+
+
+class InnerVelocitySolverWorkflow(SimpleTARDISWorkflow):
+    TAU_TARGET = np.log(2.0 / 3.0)
```
jvshields, line 24: This line doesn't do anything does it? On line 39 we're setting self.TAU_TARGET, not TAU_TARGET. 

Either way, I think it's cleaner to just do an else statement on line 40 that sets self.TAU_TARGET to log(2/3), and probably include an info logger statement saying that that target was assumed.

tardis/workflows/v_inner_solver.py
```diff
@@ -0,0 +1,360 @@
...
+        """
+        tau_integ = np.log(
+            get_tau_integ(
+                self.plasma_solver,
+                self.opacity_states["opacity_state"],
+                self.simulation_state,
+            )[self.mean_optical_depth]
+        )
+
+        interpolator = interp1d(
```
jvshields, line 64: I think scipy's interp1d is deprecated. If we're just doing 1d interpolation, we could use numpy's np.interp instead.

tardis/workflows/v_inner_solver.py
andrewfullard: When you set a value in the class before __init__, it is an attribute of the class and therefore self-accessible.

tardis/workflows/util.py
andrewfullard: I generally agree, but I am not sure if I have time to rewrite this particular function

jvshields: I just helped Jing to do this, so she'll push to this in a sec.

DeerWhale (approved):

andrewfullard (approved):

PR #2806, opened 2024-08-19
Add molecular data exposure to tardis atom data

jvshields:
Needed for STARDIS to access the molecular information ingested by carsus added in https://github.com/tardis-sn/carsus/pull/396

Currently I don't think it's worth adding a molecular_data class, but if we ever want to do more complex handling of molecules, calculating them in a similar way to atomic number densities instead of just using equilibrium constants, then it might be worth reconsidering.


andrewfullard:
Seems reasonable. Should definitely become a class eventually.

andrewfullard (dismissed):

andrewfullard (approved):

wkerzendorf:
> Seems reasonable. Should definitely become a class eventually.

I think make it a dataclass now to put all of these together in one.

tardis/io/atom_data/base.py
```diff
@@ -762,3 +780,29 @@ def get_collision_matrix(self, species, t_electrons):
             )
         )
         return c_ul_matrix + c_lu_matrix.transpose(1, 0, 2)
+
+
+@dataclass
+class MoleculeData(object):
```
wkerzendorf, line 786: no need to inherit from object anymore

andrewfullard (approved):

andrewfullard (approved):

DeerWhale (approved):

PR #2812, opened 2024-08-27
TARDISData dataclass to hold data about atoms

andrewfullard:
### :pencil: Description

**Type:** :rocket: `feature`

This class will store the unfiltered data about atoms, ions et al. for the TARDIS plasma and transport. Name suggestions welcome.

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/io/atom_data/tardis_data.py
+++ b/tardis/io/atom_data/tardis_data.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+import pandas as pd
+
+if TYPE_CHECKING:
+    from tardis.io.atom_data.base import MoleculeData
+    from tardis.io.atom_data.nlte_data import NLTEData
+
+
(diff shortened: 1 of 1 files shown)


wkerzendorf:
So the TARDIS Data class is now mirroring atomic data, correct? I would put it in the a new tardis_data.py package then.

tardis/io/atom_data/tardis_data.py
```diff
@@ -0,0 +1,35 @@
...
+
+@dataclass
+class TARDISData:
+    atom_data: pd.DataFrame
+    levels: pd.DataFrame
+    lines: pd.DataFrame
+    ionization_data: pd.DataFrame
+    nlte_data: NLTEData
+    photoionization_data: pd.DataFrame
+    two_photon_data = pd.DataFrame
```
wkerzendorf, line 21: you do `=` instead of `:`

andrewfullard:
We probably don't need this.

PR #2814, opened 2024-08-29
Move to qgridnext

sarthak-dv:
### :pencil: Description

To avoid any future errors that may be caused by the deprecated `qgrid` library we are temporarily moving to [`qgridnext` ](https://github.com/zhihanyue/qgridnext)until we find a reliable alternative.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,10 +7,10 @@ name: tests
 on:
   push:
     branches:
-    - '*'
+      - "*"
   pull_request:
     branches:
-    - '*'
+      - "*"
     types:
       - opened
       - reopened
@@ -25,18 +25,17 @@ on:
         default: false
 
 env:
-  CACHE_NUMBER: 0  # increase to reset cache manually
-  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data --cov=tardis --cov-report=xml --cov-report=html --cov-append 
+  CACHE_NUMBER: 0 # increase to reset cache manually
+  PYTEST_FLAGS: --tardis-refdata=${{ github.workspace }}/tardis-refdata --tardis-regression-data=${{ github.workspace }}/tardis-regression-data --cov=tardis --cov-report=xml --cov-report=html --cov-append
   CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
 defaults:
   run:
     shell: bash -l {0}
- 
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
-    
 
 jobs:
   tests:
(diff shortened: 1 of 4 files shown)


jamesgillanders (approved):
Looks good to me!

jamesgillanders:
@andrewfullard We think it makes sense that this benchmark test fails but we need a second opinion. If it's a problem let us know -- if it isn't, then we will merge once Jaladh has reviewed it

andrewfullard:
Seems that qgridnext doesn't exist in conda, should be a pip install instead

sarthak-dv:
> Seems that qgridnext doesn't exist in conda, should be a pip install instead

Yes, that seems to be the issue here, I've added `pip install qgridnext` in tests.yml, do I need to add it somewhere else for the benchmarks to pass?

tardis_env3.yml
```diff
@@ -38,7 +37,7 @@ dependencies:
   - matplotlib-base
   - ipywidgets
   - plotly
-  - qgrid
+  - qgridnext
```
jaladh-singhal, line 40: This won't work because qgridnext is not installable through conda. I got an error when updating/creating tardis env. Explicitly define to use pip:

```suggestion
  - pip:
    - qgridnext
```

.github/workflows/tests.yml
```diff
@@ -65,29 +64,34 @@ jobs:
...
       - name: Install package git
         if: ${{ inputs.pip_git }}
         run: |
           pip install git+https://github.com/tardis-sn/tardis.git@master
 
+      - name: Install qgridnext
+        if: ${{ !inputs.pip_git }}
+        run: |
+          pip install qgridnext
+
```
jaladh-singhal, line 87: This shouldn't be needed after the change I suggested in env file. Try removing it

jaladh-singhal (changes requested):
Thank you for this PR @sarthak-dv - I tested the widgets notebooks locally, they work now!

Although, there's a problem with dependency specification - see my comment (it's a small change)

jaladh-singhal:
@sarthak-dv benchmarks "should" pass after the suggested change.

Unrelated, @andrewfullard do we need to update conda lock files as well?

andrewfullard:
> @sarthak-dv benchmarks "should" pass after the suggested change.
> 
> Unrelated, @andrewfullard do we need to update conda lock files as well?

Yes, we will need to regenerate the environment.

.github/workflows/tests.yml
sarthak-dv: I tried removing this but the tests failed because qgridnext is not installed, this code first installs qgridnext for tests to run. If we don't need this the other option would be to enable `pip install -e . --user` line in the file but that will install all the dependencies again.

sarthak-dv:
The tests in `test_spectrum_solver.py` are showing error, I don’t think it is related to the changes made in this PR. Other than that the benchmarks build is passing after making the changes @jaladh-singhal suggested.

andrewfullard:
> The tests in `test_spectrum_solver.py` are showing error, I don’t think it is related to the changes made in this PR. Other than that the benchmarks build is passing after making the changes @jaladh-singhal suggested.

Yeah that's fine- the test failures are because of changes to regression data in support of another PR

andrewfullard:
Looks like the docs build from the lock file, so they won't work until that is updated too.

.github/workflows/tests.yml
jaladh-singhal: ok

jaladh-singhal (approved):
Thanks for the updates. @andrewfullard if you're happy, please merge it.

jamesgillanders (approved):
Looks good to me -- thanks @sarthak-dv ! I'll leave Andrew to verify the various tests failing are okay, and if so, he can merge.

PR #2818, opened 2024-09-10
Revert "Move to qgridnext"

andrewfullard:
Reverts tardis-sn/tardis#2814.

Unfortunately qgridnext does not seem to resolve the issues with qgrid. Its pip install also makes our conda lock files very difficult to use.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,10 +7,10 @@ name: tests
 on:
   push:
     branches:
-      - "*"
+    - '*'
   pull_request:
     branches:
-      - "*"
+    - '*'
     types:
       - opened
       - reopened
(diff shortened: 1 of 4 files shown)


jaladh-singhal:
> Unfortunately qgridnext does not seem to resolve the issues with qgrid.

Do you mean issues regarding installation? Or issues in rendering widgets too? 

Because I tested latter and they worked. About the former, I built my environment from tardis_env3.yml, do we absolutely need to use conda lock files for visualization use cases?

andrewfullard:
> > Unfortunately qgridnext does not seem to resolve the issues with qgrid.
> 
> @andrewfullard Do you mean issues regarding installation? Or issues in rendering widgets too?
> 
> Because I tested latter and they worked. About the former, I built my environment from tardis_env3.yml, do we absolutely need to use conda lock files for visualization use cases?

I created an environment and tested it on a jupyter server, and got the same Java errors as with qgrid. I can double check. Regardless, we have lock files because we want a reproducible environment, and that should extend to visualizations.

andrewfullard:
I tested in VS Code with qgridnext and it appears to be working, which is great. However, it does fail on the *older* notebook server style ("classicnb") with the Java errors.

andrewfullard:
After discussion- this is being closed in favour of a try-except clause in the import block of qgridnext

PR #2928, opened 2025-01-07
Cleanup SDECPlotter and LIVPlotter class

abhinavohri:
### :pencil: Description

This PR aims at cleaning up the SDECPlotter and LIVPlotter class.
1. Remove SDECData as it duplicates a lot of stuff from the simulation class.
2. Remove SDECPlotter instance from LIVPlotter class.
3. Create util functions for common methods between the two classes.
 
 Regression Data: https://github.com/tardis-sn/tardis-regression-data/pull/46
### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -152,6 +152,7 @@ class SDECPlotter:
             }
             self._keep_colour = keep_colour
         else:
+            self._full_species_list = None
             self._species_list = None
             self._species_mapped = None
             self._keep_colour = None


tardis/visualization/tools/simulation_packet_data.py
```diff
@@ -0,0 +1,359 @@
+import astropy.units as u
+import pandas as pd
+
+
+class SimulationPacketData:
```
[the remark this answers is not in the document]
KasukabeDefenceForce: Updated

andrewfullard (dismissed):

andrewfullard:
The class is duplicating a lot of information from the simulation. Could the spectrum, in particular, be separated and simply used via the spectrum class rather than copying the data into this new class? Similarly, the geometry information (v_inner, r_inner etc) is being duplicated. It may be more useful to duplicate the entire geometry class from the simulation if that is what is needed for visualization. If the geometry parts are NOT all needed, then referencing the simulation rather than duplicating the data would be best.

As far as the notebook goes, it would be most useful to demonstrate the packet tracking information that is contained in the class, since the spectrum information is already plotted elsewhere e.g. in the quickstart notebook. 

Please comment etc with your ideas for any better ways to package the data necessary for the SDEC and LIV plots and how that might look. Multiple classes may be needed, but in general it would be good to limit duplication as I noted above.

tardis/visualization/plot_util.py
```diff
@@ -79,3 +89,326 @@ def to_rgb255_string(color_tuple):
...
+                else:
+                    atomic_number = element_symbol2atomic_number(species)
+                    species_ids = [
+                        (atomic_number, ion_number)
+                        for ion_number in np.arange(atomic_number)
+                    ]
+                    requested_species_ids.append(species_ids)
+                    species_mapped[(atomic_number, 0)] = species_ids
+                    # add the atomic number to a list so you know that this element should
+                    # have all species in the same colour, i.e. it was requested like
```
andrewfullard, line 326: What does colour have to do with this?

tardis/visualization/plot_util.py
```diff
@@ -79,3 +89,326 @@ def to_rgb255_string(color_tuple):
...
+            "packet_energies": u.Quantity(
+                hdf[f"{packet_prefix}/output_energy"].to_numpy()[
+                    emitted_packet_mask
+                ],
+                "erg",
+            ),
+        }
+
+
+def parse_species_list_util(species_list):
```
andrewfullard, line 251: Does this match the Carsus parser, or is that planned for a different PR?

KasukabeDefenceForce: I would prefer a separate PR for that.

KasukabeDefenceForce: Also,one more clarification by matching the carsus parser do we mean using pyparser or just matching the output?

tardis/visualization/plot_util.py
KasukabeDefenceForce: all ions of one element will have the same colour.

andrewfullard: Okay

tardis/visualization/plot_util.py
andrewfullard: Probably just match the output and remove the functionality from Carsus. Or use the Carsus functionality in TARDIS, whichever is easier.

andrewfullard:
Looking good overall. I will ask the usual "tests?" question.

tardis/visualization/tests/test_plot_util.py
```diff
@@ -0,0 +1,187 @@
...
+            "species_mapped": plot_object,
+            "species_list": np.array(species_list_result),
+            "keep_colour": np.array(keep_colour_result),
+            "full_species_list": np.array(full_species_list),
+        }
+
+        return PlotDataHDF(**property_group)
+
+    def test_parse_species_list_util(self, regression_data, generate_parse_species_hdf):
+        expected = regression_data.sync_hdf_store(generate_parse_species_hdf)
```
andrewfullard, line 150: Replace with parametrize

tardis/visualization/conftest.py
```diff
@@ -0,0 +1,38 @@
+from tardis.base import run_tardis
+import pytest
+from copy import deepcopy
+from tardis.base import run_tardis
+
+@pytest.fixture(scope="module")
+def simulation_simple(config_verysimple, atomic_dataset):
```
andrewfullard, line 7: I believe there is already a fixture with vpacket tracking etc turned on

jvshields (approved):

PR #2944, opened 2025-01-12
tests for composition

karthik11135:
### :pencil: Description
composition tests written and also added a small test for line info

### DIFF
--- a/tardis/model/matter/tests/test_composition.py
+++ b/tardis/model/matter/tests/test_composition.py
@@ -23,7 +23,7 @@ def composition_instance(config_verysimple):
 def test_elemental_mass_fraction(composition_instance):
     elemental_mass_fraction = composition_instance.elemental_mass_fraction
     assert isinstance(elemental_mass_fraction, pd.DataFrame)
-    np.testing.assert_allclose(elemental_mass_fraction.sum(), 1.0, rtol=1e-3)
+    np.testing.assert_allclose(elemental_mass_fraction.sum(), 1.0)
 
 
 def test_elemental_number_density(composition_instance):
@@ -38,14 +38,14 @@ def test_elemental_number_density(composition_instance):
         mass_fractions.index
     )
     
-    expected_density = mass_fractions.multiply(density_value).divide(
+    expected_number_density = mass_fractions.multiply(density_value).divide(
         element_masses,
         axis=0
     )
     
     np.testing.assert_allclose(
         number_density.values,
-        expected_density.values,
+        expected_number_density.values,
         rtol=1e-10
     )
 
--- a/tardis/model/matter/tests/test_composition.py
+++ b/tardis/model/matter/tests/test_composition.py
@@ -84,3 +84,4 @@ def test_calculate_elemental_cell_masses(composition_instance):
 
     expected_masses = composition_instance.elemental_mass_fraction * (density * volume).to(u.g).value
     np.testing.assert_allclose(elemental_masses.values, expected_masses.values)
+
--- a/tardis/model/matter/tests/test_composition.py
+++ b/tardis/model/matter/tests/test_composition.py
@@ -32,7 +32,6 @@ def test_elemental_mass_fraction(test_composition_simple, regression_data):
 def test_elemental_number_density(test_composition_simple, regression_data):
     number_density = test_composition_simple.elemental_number_density
 
-    print(type(number_density), "numba")
     assert isinstance(number_density, pd.DataFrame)
     assert np.all(number_density.values >= 0)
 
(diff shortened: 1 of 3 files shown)


tardis/model/matter/tests/test_composition.py
```diff
@@ -0,0 +1,87 @@
...
+    composition, _ = parse_composition_from_config(
+        atom_data, config_verysimple, time_explosion, geometry
+    )
+    return composition
+
+
+def test_elemental_mass_fraction(composition_instance):
+    elemental_mass_fraction = composition_instance.elemental_mass_fraction
+    assert isinstance(elemental_mass_fraction, pd.DataFrame)
+    np.testing.assert_allclose(elemental_mass_fraction.sum(), 1.0, rtol=1e-3)
```
andrewfullard, line 26: Why only a tolerance of 1e-3? That seems large.

tardis/model/matter/tests/test_composition.py
```diff
@@ -0,0 +1,87 @@
...
+    assert np.all(number_density.values >= 0)
+
+    # Calculating expected number density manually
+    density_value = composition_instance.density.to(u.g / u.cm**3).value
+    mass_fractions = composition_instance.elemental_mass_fraction
+    element_masses = composition_instance.effective_element_masses.reindex(
+        mass_fractions.index
+    )
+    
+    expected_density = mass_fractions.multiply(density_value).divide(
```
andrewfullard, line 41: rename to `expected_number_density`

tardis/visualization/widgets/tests/test_line_info.py
```diff
@@ -53,6 +53,13 @@ def test_get_species_interactions(
             )
             assert species_interactions_df.shape == (expected_df_length, 1)
 
+    def test_get_middle_half_edges(self, wavelength_range, filter_mode):
```
andrewfullard, line 56: Why is this part of this PR?

tardis/model/matter/tests/test_composition.py
karthik11135: No specific reason why exactly 1e-3. thought it would be acceptable tolerance. Will remove.

tardis/visualization/widgets/tests/test_line_info.py
karthik11135: Thought a separate PR for this small test wouldnt look good. I'll create a separate one for this to keep everything clean 👍

tardis/model/matter/tests/test_composition.py
```diff
@@ -0,0 +1,87 @@
...
+import pandas as pd
+from astropy import units as u
+
+from tardis.io.model.parse_geometry_configuration import parse_geometry_from_config
+from tardis.io.model.parse_atom_data import parse_atom_data
+from tardis.io.model.parse_composition_configuration import parse_composition_from_config
+
+
+@pytest.fixture(scope="function")
+def composition_instance(config_verysimple):
```
jvshields, line 12: Take a look at some of the other fixtures we have defined throughout the code. We would prefer a more consistent naming scheme than this.

karthik11135: Have made changes. Kindly take a look. Thanks.

karthik11135:
Hi, 
Would be happy to make any other required changes. Kindly let me know. 
Thanks.

wkerzendorf:
can you use the regression data.

karthik11135:
> can you use the regression data.

Made the changes. Kindly take a look at my PR now. Thanks.

andrewfullard:
> > can you use the regression data.
> 
> Made the changes. Kindly take a look at my PR now.
> 
> Edit: I made use of regression data fixture as asked. Before running the tests I ran the tests with generate reference. Its creating files in the regression data.
> 
>     1. Is this the right approach?
> 
>     2. If so, in that case do I have to create a PR to the regression repository with those files?
> 
> 
> Thanks.

That's correct, and yes you need to create a PR in the regression data repository.

karthik11135:
> > > can you use the regression data.
> > 
> > 
> > Made the changes. Kindly take a look at my PR now.
> > Edit: I made use of regression data fixture as asked. Before running the tests I ran the tests with generate reference. Its creating files in the regression data.
> > ```
> > 1. Is this the right approach?
> > 
> > 2. If so, in that case do I have to create a PR to the regression repository with those files?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thanks.
> 
> That's correct, and yes you need to create a PR in the regression data repository and link to it here

Made a PR to the regression data repository as asked. (https://github.com/tardis-sn/tardis-regression-data/pull/44)

wkerzendorf:
I'll take this over and finish it.

PR #2945, opened 2025-01-12
from simulation test for rpacket plot

karthik11135:
### :pencil: Description

The from simulation method of rpacket plot did not have a test before. 

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/visualization/tools/tests/test_rpacket_plot.py
```diff
@@ -11,6 +11,36 @@
...
+            simulation_rpacket_tracking
+        )
+
+        assert rpacket_plotter_change.no_of_packets == 15
+
+        temp_store = simulation_rpacket_tracking.transport.transport_state.rpacket_tracker_df
+        del simulation_rpacket_tracking.transport.transport_state.rpacket_tracker_df
+        with pytest.raises(AttributeError):
+            RPacketPlotter.from_simulation(simulation_rpacket_tracking)
+        simulation_rpacket_tracking.transport.transport_state.rpacket_tracker_df = temp_store
```
atharva-2001, line 41: what does this mean?

karthik11135: When sim.transport.transport_state do not contain the attribute rpackettrackerdf, the attributeerror must be raised. I'm testing that case also. For that, I'm storing the value of that (rpackettrackerdf) and deleting it to test the error. After that, I'm assigning it back.

atharva-2001, line 41: you could have used pandas testing method here

atharva-2001: And why are the tests failing?

karthik11135: When I pushed the code, I tested with the regression data repo using tardis-regression-data flag. Unsure why they are failing. I will look into it and come up with the solution. 

I will use the necessary pandas testing method sir.

karthik11135:
@atharva-2001 
I've updated the PR to use an existing fixture instead of modifying one.

While running the tests, I encountered some Python fatal errors related to concurrency, particularly in the transport and plasma modules. These seem to come from Numba's threading layer. I was able to resolve the issue locally by disabling parallel execution in the numba_cumulative_trapezoid function in fast_array_util.py.

Also, the test_continuum.py test in the plasma module passes when run in isolation but fails when run with others. This behavior also occurs on the master branch, so it's likely unrelated to this PR. I would be glad to work on this if I could get some guidance. 

Thanks for your time!

andrewfullard:
We are working on the API for visualization packages so this test won't be relevant soon. Thanks for the contribution.

PR #2974, opened 2025-02-16
Refactor isotope mass fraction handling and configuration

wkerzendorf:
- Remove raw_isotope_abundance from Composition class
- Update mass fraction parsing to conditionally apply isotope decay
- Modify default model_isotope_time_0 to -1 s to allow skipping decay
- Update test fixtures to explicitly set model_isotope_time_0
- Simplify mass fraction parsing functions by removing redundant returns


tardis/io/configuration/schemas/model_definitions.yml
```diff
@@ -181,7 +181,7 @@ definitions:
           description: filename
         model_isotope_time_0:
           type: quantity
-          default: 0 s
+          default: -1 s
```
jvshields, line 184: Why -1? Doesn't that mean something different?

wkerzendorf: it mens we are using the exact same time as time_explosion. Can put in the description if that helps.

jvshields: I just want the description field to say something like, "-1 is a default value that will assume the isotopes are provided at the time of the requested model (i.e., the isotopes will not be decayed). If a positive value is provided, the isotopes will be for that amount of time for the requested tardis simulation."

tardis/io/model/parse_mass_fraction_configuration.py
```diff
@@ -89,18 +89,16 @@ def parse_mass_fractions_from_config(config, geometry, time_explosion):
...
-        "model_isotope_time_0", 0.0 * u.day
-    )
-    isotope_mass_fractions = IsotopicMassFraction(
-        isotope_mass_fractions, time_0=model_isotope_time_0
-    ).decay(time_explosion)
+    model_isotope_time_0 = config.model.abundances.model_isotope_time_0
+    if model_isotope_time_0 >= 0 * u.s:
+        isotope_mass_fractions = IsotopicMassFraction(
+            isotope_mass_fractions, time_0=model_isotope_time_0
+        ).decay(time_explosion)
```
jvshields, line 96: Potentially add a logger message here, either info or warning, saying that isotopes are assumed to not decay.

tardis/energy_input/tests/test_gamma_ray_channel.py
```diff
@@ -76,7 +76,7 @@ def gamma_ray_test_composition(gamma_ray_simulation_state):
     """
 
     raw_isotopic_mass_fraction = (
```
Knights-Templars: rename "raw_isotope_mass_fraction"

tardis/io/model/parse_mass_fraction_configuration.py
```diff
@@ -89,18 +89,16 @@ def parse_mass_fractions_from_config(config, geometry, time_explosion):
...
     # and not gone through the schema validator
-    raw_isotope_mass_fractions = isotope_mass_fractions
-    model_isotope_time_0 = config.model.abundances.get(
-        "model_isotope_time_0", 0.0 * u.day
-    )
-    isotope_mass_fractions = IsotopicMassFraction(
-        isotope_mass_fractions, time_0=model_isotope_time_0
-    ).decay(time_explosion)
+    model_isotope_time_0 = config.model.abundances.model_isotope_time_0
+    if model_isotope_time_0 >= 0 * u.s:
```
jvshields: This check should be changed to just if model_isotope_time_0 now I think, since the default is None instead of -1 now.

tardis/io/configuration/schemas/model_definitions.yml
```diff
@@ -199,7 +200,7 @@ definitions:
           - uniform
         model_isotope_time_0:
           type: quantity
-          default: 0 s
```
jvshields, line 202: Make consistent, with lines 184-186. I think the description should also be copied.

tardis/io/model/parse_mass_fraction_configuration.py
wkerzendorf: done - see below

PR #2981, opened 2025-02-19
v_inner_solver store plasma state fxn

erinvisser:
### :pencil: Description

**Type:** :rocket: `feature` 

Based on the function in the PlasmaStateStorerMixin class, I added a store_plasma_state method to the v_inner_solver workflow, with the additional features of the computed optical depth at each iteration and the v_inner_boundary value at each iteration. 

For this,  I initialized these new lists in __init__, updated tau_integ to be a self variable, and then called the new function at the beginning and end of run(). 

This allows you to then extract the plasma state at each iteration and plot them and their convergence.  

- [ ] Testing pipeline
- [ ] Other method (describe)


DeerWhale (approved):

tardis/workflows/v_inner_solver.py
```diff
@@ -38,13 +38,52 @@ def __init__(self, configuration, mean_optical_depth="rosseland", tau=None):
         if tau is not None:
             self.TAU_TARGET = np.log(tau)
 
+        self.iterations_w = np.full((self.total_iterations, self.simulation_state.no_of_shells), np.nan)
```
jvshields, line 41: Look at https://github.com/tardis-sn/tardis/blob/cec8b531cc56e66219fe3ca2efdbf837c008b852/tardis/simulation/base.py#L80

If this converges before max iterations, then all of these attributes will just be left with their default values, lots of nans. When it converges, probably you want to chop off these arrays after that point.

tardis/workflows/v_inner_solver.py
```diff
@@ -38,13 +38,52 @@ def __init__(self, configuration, mean_optical_depth="rosseland", tau=None):
...
+            Rosseland mean optical depth
+        """
+        self.iterations_t_rad[i, -len(t_radiative):] = t_radiative
+        self.iterations_w[i, -len(dilution_factor):] = dilution_factor
+        self.iterations_electron_densities[i, -len(electron_densities):] = electron_densities.values
+        self.iterations_t_inner[i] = t_inner
+
+        self.iterations_v_inner_boundary[i] = v_inner_boundary
+        
+        self.iterations_mean_optical_depth[i,-len(tau_integ):] = tau_integ
```
jvshields, line 85: inconsistent spacing. Are you using ruff?

tardis/workflows/v_inner_solver.py
```diff
@@ -38,13 +38,52 @@ def __init__(self, configuration, mean_optical_depth="rosseland", tau=None):
...
+        electron_densities : np.ndarray
+            electron density
+        t_inner : astropy.units.Quantity
+            temperature of inner boundary
+        v_inner_boundary : astropy.units.Quantity
+            velocity of inner boundary
+        tau_integ : np.ndarray
+            Rosseland mean optical depth
+        """
+        self.iterations_t_rad[i, -len(t_radiative):] = t_radiative
```
jvshields, line 78: Shouldn't all of these lengths be the same things? (i.e., the number of shells for this run) 

Maybe better to just define a variable called num_of_active_shells or something and just use it for all of them.

DeerWhale (dismissed):

andrewfullard (changes requested):
Add a `store_plasma_state` option to enable/disable storing the plasma state as an input to the workflow.

tardis/workflows/v_inner_solver.py
```diff
@@ -314,6 +409,16 @@ def run(self):
             logger.info(
                 f"\n\tStarting iteration {(self.completed_iterations + 1):d} of {self.total_iterations:d}"
             )
+            self.store_plasma_state(
```
jvshields, line 412: Maybe this and line 462 should be optionally called? Add something like "store_v_inner_iteration_information" as a flag to the solve_plasma function and only do these things if the flag is set. 

That does also mean you would need to have some standard way of calling this though. Maybe it can be read in from the config file.

erinvisser: What do you mean by reading it in from the config file? Would this not just be another variable taken in when initializing the object

DeerWhale (approved):

DeerWhale (approved):

tardis/workflows/v_inner_solver.py
```diff
@@ -31,20 +31,115 @@ def __init__(self, configuration, mean_optical_depth="rosseland", tau=None):
             self.convergence_strategy.v_inner_boundary
         )
 
+        self.store_plasma = self.convergence_solvers.v_inner_boundary.store_plasma
```
DeerWhale, line 34:
```suggestion
        self.store_plasma = configuration.montecarlo.convergence_strategy.v_inner_boundary.store_plasma  
```

tardis/io/configuration/schemas/montecarlo_definitions.yml
```diff
@@ -62,6 +62,10 @@ definitions:
           type: object
           additionalProperties: false
           properties:
+            store_plasma:
```
jvshields, line 65:
```suggestion
            store_iteration_properties:
```

jvshields (changes requested):

DeerWhale (dismissed):

tardis/workflows/v_inner_solver.py
```diff
@@ -37,20 +37,115 @@ def __init__(
             self.convergence_strategy.v_inner_boundary
         )
 
+        self.store_plasma = configuration.montecarlo.convergence_strategy.v_inner_boundary.store_plasma
```
DeerWhale, line 40:
```suggestion
        self.store_iteration_properties = configuration.montecarlo.convergence_strategy.v_inner_boundary.store_iteration_properties
```

tardis/workflows/v_inner_solver.py
```diff
@@ -315,6 +412,17 @@ def run(self):
             logger.info(
                 f"\n\tStarting iteration {(self.completed_iterations + 1):d} of {self.total_iterations:d}"
             )
+            if self.store_plasma:
```
DeerWhale, line 415:
```suggestion
            if self.store_iteration_properties:
```

tardis/workflows/v_inner_solver.py
```diff
@@ -354,6 +462,20 @@ def run(self):
             self.final_iteration_packet_count,
             self.virtual_packet_count,
         )
+        if self.store_plasma:
```
DeerWhale, line 465:
```suggestion
        if self.store_iteration_properties:
```

jvshields (approved):

andrewfullard (approved):

PR #3002, opened 2025-03-04
Flatten macroatom

jvshields:
This PR mostly gets rid of the nested structure of macroatom transition probability calls. Not sure if it's more reasonable to just try to incorporate Wolfgang's TransitionProbabilities class he made in the opacities restructure.


andrewfullard (changes requested):
Much nicer. The class in opacities is the replacement for the plasma version that yes, should probably be merged into the macro atom in some form eventually. 

Looks like there are some Ruff mistakes to fix in the changed files. I think test coverage of the len(j_blues) == 0 clause would be useful too, as well as transcribing the comment from Aoife.

jvshields:
> Much nicer. The class in opacities is the replacement for the plasma version that yes, should probably be merged into the macro atom in some form eventually.
> 
> Looks like there are some Ruff mistakes to fix in the changed files. I think test coverage of the len(j_blues) == 0 clause would be useful too, as well as transcribing the comment from Aoife.

The stuff that ruff is complaining about is stuff like bare excepts which I don't particularly want to mess with as part of this PR. 

I have restored the comment from Aoife, though I'm not quite sure how to cover the len(j_blues) clause without making a much larger fixture with an uninitialized or custom macroatom inputs.

andrewfullard (approved):

jvshields:
Let's turn the comment from Aoife into an issue that we point to in the code.

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -66,19 +70,41 @@ def solve_transition_probabilities(
...
-        transition_probabilities = calculate_transition_probabilities(
-            atomic_data,
-            beta_sobolev,
-            mean_intensities,
+        # I wonder why?
+        # Not sure who wrote this but the answer is that when the plasma is
+        # first initialised (before the first iteration, without temperature
+        # values etc.) there are no j_blues values so this just prevents
+        # an error. Aoife.
+        # Josh followup - I renamed j_blues to mean_intensities to be more agnostic
```
wkerzendorf, line 78: Thus we require an estimator for Jb  ji, the mean intensity at frequency ν+  ji – i.e.  in the far blue wing of the transition j → i. This can also be derived from Eq. (43).
Lucy 2003 - I think it might have been defined in Lucy 1999

andrewfullard (approved):

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -51,8 +55,9 @@ def solve_transition_probabilities(
         ----------
         atomic_data : tardis.io.atom_data.AtomData
             Atomic Data
-        mean_intensities : pd.DataFrame
-            Mean intensity of the radiation field for each shell
+        mean_intensities_lines_blue_wing : pd.DataFrame
+            Mean intensity of the radiation field of each line in the blue wing for each shell
+            For more detail see Lucy 2003, https://www.aanda.org/component/article?access=bibcode&bibcode=&bibcode=2003A%2526A...403..261LFUL
```
wkerzendorf, line 60: maybe DOI or something like that.

DeerWhale (approved):

PR #3025, opened 2025-03-21
Remove Deprecated Astropy `TestRunner`

atharva-2001:
### :pencil: Description

**Type:** :beetle: `bugfix` | :rocket: `feature` | :biohazard: `breaking change` | :vertical_traffic_light: `testing` | :memo: `documentation` | :roller_coaster: `infrastructure`

Closes https://github.com/tardis-sn/tardis/issues/2972 

### :pushpin: Resources

Examples, notebooks, and links to useful references.

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #3032, opened 2025-03-24
Remove test from __init__.py

pllim:
### :pencil: Description

This is a direct follow-up of #3025 . `test` should have never move. It should have been deleted.

xref https://github.com/tardis-sn/tardis/issues/2972


tardis/__init__.py
```diff
@@ -1,25 +1,16 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+import sys
 
-# Packages may add whatever they like to this file, but
-# should keep this content at the top.
-# ----------------------------------------------------------------------------
-import os
-
-__all__ = ['__version__', 'test']
+__all__ = ['__version__', 'run_tardis', 'yaml_load']
```
pllim, line 4: I am guessing you also want `run_tardis` and `yaml_load` in `__all__` since you specifically import them below? If not, I can remove.

pllim:
FWIW mailmap is only needed to clean up duplicate author entries, so I think we're good, bot.

tardis/__init__.py
atharva-2001: Hi @pllim, thank you for fixing my mistake. We don't use `yaml_load` as far as I know, but we do use `run_tardis`. So I think it should be good to remove the `yaml_load`.

pllim: Okay I removed yaml_load. AFAIK there is no need to import it if it is not going to be exposed, so I removed the import as well. Thanks for the review!

atharva-2001: Thank you! This should be good to merge then, right?

andrewfullard (dismissed):

andrewfullard (approved):

wkerzendorf (approved):

PR #3035, opened 2025-03-29
GSoC 2025: First Objective for Line Identification - hmperk18

hmperk18:
### :pencil: Description

Following the source code the `SDECPlotter`, I constructed a table of packet information including the last element with which it interacted. Then, I made a histogram of packet energies per element from the example TARDIS simulation configurations.  Using the distribution of energy packets, I make a first attempt at identifying features and marking the location as seen below.

image

Length of the marker lines are simply scaled by atomic number and sorted to have the longer markers at the back to identify locations of overlapping markers.

### :pushpin: Resources

This work can be found in a jupyter notebook `docs/hmlp_line_id.ipynb` along side the `quickstart.ipynb`.

- [ ] Testing pipeline
- [ ] Other method (describe)


docs/hmlp_line_id.ipynb
```diff
@@ -2400,14 +2402,20 @@
     "    mask = packets_df['last_line_interaction_atom'] == identifier\n",
     "    energies = packets_df[mask]['energies'] # <-- example of filtering the table by origin element\n",
     "\n",
-    "    ax.hist(energies, bins=100, histtype='step', color=plotter._color_list[species_counter])\n",
-    "    \n",
+    "    ax.hist(energies, bins=100, color=plotter._color_list[species_counter], zorder = -len(energies))\n",
```
hmperk18, line 2405: Looking forward to a review! I am curious if there is better choice for binning of the histogram.

hmperk18:
I have extended my notebook to include a preliminary attempt at labeling lines in the SDEC spectra. I think that my bin size selection is have an effect on peak finding and thus there is a small discrepancy between the lines and the actual spectra.  It also maybe be better to search for peaks scaled by luminosity or total energy contributed by the packets rather than the total number of packets.  Would love to hear any recommendations! 
lineLabels

wkerzendorf:
> I have extended my notebook to include a preliminary attempt at labeling lines in the SDEC spectra. I think that my bin size selection is have an effect on peak finding and thus there is a small discrepancy between the lines and the actual spectra. It also maybe be better to search for peaks scaled by luminosity or total energy contributed by the packets rather than the total number of packets. Would love to hear any recommendations! lineLabels

Peak finding will not work as the spectrum does not really consist of "lines" (they are smeared put P-Cygni profiles overlapping with each other). But it's great to try out new things in these PRs.

wkerzendorf:
I think for now having separate functionality to add for a given wavelengths ticks on the top with given labels would be great

hmperk18:
I have updated the description to include the final plot, as it best showcases the end result of my work here, but aesthetically, I believe this plot of the distribution of energy packets to be best.
image

andrewfullard:
Thank you very much for your contribution. We close GSOC-related pull requests after contributors are announced. You are welcome to contribute to TARDIS any time.

PR #3043, opened 2025-04-01
Add workflows to documentation

erinvisser:
### :pencil: Description

To display the workflow docs on the documentation, I generated a new function in conf.py that compiles all notebooks within the workflows folder into a workflows.rst file. Then, I added that new section to index.rst, so it appears between How-To Guides and FAQs on the top section of the documentation. 

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -385,11 +385,11 @@ def generate_how_to_guides_page(app):
 def generate_worflows_page(app):
     "Create workflows.rst"
     notebooks = ""
+    workflows_path = Path("workflows")
 
-    for root, dirs, fnames in os.walk("workflows/"):
-        for fname in fnames:
-            if 'workflow' in fname and fname.endswith(".ipynb") and "checkpoint" not in fname:
-                notebooks += f"\n* :doc:`{root}/{fname[:-6]}`"
+    for notebook in workflows_path.rglob("*.ipynb"):
+        if "workflow" in notebook.name and "checkpoint" not in notebook.name:
+            notebooks += f"\n* :doc:`{notebook.with_suffix('').as_posix()}`" 
 
     title = "Workflows\n*********\n"
     description = "The following pages contain the TARDIS workflows:\n\n These examples are intended to help users explore specific modules within TARDIS, with the goal of supporting their individual scientific objectives."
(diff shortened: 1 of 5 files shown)


andrewfullard:
Looking okay, the compiled notebooks in the docs are missing titles though (probably my fault :P )

erinvisser:
I can add titles (once the v_inner notebook PR is merged so i don't edit over that one)! What exactly is the difference between the standard and simple workflows?

andrewfullard:
I think this would benefit from pre-running the notebooks.

andrewfullard (approved):

docs/conf.py
```diff
@@ -382,6 +382,20 @@ def generate_how_to_guides_page(app):
...
         f.write(f"{title}\n{description}\n{notebooks}")
 
+def generate_worflows_page(app):
+    "Create workflows.rst"
+    notebooks = ""
+
+    for root, dirs, fnames in os.walk("workflows/"):
+        for fname in fnames:
+            if 'workflow' in fname and fname.endswith(".ipynb") and "checkpoint" not in fname:
+                notebooks += f"\n* :doc:`{root}/{fname[:-6]}`"
```
jvshields, line 392: I can see that this whole block of code is copied from existing things, so not necessarily your fault that the code is like this, but this entire logic block is a bit of nonsense that could be simplified heavily by using pathlib 

Here's an example from stack overflow 

from pathlib import Path
Path('my_file.mp3').suffix == '.mp3'

There are similar things to handle everything else that this is doing, like the .stem attribute

jvshields: You can try your hand at rewriting this logic with pathlib, or open an issue to keep note of it. Either is fine with me.

erinvisser: I've been having trouble building the docs locally... do you think the following does the same thing? do you want something similar done to generate_tutorials_page and generate_how_to_guides_page as well?
```
    notebooks = ""
    workflows_path = Path("workflows")

    for notebook in workflows_path.rglob("*.ipynb"):
        if "workflow" in notebook.name and "checkpoint" not in notebook.name:
            notebooks += f"\n* :doc:`{notebook.with_suffix('').as_posix()}`" 
```

andrewfullard (approved):
In the future we will want to work out a good way to demonstrate the outputs of these notebooks.

PR #3048, opened 2025-04-07
Remove legacy_solve() from OpacitySolver and migrate to solve()

jvshields:
This pr primarily removes the legacy_solve method from the opacity solver. I think it breaks the disable_line_scattering tests because beta_sobolevs were not hooked up correctly in the disable_line_scattering case. In the legacy case, beta_sobolevs were being grabbed from the plasma which did not go through opacity solver. This meant that disable_line_scattering in the tests was being ignored. 

This means we probably need to regenerate the regression data for this change.


andrewfullard:
You could do a regression data comparison to see what changes.

andrewfullard:
@jvshields this should be looked at again

andrewfullard (dismissed):

tardis/opacities/tests/test_opacity_solver.py
```diff
@@ -50,10 +50,11 @@ def test_opacity_solver(
             legacy_plasma.stimulated_emission_factor,
             beta_sobolev=actual.beta_sobolev,
         )
-        pdt.assert_frame_equal(
-            macro_atom_state.transition_probabilities,
-            legacy_plasma.transition_probabilities,
-        )
+        if not disable_line_scattering:
```
jvshields, line 53: Note this change from the most recent commit!

Let me know if there's a better way to do this. The fact that this wasn't here before but was still passing is basically showing that the macro atom was ignoring disable_line_scattering, since the plasma does not have access to that flag. 

I don't think it's worth making the plasma property be able to access that flag so right now this just skips it.

andrewfullard (approved):

PR #3049, opened 2025-04-09
Fixes the ruff toml issues

### LINKED ISSUE #3046: Issue with ruff and TOML rules
```
[Error - 5:22:48 PM] Ruff: Lint failed (ruff failed
  Cause: Failed to parse /Users/wkerzend/python/tardis/.ruff.toml
  Cause: TOML parse error at line 3, column 1
  |
3 | lint.ignore = [
  | ^^^^
Unknown rule selector: `conda-osx-arm64.lock`

)```

### PR DESCRIPTION
**Pull request by Sonu0305**
### :pencil: Description

**Type:** :beetle: `bugfix` 

- Resolved the scope of `exclude` key

Fixes #3046 

I have tested this too,

```python
(tardis) swayam-shah@Swayam:/mnt/c/Users/admin/Desktop/tardis$ ruff check --select ANN001 tardis/constants.py
warning: The following rules have been removed and ignoring them has no effect:
    - ANN101
    - ANN102
    - PT004

All checks passed!
``` 

### :pushpin: Resources

- [ ] Testing pipeline
- [x] Other method (describe)

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,5 +1,8 @@
 # Ruff configuration file adapted from astropy
 extend = "pyproject.toml"
+
+extend-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
+
 lint.ignore = [
     # NOTE: to find a good code to fix, run:
     # ruff --select="ALL" --statistics astropy/<subpackage>
@@ -214,8 +217,6 @@ lint.ignore = [
 lint.unfixable = [
     "E711"  # NoneComparison. Hard to fix b/c numpy has it's own None.
 ]
-
-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
 
 [lint.extend-per-file-ignores]
 "docs/*" = []
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -215,7 +215,8 @@ lint.unfixable = [
     "E711"  # NoneComparison. Hard to fix b/c numpy has it's own None.
 ]
 
+exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
+
 [lint.extend-per-file-ignores]
 "docs/*" = []
 
-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,6 @@ verbose = 0
 [tool.ruff]
 line-length = 80
 lint.select = ["ALL"]
-exclude=[]
 lint.ignore = [  # NOTE: non-permanent exclusions should be added to `.ruff.toml` instead.
 
     # flake8-builtins (A) : shadowing a Python built-in.


andrewfullard (dismissed):

.ruff.toml
```diff
@@ -215,7 +215,8 @@ lint.unfixable = [
     "E711"  # NoneComparison. Hard to fix b/c numpy has it's own None.
 ]
 
+exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
```
atharva-2001, line 218: Thanks, do you know why this worked?

Sonu0305: Yes, the `exclude` key seemed to be placed out of the scope of `[tool.ruff]` section.

atharva-2001: but this is inside ruff.toml right? [tool.ruff] is inside pyproject.toml right?

Sonu0305: One more possible - tested & working - fix I found out is by using `extend-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]`, placing this at the very start of the `ruff.toml`.

Hence, sharing the diff:

```diff
diff --git a/.ruff.toml b/.ruff.toml
index 440a5a81ef..72ef786449 100644
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,5 +1,8 @@
 # Ruff configuration file adapted from astropy
 extend = "pyproject.toml"
+
+extend-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
+
 lint.ignore = [
     # NOTE: to find a good code to fix, run:
     # ruff --select="ALL" --statistics astropy/<subpackage>
@@ -215,8 +218,6 @@ lint.unfixable = [
     "E711"  # NoneComparison. Hard to fix b/c numpy has it's own None.
 ]
 
-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
-
 [lint.extend-per-file-ignores]
 "docs/*" = []
```

Reference: [ruff docs (extend-exclude)](https://docs.astral.sh/ruff/settings/#__tabbed_5_2)

atharva-2001: does it work if you remove the exclude key from pyproject.toml then?

Sonu0305: > does it work if you remove the exclude key from pyproject.toml then?

Yes, it works, but compulsorily along with my previous [commit](8e1aeabb81df0994982d51a3ffc3dd8a1f723405).

atharva-2001: okay, then just removing the exclude key from pyproject.toml doesn't work?

Sonu0305: > okay, then just removing the exclude key from pyproject.toml doesn't work?

No, just removing it from pyproject.toml doesn't work.

Sonu0305: > here it says-
> 
> > Note that you'll typically want to use [extend-exclude](https://docs.astral.sh/ruff/settings/#extend-exclude) to modify the excluded paths.
> 
> So we should go with the extend-exclude

Correct, I'll make the necessary changes.

atharva-2001: I'd suggest removing the empty exclude key from pyproject- since the docs say we should not use it.

Sonu0305: > I'd suggest removing the empty exclude key from pyproject- since the docs say we should not use it.

Correct, Done.

.ruff.toml
```diff
@@ -1,5 +1,8 @@
 # Ruff configuration file adapted from astropy
 extend = "pyproject.toml"
+
+extend-exclude = ["conda-osx-arm64.lock", "conda-linux-64.lock", "conda-lock.yml"]
```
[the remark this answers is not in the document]
Sonu0305: Yes, but it only works correctly if it is placed after [L2](https://github.com/Sonu0305/tardis/blob/fix%233046/.ruff.toml#L2) (i.e. extend variable) and before [L221](https://github.com/Sonu0305/tardis/blob/fix%233046/.ruff.toml#L221). 

Also, the `extend` variable that u mentioned serves a different purpose over here, hence can't be clubbed together, need to be separate.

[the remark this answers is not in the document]
atharva-2001: is it not possible to move the pyproject.toml to extend exclude? what happens then? what purpose does that extend variable serve?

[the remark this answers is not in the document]
Sonu0305: > is it not possible to move the pyproject.toml to extend exclude? what happens then? what purpose does that extend variable serve?

No, we can't move `pyproject.toml` to `extend-exclude`

`extend` variable basically tells ruff to inherit all configurations from pyproject.toml

atharva-2001 (approved):
Thanks!

andrewfullard (approved):

PR #3050, opened 2025-04-14
Tests for workflows: Simple, Standard, and Inner Velocity Solver

erinvisser:
### :pencil: Description

Added tests to cover the Simple and Standard TARDIS Workflows using the [config_verysimple](https://github.com/tardis-sn/tardis/blob/2eeb45f9002c32f81d73d88abfe0000ed97bfe81/tardis/conftest.py#L206) and [simulation_verysimple](https://github.com/tardis-sn/tardis/blob/2eeb45f9002c32f81d73d88abfe0000ed97bfe81/tardis/conftest.py#L228) as defined in conftest.py.

Added tests to cover the v_inner_solver workflow. Created a new config file,  tardis_config_v_inner.yml, to run simulation including parameters only required for this workflow. 

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/workflows/tests/test_workflows.py
+++ b/tardis/workflows/tests/test_workflows.py
@@ -1,0 +1,28 @@
+import pandas as pd
+import pytest
+
+from tardis.io.configuration.config_reader import Configuration
+from tardis.simulation import Simulation
+from tardis.workflows.v_inner_solver import InnerVelocitySolverWorkflow
+from tardis.workflows.simple_tardis_workflow import SimpleTARDISWorkflow
+from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
+
+
+@pytest.fixture(scope="function")
+def v_inner_config(example_configuration_dir):
+    return Configuration.from_yaml(
+        example_configuration_dir / "tardis_config_v_inner.yml"
+    )
+
+@pytest.fixture(scope="function")
+def run_v_inner(v_inner_config, atomic_data_fname):
+    sim = InnerVelocitySolverWorkflow(v_inner_config)
+    sim.run()
+    return sim
+
+def test_v_inner_solver_workflow(run_v_inner):
+    actual = getattr(run_v_inner, 'completed_iterations')
+    if hasattr(actual, "value"):
+        actual = actual.value
+    expected = 10 
+    assert actual == expected
--- a/tardis/workflows/tests/test_workflows.py
+++ b/tardis/workflows/tests/test_workflows.py
@@ -1,12 +1,18 @@
-import pandas as pd
+from pathlib import Path
+
 import pytest
 
 from tardis.io.configuration.config_reader import Configuration
-from tardis.simulation import Simulation
+# from tardis.simulation import Simulation
 from tardis.workflows.v_inner_solver import InnerVelocitySolverWorkflow
 from tardis.workflows.simple_tardis_workflow import SimpleTARDISWorkflow
 from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
 
+@pytest.fixture(scope="module")
+def config(example_configuration_dir):
+    return Configuration.from_yaml(
+        example_configuration_dir / "tardis_configv1_verysimple.yml"
+    )
 
 @pytest.fixture(scope="function")
 def v_inner_config(example_configuration_dir):
@@ -21,8 +27,20 @@ def run_v_inner(v_inner_config, atomic_data_fname):
     return sim
 
 def test_v_inner_solver_workflow(run_v_inner):
-    actual = getattr(run_v_inner, 'completed_iterations')
+    actual = getattr(run_v_inner, "completed_iterations")
     if hasattr(actual, "value"):
         actual = actual.value
     expected = 10 
     assert actual == expected
+
+def test_simple_tardis_workflow(config):
+    sim = SimpleTARDISWorkflow(config)
+    sim.run()
+    assert hasattr(sim, "completed_iterations")
+    assert sim.completed_iterations == 10 # placeholder value
+
+def test_standard_tardis_workflow(config):
+    sim = StandardTARDISWorkflow(config)
+    sim.run()
+    assert hasattr(sim, "completed_iterations")
+    assert sim.completed_iterations == 10 # placeholder value
--- a/tardis/workflows/tests/test_workflows.py
+++ b/tardis/workflows/tests/test_workflows.py
@@ -1,46 +1,59 @@
-from pathlib import Path
-
 import pytest
+import numpy as np
+from astropy import units as u
 
 from tardis.io.configuration.config_reader import Configuration
-# from tardis.simulation import Simulation
 from tardis.workflows.v_inner_solver import InnerVelocitySolverWorkflow
 from tardis.workflows.simple_tardis_workflow import SimpleTARDISWorkflow
 from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
 
 @pytest.fixture(scope="module")
-def config(example_configuration_dir):
-    return Configuration.from_yaml(
-        example_configuration_dir / "tardis_configv1_verysimple.yml"
-    )
+def simulation_simple_workflow(config_verysimple, atomic_data_fname):
+    sim = SimpleTARDISWorkflow(config_verysimple, csvy=True)
+    sim.run()
+    return sim
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
+def simulation_standard_workflow(config_verysimple, atomic_data_fname):
+    sim = StandardTARDISWorkflow(config_verysimple, csvy=True)
+    sim.run()
+    return sim
+
+@pytest.fixture(scope="module")
 def v_inner_config(example_configuration_dir):
     return Configuration.from_yaml(
         example_configuration_dir / "tardis_config_v_inner.yml"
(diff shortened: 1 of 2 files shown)


tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,59 @@
...
+    actual = getattr(simulation_v_inner_workflow.simulation_state, key)
+    if not actual.isscalar:
+        actual = actual[-1]    
+    assert np.allclose(actual.to_value(), expected.to_value())
+
+@pytest.mark.parametrize("key",["t_radiative","dilution_factor","t_inner"])
+def test_simple_tardis_workflow(key, simulation_simple_workflow, simulation_verysimple):
+    actual = getattr(simulation_simple_workflow.simulation_state, key)
+    actual = actual.to_value() if hasattr(actual, "to_value") else actual
+    expected = getattr(simulation_verysimple.simulation_state, key)
```
jvshields, line 49: I don't think expected vs actual is a great way of phrasing these things. That makes more sense comparing something like a new value to regression data. Here I think you're comparing different workflows together, which is a useful test, but they're both freshly calculated.

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,59 @@
...
+import numpy as np
+from astropy import units as u
+
+from tardis.io.configuration.config_reader import Configuration
+from tardis.workflows.v_inner_solver import InnerVelocitySolverWorkflow
+from tardis.workflows.simple_tardis_workflow import SimpleTARDISWorkflow
+from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
+
+@pytest.fixture(scope="module")
+def simulation_simple_workflow(config_verysimple, atomic_data_fname):
```
jvshields, line 11: Maybe this should be called something like "workflow_simple" rather than "simulation_simple_workflow" They can have different attributes so keying in somebody else who might use this object that it is a different thing than a simulation object is important.

Similarly I'd probably rename sim to workflow as well. 

These comments apply to all of these tests too, but should be an easy enough find and replace

andrewfullard:
The tests seem to be failing to find the atomic data. Not sure if this is an issue with the workflows or the tests.

jvshields:
Should investigate why the tests are failing because they can't find the atom data.

tardis/workflows/tests/test_workflows.py
DeerWhale: Updated the phrasing to be more informative.

tardis/workflows/tests/test_workflows.py
DeerWhale: Updated the naming to emphasize the workflow.

DeerWhale:
> The tests seem to be failing to find the atomic data. Not sure if this is an issue with the workflows or the tests. I also wonder if we can just compare to the simulation regression data instead of the verysimple fixture (or if that is already happening anyway)

That's a good idea, now the **standard_workflow** is compared to the regression data produced for the [test_simulation.py](https://github.com/tardis-sn/tardis/blob/master/tardis/simulation/tests/test_simulation.py), the iteration values are consistent, but the estimators and output energies are somewhere different, looks like the run_tardis recompute the opacity state in the run_final while the workflow only recompute the MC state after convergence. 

To match with the run_tardis, added the step recompute the opacity state in simple and standard workflow after convergence, now all test pass comparing the simulation regression data.

Since the simple workflow doesn't have iteration values logged by default, it is currently cross compared with the standard workflow results. 

The v_inner test will fail for now till the newly generated regression data PR is merged.

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,181 @@
...
+from astropy import units as u
+
+from tardis.io.configuration.config_reader import Configuration
+from tardis.workflows.v_inner_solver import InnerVelocitySolverWorkflow
+from tardis.workflows.simple_tardis_workflow import SimpleTARDISWorkflow
+from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
+
+
+@pytest.fixture(scope="module")
+def config_simulation(example_configuration_dir, atomic_data_fname):
```
wkerzendorf: let's use the existing fixtures for the config - I think it's called config very simple.

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,181 @@
...
+    workflow = StandardTARDISWorkflow(config_simulation)
+    workflow.run()
+    return workflow
+
+
+@pytest.fixture(scope="function")
+def simulation_regression_data(regression_data):
+    """Fixture to access simulation regression data for test_simulation.py"""
+
+    class SimulationRegressionData:
```
wkerzendorf: why is there a class needed? can we just return the regression_file

tardis/workflows/tests/test_workflows.py
DeerWhale: the test_simulation used a modified version of that yml, I added a session scope fixture in test_simulation.py to make sure we use the same config as that, is that okay?

tardis/workflows/tests/test_workflows.py
DeerWhale: That's true, removed the class and cahnged it to directly read the red file in the test function.

andrewfullard (dismissed):

tardis/workflows/simple_tardis_workflow.py
```diff
@@ -493,8 +493,9 @@ def run(self):
             logger.error(
                 "\n\tITERATIONS HAVE NOT CONVERGED, starting final iteration"
             )
+        self.opacity_states = self.solve_opacity()
```
jvshields, line 496: If you want to do this also change line 469 and 472 in this file.

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,146 @@
...
+    [
+        "t_radiative",
+        "dilution_factor",
+        "t_inner",
+    ],
+)
+def test_simple_tardis_workflow(
+    simple_workflow_one_loop, standard_workflow_one_loop, attr
+):
+    # this test reply on that the standard workflow pass the test comparing with the regression data
```
jvshields: Is this supposed to be "rely on?" Looks like a typo that kinda changes meaning

jvshields: The comment is also generally a bit unclear though. I don't know why the test would need another test to pass specifically. Can it be a bit more informative?

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,146 @@
...
+    attr_simple_workflow = getattr(
+        simple_workflow_one_loop.simulation_state, attr
+    )
+    attr_standard_workflow = getattr(
```
jvshields, line 121: I'm not a fan of sticking standard worklow test things in the "test_simple_tardis_workflow" test. Does the name need to be changed?

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,146 @@
...
+        standard_workflow_one_loop.simulation_state, attr
+    )
+    if hasattr(attr_simple_workflow, "value"):
+        attr_simple_workflow = attr_simple_workflow.value
+    if hasattr(attr_standard_workflow, "value"):
```
jvshields, line 126: Same as last comment

tardis/workflows/tests/test_workflows.py
```diff
@@ -0,0 +1,146 @@
...
+        "iterations_mean_optical_depth",
+        "iterations_v_inner_boundary",
+    ],
+)
+def test_v_inner_solver_workflow(v_inner_workflow, attr, regression_data):
+    attr_data = getattr(v_inner_workflow, attr)
+    if hasattr(attr_data, "value"):
+        attr_data = attr_data.value
+    attr_data = pd.DataFrame(attr_data)
+    ref_data = regression_data.sync_dataframe(attr_data)
```
jvshields, line 148: I don't think this is how you should use this method? Sync_dataframe() takes a key which I don't think is what you're doing here.

tardis/workflows/tests/test_workflows.py
jvshields: Maybe call it "test_simple_against_standard_workflow"

tardis/workflows/tests/test_workflows.py
DeerWhale: changed the name now to test_simple_tardis_workflow_again_standard_workflow to be more informative

tardis/workflows/tests/test_workflows.py
DeerWhale: That's a much better name! Updated it.

tardis/workflows/simple_tardis_workflow.py
DeerWhale: Good catch, changed in both simple and standard workflow.

DeerWhale:
The test still failing because silly me generated the regression data in the wrong path, should be passing after merging this PR on regression-data repo: https://github.com/tardis-sn/tardis-regression-data/pull/60

andrewfullard:
Weirdly Mac is failing, but Ubuntu passing

andrewfullard:
Likely an issue in a sort function somewhere. Need to use `kind="stable"` for ALL sorting algorithms.

PR #3058, opened 2025-04-22
Docs build and logger widget fixes

atharva-2001:
### :pencil: Description

This PR fixes several different issues-
- [X] Fixes pyzmq errors that were due to unclosed HDF files
- [X] Fixes logging behavior outside run_tardis(dumps logs to normal cell output instead of in the widget)
- [X] Upgrades several actions to newer versions
- [X] Fixes the giant text block in docs and tests(when ran with the -s flag)
- [X] Excludes .lock, .rst, .yml files from the ruff check in the git command- the ruff by default doesn't check non python files, but the git command used to provide them, also it appears that the files provided will take priority over the ruff config. check will fail in this PR since its a pull_request_target event
-

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/io/logger/logger.py
```diff
@@ -149,6 +149,7 @@ def emit(self, record):
             stream_handler = logging.StreamHandler()
             stream_handler.setFormatter(logging.Formatter("%(name)s [%(levelname)s] %(message)s (%(filename)s:%(lineno)d)"))
             stream_handler.emit(record)
+            stream_handler.close()
```
andrewfullard, line 152: Does this mean the function opens and closes a StreamHandler for every log event?

atharva-2001: I'm sorry, I fixed it in https://github.com/tardis-sn/tardis/pull/3058/commits/3eb595c9fff6a83f7becc16289a4eb19045cc59d !

tardis/io/logger/logger.py
```diff
@@ -136,6 +136,9 @@ def __init__(self, log_outputs, colors, display_widget=True, display_handle=None
         self.display_widget = display_widget
         self.display_handle = display_handle
         self.logger_widget = logger_widget
+        if not self.display_widget or self.display_handle is None:
+            self.stream_handler = logging.StreamHandler()
```
andrewfullard, line 140: This is better- does the stream handler still need to be closed?

atharva-2001: I need to apply the formatting however, will let you know once done

andrewfullard:
Please update the PR title and description so we can track changes in this more reliably

.github/workflows/codestyle.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       
       - name: Show entire output pull request
         if: github.event_name == 'pull_request_target'
-        run: ruff check --output-format=concise $(git --no-pager diff --name-only origin/master HEAD --) | tee ruff_full.txt
+        run: ruff check --output-format=concise $(git --no-pager diff --name-only origin/master HEAD -- | grep -v -E '\.(lock|yml|rst)$') | tee ruff_full.txt
```
atharva-2001, line 46: right now it checks- "all files changed in this PR except files ending in .lock, .rst and .yml". Is it better to change it "all python files changed in this PR?"

andrewfullard (approved):

jvshields (approved):
Somewhat odd to initialize the widget handler and then close it later if it's not requested, rather than just not initialize the widget part of things in that case. Maybe we'll want to investigate doing that better in the future.

Issue #3061, opened 2025-04-28
Refactor logger widget to only be initialized if required


andrewfullard:
Currently the widget is created then torn down if it's not needed. It shouldn't be created if we don't need it.

PR #3062, opened 2025-04-28
Logger Widget Refactor

atharva-2001:
### :pencil: Description

Write a complete description of your changes, including the necessary context or any piece of information required to understand your work.

Closes https://github.com/tardis-sn/tardis/issues/3061 

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/base.py
+++ b/tardis/base.py
@@ -117,7 +117,7 @@ def run_tardis(
 
     simulation.run_convergence()
     simulation.run_final()
-    if logger_widget:
+    if logger_widget and tardislogger:
         tardislogger.remove_widget_handler()
         tardislogger.setup_stream_handler()
     return simulation
--- a/tardis/io/logger/logger.py
+++ b/tardis/io/logger/logger.py
@@ -3,10 +3,12 @@ import os
 import re
 import sys
 from dataclasses import dataclass, field
+
 import pandas as pd
-from tardis.io.logger.colored_logger import ColoredFormatter
 import panel as pn
 from IPython.display import display
+
+from tardis.io.logger.colored_logger import ColoredFormatter
 
 PYTHON_WARNINGS_LOGGER = logging.getLogger("py.warnings")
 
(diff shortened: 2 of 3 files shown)


tardis/io/logger/logger.py
```diff
@@ -35,17 +30,18 @@ def get_environment():
...
     height : int, optional
         The height of the pane in pixels.
-        
+
     Returns
     -------
     panel.pane.HTML
         A Panel HTML pane configured for logging output.
     """
+    import panel as pn
```
KasukabeDefenceForce, line 44: this can be removed no?

tardis/io/logger/logger.py
```diff
@@ -60,24 +56,43 @@ def create_output_widget(height=300):
...
+def initialize_widget_components():
+    """Initialize and return the widget components.
+
+
+    Returns
+    -------
+    tuple
+        A tuple containing (environment, log_outputs, logger_widget)
+    """
+    if 'GITHUB_ACTIONS' not in os.environ:
```
jvshields, line 68: This seems a bit odd to me. I don't think we want to let github action things creep into the tardis codebase to decide configurations. I would think we would just want a configuration for showing everything, or not showing the logger, and maybe some configuration in between, but I would rather these choices be made just with some argument rather than the function just checking whether it's a github action.

tardis/base.py
```diff
@@ -117,7 +117,7 @@ def run_tardis(
 
     simulation.run_convergence()
     simulation.run_final()
-    if logger_widget:
+    if logger_widget and tardislogger:
```
jvshields, line 120: Is there a reason both of these should be specified? Would tardislogger ever exist, but not logger widget?

tardis/io/logger/logger.py
```diff
@@ -139,19 +156,19 @@ def __init__(self, log_outputs, colors, display_widget=True, display_handle=None
         self.display_handle = display_handle
         self.logger_widget = logger_widget
         self.stream_handler = None
-        if not self.display_widget or self.display_handle is None:
+        if not self.display_widget or self.display_handle is None or self.log_outputs is None:
```
jvshields, line 159: This seems like it could lead to unintended or confusing behavior. Maybe When the class is initialized, if a user is asking to display the widget but there's no outputs, maybe a warning should appear to let them know that the logger isn't showing because there are no outputs. 

Also possible I don't completely understand what's going on here, but I could use a bit of an explanation why a lot of these conditionals are just having an extra condition added onto them.

tardis/base.py
atharva-2001: yes this will be the case when running tardis via command line.

wkerzendorf:
talk to the panel people

atharva-2001:
Closing since changes here will be fixed by https://github.com/tardis-sn/tardis/pull/3123

Issue #3073, opened 2025-05-08
TARDIS install version shows 0.0.0


andrewfullard:
**Describe the bug**

After installing TARDIS following our instructions, run `conda list` to see the packages installed. TARDIS will show as version 0.0.0. It should instead show the version number.

atharva-2001:
https://github.com/tardis-sn/tardis/pull/3122
Should fix it

PR #3086, opened 2025-05-18
Fix tardis version fallback

KasukabeDefenceForce:
### :pencil: Description
This PR aims to fix the tardis version.
Solution:
https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme
Solves: https://github.com/tardis-sn/tardis/issues/3073

image

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,8 @@ requires = ["setuptools",
             "wheel"]
 build-backend = "setuptools.build_meta"
 
+[tool.setuptools_scm]
+
 [tool.interrogate]
 color = true
 #exclude = ["docs", "build"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -29,6 +29,8 @@ requires = ["setuptools",
             "wheel"]
 build-backend = "setuptools.build_meta"
 
+[tool.setuptools_scm]
```
atharva-2001, line 32: this block is empty- does it really need to be here?

KasukabeDefenceForce: yep, an empty block is fine if we dont want to add any extra settings

jvshields (approved):

PR #3092, opened 2025-05-22
Add assertion check in v_inner_solver workflow

DeerWhale:
In the current status of TARDIS, if using csvy in the v_inner_solver workflow, the initial v_inner_boundary has to be the first shell, otherwise it will have a dimension conflict, since the plasma solver and transport solver does not change its dimension after the initialization, the v_inner_solver update the sliced dimension only in each iteration. 

Hence, added an assertion error on this. 

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/workflows/simple_tardis_workflow.py
+++ b/tardis/workflows/simple_tardis_workflow.py
@@ -49,6 +49,11 @@ class SimpleTARDISWorkflow(WorkflowLogging):
             self.simulation_state = SimulationState.from_csvy(
                 configuration, atom_data=atom_data
             )
+            assert (
+                self.simulation_state.v_inner_boundary
+                == self.simulation_state.geometry.v_inner[0]
+            ), "If using csvy in the v_inner_solver workflow, the initial v_inner_boundary must start with the first shell in the current status."
+
         else:
             self.simulation_state = SimulationState.from_config(
                 configuration,


tardis/workflows/simple_tardis_workflow.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, configuration, csvy=False):
             self.simulation_state = SimulationState.from_csvy(
                 configuration, atom_data=atom_data
             )
+            assert (
+                self.simulation_state.v_inner_boundary
+                == self.simulation_state.geometry.v_inner[0]
+            ), "If using csvy in the v_inner_solver workflow, the initial v_inner_boundary must start with the first shell in the current status."
```
andrewfullard, line 55:
```suggestion
            ), "If using csvy density input in the v_inner_solver workflow, the initial v_inner_boundary must start at the first shell."
```

tardis/workflows/simple_tardis_workflow.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, configuration, csvy=False):
             self.simulation_state = SimulationState.from_csvy(
                 configuration, atom_data=atom_data
             )
+            assert (
+                self.simulation_state.v_inner_boundary
```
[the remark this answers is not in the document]
DeerWhale: changed to np.isclose

andrewfullard (dismissed):

jvshields (approved):

tardis/workflows/simple_tardis_workflow.py
```diff
@@ -49,6 +49,13 @@ def __init__(self, configuration, csvy=False):
             self.simulation_state = SimulationState.from_csvy(
                 configuration, atom_data=atom_data
             )
+            assert np.isclose(
+                self.simulation_state.v_inner_boundary.to(u.km / u.s).value,
+                self.simulation_state.geometry.v_inner[0].to(u.km / u.s).value,
+            ), (
+                "If using csvy density input in the v_inner_solver workflow, the initial v_inner_boundary must start at the first shell."
```
jvshields: I think it's kind of weird to have this statement live in the simple tardis workflow, if it's specifically about the v_inner_solver workflow. Could this check be moved into that workflow instead? I think it's bad precedent to have workflow specific things live in the simple workflow.

DeerWhale: That's true! I'm moving this to the v_inner_solver instead

jvshields: Reopening this because you said you would do this but it is still not done. Maybe forgotten about.

wkerzendorf (approved):

jvshields:
@wkerzendorf I see you enabled automerge, but I'd really rather have my comment addressed for this PR and the warning moved to the right workflow.

tardis/workflows/simple_tardis_workflow.py
jvshields: Bumping this so I can leave a request changed review

jvshields (changes requested):

tardis/workflows/simple_tardis_workflow.py
DeerWhale: After testing, found that this issue is not unique to the v_inner workflow, with non first-shell v_inner, all workflows won't work, actually the run_tardis as well.  See related issue #3129

jvshields (approved):

jvshields (approved):

andrewfullard (approved):

PR #3100, opened 2025-05-29
Use lineid plot to make spectral line markers

hmperk18:
### :pencil: Description

**Type:** :rocket: `feature` |  :memo: `documentation`

I constructed a wrapper function that utilizes the `lineid_plot` package to add spectral line markers to SDEC Plots.  An example is included in the SDEC notebook and is shown below.

The package can be found [here](https://github.com/phn/lineid_plot) and documentation can be found [here](https://pythonhosted.org/lineid_plot/). The wrapper allows the user to supply kwargs to the `lineid_plot.plot_line_ids()` function, but also include 3 preset configurations that can be specified with `style='top'/'inside'/'along'` parameter. 

image

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,115 @@
+try:
+    import lineid_plot
+    from lineid_plot import get_line_flux
+
+except ImportError:
+    print("Please pip install lineid_plot to use the lineid_plotter function")
```
Rodot-, line 6: Add a variable that is set to True/False depending on whether lineid_plot is installed.  That way it can be checked in tests and in the function

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,115 @@
...
+        axes from generate_plot_mpl with the lines plotted.
+    """
+
+    if plotter_kwargs is None:
+        plotter_kwargs = {}
+
+    if lineid_kwargs is None:
+        lineid_kwargs = {}
+
+    ax = plotter.generate_plot_mpl(**plotter_kwargs)
```
andrewfullard: I know you have this run for SDEC only, but it would be good to see if this works for other plotting objects we have

Rodot-: Could we instead have this function just take in the ax, waves, fluxes, etc. then have another function that takes a plotter instance and generates the line_id plot?

tardis/visualization/tools/tests/test_lineid_plotter.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+from tardis.visualization.tools.sdec_plot import SDECPlotter
+from tardis.visualization.tools.lineid_plotter import lineid_plotter
+from matplotlib.testing.compare import compare_images
```
Rodot-, line 3: Add in a check for whether or not the lineid_plot package is installed and use pytest.mark.skip to skip the test if the module is not available

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,115 @@
...
+            spec,
+            wavelengths,
+            line_labels,
+            ax=ax,
+            arrow_tip=arrow_loc,
+            box_loc=box_loc,
+            **lineid_kwargs,
+        )
+
+    elif style == "along spectrum":
```
Rodot-: It might be good to switch this option to just "along", since one-word strings are usually easier to remember.

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,115 @@
...
+            spec,
+            wavelengths,
+            line_labels,
+            ax=ax,
+            arrow_tip=arrow_loc,
+            box_loc=box_loc,
+            **lineid_kwargs,
+        )
+
+    elif style == "along":
```
Rodot-, line 88: Make sure the docstring and ValueError reflect this change!

tardis/visualization/tools/lineid_plotter.py
hmperk18: do I need to use the flag in the plotter function? Or will the import error raise the issue before the user gets to use the function
I currently have the flag just in the testing code

tardis/visualization/tools/tests/test_lineid_plotter.py
```diff
@@ -0,0 +1,73 @@
...
+            pytest.skip("Skipping test to generate reference data")
+        else:
+            expected = str(
+                regression_data.absolute_regression_data_dir
+                / f"{regression_data.fname_prefix}.png"
+            )
+            actual = str(tmp_path / f"{regression_data.fname_prefix}.png")
+            compare_images(expected, actual, tol=0.001)
+    else:
+        pytest.skip("lineid_plot is not installed, skipping test")
```
jaladh-singhal: this is good but instead of skipping inline, you can skip it on module level by using `pytest.mark.skipif`

tardis/visualization/tools/tests/test_lineid_plotter.py
```diff
@@ -0,0 +1,73 @@
...
+    "wavelengths, labels, style",
+    [
+        ([3951, 6355, 8567], ["Ca II", "Si II", "Ca III"], "top"),
+        ([3951, 6355, 8567], ["Ca II", "Si II", "Ca III"], "inside"),
+        ([3951, 6355, 8567], ["Ca II", "Si II", "Ca III"], "along"),
+        pytest.param(
+            [3951, 6355, 8567],
+            ["Ca II", "Si II", "Ca III"],
+            "?",
+            marks=pytest.mark.xfail,
```
jaladh-singhal, line 42: this is great!

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,114 @@
...
+    -------
+    matplotlib.axes._subplots.AxesSubplot
+        axes from generate_plot_mpl with the lines plotted.
+    """
+
+    if plotter_kwargs is None:
+        plotter_kwargs = {}
+
+    if lineid_kwargs is None:
+        lineid_kwargs = {}
```
jaladh-singhal, line 52: why not just set `{}` instead of `None` in function definition?

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if style == "top" or style == None:
```
jaladh-singhal: you don't really have to check for `None` here since `"top"` is default. If user set `None` explicitly, let it raise ValueError

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,114 @@
...
+    import lineid_plot
+    from lineid_plot import get_line_flux
+
+except ImportError:
+    print("Please pip install lineid_plot to use the lineid_plotter function")
+    raise
+
+
+def lineid_plotter(
+    ax,
```
jaladh-singhal, line 11: I don't see this defined in docstring, perhaps you meant to name this parameter `plotter` instead?

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,114 @@
...
+    if plotter_kwargs is None:
+        plotter_kwargs = {}
+
+    if lineid_kwargs is None:
+        lineid_kwargs = {}
+
+    if style == "top" or style == None:
+        fig, ax = lineid_plot.plot_line_ids(
+            spectrum_wavelengths, spectrum_data, line_wavelengths, line_labels, ax=ax, **lineid_kwargs
+        )
```
jaladh-singhal: I'm not sure why `ax` is a parameter but also the return value? Perhaps, you meant to pass plotter_ax? Avoid duplicate names

tardis/visualization/tools/lineid_plotter.py
```diff
@@ -0,0 +1,114 @@
...
+        arrow_loc = ax.transData.inverted().transform(
+            ax.transAxes.transform((0, 0.8))
+        )[1]
+
+        fig, ax = lineid_plot.plot_line_ids(
+            spectrum_wavelengths,
+            spectrum_data,
+            line_wavelengths,
+            line_labels,
+            ax=ax,
```
jaladh-singhal, line 82: same here and all the calls below

jaladh-singhal (changes requested):
@hmperk18 this is great! it is pretty clear you know how to write testable code!

I have left some nitpicky comments to improve code quality. Let me know if anything is unclear.

docs/analyzing_tardis/visualization/how_to_sdec_plot.ipynb
```diff
@@ -363,6 +363,34 @@
     "plotter.generate_plot_mpl(show_modeled_spectrum=False)"
```
jaladh-singhal, line 363: <p>Consider changing the heading to "Marking locations of spectral lines" - we try to keep headings short because they appear in TOC in sidebar. You're explaining <code>lineid_plot</code> in description already </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/3100/discussion/#comment-2122390686'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='32'/>

jaladh-singhal, line 363: <pre class="ql-syntax" spellcheck="false">Line #7.    lineid_plotter(plotter, line_waves, line_labels, style='along',
</pre><p>can you add one more cell below to show other styles?</p><p> </p><p>From your PR description, this should be here in the tutorial text</p><blockquote>The wrapper allows the user to supply kwargs to the&nbsp;<code>lineid_plot.plot_line_ids()</code>&nbsp;function, but also include 3 preset configurations that can be specified with&nbsp;<code>style='top'/'inside'/'along'</code>&nbsp;parameter.</blockquote>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tardis-sn/tardis/pull/3100/discussion/#comment-2122390689'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='33'/>

tardis/visualization/tools/lineid_plotter.py
Rodot-: I think leaving it how it is should be fine

tardis/visualization/tools/lineid_plotter.py
Rodot-: Pass by reference issues, using None is a better way to do it so that the default object doesn't persist.

tardis/visualization/tools/lineid_plotter.py
hmperk18: I missed updating the docstring like you mentioned, so it does take in an ax object and then modifies it with the `lineid_plot` package functions, but it is still the same object throughout so I used the same `ax` variable. Is that okay, or should it be renamed throughout?

tardis/visualization/tools/lineid_plotter.py
hmperk18: I fixed this to state that it uses the axes object, not the plotter, as I want it to be usable on non-sdec plotter plots.

tardis/visualization/tools/tests/test_lineid_plotter.py
```diff
@@ -0,0 +1,77 @@
...
+        ),
+        pytest.param( # different length arrays
+            [3951, 6355],
+            ["Ca II", "Si II", "Ca III"],
+            "top",
+            marks=pytest.mark.xfail,
+        ),
```
hmperk18, line 49: I wanted to check that the package does indeed raise an error when they are different lengths. It does, but should it be in our tests?

tardis/visualization/tools/lineid_plotter.py
jaladh-singhal: Ah, I see - then it's fine to leave it named 'ax'. It was not immediately obvious to me - so perhaps leave a comment explaining you're modifying the same passed ax object.

docs/analyzing_tardis/visualization/how_to_sdec_plot.ipynb
jaladh-singhal: <p>much clear! thanks</p>

tardis/visualization/tools/tests/test_lineid_plotter.py
jaladh-singhal: never hurts to add more tests but not a dealbreaker

jaladh-singhal (approved):
Thanks for addressing the comments. Ready to merge once we add the optional dependency for tests to run instead of skipping.

hmperk18:
The tests depend on the regression data in [`tardis-regression-data`PR #52](https://github.com/tardis-sn/tardis-regression-data/pull/52). I do not have permission to add reviewers there, so please check that out :)

jaladh-singhal (approved):
Ready to be merged once https://github.com/tardis-sn/tardis/pull/3119 is merged and this is rebased to retrigger the tests (that don't get skipped)

PR #3108, opened 2025-06-01
restructure the helper functions for the sdec plot

wkerzendorf:
Update of the SDec plot.


abhinavohri:
I have restructured these functions inside of the plotter class:
https://github.com/tardis-sn/tardis/blob/master/tardis/visualization/tools/sdec_plot.py#L541

If this looks fine I can do the same structure here.

andrewfullard (approved):
Looks good, much easier to read.

Issue #3054, opened 2025-04-16
Continuum estimator attributes are incorrectly accessed when the plasma properties are updated


Rodot-:
in `Simulation.advance_state`

`estimated_continuum_properties.photo_ion_coeff` should be `estimated_continuum_properties.photo_ionization_rate_coefficient`
`estimated_continuum_properties.stim_recomb_estimator` should be `estimated_continuum_properties.stimulated_recombination_rate_factor`

**To Reproduce**

Run a model with continuum processes enabled.  This will crash with a traceback to the attribute error described in this issue.  Fixing this bug will then cause tardis to mysteriously crash, and I will explain why in another issue (hint: there's once again NaNs in the transition_probabilities)

**Screenshots**

**System**
 - OS:
    - [x] GNU/Linux
    - [x] macOS

Rodot-:
Also, the `self.atom_data` passed to the continuum solver in the simulation should be `self.plasma.atomic_data`

This is all legacy stuff of course, but it is a couple lines to fix to make to continuum processes function for the time being

jvshields:
Fixed in #3116

PR #3116, opened 2025-06-03
Macroatom restructure

jvshields:
New macroatom that computes transition probabilities and the associated metadata required by tardis much closer to the macroatom. This means it doesn't require any of the "macroatom" parts of the atomic data itself, but just the atomic lines and levels. 

Transition probabilities are a little bit different due to different normalization rourtines primarily. Here's a plot showing  differences using only Helium.  

[macro_diff.pdf](https://github.com/user-attachments/files/20802454/macro_diff.pdf)

Also it is worth noting that the new macroatom is sorted very differently from the previous one, which I wasn't find a way to reproduce only using the new macroatom. The new macroatom state contains methods to sort to the previous one, but requires you give it an old macroatom state to match to. It also contains a method to convert to the legacy format. 

One major upside of this is that the new macroatom nicely uniquely specifies each transition in the index, where the previous macroatom did not have uniquely specified transitions unless you went back and looked at both the line associated with the transition, as well as the transition type. 

As a final note, this PR only includes the non-continuum transitions, and functions as a bound-bound macroatom solver. A future PR will add extra transitions, which could be implemented as a continuum macro atom solver, but it should work to output the same macroatom state class. 

I've added new regression data for this PR in https://github.com/tardis-sn/tardis-regression-data/pull/53


tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -145,7 +145,113 @@ def solve(
...
+    def __init__(self, levels, lines):
+        self.levels = levels
+        self.lines = lines
+
+
+    def solve(self, mean_intensities_blue_wing, beta_sobolevs, stimulated_emission_factors, normalize=True):
+        """Solves the transition probabilities for the macro atom.
+        Parameters
+        ----------
+        mean_intensities_blue_wing : pd.DataFrame
```
Rodot-, line 182: Can you put in the LaTeX of the variable name used in the reference in the docstring?

tardis/opacities/macro_atom/macroatom_state.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         self.transition_probabilities = transition_probabilities
         self.transition_type = transition_type
         self.destination_level_id = destination_level_id
-        self.transition_line_id = transition_line_id
+        self.transition_line_id = transition_line_id  # THESE ARE NOT TRANSITION LINE IDS, THEY ARE INTEGER LOCATIONS OF THE LINES. IN THE MACRO ATOM DATA THEY WERE LINES_IDX
```
andrewfullard: What does "integer location" mean?

jvshields: It means where you would access the dataframe with .iloc[]. Would you call this something like index location instead? I wasn't sure what the right term was.

andrewfullard: I think index location is correct

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -141,11 +150,180 @@ def solve(
...
+                internal_up_metadata,
+            ]
+        )
+
+        if normalize:
+            # Normalize the probabilities by source.
+            probabilities_df = probabilities_df.div(
+                probabilities_df.groupby("source").transform("sum")
+            )
+            probabilities_df.replace(
```
Rodot-: Is it guaranteed that no transitions end up in this block during iteration?

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -0,0 +1,160 @@
...
+def line_transition_internal_up(
+    line_f_lus,
+    line_nus,
+    energies_lower,
+    mean_intensities_blue_wing,
+    beta_sobolevs,
+    stimulated_emission_factors,
+    transition_a_i_l_u_array,
+    line_ids,
+):
```
Rodot-, line 17: Can you add some documentation for this function?

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def line_transition_internal_down(
+    line_f_uls,
+    line_nus,
+    energies_lower,
+    beta_sobolevs,
+    transition_a_i_l_u_array,
+    line_ids,
+):
```
Rodot-, line 112: Ditto

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -0,0 +1,160 @@
...
+
+def line_transition_emission_down(
+    line_f_uls,
+    line_nus,
+    energies_upper,
+    energies_lower,
+    beta_sobolevs,
+    transition_a_i_l_u_array,
+    line_ids,
+):
```
Rodot-, line 203: again

tardis/opacities/macro_atom/macroatom_solver.py
jvshields: Not sure exactly what you mean, but the blocks that this affects are the ones with transition probabilities that sum to 0, so the groupby division nans. Though I guess I could handle that by just giving the .div a fill value which is a little cleaner. To be sure about what you're asking, I could add a check to see if there are any nans before the division, but I don't believe any of those exist.

jvshields:
> I think it would be a good idea to have a simple larger contained class for the macro atom that takes in individual macroatom states and aggregates them. For now this can just be a simple dataclass or similar that returns back the BoundBoundMacroAtom

I'm not sure what you mean by this exactly. I think downstream any assembling of the macroatom will behave the same, because it will just end up sampling from the transitions, so if you feed it something with a subset of the transitions (i.e., scattering or downbranch) the structure of the macroatom won't be any different. Is there a reason you think you want a larger macroatom state that then makes the distinction between which it is assembled? At most I feel like you'd want a enum that just keeps track of which type of macroatom it is, but completely different types seems like overkill to me.

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -141,11 +150,178 @@ def solve(
...
+            Stimulated emission factors for the lines.
+        normalize : bool, optional
+            Whether to normalize the transition probabilities to unity. Default is True.
+        Returns
+        -------
+        MacroAtomState
+            A MacroAtomState object containing the transition probabilities, transition metadata, and a mapping from line IDs to macro atom level upper indices.
+        """
+
+        f_ul = self.lines.f_ul.values.reshape(-1, 1)
```
andrewfullard: `oscillator_strength_ul`? Unless a better suggestion.

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -141,11 +150,178 @@ def solve(
...
+            .reindex(self.lines.index.droplevel("level_number_lower"))
+            .values
+        )
+        energies_lower = (
+            self.levels[["energy"]]
+            .rename(columns={"energy": "level_number_lower"})
+            .reindex(self.lines.index.droplevel("level_number_upper"))
+            .values
+        )
+        transition_a_i_l_u_array = (
```
andrewfullard: Maybe a comment that this stands for atomic ion lower upper

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -141,11 +150,178 @@ def solve(
...
+
+        macro_atom_transition_metadata = pd.concat(
+            [
+                emission_down_metadata,
+                internal_down_metadata,
+                internal_up_metadata,
+            ]
+        )
+
+        if normalize:
```
andrewfullard: Check if this needs to be optional or not

andrewfullard (changes requested):

jvshields:
@andrewfullard, fyi, updated some legacy keys in simulation.base to close https://github.com/tardis-sn/tardis/issues/3054

PR #3126, opened 2025-06-09
Composition/to xarray

wkerzendorf:
new to_xarray functionality for composition

### DIFF
--- a/tardis/model/matter/composition.py
+++ b/tardis/model/matter/composition.py
@@ -6,6 +6,8 @@ from radioactivedecay.decaydata import DEFAULTDATA as RD_DEFAULT_DATA
 
 from tardis.model.matter.decay import IsotopicMassFraction
 
+import xarray as xr
+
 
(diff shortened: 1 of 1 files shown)


tardis/model/matter/composition.py
```diff
@@ -100,6 +100,38 @@ def assemble_isotope_masses(self):
 
         return isotope_mass_df
 
+    def to_xarray(self):
+        try:
+            import xarray as xr
+        except ImportError:
+            raise ImportError(
+                "xarray is required to use the Composition.to_xarray method. "
+                "Please install it using 'pip install xarray'."
```
jvshields: Would we rather tell people to conda install it?

wkerzendorf: sure we can

wkerzendorf:
```suggestion
                "Please install it using 'conda install xarray'."
```

PR #3138, opened 2025-06-16
Make `run_tardis` use the workflow

atharva-2001:
### :pencil: Description

Change `run_tardis` function to use the Standard TARDIS Workflow instead of the simulation class.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/base.py
```diff
@@ -96,30 +92,29 @@ def run_tardis(
...
+        enable_virtual_packet_logging=virtual_packet_logging,
+        log_level=log_level,
+        specific_log_level=specific_log_level,
         show_progress_bars=show_progress_bars,
-        **kwargs,
+        show_convergence_plots=show_convergence_plots,
+        convergence_plots_kwargs=convergence_plots_kwargs,
     )
-    for cb in simulation_callbacks:
-        simulation.add_callback(*cb)
```
andrewfullard, line 119: We need to decide what to do about callbacks

tardis/base.py
```diff
@@ -96,30 +92,29 @@ def run_tardis(
 
     logger_widget, tardislogger = logging_state(log_level, tardis_config, specific_log_level, display_logging_widget)
 
-    if atom_data is not None:
```
andrewfullard, line 101: I think it would make sense to modify the workflow to allow atomic data overrides.

atharva-2001: I see, sure I will change it.

andrewfullard: Better option is to override the configuration atom data string

PR #3149, opened 2025-06-18
Refactor imports to use HDFWriterMixin from hdf_writer_mixin module

wkerzendorf:
Update all instances of HDFWriterMixin import to use the correct module path, ensuring consistency and clarity in the codebase.

Also - which the AIs missed. add a format keyword so we can save categoricals.


andrewfullard:
Looks like there are some bugs in the quantity reading.

andrewfullard (dismissed):
Approved pending tests passing

andrewfullard (approved):

PR #3152, opened 2025-06-19
add plugin to display writing symbol and not skip test

jvshields:
This is should be merged with https://github.com/tardis-sn/tardisbase/pull/17 to not break things.

This adds a plugin to change the behavior of the --generate-reference flag. Instead of calling pyest.skip() which immediately terminates the test, the test should continue with this and display a W and log the test with the "written" status.

This should specifically enable the behavior where you can do regression_data.sync_dataframe() on a test, and pass a different key, so that multiple dataframes can be saved to the same hdf file for the same test.

I believe np.save by default will also just append to the file sequentially and also read them in the same way (see https://numpy.org/doc/stable/reference/generated/numpy.save.html) so sync_ndararay might just work with no modifications.

### DIFF
--- a/tardis/conftest.py
+++ b/tardis/conftest.py
@@ -4,8 +4,6 @@ from pathlib import Path
 
 import pytest
 from astropy.version import version as astropy_version
-import tardisbase.testing
-import tardisbase.testing
 
 from tardis import run_tardis
 from tardis.io.configuration.config_reader import Configuration


tardis/conftest.py
```diff
@@ -85,6 +88,10 @@ def pytest_configure(config):
         "markers",
         "ignore_generate: mark test to not generate new reference data",
     )
+    config.pluginmanager.register(
```
wkerzendorf, line 91: normally they are a single character normally? No?

andrewfullard: I think that is defined in tardisbase

jvshields: Not sure what you mean by "they"

The plugin that gets registered has a unique symbol (the character that gets printed in the test block) and a message to display when it triggers (like the x passed, y skipped, z failed). And yeah, the plugin is made in tardisbase and the PR there.

Not sure if I answered your question.

andrewfullard:
Make sure these changes are propagated to stardis and carsus if necessary

jvshields:
> Make sure these changes are propagated to stardis and carsus if necessary

I mentioned this on the other PR, but stardis and carsus are not plugged in to tardisbase yet, so that would need to happen first. They both have their own regression data objects defined in their own util files.

PR #3158, opened 2025-06-24
mention intel lockfile and tabs for different platforms to install docs

hmperk18:
### :pencil: Description

Adds the intel cpu lockfile to the installation instructions and uses [`sphinx-tabs`](https://sphinx-tabs.readthedocs.io/en/latest/) to have tabs per platform.

Screenshot 2025-06-23 at 7 29 53 PM

The intel cpu tab includes a warning about that lack of testing for its lockfile.
Screenshot 2025-06-23 at 7 30 11 PM

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -89,6 +89,7 @@ extensions = [
     "matplotlib.sphinxext.plot_directive",
     "sphinxcontrib.jquery",
     "nbsphinx",
+    "sphinx_tabs.tabs",
     "numpydoc",
     "recommonmark",
 ]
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -89,6 +89,7 @@
     "matplotlib.sphinxext.plot_directive",
     "sphinxcontrib.jquery",
     "nbsphinx",
+    "sphinx_tabs.tabs",
```
atharva-2001, line 92: I guess this means resolving the lockfiles? This is a new dependency right?

hmperk18: Yes, its a new dependency that allows tabs in the docs, so I've left it as a draft until we want to deal with adding that. Tardisbase [PR #19](https://github.com/tardis-sn/tardisbase/pull/19) is related.

Issue #3079, opened 2025-05-12
[Documentation] Missing Hyperlinks in tutorials, how to guides, workflow page


DeerWhale:
The documentation pages leading to tutorial, how to guides, workflows in the main TARDIS doc page (eg., https://tardis-sn.github.io/tardis/tutorials.html#) is missing the hyperlinks to the notebooks. 

<img width="347" alt="Image" src="https://github.com/user-attachments/assets/b3d12e9b-6d5d-4623-a2d4-ab693387da90" />

PR #3184, opened 2025-07-10
Fix missing hyperlink for the tutorial notebooks in doc

DeerWhale:
Fixing Issue #3079 that the notebook links are missing in the documentation page. 

Meantime also added simple title description for two of the new workflow notebooks. 

For the v_inner_workflow, the formal integral is commented for now due to dimensionality incompatible issue (probably because the v_inner is not at first shell.).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -374,7 +374,7 @@ def generate_tutorials_page(app):
 
     for notebook in io_path.rglob("*.ipynb"):
         if "tutorial_" in notebook.name and "checkpoint" not in notebook.name:
-            notebooks += f"\n* :doc:`{notebook.stem}`"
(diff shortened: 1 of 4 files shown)


jvshields:
I think all of the notebook changes are not supposed to be part of this PR?

andrewfullard:
Unfortunately the docs will still fail due to the formal integral issue.

DeerWhale:
> Unfortunately the docs will still fail due to the formal integral issue.

It seems only break the one in v_inner workflow so I commented out for now, the rest was able to run and I was able to built a working doc locally.

jvshields:
Build docs workflow did pass and it does look like the hyperlinks work now.

andrewfullard (approved):

Issue #3191, opened 2025-07-15
Make SDEC plot work with workflows


wkerzendorf:
@jvshields has some comments

jvshields:
As of the connector we had to do something like this

`class DummySimForPlots:
    class DummyTransport:
        def __init__(self, transport_state):
            self.transport_state = transport_state`

`    def __init__(self, plasma, spectrum_solver, simulation_state, transport_state):
        self.plasma = plasma
        self.spectrum_solver = spectrum_solver
        self.simulation_state = simulation_state
        self.transport = self.DummyTransport(transport_state)`

`dummy_sim = DummySimForPlots(
    workflow.plasma_solver,
    workflow.spectrum_solver,
    workflow.simulation_state,
    workflow.transport_state,
)`

`plotter = SDECPlotter.from_simulation(dummy_sim)`


It would be great to just have an SDECPlotter.from_workflow() that knows where to find the data it needs already. 
Similarly with the LIVplot

PR #3198, opened 2025-07-16
add from workflow method to sdec and liv plot

abhinavohri:
### :pencil: Description

This PR adds a from_workflow method to initialize the the SDEC and LIV plotter classes directly from a workflow.
Fixes: https://github.com/tardis-sn/tardis/issues/3191

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/visualization/tools/liv_plot.py
+++ b/tardis/visualization/tools/liv_plot.py
@@ -102,7 +102,7 @@ class LIVPlotter:
 
         Parameters
         ----------
-        workflow : SimpleTARDISWorkflow
+        workflow : A TARDIS workflow object
 
         Returns
         -------
--- a/tardis/visualization/tools/sdec_plot.py
+++ b/tardis/visualization/tools/sdec_plot.py
@@ -148,7 +148,7 @@ class SDECPlotter:
 
         Parameters
         ----------
-        workflow : SimpleTARDISWorkflow
+        workflow : A TARDIS workflow object
 
         Returns
         -------


tardis/visualization/tools/liv_plot.py
```diff
@@ -95,6 +95,33 @@ def from_hdf(cls, hdf_fpath):
 
         return plotter
 
+    @classmethod
+    def from_workflow(cls, workflow):
+        """
+        Create an instance of the plotter from a StandardTARDISWorkflow.
```
jvshields: Is this specific to the StandardTARDISWorkflow? As far as I can tell it should work with the Simple workflow as well.

tardis/visualization/tools/sdec_plot.py
```diff
@@ -141,6 +141,37 @@ def from_hdf(cls, hdf_fpath):
 
         return plotter
 
+    @classmethod
+    def from_workflow(cls, workflow):
```
jvshields, line 145: This looks exactly like what we had in mind, but again I would check to see if it's specific to the Standard workflow, and if not change the docstring.

tardis/visualization/plot_util.py
```diff
@@ -136,7 +136,16 @@ def extract_and_process_packet_data(simulation, packets_mode):
         Dictionary containing raw packet data, the full DataFrame `packets_df`,
         and a filtered `packets_df_line_interaction` with line interaction info.
     """
-    transport_state = simulation.transport.transport_state
+    if hasattr(simulation, "transport_state"): # for workflows
```
jvshields, line 139: This is a bit hacky. Maybe fine in the short term but if there are other places where we have to do a similar thing, I wonder if this is telling us we should have a flag somewhere that just tells us whether it's a workflow or not.

tardis/visualization/plot_util.py
```diff
@@ -136,7 +136,16 @@ def extract_and_process_packet_data(simulation, packets_mode):
...
+        transport_state = simulation.transport_state
+    else:
+        transport_state = simulation.transport.transport_state
+
+    if hasattr(simulation, "plasma"):
```
andrewfullard: Could probably merge this if-else with the above one

tardis/visualization/tools/sdec_plot.py
abhinavohri: it will work with simple tardis workflow as well.

andrewfullard (approved):

tardis/visualization/tools/sdec_plot.py
```diff
@@ -141,6 +141,37 @@ def from_hdf(cls, hdf_fpath):
...
+        """
+        Create an instance of SDECPlotter from a TARDIS workflow.
+
+        Parameters
+        ----------
+        workflow : SimpleTARDISWorkflow
```
jvshields: This docstring is still wrong here then right? I'm not sure how we call these things generically but something like TARDISWorkflow I guess.

tardis/visualization/tools/liv_plot.py
```diff
@@ -95,6 +95,33 @@ def from_hdf(cls, hdf_fpath):
...
         return plotter
 
+    @classmethod
+    def from_workflow(cls, workflow):
+        """
+        Create an instance of the plotter from a TARDIS workflow.
+
+        Parameters
+        ----------
+        workflow : SimpleTARDISWorkflow
```
jvshields: Also said this below. Probably should be more generic like TARDISWorkflow.

andrewfullard (approved):

PR #3204, opened 2025-07-23
integrated tau calculation minor type check for tau sobolev

DeerWhale:
Small change to avoid error when opacity_state.tau_sobolev is np.ndararry, it has no attribute of values.


DeerWhale:
this is the right way to go, opening a new PR to add the non-numba opacity state as a attribute to the simulation object instead.

The issue was the sim.transport.transport_state.opacity_state is acutally the opacity_state_numba, which is not the one we want to use.

PR #3205, opened 2025-07-23
Add the non-numba opacity_state as a attribute of the simulation object

DeerWhale:
When calculating the integrated tau using the `tardis.workflows.util.get_tau_integ`, the opacity_state should be the non-numba version, in which the opacity_state.tau_sobolev is a dataframe with atomic number, ionization number and level number as indexes. 

However, currently the non-numba opacity_state is not a attribute of the simulation object, and the only opacity state we can access is the `simulation.transport.transport_state.opacity_state`, which is a numba version. 

In this PR, the quick fix is to add the opacity_state as a attribute of the simulation class. The `simulation.transport.transport_state.opacity_state` should be renamed to be more clear about this, see issue #3206

### DIFF
--- a/tardis/simulation/base.py
+++ b/tardis/simulation/base.py
@@ -422,7 +422,7 @@ class Simulation(PlasmaStateStorerMixin, HDFWriterMixin):
             self.plasma.beta_sobolev = None
             macro_atom_state = None
 
-        opacity_state = self.opacity.legacy_solve(self.plasma)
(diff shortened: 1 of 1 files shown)


wkerzendorf:
thanks - please make sure that there is an issue - just very briefly saying that there are various attributes - opacity, 2x opacity_state that are flying around not clearly marked

PR #3219, opened 2025-08-04
Apply diatixis across the docs

hmperk18:
### :pencil: Description

Diataxis rework


andrewfullard:
Analysing monte carlo packets is in index twice

andrewfullard:
Rewrite the front page docs org bullet points

jvshields:
I think there needs to be some more grouping. Some of it was removed and I think this structure is better, but I would try to include that previous information still. Like under how-to-guides I think we should restore subsections for say, "How to run tardis" and "Analyzing tardis" 

New structure is quite good and I think we should use it. I would just argue that without some subsections it's a bit hard for somebody unfamiliar with the information to find what they're looking for without reading every single thing within a section, and it also won't be clear to somebody reading whether they should be looking under tutorials or how-to-guides so they'll probably have to look at even more things.

docs/analyzing_tardis/index.rst
```diff
@@ -1,12 +1,28 @@
 *******************************
-Analyzing TARDIS
+Analysis How-To Guides
 *******************************
 
-This section provides an overview of how to analyze the outputs generated by TARDIS.
+**These are problem-oriented guides for analyzing TARDIS simulation outputs.**
```
Knights-Templars: Problem-oriented sounds "students/learners" will work on practical problems and we do not provide any SN problem in this documentation. 
Maybe rephrase as-

These guides provide a deeper dive into TARDIS simulated spectra.

docs/analyzing_tardis/index.rst
```diff
@@ -1,12 +1,28 @@
...
 .. toctree::
     :maxdepth: 2
-    :hidden:
 
-    analyzing_tardis/spectrum/spectrum-visualization-notebook
+    spectrum/spectrum-visualization-notebook
+    visualization/index
+
+.. note::
+   **New to TARDIS analysis?** Start with the spectrum analysis tutorial in our :doc:`../tutorials`.
```
Knights-Templars: This is taking me to configuring logging output and custom callbacks. Maybe that was the intention, but it says spectrum analysis tutorial.

docs/resources/zreferences.rst
```diff
@@ -17,9 +17,36 @@ Glossary
     Chianti
         CHIANTI consists of a critically evaluated set of up-to-date atomic data, together with user-friendly programs written in Interactive Data Language (IDL) and Python to calculate the spectra from astrophysical plasmas.
     
+    Estimator
+        A statistical tool or algorithm used in Monte Carlo simulations to calculate physical quantities (like radiation field properties) from the random sampling of particle interactions.
```
jvshields: This doesn't seem quite right to me. I think it is better as

...simulations to calculate physical quantities. In the case of TARDIS this is usually the estimation of radiation field properties based on the behavior and distribution of photon packets.

jvshields:
```suggestion
        A statistical tool or algorithm used in Monte Carlo simulations to calculate physical quantities. In the case of TARDIS this is usually the estimation of radiation field properties based on the behavior and distribution of photon packets.
```

docs/resources/zreferences.rst
```diff
@@ -17,9 +17,36 @@ Glossary
...
+        A comprehensive database of atomic line data compiled by Robert Kurucz, widely used in stellar and supernova atmosphere modeling for opacity calculations.
+    
     Meta-Stable
         Metastability is the condition of a system where the system has stability, but is not as stable as in the system's state of least energy.
     
+    Monte Carlo
+        A computational method that uses random sampling to solve complex physical problems, particularly useful for simulating particle transport and radiative transfer.
+    
+    Nebular
+        Relating to the nebular phase of supernovae, typically occurring weeks to months after explosion when the ejecta becomes optically thin and emission lines dominate the spectrum.
```
jvshields:
```suggestion
        Relating to the nebular phase of supernovae, typically occurring weeks to months after explosion when the ejecta becomes optically thin and emission lines dominate the spectrum. 
        In TARDIS, this may also refer to adjustments to LTE assumptions to handle lower density environments, which are often applicable to supernova ejecta. 
```

docs/resources/zreferences.rst
```diff
@@ -17,9 +17,36 @@ Glossary
...
+        A simplified opacity model where the absorption coefficient is assumed to be independent of photon wavelength or frequency, making calculations more tractable.
+    
+    HDF (Hierarchical Data Format)
+        A file format designed for storing and organizing large amounts of scientific data in a hierarchical structure, commonly used with the .h5 extension.
+    
+    Kurucz
+        A comprehensive database of atomic line data compiled by Robert Kurucz, widely used in stellar and supernova atmosphere modeling for opacity calculations.
+    
     Meta-Stable
         Metastability is the condition of a system where the system has stability, but is not as stable as in the system's state of least energy.
```
jvshields:
```suggestion
        Metastability is the condition of a system where the system has stability, but is not as stable as in the system's state of least energy. 
        In atomic physics, this is term is usually used to describe excitation states that have long spontaneous emission timescales, which corresponds to low oscillator strengths of transitions away from the metastable state. 
```

jvshields (dismissed):

docs/analyzing_tardis/index.rst
```diff
@@ -1,12 +1,28 @@
 *******************************
-Analyzing TARDIS
+Analysis How-To Guides
```
DeerWhale: Perhaps "Analyze Simulation Outputs"  is more straightforward here?

jvshields (approved):

docs/index.rst
```diff
@@ -36,6 +36,19 @@ Please contact the TARDIS team via the `GitHub page
 <https://github.com/tardis-sn/tardis>`_ if you have questions or need
 assistance.
 
+------------------------------
+Documentation Structure Guide
```
DeerWhale, line 40: Thanks Andrew! One small thing, can this be linked to the left hand side bar before the quick start too?

DeerWhale (approved):

PR #3070, opened 2025-05-06
Ionization calculations

andrewfullard:
### :pencil: Description

**Type:** | :rocket: `feature`

This PR adds photoionization and recombination rates calculations and population solvers to replace the current NLTE implementation in the near future. The classes are designed to work as standalone modules where possible.

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #3220, opened 2025-08-04
Continuum transition probabilities

Rodot-:
### :pencil: Added computation for the continuum transition probabilities for the macroatom to the macroatom transition probabilities

**Type:** ::roller_coaster: `infrastructure`

This creates more generic functions for computing the recombination, photoionization, and collisional transition probabilities from the macroatom that don't require the full plasma to compute.

depends on #3070 (sort of, it is independent but is really only useful with that PR)

### :pushpin: Resources

None right now

- [ ] Testing pipeline
- [ ] Other method (describe)
- [x] My changes haven't been tested (And is marked draft)

### DIFF
--- a/tardis/opacities/macro_atom/macroatom_transitions.py
+++ b/tardis/opacities/macro_atom/macroatom_transitions.py
@@ -307,7 +307,7 @@ def set_index(p, photo_ion_idx, transition_type=0, reverse=True):
         p = p.set_index(index, drop=True)
         return p
 
-def recombination(alpha_sp, nu_i, energy_i, photoionization_index):
+def continuum_transition_recombination(alpha_sp, nu_i, energy_i, photoionization_index):
     '''Unnormalized probabilities of radiative recombination
     alpha_sp : pandas.Series, dtype float
         Rate coefficient for spontaneous recombination from `k` to level `i`
@@ -352,7 +352,7 @@ def recombination(alpha_sp, nu_i, energy_i, photoionization_index):
     return p_recombination, recombination_metadata
 
 
-def photoionization(gamma_corr, energy_i, photo_ion_idx):
+def continuum_transition_photoionization(gamma_corr, energy_i, photo_ion_idx):
     '''
     Photoionization probability unnormalized
         gamma_corr : pandas.Series, dtype float
@@ -386,7 +386,7 @@ def photoionization(gamma_corr, energy_i, photo_ion_idx):
 
 
 
-def collisional(
+def continuum_transition_collisional(
     self,
     coll_exc_coeff,
     coll_deexc_coeff,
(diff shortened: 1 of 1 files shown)


tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -285,3 +288,160 @@ def line_transition_emission_down(
...
+
+
+def continuum_transition_photoionization(gamma_corr, energy_i, photo_ion_idx):
+    '''
+    Photoionization probability unnormalized
+        gamma_corr : pandas.Series, dtype float
+            Corrected photoionization rate coefficeint from level `i` to `k`
+    This method may need an additional transition type
+     '''
+    p_photoionization = gamma_corr.multiply(energy_i, axis=0)
```
Rodot-, line 362: Corrected_photoionization_rate

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -285,3 +288,160 @@ def line_transition_emission_down(
...
+        ]
+    )
+
+    return p_photoionization, photoionization_metadata
+
+
+
+
+def continuum_transition_collisional(
+    self,
```
jvshields, line 390: This needs to be fixed right? Maybe this used to belong to a class but now it's a function so there shouldn't be a self.

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -285,3 +288,160 @@ def line_transition_emission_down(
         ]
     )
     return p_emission_down, emission_down_metadata
+
+
+
+def set_index(p, photo_ion_idx, transition_type=0, reverse=True):
```
jvshields, line 294: p -> transition_probabilities_dataframe 

Probably verify that it's a dataframe always before making that change though.

tardis/opacities/macro_atom/macroatom_transitions.py
jvshields: Also I think maybe this isn't the right way to do this but maybe you just haven't gotten to it yet. If they're prepared like the line above in 384 we can probably just feed them into a new continuum macroatom class.

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -285,3 +288,160 @@ def line_transition_emission_down(
...
+        idx = photo_ion_idx.loc[p.index]
+        transition_type = transition_type * np.ones_like(
```
jvshields, line 296: Also on reading this and actually thinking about it, I think everything to do with transition_type could be accomplished with just a 
```p["transition_type"] = transition_type```
and then doing a reset index onto the ones you want.

tardis/opacities/macro_atom/macroatom_transitions.py
jvshields: Also, I'm realizing that this basically just maps things to the structure of photo_ion_idx so that one has to come first. I don't know if that's what we want to do but if so it should probably be specified either in the name or at least the docstring.

andrewfullard:
Josh taking this over on his own branch.

PR #3223, opened 2025-08-06
Standardize sorting with a global

jvshields:
This makes everywhere in tardis use the same sorting algorithm set as a global. Everywhere was using quicksort by default which we don't really want, but this enforces that everywhere so we can make another pr that changes everywhere to stable so we can see those differences.

The regression data in the PR below makes the v_inner_workflow tests pass for this PR (at least on mac) by generating regression data from this PR. 
https://github.com/tardis-sn/tardis-regression-data/pull/72


jvshields:
I think we expect this to change the regression data, but probably should be regenerated for this if we're happy with it for more consistency between platforms.

tardis/workflows/util.py
```diff
@@ -25,7 +26,7 @@ def get_tau_integ(plasma, opacity_state, simulation_state, bin_size=10):
     """
     index = plasma.atomic_data.lines.nu.index
     freqs = plasma.atomic_data.lines.nu.values * u.Hz
-    order = np.argsort(freqs, kind="stable")
+    order = np.argsort(freqs, kind=SORTING_ALGORITHM)
```
jvshields, line 29: This is the one change that breaks the test, because it's the one place stable was applied.

DeerWhale (approved):

PR #3240, opened 2025-08-18
Use new macroatom

jvshields:
This hooks up the new macroatom to the simple workflow. It should be easy enough to do this on run_tardis too.

The problem is that, in order to get the tests to pass, we need to sort it to match the previous macro atom, which was not unique. I'm having this PR up to show that tests pass, but before we merge this I want to strip out the sort_to_legacy() method from the macro_atom_state, since it necessitates you also compute the old macro atom, so this really computes the macroatom twice. 

The point of this is mostly to show that all of the workflow tests pass if you DO do that sorting, so that we can then actually regenerate the regression data with the new macroatom that has a unique sorting method built in and will actually be fully reproducible. 

I've also said this implicitly here, but the tests do not pass if you don't do the sorting to match the old version. I think that this is basically because using a differently sorted macroatom is akin to using a different seed, so we get montecarlo noise level differences when switching to the new macroatom.


andrewfullard (dismissed):

jvshields:
This has been updated to plug into the entire tardis test structure with run_tardis. 

Shockingly, it seems like the only place the new transition probabilities actually matter for our tests are in the one failing test_montecarlo_main_loop_vpacket_log test. Not quite sure why here and nowhere else, though. The real packet test passes fine with similar tolerances.

PR #3250, opened 2025-08-20
Test the Cuda Formal Integral

hmperk18:
### :pencil: Description

**Type:** :beetle: `bugfix` 

Fix how the `FormalIntegralSolver` handles choosing an implementation and the variable names in the cuda `test_full_formal_integral` such that the cuda implementation is used to compute `L_cuda` and the numba implementation is used to compute `L_numba`. 

Additionally, update the cuda implementation to initialize and use the `z` and `shell_id` arrays correctly


hmperk18:
Actually using the `cuda` version causes the tests to fail, currently. :(

hmperk18:
however, issue #3251 may be related

tardis/spectrum/formal_integral/formal_integral_cuda.py
```diff
@@ -496,11 +496,10 @@ def formal_integral(
             self.geometry.r_outer[size_shell - 1], N
         )  # array(float64, 1d, C)
         z = np.zeros(
-            (inu_size, 2 * size_shell), dtype=np.float64
-        )  # array(float64, 1d, C)
+        )  # array(float64, 2d, C)
```
Rodot-, line 500: Is this meant to be empty?

Rodot- (approved):
LGTM

PR #3253, opened 2025-08-21
Speedup and implement new macroatom

jvshields:
This PR fully migrates to the new macroatom, which will need a regeneration of the regression data again because the macroatom data will be differently sorted from the stuff previously output by CARSUS. 

With this PR, we should be able to purge the macroatom data from carsus and tardis, because it uses the new system that calculates transition probabilities when they're actually needed and the full macroatom is generated. 

There's also general fixes including using the MacroAtomTransitionType enum.

And finally this also makes the macro atom metadata not recalculate on subsequent iterations. It calculates the metadata and sorts everything on the first iteration, and then just recalculates transition probabilities for each iteration after that. 

Corresponding data is in https://github.com/tardis-sn/tardis-regression-data/pull/74

### DIFF
--- a/tardis/tests/test_tardis_full_formal_integral.py
+++ b/tardis/tests/test_tardis_full_formal_integral.py
@@ -79,9 +79,7 @@ class TestTransportSimpleFormalIntegral:
 
     def test_j_blue_estimators(self, simulation, request):
         regression_data = RegressionData(request)
-        j_blue_estimator = (
-            simulation.transport.transport_state.radfield_mc_estimators.j_blue_estimator
-        )
+        j_blue_estimator = simulation.transport.transport_state.radfield_mc_estimators.j_blue_estimator
         expected = regression_data.sync_ndarray(j_blue_estimator)
         npt.assert_allclose(j_blue_estimator, expected)
 
@@ -90,11 +88,15 @@ class TestTransportSimpleFormalIntegral:
         luminosity = simulation.spectrum_solver.spectrum_real_packets.luminosity
         expected = regression_data.sync_ndarray(luminosity.cgs.value)
         expected = u.Quantity(expected, "erg /s")
-        assert_quantity_allclose(luminosity, expected, rtol=1e-11, atol=0*u.erg/u.s)
+        assert_quantity_allclose(
+            luminosity, expected, rtol=1e-11, atol=0 * u.erg / u.s
+        )
 
     def test_spectrum_integrated(self, simulation, request):
         regression_data = RegressionData(request)
         luminosity = simulation.spectrum_solver.spectrum_integrated.luminosity
         expected = regression_data.sync_ndarray(luminosity.cgs.value)
         expected = u.Quantity(expected, "erg /s")
-        assert_quantity_allclose(luminosity, expected, rtol=1e-11, atol=0*u.erg/u.s)
+        assert_quantity_allclose(
+            luminosity, expected, rtol=1.5e-11, atol=0 * u.erg / u.s
+        )  # Increased atol to 1.5e-11 because of mac/linux differences


tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -229,73 +258,91 @@ def solve(
...
 
+    def _solve_first_macroatom_iteration(
+        self,
+        mean_intensities_blue_wing: pd.DataFrame,
+        beta_sobolevs: pd.DataFrame,
+        stimulated_emission_factors: np.ndarray,
+        lines_level_upper: pd.MultiIndex,
+    ) -> tuple[pd.DataFrame, pd.DataFrame, pd.Series, pd.Series]:
+        """Handle the first iteration of the solve method.
+        Fully computes all metadata for the macroatom and adds it to the class with the computed_metadata attribute.
```
andrewfullard: Could use docstrings for the arguments

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -340,21 +387,148 @@ def solve(
...
+        )
+
+    def _solve_next_macroatom_iteration(
+        self,
+        mean_intensities_blue_wing: pd.DataFrame,
+        beta_sobolevs: pd.DataFrame,
+        stimulated_emission_factors: np.ndarray,
+    ) -> tuple[pd.DataFrame, pd.DataFrame, pd.Series, pd.Series]:
+        """Handle subsequent iterations of the solve method.
+        Uses precomputed metadata and only recalculates the probabilities.
```
andrewfullard: Could use argument docstring

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -244,44 +291,66 @@ def line_transition_emission_down(
 
     The function uses the constant P_EMISSION_DOWN to mark the transition type.
```
andrewfullard: No longer true?

tardis/opacities/opacity_state.py
andrewfullard, line 1: Appear to just be whitespace changes?

tardis/visualization/plot_util.py
andrewfullard, line 1: Just whitespace changes?

tardis/opacities/tests/test_opacity_state_numba.py
```diff
@@ -63,32 +65,3 @@ def test_opacity_state_to_numba(
         npt.assert_allclose(actual.transition_type, empty)
         npt.assert_allclose(actual.destination_level_id, empty)
         npt.assert_allclose(actual.transition_line_id, empty)
-    else:
```
andrewfullard, line 66: Are these tests no longer relevant?

jvshields: I would argue that they are not, since the next step is to drop the carsus and plasma macro atom data. 

They used to be the same because the meta attributes were copied directly over from the plasma metadata.

andrewfullard:
Would be nice if the regression comparison would work, but I guess the failing tests need to pass and require updating the regression data anyway? Still, the results from Riddhi's tool may be useful.

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -229,73 +258,127 @@ def solve(
             A MacroAtomState object containing the transition probabilities, transition metadata,
             and a mapping from line IDs to macro atom level upper indices.
         """
-        oscillator_strength_ul = self.lines.f_ul.values.reshape(-1, 1)
-        oscillator_strength_lu = self.lines.f_lu.values.reshape(-1, 1)
-        nus = self.lines.nu.values.reshape(-1, 1)
-        line_ids = self.lines.line_id.values
+        is_first_iteration = not hasattr(self, "computed_metadata")
+        lines_level_upper = self.lines.index.droplevel("level_number_lower")
```
DeerWhale: does this lines_level_upper change during iterations? If not then why not also put in the pre-compute function?

jvshields:
> Would be nice if the regression comparison would work, but I guess the failing tests need to pass and require updating the regression data anyway? Still, the results from Riddhi's tool may be useful.

Yup. The notebook seems like it's failing though. I'm getting an error in packet_source/tests/test_weighted_integration.py because there's a test in there that appears to not use the sync stuff. It's hardcoded and apparently that test failing causes the notebook not to execute so that's a bit annoying.

tardis/transport/montecarlo/packet_source/tests/test_weighted_integration.py
```diff
@@ -2,10 +2,12 @@
 
 import numpy.testing as npt
 import pandas as pd
+import pytest
 
 from tardis.simulation import Simulation
 
 
+@pytest.mark.xfail(reason="Relies on later test")
```
jvshields, line 10: We'll want to get rid of this before merging I think, because with the new regression data this test should actually pass.

andrewfullard (dismissed):

andrewfullard (approved):

Issue #3151, opened 2025-06-19
make sure that TARDIS, STARDIS, Carsus have the correct tardisbase lockfiles


wkerzendorf:
as discussed with Atharva

wkerzendorf:
make sure that the github actions use the per repository lock files - not the lock files from tardis base.

PR #3269, opened 2025-08-27
Use lockfiles from TARDIS instead of from TARDISBase

atharva-2001:
### :pencil: Description

https://github.com/tardis-sn/tardis/issues/3151#issuecomment-3224558294
- [X] Requires https://github.com/tardis-sn/tardis-actions/pull/7 to be merged before. 

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -102,7 +102,7 @@ jobs:
           atom-data-sparse: true
 
       - name: Setup environment
-        uses: tardis-sn/tardis-actions/setup-env@lockfile-path
+        uses: tardis-sn/tardis-actions/setup-env@main
         with:
           os-label: linux-64
           lockfile-path: ./conda-linux-64.lock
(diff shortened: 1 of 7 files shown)


.github/workflows/build-docs.yml
```diff
@@ -105,6 +105,7 @@ jobs:
         uses: tardis-sn/tardis-actions/setup-env@main
         with:
           os-label: linux-64
+          lock-file-url-prefix: "https://raw.githubusercontent.com/tardis-sn/tardis/master"
```
wkerzendorf: is there a way to use the local lock file? I don't think we should download. It should pass with the one that is in the repo not the one on master

atharva-2001: It doesn't exist yet, but I can add it- https://github.com/tardis-sn/tardis-actions/pull/7

PR #3277, opened 2025-09-01
Viz modules `from_workflow` tests

atharva-2001:
### :pencil: Description

Tests the `from_workflow` method in the SDEC plot and the liv plot using the existing regression data.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/visualization/tools/tests/test_liv_plot.py
+++ b/tardis/visualization/tools/tests/test_liv_plot.py
@@ -90,11 +90,11 @@ class TestLIVPlotter:
                 for item in sublist
             ]
             data = regression_data.sync_ndarray(plot_object)
-            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-12)
+            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-15)
         else:
             plot_object = getattr(plotter, attribute)
             data = regression_data.sync_ndarray(plot_object)
-            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-12)
+            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-15)
 
     @pytest.fixture(scope="class", params=combinations)
     def plotter_prepare_plot_data(self, request, plotter):
@@ -158,12 +158,12 @@ class TestLIVPlotter:
             if all(isinstance(item, u.Quantity) for item in plot_object):
                 plot_object = [item.value for item in plot_object]
             data = regression_data.sync_ndarray(plot_object)
-            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-12)
+            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-15)
         else:
             plot_object = getattr(plotter_prepare_plot_data, attribute)
             plot_object = plot_object.value
             data = regression_data.sync_ndarray(plot_object)
-            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-12)
+            np.testing.assert_allclose(plot_object, data, atol=0, rtol=1e-15)
 
     @pytest.fixture(scope="function", params=combinations)
     def plotter_generate_plot_mpl(self, request, plotter):
@@ -268,7 +268,7 @@ class TestLIVPlotter:
                 np.testing.assert_allclose(
                     expected_values,
                     actual_values,
-                    atol=0, rtol=1e-12
+                    atol=0, rtol=1e-15
                 )
             else:
                 assert np.array_equal(expected_values, actual_values)
@@ -284,12 +284,12 @@ class TestLIVPlotter:
                 np.testing.assert_allclose(
                     data.get_xydata(),
                     expected.get("plot_data_hdf/" + "data" + str(index1)),
-                    atol=0, rtol=1e-12
+                    atol=0, rtol=1e-15
                 )
                 np.testing.assert_allclose(
                     data.get_path().vertices,
                     expected.get("plot_data_hdf/" + "linepath" + str(index1)),
-                    atol=0, rtol=1e-12
+                    atol=0, rtol=1e-15
                 )
             if isinstance(data, PolyCollection):
                 for index2, path in enumerate(data.get_paths()):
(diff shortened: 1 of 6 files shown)


tardis/visualization/tools/tests/test_sdec_plot.py
```diff
@@ -399,3 +401,210 @@ def test_make_colorbar_labels(self, plotter):
...
+    @pytest.fixture(scope="class")
+    def sdec_regression_data(self, tardis_regression_path):
+        return tardis_regression_path / "tardis/visualization/tools/tests/test_sdec_plot/test_sdec_plotter"
+
+    def test_from_workflow_vs_from_simulation_data_consistency(
+        self, plotter, plotter_from_workflow
+    ):
+        # Test key attributes are equal
+        np.testing.assert_allclose(
+            plotter.t_inner.value, plotter_from_workflow.t_inner.value, rtol=1e-10
```
andrewfullard: atol=0, rtol small as possible

tardis/visualization/tools/tests/test_liv_plot.py
```diff
@@ -455,13 +455,148 @@ def test_generate_plot_ply(
...
+        
+        # Save actual image
+        actual_image_path = tmp_path / f"test_mpl_image_workflow_{param_idx}.png"
+        fig.figure.savefig(actual_image_path)
+        
+        # Path to expected image
+        expected_image_path = liv_regression_data / f"test_mpl_image__plotter_generate_plot_mpl{param_idx}__.png"
+        
+        # Compare images
+        compare_images(str(expected_image_path), str(actual_image_path), tol=0.001)
```
wkerzendorf: write in scientific notation

tardis/visualization/tools/tests/test_sdec_plot.py
```diff
@@ -399,3 +403,195 @@ def test_make_colorbar_labels(self, plotter):
...
+        
+        # Save actual image
+        actual_image_path = tmp_path / f"test_mpl_image_workflow_{param_idx}.png"
+        fig.figure.savefig(actual_image_path)
+        
+        # Path to expected image
+        expected_image_path = sdec_regression_data / f"test_mpl_image__plotter_generate_plot_mpl{param_idx}__.png"
+        
+        # Compare images
+        compare_images(str(expected_image_path), str(actual_image_path), tol=0.001)
```
andrewfullard: `1e-3` for all image comparisons

tardis/visualization/tools/tests/test_liv_plot.py
```diff
@@ -7,9 +8,14 @@
 from matplotlib.lines import Line2D
 from matplotlib.testing.compare import compare_images
 
+import pandas as pd
+
 from tardis.visualization.tools.liv_plot import LIVPlotter
+from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
 from tardisbase.testing.regression_data.regression_data import PlotDataHDF
 
+RTOL_VALUE=1e-12
```
wkerzendorf:
```suggestion
RELATIVE_TOLERANCE_LIV=1e-12
```

tardis/visualization/tools/tests/test_sdec_plot.py
```diff
@@ -12,6 +13,9 @@
 
 from tardisbase.testing.regression_data.regression_data import PlotDataHDF
 from tardis.visualization.tools.sdec_plot import SDECPlotter
+from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
+
+RTOL_VALUE=1e-12
```
andrewfullard:
```suggestion
RELATIVE_TOLERANCE_SDEC=1e-12
```

atharva-2001: Done in https://github.com/tardis-sn/tardis/pull/3277/commits/5628ff202e1a34b6e10713e42609d62a133f75ba

tardis/visualization/tools/tests/test_liv_plot.py
atharva-2001: Done in https://github.com/tardis-sn/tardis/pull/3277/commits/5628ff202e1a34b6e10713e42609d62a133f75ba

tardis/visualization/tools/tests/test_liv_plot.py
atharva-2001: Done in https://github.com/tardis-sn/tardis/pull/3277/commits/5628ff202e1a34b6e10713e42609d62a133f75ba

tardis/visualization/tools/tests/test_sdec_plot.py
atharva-2001: Done in https://github.com/tardis-sn/tardis/pull/3277/commits/5628ff202e1a34b6e10713e42609d62a133f75ba

tardis/visualization/tools/tests/test_sdec_plot.py
```diff
@@ -399,3 +405,195 @@ def test_make_colorbar_labels(self, plotter):
...
+            packets_mode, packet_wvl_range, distance, nelements
+        )
+        plotter_from_workflow._param_idx = param_idx
+        return plotter_from_workflow
+
+    def test_calculate_plotting_data_workflow_vs_regression(
+        self, plotter_calculate_plotting_data_from_workflow, sdec_regression_data
+    ):
+        param_idx = plotter_calculate_plotting_data_from_workflow._param_idx
+        regression_file = sdec_regression_data / f"test_calculate_plotting_data__plotter_calculate_plotting_data{param_idx}__.h5"
```
jvshields, line 484: I am very slightly worried about this because it seems like something we would have to manually update and wouldn't be generated by the normal regression data thing we do. Is that true? And if so would it be possible to stick it on to one of our sync methods somehow?

atharva-2001: Right now it compares to the existing saved regression data so that generating new one is not needed. Adding regression data would mostly mean duplicating the items already saved. Using the sync method will mean generating new data.I don't think we would want new data for a very similar functionality, would we? The workflows will always produce the same data the simulation produced with the same configuration right?

jvshields: Oh I see now. Yes, piggybacking off existing non-workflow regression data is fine and good and I'm happy with that. I will instead just request you add a comment at the top of the test that points this out explicitly because we will run into problems generating regression data if this test ever runs before that one. Currently that shouldn't happen since the workflow tests are run last, but we've moved tests around before and it'd be nice to know more easily if something breaks as a result.

PR #3278, opened 2025-09-02
Refactor Monte Carlo transport packet tracking

wkerzendorf:
thia ia based on the previous restructure PR


docs/tutorials/tardis_example.yml
andrewfullard, line 1: Use standard tardis example dl

connor-mcclellan:
This PR is missing changes that were merged into `master` before a re-generation of the `tardis-regression-data` (commit [#76](https://github.com/tardis-sn/tardis-regression-data/commit/d0d32c847361968eb8e3ef43b972b8377d757afd)) which was causing the `simulation` tests to fail since the expected values differed from actual by ~1%.

[4e26f22](https://github.com/tardis-sn/tardis/pull/3278/commits/4e26f220ae19984007d326fa29847c77ece64ce2) is a merge from `master`  into this PR to include those changes. I ran the tests locally to confirm the `simulation` tests are now passing. The only merge conflict here was a tutorial notebook.

Other tests are still failing: some in `transport/montecarlo/`, `transport/montecarlo/packets/`, and a lot of the `visualization` tests. Looking into those next.

connor-mcclellan:
There appears to be a real issue with the rpacket tracker which is causing the SDEC plots to fail non-trivially. The luminosity emitted after the changes doesn't match the regression data --- there is always less total luminosity from emission and absorption.

I suspect the cause is filtering out some packets that were previously counted here (which would account for the missing luminosity compared with the regression data), but haven't been able to track it down yet. But, this PR shouldn't move forward until this is sorted out.

There is a notebook demonstrating this issue at [/tardis/visualization/tools/tests/debug_sdec.ipynb](https://github.com/tardis-sn/tardis/blob/3eababf73174e5a6b3940f44770fdac548335590/tardis/visualization/tools/tests/debug_sdec.ipynb) (should probably be removed before merging if we figure out the cause)

tardis/visualization/tools/tests/test_sdec_plot.py
```diff
@@ -136,8 +143,7 @@ def test_parse_species_list(self, regression_data, plotter, attribute):
         else:
             np.testing.assert_allclose(getattr(plotter, attribute), data, atol=0, rtol=RELATIVE_TOLERANCE_SDEC)
 
-    @pytest.fixture(scope="class", params=combinations,
-        ids=lambda p: f"distance={p[0]}, packets_mode={p[3]}, nelements={p[4]}, show_modeled_spectrum={p[5]}")
+    @pytest.fixture(scope="class", params=combinations)
```
wkerzendorf, line 144: @atharva-2001 - make sure to put the ids parameter into an issue. so we add this in the next PR.

atharva-2001: Sure- https://github.com/tardis-sn/tardis/issues/3340

tardis/transport/montecarlo/packets/trackers/tracker_last_interaction.py
```diff
@@ -0,0 +1,239 @@
...
+        self.after_nu = r_packet.nu
+        self.after_energy = r_packet.energy
+        self.after_mu = r_packet.mu
+        self.interactions_count += 1
+        # Update general tracking
+        self.r = r_packet.r
+        self.nu = r_packet.nu
+        self.energy = r_packet.energy
+        self.shell_id = r_packet.current_shell_id
+        self.interaction_type = InteractionType.CONTINUUM_PROCESS  # Set interaction type directly
```
andrewfullard, line 193: This always assumes bound-free, correct? What about free-bound?

PR #3282, opened 2025-09-03
Implement downbranch into new macroatom.

jvshields:
I pulled out the downbranch implementation from PR #3240 to make it easier to review in isolation. 

Not hooked up to anything in this PR. 

Both files touched also got numpydoc docstrings and type hints.


tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -223,45 +224,59 @@ def solve(
...
-
-        macro_atom_transition_metadata = pd.concat(
-            [
-                emission_down_metadata,
-                internal_down_metadata,
-                internal_up_metadata,
-            ]
-        )
+        if (
+            self.line_interaction_type == "downbranch"
```
wkerzendorf:
```suggestion
            self.line_interaction_type == LINE_INTERACTION_TYPE.DOWNBRANCH
```

jvshields: Opened issue #3286 for this.

tardis/opacities/macro_atom/macroatom_state.py
```diff
@@ -1,6 +1,9 @@
+from __future__ import annotations
```
andrewfullard: This *probably* shouldn't be needed now that we have python>=3.13 in pyproject.toml

jvshields: I'm reading a bit more into it, but without it there the from_legacy_plasma method complains that LegacyMacroAtomState isn't defined yet you can't import tardis.

jvshields: https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class 
and https://peps.python.org/pep-0749/
Apparently this will probably just work in 3.14, but not yet. So we do need it for now for classmethods like this.

andrewfullard: booo

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -16,33 +17,36 @@
...
-from tardis.configuration.sorting_globals import SORTING_ALGORITHM
-
 
 class LegacyMacroAtomSolver:
     initialize: bool = True
     normalize: bool = True
 
-    def __init__(self, initialize=True, normalize=True):
-        """Solver class for Macro Atom related opacities
+    def __init__(self, initialize: bool = True, normalize: bool = True) -> None:
```
wkerzendorf, line 26: is there any point in not normalizing them?

jvshields: No but that's the legacy solver that I'm not touching. The docstrings were just updated because it's in the file that I touched.

tardis/opacities/macro_atom/macroatom_state.py
wkerzendorf: remove

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -223,45 +254,59 @@ def solve(
...
-
-        macro_atom_transition_metadata = pd.concat(
-            [
-                emission_down_metadata,
-                internal_down_metadata,
-                internal_up_metadata,
-            ]
-        )
+        if (
+            self.line_interaction_type == "downbranch"
```
wkerzendorf: makes this `in ['downbranch', 'macroatom']`

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -289,6 +334,17 @@ def solve(
             ],
             inplace=True,
         )
+        source_to_index = {
+            source: idx
+            for idx, source in enumerate(
+                macro_atom_transition_metadata.source.unique()
+            )
+        }
+        macro_atom_transition_metadata["destination_level_idx"] = (
```
andrewfullard, line 346: Add descriptive comment about the -99 magic number

andrewfullard (approved):

PR #3288, opened 2025-09-04
Cuda Tests

atharva-2001:
### :pencil: Description

**Type:**  :rocket: `feature` | :roller_coaster: `infrastructure`

- [X] Adds cuda tests, currently only for Cuda formal integral tests
- [X] Does not use setup env action since mamba needs to be cleaned up. 
- [X] These tests will only run if the test formal integral label is applied.

### :pushpin: Resources

- [X] Testing pipeline
- [ ] Other method (describe)


.github/workflows/tests.yml
```diff
@@ -44,6 +44,33 @@ jobs:
       regression-data-repo: tardis-sn/tardis-regression-data
       allow_lfs_pull: ${{ github.ref == 'refs/heads/master' || contains(github.event.pull_request.labels.*.name, 'git-lfs-pull') }}
 
+  ubuntu-tests:
```
andrewfullard: This should be renamed to cuda-tests or similar.

.github/workflows/cuda-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+jobs:
+  test-cache:
+    if: github.repository_owner == 'tardis-sn' && contains(github.event.pull_request.labels.*.name, 'test-formal-integral')
+    uses: ./.github/workflows/lfs-cache.yml
+    with:
+      atom-data-sparse: false
+      regression-data-repo: tardis-sn/tardis-regression-data
+      allow_lfs_pull: ${{ github.ref == 'refs/heads/master' || contains(github.event.pull_request.labels.*.name, 'git-lfs-pull') }}
+
+  build:
```
andrewfullard: I don't think `build` is a good name for this job. Perhaps `run-tests` or similar.

wkerzendorf:
as mentioned before - let's not call them cuda tests but full tests that simply run on our moria runner and that means that all tests (including the ones that require cuda will run),

.github/workflows/cuda-tests.yml
atharva-2001: Fixed in https://github.com/tardis-sn/tardis/pull/3288/commits/22c8161358e2acb761c155191872ab82c9d80377

.github/workflows/tests.yml
atharva-2001: I changed it to run-tests in https://github.com/tardis-sn/tardis/pull/3288/commits/22c8161358e2acb761c155191872ab82c9d80377

andrewfullard (approved):

.github/workflows/full-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+          environment-file: conda-linux-64.lock
+          environment-name: tardis
+          micromamba-binary-path: ${{ runner.temp }}/bin/micromamba
+          post-cleanup: all
+
+      - name: Install tardis editable
+        run: |
+          pip install -e ".[tardisbase]"
+
+      - name: Install qgridnext and lineid_plot
```
jvshields, line 67: Check qgridnext install

PR #3318, opened 2025-09-30
Continuum transitions

jvshields:
When trying to run the simulation crashes with 
```PlasmaException: Negative values in CorrPhotoIonRateCoeff.  Try raising the number of montecarlo packets.```
This bug MAY be due to an incorrect calculation of CorrPhotoIonRateCoeff. It was investigated briefly by Connor and it seems like CorrPhotoIonRateCoeff or the values used to calculate it differ from analytic values calculated in some of the solvers used for Andrew's ionization and excitation calculations, and may resolve with that.

### DIFF
--- a/tardis/opacities/macro_atom/tests/test_continuum_transitions.py
+++ b/tardis/opacities/macro_atom/tests/test_continuum_transitions.py
@@ -527,7 +527,3 @@ class TestEdgeCases:
 
         result = probability_recombination_internal(neg_df, pos_series)
         assert (result < 0).all().all()  # Result should be negative
-
-
-if __name__ == "__main__":
-    pytest.main([__file__])

### CODE REVIEW

File: tardis/opacities/macro_atom/macroatom_transitions.py [hunk 1]
```diff
@@ -354,3 +357,307 @@ def probability_emission_down(
...
+            "destination": destinations,
+            "transition_type": MacroAtomTransitionType.RECOMB_INTERNAL,
+            "transition_line_idx": -99,
+        },
+        index=p_recomb_internal.index,
+    )
+    return p_recomb_internal, recombination_internal_metadata
+
+
+def probability_recombination_internal(
```
DeerWhale: this function and the next two functions missing doc strings

tardis/opacities/macro_atom/macroatom_transitions.py
```diff
@@ -354,3 +357,307 @@ def probability_emission_down(
...
+    #     cool_rate_adiabatic = cooling_rate_series2dataframe(
+    #         cool_rate_adiabatic, destination_level_idx="adiabatic"
+    #     )
+    #     return cool_rate_adiabatic
+
+
+def continuum_transition_collision_deexc_deactivate(
+    coll_deexc_coeff, yg_idx, electron_densities, delta_E_yg
+):
+    p_deexc_deactivation = (coll_deexc_coeff * electron_densities).multiply(
```
DeerWhale: This looks like the same as the first line in the function above, is this needed elsewhere?

tardis/opacities/macro_atom/tests/test_continuum_transitions.py
```diff
@@ -0,0 +1,533 @@
...
+
+        # Negative coefficients should be allowed (could represent different physics)
+        neg_df = pd.DataFrame([[-1e-12]], index=index, columns=[0])
+        pos_series = pd.Series([10.0], index=index)
+
+        result = probability_recombination_internal(neg_df, pos_series)
+        assert (result < 0).all().all()  # Result should be negative
+
+
+if __name__ == "__main__":
```
andrewfullard: Probably some LLM nonsense ;)

jvshields: Woops, definitely missed that one.

jvshields:
I think the benchmarks are failing because the imports in the benchmark file got changed, so they can't find the function to import in the old commit checkout.

tardis/opacities/macro_atom/tests/__init__.py
connor-mcclellan, line 1: Just confirming that this is included intentionally - is this needed for pytest, or are these macroatom tests imported as a module somewhere I'm not seeing?

jvshields: Reasonable question. I didn't think too much about it since most of our test directories include one. I don't think we usually import tests, but I also don't remember off the top of my head if it's important for any pytest or fixture functionality.

connor-mcclellan (approved):
Looks good to me!

PR #3326, opened 2025-10-17
tests: compress regression artifacts (HDF/NPZ) to reduce CI storage

NobleCoder69:
This PR reduces the size of generated regression artifacts on CI by writing compressed HDF/NPZ where possible.\n\nChanges\n- add helper `tardis/tests/regression_storage.py` providing:\n  - `save_npz`: np.savez_compressed wrapper\n  - `to_hdf_compressed`: uses domain to_hdf when available, otherwise pandas with compression\n- update test helpers writing HDF buffers to use compressed writes:\n  - `tardis/spectrum/tests/test_spectrum.py`\n  - `tardis/transport/montecarlo/tests/test_base.py`\n  - `tardis/model/tests/test_density.py`\n  - `tardis/model/tests/test_base.py`\n\nNotes\n- No behavioral changes to assertions; only storage method differs.\n- Domain HDFWriter mixin already compresses (blosc). Fallback path uses pandas with zlib.\n- Compression level set to 9 by default; can be tuned later if runtime becomes an issue.\n\nFollow-ups (separate PR)\n- Migrate checked-in regression .npy to .npz and update loaders.\n- Consider storing only required fields instead of whole objects where feasible.\n- Optional: add env knobs for compression level/precision for lossy size trade-offs.\n

### DIFF
--- a/tardis/model/tests/test_base.py
+++ b/tardis/model/tests/test_base.py
@@ -472,7 +472,9 @@ class TestModelStateFromNonUniformAbundances:
 
 @pytest.fixture(scope="module", autouse=True)
 def to_hdf_buffer(hdf_file_path, simulation_verysimple):
-    simulation_verysimple.simulation_state.to_hdf(hdf_file_path, overwrite=True)
+    from tardis.tests.regression_storage import to_hdf_compressed
+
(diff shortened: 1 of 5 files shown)


andrewfullard:
Thanks for your contribution. Since you did not follow our automated responses regarding mailmap and copilot review, we are closing this PR.

PR #3346, opened 2025-11-20
Add script to convert atomic data from the format used in Christian's code to modern TARDIS

connor-mcclellan:
### :pencil: Description

**Type:** :rocket: `feature` :memo: `documentation`

Christian's version of TARDIS uses an older format for atomic data. This script converts the `merged_mod_20SNG_forbidden_yg_fix_H30_cmfgen_yg.h5` atomic data file (and separate photoionization data file `tardis/continuum/data/photoionization_data_H30_He.h5`) to a format compatible with the `master` branch of TARDIS using [`tardis-regression-data/atom_data/nlte_atom_data/TestNLTE_He_Ti.h5`](https://github.com/tardis-sn/tardis-regression-data/blob/main/atom_data/nlte_atom_data/TestNLTE_He_Ti.h5) as a template for the correct table formatting and column data types.

This is certainly a one-time patch solution for this single file and shouldn't be used on other atom data files. The only portable pieces of this script are potentially the `multiindex_port` and `simple_port` functions which convert `h5py.File` datasets into `pd.DataFrame` objects fairly reliably, converting multi-indices and column data types appropriately.

- [ ] Testing pipeline
- [ ] Other method (describe)
- [X] My changes can't be tested (explain why)

Will depend on comparisons between the deprecated version of TARDIS and more recent TARDIS to verify that the atom data has been ported correctly.

### DIFF
--- a/docs/how-to/convert_atomic_data.ipynb
+++ b/docs/how-to/convert_atomic_data.ipynb
@@ -12,7 +12,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "3a629c49",
    "metadata": {},
    "outputs": [],
@@ -67,7 +67,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "id": "2cbdc264",
    "metadata": {},
    "outputs": [],
@@ -154,7 +154,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "id": "065f3239",
    "metadata": {},
    "outputs": [],
@@ -217,7 +217,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 4,
    "id": "0fcf0236",
    "metadata": {},
    "outputs": [],
@@ -282,7 +282,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
    "id": "a60ded02",
    "metadata": {},
    "outputs": [],
@@ -314,7 +314,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 6,
    "id": "b40155a2",
    "metadata": {},
    "outputs": [],
@@ -327,7 +327,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 7,
    "id": "11408b63",
    "metadata": {},
    "outputs": [],
@@ -374,10 +374,21 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 8,
    "id": "380291d3",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "['/Y/g']"
+      ]
+     },
+     "execution_count": 8,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
    "source": [
     "old_df.keys()"
    ]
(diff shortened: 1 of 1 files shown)


andrewfullard:
Looks like the PR needs a rebase!

connor-mcclellan:
Whoops--- should be fixed now

andrewfullard (commented):
Could you convert this to a notebook with markdown commentary?

connor-mcclellan:
> Could you convert this to a notebook with markdown commentary?

Sure- I'll leave the .py script in there too so it can be run straight from the command line if desired.

jvshields:
I've looked at this a bit. I would rather this be stuck in a util folder, rather than directly in docs. If there's a forthcoming notebook, that could probably go in the how-to docs folder, but this script probably shouldn't live in docs directly.

Also, there's some weirdness in that this imports carsus, and tardis doesn't actually have carsus as a dependency, so you won't be able to run this with the tardis environment. That's probably fine since this is mostly just a one-time script we want, but maybe worth noting somewhere.

connor-mcclellan:
> Also, there's some weirdness in that this imports carsus, and tardis doesn't actually have carsus as a dependency, so you won't be able to run this with the tardis environment. That's probably fine since this is mostly just a one-time script we want, but maybe worth noting somewhere.

Good point - the usage of carsus is limited to two utility functions in  `carsus.util` to update the checksum: `hash_pandas_object` and `serialize_pandas_object`. These are super short functions (https://github.com/tardis-sn/carsus/blob/master/carsus/util/hash.py) and can be split off and included in this script so we don't have a random carsus dependency. I suspect @andrewfullard just did this out of convenience in his notebook demonstrating how to reconstruct the metadata, so I'll fix that before we merge this

tardis/scripts/convert_atomic_data.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # Check datatypes of columns, convert if necessary
+    for col in templatedata[templatekey].columns:
+        newdata[col] = newdata[col].astype(templatedata[templatekey][col].dtype)
+
+    return newdata
+
+
+# Load data in old format - use h5py for now, not pandas
+filename = "merged_mod_20SNG_forbidden_yg_fix_H30_cmfgen_yg.h5"
```
jvshields: This filename and the ones below should probably just be command line arguments, or this could be handled in a notebook which I remember andrew suggesting.

andrewfullard:
One small issue is that a number of dataframes end up with non-integer index values, which causes issues later.

jvshields:
I think this also still needs this field to be the array of collision temperatures 
`atomic_data.collision_data_temperatures`
otherwise the tardis atom data complains.

tardis/scripts/convert_atomic_data.py
```diff
@@ -0,0 +1,254 @@
...
+    Parameters
+    ----------
+    pd_object : pandas.Series or pandas.DataFrame
+        Pandas object to be serialized with Pickle.
+
+    Returns
+    -------
+    Pickle serialized Python object.
+    """
+    return pickle.dumps(pd_object)
```
jvshields, line 36: I guess this comes from carsus so not exactly your problem to fix, but we pretty strongly try to avoid pickling objects where possible.

connor-mcclellan:
> One small issue is that a number of dataframes end up with non-integer index values, which causes issues later.

Aha - I thought I handled this correctly but a couple dataframes snuck through. Apparently `isinstance(<something>, pd.Index)` returns true for both `pd.Index` and `pd.MultiIndex`, so since I need to handle these cases separately I should differentiate it based on `isinstance(<something>, pd.MultiIndex)` instead. Good catch

tardis/scripts/convert_atomic_data.py
connor-mcclellan: This is just used to calculate the checksum for each dataframe in the file. Probably just a quick way of converting an object to bytes so it can be summed up.

https://github.com/tardis-sn/tardis/pull/3346/files#diff-2a4435b796d518c6b4bfd50499b079beb820c6637ab83a6cd7609774f23cd92bR204-R211
image

tardis/scripts/convert_atomic_data.py
```diff
@@ -0,0 +1,241 @@
+"""
+Quick-and-dirty script for converting an older atomic data format to one compatible with modern TARDIS.
```
jvshields, line 2:
```suggestion
Quick-and-dirty script for converting an older atomic data format to one compatible with modern TARDIS.

Arguments are {old_atomdata_to_convert) {old_photoionization_data_to_convert} {new_atomdata_as_template} {output_atomdata}
```

jvshields:
Looks good. I suggested a change to add the script arguments to the header comment, but other than that I'm happy with it.

andrewfullard (approved):

andrewfullard:
This breaks the docs build. Options: find a fix, or commit the pre-run notebook so that the docs build does not run it.

andrewfullard (approved):

PR #3348, opened 2025-11-26
Continuum solver CT19 plasma

andrewfullard:
### :pencil: Description

**Type:**  :rocket: `feature` 

A transplant of chvogl's plasma suitable for IIP.


andrewfullard:
Move to `plasma/equilibrium/ctardis`. Remove h5 files

tardis/iip_plasma/base.py
```diff
@@ -181,44 +197,44 @@ def _resolve_update_list(self, changed_properties):
         descendants_ob = list(set(descendants_ob))
         sort_order = list(nx.topological_sort(self.graph))
 
-        descendants_ob.sort(key=lambda val: sort_order.index(val) )
+        descendants_ob.sort(key=lambda val: sort_order.index(val))
 
-        logger.debug('Updating modules in the following order:'.format(
-            '->'.join(descendants_ob)))
+        logger.debug(
+            "Updating modules in the following order:"
```
jvshields: This seems like it dropped something and is no longer a sensible debug logger message.

tardis/iip_plasma/base.py
```diff
@@ -181,44 +197,44 @@ def _resolve_update_list(self, changed_properties):
...
 
         return descendants_ob
 
     def write_to_dot(self, fname, latex_label=True):
-#        self._update_module_type_str()
+        #        self._update_module_type_str()
 
         try:
-            import pygraphviz
+            pass
```
jvshields: This should never hit the except block now. Is it worth just removing that?

tardis/iip_plasma/continuum/base_continuum_data.py
```diff
@@ -155,44 +240,54 @@ def from_hdf5(cls, fname, atom_data):
...
+        ]
         # TODO: Fixme! Have to also use ionization stage here
-        photoionization_data = photoionization_data.loc[pd.IndexSlice[selected_atoms,:,:], :]
-        #photoionization_data.drop([(1,0,i) for i in range(14, 20)], inplace=True)
-        photoionization_data.loc[(1,0,0),'x_sect'] *= 0.0
+        photoionization_data = photoionization_data.loc[
+            pd.IndexSlice[selected_atoms, :, :], :
+        ]
+        # photoionization_data.drop([(1,0,i) for i in range(14, 20)], inplace=True)
+        photoionization_data.loc[(1, 0, 0), "x_sect"] *= 0.0
```
jvshields, line 252: This seems concerning. Not something to change here since it comes from the old plasma, but something to ask Christian about if we get a hold of him.

tardis/iip_plasma/continuum/collisional_processes.py
```diff
@@ -1,10 +1,12 @@
+import astropy.constants as const
```
jvshields: Could be worth standardizing things with tardis constants, or maybe making a global sets where to import from.

docs/physics_walkthrough/plasma/equilibrium/hydrogen_continuum_ctardis_plasma.ipynb
```diff
@@ -0,0 +1,596 @@
...
+   "outputs": [],
+   "source": [
+    "from tardis.iip_plasma.continuum.base_continuum_data import ContinuumData\n",
+    "\n",
+    "atom_data.continuum_data = ContinuumData(\n",
+    "               atom_data, selected_continuum_species=[(1, 0)]\n",
+    "           )\n",
+    "\n",
+    "atom_data.Yg_data = atom_data.yg_data\n",
+    "atom_data.Yg_data.columns = list(atom_data.collision_data_temperatures)\n",
```
connor-mcclellan: I can see that functions in `tardis/iip_plasma/properties/continuum.py` and `tardis/iip_plasma/properties/atomic.py` reference `Yg_data` and not `yg_data` here, which makes this necessary, but I wonder if it's worth refactoring if it's just in those two locations (and also confined to the IIp plasma).

Christian's original atomic data also already had temperatures as column names, but I set them to indices instead for consistency with the template data. If no problems are caused by having temperatures instead of indices in the column names, though, I could change the conversion script to keep them there and that would make setting the columns to temperatures unnecessary here.

tardis/iip_plasma/continuum/radiative_processes.py
```diff
@@ -325,23 +435,45 @@ class FreeFree(PhysicalContinuumProcess):
...
     def _calculate_chi_ff_factor(self):
         ionic_charge_squared = np.square(self._get_ionic_charge())
-        ff_gaunt_factor = self._get_ff_gaunt_factor(self.ion_number_density.index)
-        chi_ff_helper = 3.69255e8 * self.electron_densities / np.sqrt(self.t_electrons)
-        chi_ff_factor = self.ion_number_density.multiply(ionic_charge_squared * ff_gaunt_factor, axis=0).sum().values
+        ff_gaunt_factor = self._get_ff_gaunt_factor(
+            self.ion_number_density.index
+        )
+        chi_ff_helper = (
+            3.69255e8 * self.electron_densities / np.sqrt(self.t_electrons)
```
jvshields, line 475: Spooky constant. If there's any idea where it came from or some description, adding it would be nice.

tardis/iip_plasma/properties/atomic.py
```diff
@@ -27,19 +40,28 @@ class Levels(BaseAtomicDataProperty):
...
+        "\\textrm{levels}",
+        "\\epsilon_{\\textrm{k}}",
+        "\\textrm{metastability}",
+        "g",
+    )
 
     def _filter_atomic_property(self, levels, selected_atoms):
-        return levels
-        # return levels[levels.atomic_number.isin(selected_atoms)]
+        return levels.loc[(selected_atoms), :]
```
jvshields, line 53: Big code difference. This is probably the intention but could change something somewhere else.

jvshields: Is it a worry that the lines one immediately below this on new line 85 doesn't change the same way?

tardis/iip_plasma/properties/atomic.py
```diff
@@ -55,201 +77,239 @@ class Lines(BaseAtomicDataProperty):
...
+
+    # Would like for lines to just be the line_id values, but this is necessary because of the filtering
+    # error noted below.
+    outputs = ("lines", "nu", "f_lu", "wavelength_cm", "lines_multi_index")
 
     def _filter_atomic_property(self, lines, selected_atoms):
         # return lines[lines.atomic_number.isin(selected_atoms)]
         return lines
 
     def _set_index(self, lines):
```
jvshields, line 89: Er, this doesn't actually do anything at all with setting the index, does it? Just calling attention to it so somebody else also looks at it.

tardis/iip_plasma/properties/atomic.py
```diff
@@ -55,201 +77,239 @@ class Lines(BaseAtomicDataProperty):
...
-                                 index=levels)
-        lines_index = lines.set_index(
-            ['atomic_number', 'ion_number',
-             'level_number_lower']).index
-        return np.array(levels_index.ix[lines_index])
+        levels_index = pd.Series(
+            np.arange(len(levels), dtype=np.int64), index=levels
+        )
+        lines_index = lines.index.droplevel("level_number_upper")
+        return np.array(levels_index.loc[lines_index])
```
jvshields, line 123: I can't completely tell what this is doing, but this might be an .iloc and not a .loc. I think it's worth looking more into or double checking.

tardis/iip_plasma/continuum/radiative_processes.py
```diff
@@ -0,0 +1,484 @@
...
+        corrected_photoion_coeff = pd.DataFrame(tmp)
+        return corrected_photoion_coeff
+
+    def _calculate_rate_coefficient_from_estimator(self):
+        index = self._get_estimator_index()
+        lte_nonlte_level_pop_ratio = self._get_lte_nonlte_level_pop_ratio(index)
+        corrected_photoion_coeff = (
+            self.photo_ion_estimator
+            - lte_nonlte_level_pop_ratio * self.stim_recomb_estimator
+        )
```
connor-mcclellan, line 133: These values have the potential to be negative with poor estimator statistics --- should raise an exception (perhaps with the error message "try increasing the number of monte carlo packets", like in Jack's code) if there are any negative values

tardis/iip_plasma/properties/properties.py
connor-mcclellan, line 1: ?

Is this supposed to be here?

tardis/iip_plasma/properties/continuum.py
```diff
@@ -300,29 +415,40 @@ def calculate(self, photo_ion_cross_sections, t_electrons):
...
+        ).first()
+        factor = self._calculate_factor(
+            collion_coeff["nu"].values, t_electrons, index=collion_coeff.index
+        )
+        collion_coeff = 1.55e13 * collion_coeff["x_sect"]
         ion_number = collion_coeff.index.get_level_values(1).values
-        collion_coeff.ix[ion_number == 0] *= 0.1
-        collion_coeff.ix[ion_number == 1] *= 0.2
-        collion_coeff.ix[ion_number >= 2] *= 0.3
+        collion_coeff.iloc[ion_number == 0] *= 0.1
```
jvshields: I think these aren't incorrect but I would get rid of the iloc. It confused me on first couple reads and is unnecessary.

tardis/iip_plasma/continuum/base_continuum_data.py
andrewfullard: It's something to do with the ground state of Hydrogen being really weird

tardis/iip_plasma/continuum/collisional_processes.py
andrewfullard: Good catch, this may be the cause of some differences

tardis/iip_plasma/properties/atomic.py
andrewfullard: This is a change to match the current TARDIS plasma so that the levels are correctly filtered, but it may be unnecessary.

tardis/iip_plasma/properties/atomic.py
andrewfullard: Correct. I think it's an artefact of how the `BaseAtomicDataProperty` works

tardis/iip_plasma/properties/ion_population.py
```diff
@@ -203,22 +273,27 @@ class IonNumberDensity(ProcessingPlasmaProperty):
...
+        "N_{i,j}",
+        "n_{e}",
+    )
 
     def __init__(self, plasma_parent, ion_zero_threshold=1e-20):
         super(IonNumberDensity, self).__init__(plasma_parent)
         self.ion_zero_threshold = ion_zero_threshold
         self.block_ids = None
 
     def update_helium_nlte(self, ion_number_density, number_density):
```
jvshields, line 288: So this function just sets all helium to singly ionized? Again not something to change but something I wanted to highlight.

andrewfullard: This is the "helium treatment" which is unused for IIP

tardis/iip_plasma/properties/ion_population.py
```diff
@@ -231,180 +306,281 @@ def calculate_with_n_electron(self, phi, partition_function,
...
-                if 'HeliumNLTE' in \
-                    self.plasma_parent.plasma_properties_dict.keys():
-                    ion_number_density = \
-                        self.update_helium_nlte(ion_number_density,
-                        number_density)
+                phi, partition_function, number_density, n_electron
+            )
+            if hasattr(self.plasma_parent, "plasma_properties_dict"):
+                if (
+                    "HeliumNLTE"
```
jvshields, line 343: As per previous comment, setting "HeliumNLTE" as a plasma property will just make all helium singly ionized.

tardis/iip_plasma/continuum/radiative_processes.py
andrewfullard: I think I know what it is- see https://github.com/andrewfullard/tardis/blob/b87db382b06b5c0f29364a2d00cb31ce35543c16/tardis/opacities/opacities.py#L25

tardis/iip_plasma/properties/ion_population.py
jvshields: Also I think these ones should actually be locs instead of ilocs, right? They should point to the atomic number and the ion number, not the integer index location.

docs/physics_walkthrough/plasma/equilibrium/hydrogen_continuum_ctardis_plasma.ipynb
andrewfullard: I think renaming yg_data in the iip_plasma is fine, yeah.

I want to avoid using the old data as the "correct" version as much as possible, so let's keep the index-based columns and temperature in metadata.

tardis/iip_plasma/properties/properties.py
andrewfullard: Something from old old TARDIS I expect. I'll remove it.

tardis/iip_plasma/properties/level_population.py
```diff
@@ -1,127 +1,184 @@
...
-            level_boltzmann_factor, ion_number_density, levels,
-            partition_function)
+            level_boltzmann_factor,
+            ion_number_density,
+            levels,
+            partition_function,
+        )
         if helium_population is not None:
-            level_number_density.ix[2].update(helium_population)
+            level_number_density.iloc[2].update(helium_population)
```
jvshields: I think this should be .loc too maybe?

tardis/iip_plasma/properties/ion_population.py
andrewfullard: Probably. But I don't think this code will ever be used for the intended purpose of this plasma

tardis/iip_plasma/properties/nlte.py
```diff
@@ -4,143 +4,230 @@
...
+        delta,
+        zeta_data,
+        number_density,
+        partition_function,
+    ):
         """
         Updates all of the helium level populations according to the helium NLTE recomb approximation.
         """
-        helium_population = level_boltzmann_factor.ix[2].copy()
+        helium_population = level_boltzmann_factor.iloc[2].copy()
```
jvshields: Another one that looks like it should be .loc to me.

tardis/iip_plasma/properties/nlte.py
```diff
@@ -4,143 +4,230 @@
...
+        he_one_population = self.calculate_helium_one(
+            g_electron,
+            beta_rad,
+            ionization_data,
+            level_boltzmann_factor,
+            electron_densities,
+            g,
+            w,
+        )
+        helium_population.iloc[0].update(he_one_population)
```
jvshields: Not as sure about any of these ones because I'm not sure how helium_population is structured, but I think .ix tries .loc by default and falls back on .iloc so if these run with .loc then maybe those are right.

tardis/iip_plasma/properties/nlte.py
```diff
@@ -4,143 +4,230 @@
...
         """
         Calculates the He I level population values, in equilibrium with the He II ground state.
         """
-        return level_boltzmann_factor.ix[2,0].mul(
-            g.ix[2,0], axis=0) * (1./(2*g.ix[2,1,0])) * \
-            (1/g_electron) * (1/(w**2)) * np.exp(
-            ionization_data.ionization_energy.ix[2,1] * beta_rad) * \
-            electron_densities
+        return (
+            level_boltzmann_factor.iloc[2, 0].mul(g.iloc[2, 0], axis=0)
```
jvshields: More level_boltzmann_factor that I think should be referenced by loc. 

Line 159 as well.

tardis/iip_plasma/properties/nlte.py
```diff
@@ -4,143 +4,230 @@
...
         """
         zeta = PhiSahaNebular.get_zeta_values(zeta_data, 2, t_rad)[1]
-        he_three_population = (2 / electron_densities) * \
-            (float(g.ix[2,2,0])/g.ix[2,1,0]) * g_electron * \
-            np.exp(-ionization_data.ionization_energy.ix[2,2] * beta_rad) \
-            * w * (delta.ix[2,2] * zeta + w * (1. - zeta)) * \
-            (t_electrons / t_rad) ** 0.5
+        he_three_population = (
+            (2 / electron_densities)
+            * (float(g.iloc[2, 2, 0]) / g.iloc[2, 1, 0])
```
jvshields: gs should be .loc too? But this function doesn't seem to have a return statement so probably doesn't matter for any current bugs.

tardis/iip_plasma/properties/nlte.py
```diff
@@ -151,74 +238,116 @@ def calculate(self, ion_number_density, electron_densities, t_electrons, w,
...
+            helium_lines = helium_lines[helium_lines["ion_number"] == 0]
         for zone, _ in enumerate(electron_densities):
-            with open('He_NLTE_Files/discradfield_{}.txt'.format(zone), 'w') \
-                    as output_file:
+            with open(
+                f"He_NLTE_Files/discradfield_{zone}.txt", "w"
+            ) as output_file:
                 j_blues = pd.DataFrame(j_blues, index=lines.index)
-                helium_j_blues = j_blues[zone].ix[helium_lines.index]
+                helium_j_blues = j_blues[zone].iloc[helium_lines.index]
```
jvshields: This looks like another .loc to me.

tardis/iip_plasma/continuum/radiative_processes.py
andrewfullard: Added. I don't think this class is ever used though.

tardis/iip_plasma/properties/nlte.py
```diff
@@ -151,74 +238,116 @@ def calculate(self, ion_number_density, electron_densities, t_electrons, w,
...
                 for value in helium_lines.index:
```
jvshields, line 259: Everything in here is also looping through the index which should be .loc references.

tardis/iip_plasma/properties/nlte.py
```diff
@@ -151,74 +238,116 @@ def calculate(self, ion_number_density, electron_densities, t_electrons, w,
...
+                "He_NLTE_Files/shellconditions_current.txt",
+                f"He_NLTE_Files/shellconditions{zone}.txt",
+            )
+            os.rename(
+                "He_NLTE_Files/discradfield_current.txt",
+                f"He_NLTE_Files/discradfield{zone}.txt",
+            )
+            os.rename("debug_occs.dat", f"He_NLTE_Files/occs{zone}.txt")
+        # Reading in populations from files
+        helium_population = level_boltzmann_factor.iloc[2].copy()
```
jvshields: Here too. Maybe you've checked this whole file again by now but a lot of the references in here seem wrong to me. I see a couple more below this too.

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -125,131 +159,248 @@ def __init__(self, plasma_parent, classical_nebular=False,
...
+                    beta_sobolevs,
+                    previous_electron_densities,
+                )
+            )
+
+            for i in range(len(t_electrons)):
+                level_boltzmann_factor = np.linalg.solve(
+                    rates_matrix[:, :, i], x
+                )
+                general_level_boltzmann_factor[i].iloc[species] = (
```
jvshields: Could be another one?

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -259,179 +410,282 @@ def _main_nlte_cont_calculation(self, atomic_data, nlte_data,
...
-                    initial = self.plasma_parent.level_number_density[i].ix[species].values
+                    initial = (
+                        self.plasma_parent.level_number_density[i]
+                        .iloc[species]
+                        .values
+                    )
                     initial /= initial.sum()
                     initial = np.hstack([initial, np.array(next_ion_density)])
 
+                    import pdb
```
jvshields: Should definitely take this one out before the merge.

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -259,179 +410,282 @@ def _main_nlte_cont_calculation(self, atomic_data, nlte_data,
...
+                        self.plasma_parent.number_density.loc[species[0], i],
+                        gamma_vec,
+                        species,
+                        phis,
+                    )
 
-                    initial = self.plasma_parent.level_number_density[i].ix[species].values
+                    initial = (
+                        self.plasma_parent.level_number_density[i]
+                        .iloc[species]
```
jvshields: Another one to maybe double check, but not sure how species looks at this point in the code.

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -259,179 +410,282 @@ def _main_nlte_cont_calculation(self, atomic_data, nlte_data,
...
 
                     level_boltzmann_factor = res.x[:-1]
                     ion_ratio[i] = res.x[-1]
 
-                    general_level_boltzmann_factor[i].ix[species] = \
+                    general_level_boltzmann_factor[i].iloc[species] = (
                         level_boltzmann_factor
+                    )
+
+                    import pdb
```
jvshields: More pdbs to remove.

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -259,179 +410,282 @@ def _main_nlte_cont_calculation(self, atomic_data, nlte_data,
...
-                            raise PlasmaNLTEExcitationError('level_boltzmann_factor is negative - aborting')
+                            raise PlasmaNLTEExcitationError(
+                                "level_boltzmann_factor is negative - aborting"
+                            )
 
                     level_boltzmann_factor = res.x[:-1]
                     ion_ratio[i] = res.x[-1]
 
-                    general_level_boltzmann_factor[i].ix[species] = \
+                    general_level_boltzmann_factor[i].iloc[species] = (
```
jvshields: Also could be a .loc

tardis/iip_plasma/properties/radiative_properties.py
```diff
@@ -34,52 +52,75 @@ def __init__(self, plasma_parent=None, nlte_species=None):
 
     def get_g_lower(self, g, lines_lower_level_index):
         if self._g_lower is None:
-            g_lower = np.array(g.ix[lines_lower_level_index],
-                                     dtype=np.float64)
+            g_lower = np.array(
+                g.iloc[lines_lower_level_index], dtype=np.float64
```
jvshields, line 56: I would have figured this and 64 should be .locs too

jvshields:
Still need the build docs to be fixed, but otherwise happy with this.

PR #3349, opened 2025-12-03
IIP workflow

jvshields:
This PR just sets up the iip workflow, assuming continuum transitions. The plasma is sort of hooked up, but still has a bug that makes the simulation not run past the first iteration. We will probably merge this and then plop in christian's plasma in the short term.


andrewfullard:
Time for a rebase :)

PR #3359, opened 2025-12-10
Move Arepo snapshot reader and associated tests

wkerzendorf:
Moving in line with the other restructure

### DIFF
--- a/tardis/io/model/arepo/tests/test_arepo_parser.py
+++ b/tardis/io/model/arepo/tests/test_arepo_parser.py
@@ -1,5 +1,4 @@
 import json
-import os
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


andrewfullard:
This needs to be rebased

tardis/io/model/arepo/arepo.py
jvshields, line 1: This file was moved, but also it looks like a lot of the logic is lifted into the data.py and readers.py files. Double check that there's not code duplication.

tardis/io/model/arepo/tests/test_arepo_parser.py
```diff
@@ -0,0 +1,233 @@
...
+        xnuc_prof_p,
+        arepo_data.time,
+        example_model_file_dir / "arepo_parser_test.csvy",
+        nshells=20,
+    )
+
+    with open(testfile) as file:
+        csvy_lines = file.readlines()
+
+    print(testfile)
```
wkerzendorf: needs to be removed.

andrewfullard (approved):

PR #3362, opened 2025-12-12
Adds geometry mesh classes for TARDIS model IO

wkerzendorf:
Introduces a comprehensive geometry system with mesh classes that handle
1D radial finite-volume meshes and velocity-based homologous expansion.

Provides type-safe quantity handling with astropy units and clear location
semantics to distinguish between interface and volume properties.

Enables easy conversion between velocity and spatial representations,
establishing the foundation for TARDIS supernova model spatial structure.

Includes demonstration notebook showing practical usage for Type Ia
supernova mesh construction and visualization.


andrewfullard:
The mesh stuff is nice, but the docs changes seem like overkill

PR #3364, opened 2025-12-12
Fix typo in SDEC tests causing name collision and incorrect regression data

### LINKED ISSUE #3363: False positive tests results for SDEC `test_generate_plot_mpl`
**Describe the bug**
There is a re-definition of the SDEC test for `test_generate_plot_mpl` which uses data for `plotter_generate_plot_ply`--- probably a copy-paste error at some point. This is currently the test used in `master`, meaning the regression data for these tests is incorrect.

**Screenshots**
Unused test that actually uses the correct `plotter_generate_plot_mpl` function:
Image

Test that overwrites `test_generate_plot_mpl`:
Image

The notable consequence here is that `test_generate_plot_mpl` actually tests `test_generate_plot_ply` and `test_generate_plot_ply` doesn't exist, so neither of these tests will produce the correct regression data. This name collision needs to be fixed and new regression data needs to generated for the SDEC plots.

**System**
 - OS:
    - [X] GNU/Linux
    - [ ] macOS

connor-mcclellan:
### :pencil: Description

**Type:** :beetle: `bugfix`

Fixes #3363.

- [X] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/visualization/tools/tests/test_sdec_plot.py
+++ b/tardis/visualization/tools/tests/test_sdec_plot.py
@@ -276,7 +276,7 @@ class TestSDECPlotter:
             # save artists which correspond to element contributions
             if isinstance(data, PolyCollection):
                 for index2, path in enumerate(data.get_paths()):
-                    np.testing.assert_almost_equal(
+                    np.testing.assert_allclose(
                         path.vertices,
                         expected.get(
                             "plot_data_hdf/"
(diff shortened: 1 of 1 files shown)


connor-mcclellan:
Updated regression data for this PR: https://github.com/tardis-sn/tardis-regression-data/pull/82

connor-mcclellan:
Tests passing with most recent regression data from https://github.com/tardis-sn/tardis-regression-data/pull/83

PR #3365, opened 2025-12-13
Fixes W7 density solver implementation and reorganizes tests

wkerzendorf:
Corrects the W7 density profile calculation to properly implement the Branch85 W7 model using a power law with exponent -7 and time scaling.

Adds velocity_0 parameter to W7DensitySolver and imports calculate_density_after_time function for accurate time evolution scaling.

Updates parse_density_solver_from_density_config to return W7DensitySolver instead of PowerLawDensitySolver for branch85_w7 type, ensuring proper W7-specific implementation.

Reorganizes test suite by splitting monolithic test file into focused modules for better maintainability and adds comprehensive integration tests using actual configuration files.


andrewfullard:
Interesting. The W7 solver could probably just inherit from the power law solver class. The LLM seems to have correctly caught the other LLM's mistakes.

Also, is the current W7 actually incorrect?

PR #3367, opened 2025-12-15
git-lfs-pull label fix

atharva-2001:
### :pencil: Description

Removes the script that removed the label to save LFS
Similar- https://github.com/tardis-sn/tardis/pull/3367 

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,8 +15,6 @@ on:
       - opened
       - reopened
       - synchronize
-      - labeled
-
   workflow_call:
     inputs:
       pip_git:


.github/workflows/tests.yml
```diff
@@ -15,6 +15,8 @@ on:
       - opened
       - reopened
       - synchronize
+      - labeled
```
atharva-2001: I added this since people will not have to do an empty commit for example to run the workflow after applying the label, but I guess this is a rare occurrence when such feature is needed? Additional conditional statements to further fit this feel too much complexity added to the code. Removing it, let me know if it's required.

PR #3385, opened 2026-01-13
Fix ASV benchmark environment with pinned dependencies (#3144)

### LINKED ISSUE #3144: Benchmarks don't use lockfiles
**Describe the bug**

The benchmarks don't use lockfiles(they are not supported by ASV). 
The second best thing we could do is use a pinned environment to best replicate the lock file?

See- https://github.com/tardis-sn/tardis/blob/master/asv.conf.json

BhoomiAgrawal12:
### :pencil: Description

**Type:** :beetle: `bugfix`

This PR resolves the benchmark environment setup issue where ASV was referencing a non-existent `env.yml` file. More importantly, it ensures reproducible benchmark results by pinning all dependency versions.

**Fixes #3144**

#### The Problem

1. **Missing file**: The ASV configuration (`asv.conf.json`) was pointing to an `env.yml` file that didn't exist in the repository
2. **Reproducibility gap**: Even if the file existed, unpinned dependency versions would cause benchmark variability when packages updated

This meant  benchmarks couldn't reliably detect TARDIS performance regressions because we couldn't separate actual code changes from dependency version changes.

#### Fix

I've implemented an auto-generation approach that creates `env.yml` with exact pinned versions extracted from our existing `conda-lock.yml`:

```bash
python generate_benchmark_env.py
```

This generates an environment file with:
- **489 packages** with exact versions (e.g., `numpy=1.26.4` instead of just `numpy`)
- Extracted from the `linux-64` platform in `conda-lock.yml` (our standard CI platform)
- Standard conda environment format that ASV understands natively

#### Files Changed

- `env.yml` - The ASV environment file (auto-generated, don't edit manually!)
- `generate_benchmark_env.py` - Script to regenerate env.yml from conda-lock.yml
- `docs/contributing/development/benchmarks.rst` - Added "Maintaining the Benchmark Environment" section

- [x] **Testing pipeline**: All verification checks passing
  -  `env.yml` validates as correct YAML syntax
  -  `asv.conf.json` correctly references `env.yml`
  -  Generation script executes successfully
  -  All 489 packages extracted correctly from conda-lock.yml

- [x] **Manual testing**:
  - Verified the auto-generation script produces consistent output
  - Confirmed documentation is accurate and helpful

- [ ] My changes can't be tested

  - Updated `docs/contributing/development/benchmarks.rst` with maintenance section
  - Included clear inline documentation in all files

---

### :memo: Additional Notes for Reviewers

**Auto-generated file**: `env.yml` is **auto-generated** - to check the implementation, run `python generate_benchmark_env.py` and it should produce identical output.

**Maintenance**: When `conda-lock.yml` is updated in the future, just run:
```bash
python generate_benchmark_env.py
```


env.yml
```diff
@@ -0,0 +1,503 @@
...
+
+name: tardis-benchmark
+channels:
+- conda-forge
+dependencies:
+- _libgcc_mutex=0.1
+- _openmp_mutex=4.5
+- accessible-pygments=0.0.5
+- adwaita-icon-theme=48.0
+- aiobotocore=2.22.0
```
atharva-2001: we already have lockfiles, idea was to use lockfiles and not environment files, I guess

atharva-2001: This PR resolves the benchmark environment setup issue where ASV was referencing a non-existent env.yml file. 
How do the benchmarks run then?

generate_benchmark_env.py
```diff
@@ -0,0 +1,86 @@
...
+                channels.append(ch['url'])
+    
+    if not channels:
+        channels = ['conda-forge']
+    
+    # Extract packages for linux-64 platform (standard CI platform)
+    packages = []
+    if 'package' in lock_data:
+        for pkg in lock_data['package']:
+            if pkg.get('platform') == 'linux-64' and pkg.get('manager') == 'conda':
```
atharva-2001: a set of bash commands could have been simpler?

env.yml
atharva-2001: I'd suggest running the benchmarks locally first(that would be the "test" in the testing pipeline)

atharva-2001: furthermore, these pins can have more metadata associated with them, if you could get those it would be better(with working benchmarks)

atharva-2001:
Thanks for the contribution! I left a few comments.

generate_benchmark_env.py
BhoomiAgrawal12: You're right! bash script would definitely be simpler for this parsing task. I ll make the required change.

env.yml
BhoomiAgrawal12: Thanks for the feedback @atharva-2001 

I initially went with generating `env.yml` because ASV's `conda_environment_file` parameter specifically requires a standard conda environment file format,  it doesn't support conda-lock.yml directly. From the [ASV documentation](https://asv.readthedocs.io/en/stable/asv.conf.json.html#conda-environment-file).

However, I see what you're suggesting is we can bypass the `conda_environment_file` approach entirely and instead use conda-lock's installation commands directly

The better approach would be to remove the `conda_environment_file` parameter from `asv.conf.json` and use custom build/install commands that leverage `conda-lock install` directly

Something like:
```json
"build_command": [
    "conda-lock install --name {env_name} conda-lock.yml",
    "python -m pip wheel --no-deps -w {build_cache_dir} {build_dir}"
],
"install_command": [
    "in-dir={env_dir} python -m pip install {wheel_file}"
]
```
Would you like me to try this approach? I can:
- Update `asv.conf.json` to use conda-lock install commands
- Remove the generation script and env.yml entirely

Let me know if this direction makes sense or if you had a different approach in mind! Happy to adjust.

atharva-2001: It would be nice if you could investigate this.

BhoomiAgrawal12:
@atharva-2001, Thankyou so much for the feedback on this PR. I spent some time digging into whether we could use `conda-lock.yml` directly with ASV and I found that when I tried to use conda-lock.yml directly it fails because ASV needs the environment already created before it runs build_command. When using custom conda_create_command wrapper it  causes lock conflicts because ASV still tries to use conda in the background. The root issue is that ASV's conda plugin has this fixed flow:
```
1. Read conda_environment_file
2. Run conda env create -f <that_file>
3. Run build_command in the created environment
```
There's no hook to customize step 2,  so a file is needed that `conda env create` can actually read.
The solution to this is to generate a env file explicitly that might me the best of both. It now includes  metadata fields like URL, MD5 hash, build string and platform.

So while it's not *every* field from conda-lock.yml (like dependencies between packages), it captures the critical reproducibility info **exactly which binary to download and how to verify it**.

### Testing Results
Ran the benchmarks locally and they work! 
```
Environment created successfully
Python 3.13.3 installed (489 packages)
TARDIS installed without errors
Benchmarks running:
   - time_compton_opacity_calculation: 188±0ms ✓
   - time_pair_creation_opacity_calculation: 152±0ms ✓
   - time_photoabsorption_opacity_calculation: 142±0ms ✓
```
(Some benchmarks failed locally because I don't have the atomic data files, but that's expected I think, the env.yml itself works perfectly)

I have created `update_script.sh` to automate the env file creation as per your suggestion!

##### One Thing I Noticed 

The benchmark workflow currently downloads `env.yml` from `tardisbase`:

```yaml
- name: Download lockfile for benchmarks
  run: wget -q https://raw.githubusercontent.com/tardis-sn/tardisbase/refs/heads/master/env.yml
```

This means **my updated env.yml won't be tested in CI** unless  env.yml in `tardisbase` is also updated with explicit format or workflow is changed to use the repo's env.yml

I didn't want to change the workflow in my PR without discussing with you first, since it affects the whole project. Would you prefer:
- **A:** Update env.yml in tardisbase with explicit format
- **B:** Change workflow to use repo's env.yml (I can add that to this PR)
- **C:** Something else?

Let me know what you think :)

UPDATE_SCRIPT.sh
```diff
@@ -0,0 +1,15 @@
+# This script shows how to update env.yml from conda-lock.yml
+# Run this whenever conda-lock.yml is updated
+
+set -e
+
+echo "Generating env.yml from conda-lock.yml using explicit format..."
+/home/bhoomi/miniforge3/bin/conda-lock render --kind explicit
```
atharva-2001: https://github.com/conda/conda-lock?tab=readme-ov-file#pre-10-compatible-usage-explicit-per-platform-locks

why do we need the explicit flag?

UPDATE_SCRIPT.sh
```diff
@@ -0,0 +1,15 @@
...
+
+mv conda-linux-64.lock env.yml
```
atharva-2001: this just renames? so its not actually an environment yml? why? plus its platform specific

UPDATE_SCRIPT.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo " env.yml updated with full package URLs and hashes"
+
+rm -f conda-osx-64.lock conda-osx-arm64.lock conda-linux-64.lock.yml conda-osx-64.lock.yml conda-osx-arm64.lock.yml
```
atharva-2001: why?

generate_benchmark_env.py
atharva-2001, line 1: > This will create/update env.yml in the repository root.

we dont want to update the environment multiple times. we use lockfiles just for this reason. what do you think would be a better way? 

also, what purpose does this file serve then?

update_script.sh
```diff
@@ -0,0 +1,15 @@
+# This script shows how to update env.yml from conda-lock.yml
+# Run this whenever conda-lock.yml is updated
+
+set -e
+
+echo "Generating env.yml from conda-lock.yml"
+conda-lock render --kind explicit
+
+mv conda-linux-64.lock env.yml
```
atharva-2001: similar thing I guess, there are two files now

UPDATE_SCRIPT.sh
BhoomiAgrawal12: I look on the documentation, mentioning explicit is redundant as it is default now.

UPDATE_SCRIPT.sh
BhoomiAgrawal12: hey @atharva-2001, explicit is causing the platform specific issue , I generated it specifically for linux because our CI workflows for ASV benchmarks use linux platform.
I saw a better choice of using `--kind env` as flag which generates the env file without any platform specific url that makes it more readable.

while the `--kind env` format is more portable than explicit URLs, the **packages themselves are still platform specific**. I mean:
- Using `-p linux-64` extracts Linux-specific packages
- macOS would need different packages

Checking on tardis `.github/workflows/benchmarks.yml` it runs on ubuntu-latest. So while it's technically "platform-specific packages", this is appropriate because **ASV only runs on Linux in your CI**. if you later want Mac support, we can generate multiple files

### What  happening earlier was:
`conda-lock render --kind explicit` reads conda-lock.yml which contains all platforms and generate seperate files for each platform and then I was just renaming the linux one because our CI workflow use that only.

### Noting Point:
In latest commit using `-p linux-64` gives us a single file that works for CI but it's not truly "cross-platform" , Mac developers would need a different file although the `--kind env` format is PORTABLE (conda can read it anywhere), but the content is platform specific itself in conda-lock file.

UPDATE_SCRIPT.sh
BhoomiAgrawal12: Here in the new commit, with `--kind env`, we don't get multiple platform files anymore, just one env.yml and thus no renaming and cleanup needed.
If you later want to support local Mac benchmarking, we could generate all three platform files, but currently I think that's not needed.

generate_benchmark_env.py
BhoomiAgrawal12: I think there is a clarification needed, what I understood from the issue #3144, it points out that ASV references `env.yml` in `asv.conf.json` .But that file doesn't exist in the tardis repo and currently CI downloads it from tardisbase (different repo). 
tardisbase already provides env.yml for the entire ecosystem. However, my PR generates env.yml specifically from tardis's own conda-lock.yml to ensure benchmark dependencies match tardis's exact dependencies.

I think using env.yml from tardisbase also creates confusion for developers trying to run benchmarks locally. But would like to know your call.
There are two options:
 1. Either we use tardisbase env only as single source of truth but in that case developers locally will face problem to run benchmarks. In that case I will close the pr.
2. Generate local env.yml from tardis's own conda-lock.yml as done in this pr.

Please let me know @atharva-2001 what aligns with TARDIS ecosystem's architecture!

connor-mcclellan:
Hello @BhoomiAgrawal12, are you applying to Google Summer of Code?

BhoomiAgrawal12:
yes @connor-mcclellan , I am interested in applying for GSOC.

connor-mcclellan:
Thank you for your contribution - this isn't what we're looking for. If you have any questions please refer to our ideas page and LLM usage guidelines.

Issue #3381, opened 2026-01-05
Skip stardis tests in our development pipeline, or make an expected failur


jvshields:
We don't really have time to go actually make stardis work, but the stardis tests are failing and we expect them to at this point, and it would be better for tardis development to change this behavior so we can more easily see if a given pr actually breaks tardis (the tardis tests or benchmarks or docs build).

PR #3395, opened 2026-01-25
stardis-tests is no longer required to pass CI

SS-9098:
### :pencil: Description

Changed `stardis-test.yml` so that it is not required to pass Github actions CI. This allows better review of PRs by focusing on other tests as this test was always expected to fail. Rather then removing the test altogether, the test still exists but the workflow continues even on failing this.
Note that this change only effects github actions workflow and not local unit tests. 

Fixes: #3381 

- [ ] Testing pipeline
- [x] Other method - If CI passes even after failing this test, then the change works.


atharva-2001:
I don't think this is a preferred solution. This is something we're working on internally.
Please work on other issues. Thanks for contributing!

PR #3400, opened 2026-01-27
[GSOC] Added SDECPlotter support to LineInfoWidget for displaying additional spectrum analysis data

### LINKED ISSUE #2023: Integrate SDEC plot in Line Info Widget
The Line Info Widget should have an option to show SDEC plot besides real and virtual spectra (maybe by using a button group or function parameter). This will make analyses more easier.

Besides, it should allow to overplot observed spectrum on modelled spectrum (an in-built feature in SDEC plot), so the integration should take that into consideration.

SS-9098:
### :pencil: Description

**Type:** : :rocket: `feature`

The Line Info Widget has been extended to optionally overlay data from the SDEC Plot, providing richer spectral context and improved analysis capabilities.
- Virtual or real spectrum overlay
- Blackbody photosphere curve
- No interaction packets
- Electron scatter only packets
- observed spectrum

Each plot component can be independently toggled using checkboxes in the widget UI.

Usage Details

- A new parameter,` show_sdec=True`, enables the SDEC Plot overlay on the Line Info Widget.
SDEC plots are generated using the `SDECPlotter`class.

- If `show_sdec` is not provided (or set to `False`), the widget behaves as before.

- The observed spectrum can be displayed by passing the `observed_spectrum` parameter. This accepts the same arguments as the corresponding parameter in SDECPlotter and renders an equivalent plot.

- The packet mode for the SDEC plot can be set to real or virtual via `sdec_packets_mode`.

Example Usage:

https://github.com/user-attachments/assets/82e1fad8-2f90-45c0-927b-5eb2fe57beb0

Resolves #2023

- [ ] Testing pipeline
- [x] Other method (Check its functionality through local Jupyter notebook server)

### AI and LLM Usage

Claude Opus 4.6 was used to help draft the plotting functions.
I confirm that I fully understand and can explain all AI-assisted contribution.


tardis/visualization/widgets/line_info.py
```diff
@@ -422,56 +445,78 @@ def get_middle_half_edges(arr):
...
+        )
+
+        self.line_renderers["photosphere"] = p.line(
+            self.sdec_plotter.plot_wavelength.value,
+            self.sdec_plotter.photosphere_luminosity.value,
+            legend_label="Blackbody Photosphere",
+            color="orange",
+            line_dash="dotted",
+        )
+
```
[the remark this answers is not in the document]
SS-9098: added parameter to set packets mode to real or virtual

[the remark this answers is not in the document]
SS-9098: Added a helper function in `sdec_plot.py`. private methods are no longer used in `line_info.py`

tardis/visualization/widgets/line_info.py
```diff
@@ -598,6 +643,24 @@ def _update_last_line_counts(self, species, filter_mode, group_mode):
...
+    def _on_toggle_line(self, event):
+        """Toggle visibility of spectrum lines based on widget state."""
+        toggle_map = {
+            "Real Packets": "real_packets",
+            "Virtual Packets": "virtual_packets",
+            "Virtual Spectrum": "virtual_spectrum",
+            "Blackbody Photosphere": "photosphere",
+            "No Interaction": "no_interaction",
+            "Electron Scatter Only": "electron_scatter",
+            "Observed Spectrum": "observed",
```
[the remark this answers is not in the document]
SS-9098: Added option to overlay observed spectrum

andrewfullard:
The real value of the SDEC-style plot is to display the elements that are contributing to the  spectrum. See the second plot here: https://tardis-sn.github.io/tardis/analyzing_tardis/visualization/how_to_sdec_plot.html

SS-9098:
I have added functionality to show species contribution too when sdec is shown.


image

andrewfullard:
That doesn't look like it aligns with the real spectrum, which it should.

tardis/visualization/widgets/line_info.py
```diff
@@ -507,6 +600,103 @@ def plot_spectrum(
 
         return pn.pane.Bokeh(p)
 
+    def _plot_species_contributions(self, p):
+        import matplotlib.pyplot as plt
```
andrewfullard: Why is the import inside the function?

SS-9098:
> That doesn't look like it aligns with the real spectrum, which it should.

This seems to happen because I plotted the species contribution independently rather than stacking it, which in hindsight should have been obvious after seeing how it was implemented in sdec_plot.py.

After stacking and testing I got this result:

image

I will clean up the code a bit and commit this change

tardis/visualization/widgets/line_info.py
SS-9098: Oh, I left that by accident while testing cmap.

Refactored now!

andrewfullard (dismissed):

tardis/visualization/widgets/line_info.py
```diff
@@ -73,9 +83,50 @@ def __init__(
...
+        self.observed_spectrum = observed_spectrum
+
+        # Store renderers for toggle functionality
+        self.line_renderers = {}
+
+        self.element_checkboxes = {}
+        self.checkbox_real_packets = pn.widgets.Checkbox(name="Real Packets", value=True)
+        self.checkbox_virtual_packets = pn.widgets.Checkbox(name="Virtual Packets", value=True)
+        if self.show_sdec:
+            self.checkbox_sdec_spectrum = pn.widgets.Checkbox(name="SDEC Spectrum", value=True)
```
jaladh-singhal: The "SDEC spectrum" in theory is same as "Real Packets" (in its `mode="real"`) and "Virtual Packets" (in its `mode="virtual"`) - you can also notice that by quick visual inspection. They should be mutually exclusive instead of we duplicating/overplotting them.

tardis/visualization/widgets/line_info.py
```diff
@@ -73,9 +83,50 @@ def __init__(
...
+            self.checkbox_photosphere = pn.widgets.Checkbox(name="Blackbody Photosphere", value=True)
+            self.checkbox_no_interaction = pn.widgets.Checkbox(name="No Interaction", value=True)
+            self.checkbox_electron_scatter = pn.widgets.Checkbox(name="Electron Scatter Only", value=True)
+        if self.observed_spectrum is not None:
+            self.checkbox_observed_spectrum = pn.widgets.Checkbox(name="Observed Spectrum", value=True)
```
jaladh-singhal: I don't think we want all these checkboxes - we can just expose the API of SDEC Plotter in `show_sdec` mode.

tardis/visualization/widgets/line_info.py
```diff
@@ -422,56 +486,90 @@ def get_middle_half_edges(arr):
...
-        virt_wavelength : astropy.Quantity
-            Wavelength values of a virtual spectrum, having unit of Angstrom
-        virt_luminosity_density_lambda : astropy.Quantity
-            Luminosity density lambda values of a virtual spectrum, having unit
-            of (erg/s)/Angstrom
-
-        Returns
-        -------
-        plotly.graph_objects.FigureWidget
+        Produce a Bokeh figure by plotting the spectrum of model with SDEC data.
```
jaladh-singhal: Ah! Thanks for updating this docstring - it was originally designed to be a plotly figure but it chaned to bokeh figure recently.

tardis/visualization/widgets/line_info.py
```diff
@@ -422,56 +486,90 @@ def get_middle_half_edges(arr):
...
+            observed_flux = self.observed_spectrum[1]
+            self.line_renderers["observed"] = p.line(
+                observed_wavelength.value,
+                observed_flux.value,
+                legend_label="Observed Spectrum",
+                color="black",
+            )
+
+        if self.show_sdec and self.sdec_plotter is not None:
+            self.sdec_plotter.prepare_plot_data(
```
jaladh-singhal: Instead of constructing SDEC plot in bokeh/Panel here, we should do that in `SDECPlotter` and call it here. Then `plot_spectrum` gets invoked when `show_sdec=False` and `plot_sdec_spectrum` will invoke `SDECPlotter.generate_plot_bokeh` when `show_sdec=True`.

tardis/visualization/widgets/line_info.py
```diff
@@ -507,6 +605,153 @@ def plot_spectrum(
 
         return pn.pane.Bokeh(p)
 
+    def _plot_species_contributions(self, p):
```
jaladh-singhal: so will this and it's related functions move to `SDECPlotter` - the goal is separation of concerns: this SDEC-style plotting mechanics is SDECPlotter's concern, not of LineInfoWidget.

jaladh-singhal (changes requested):
This is a great start @SS-9098 but the design of code is not quite right.

I should have added more details when creating #3400. Also we switched LineInfoWidget to use bokeh/panel from plotly since this issue was opened so the design has more gaps than before. As of now, the following makes most sense to me:

Existing behavior: `LineInfoWidget.from_simulation(sim)` -> creates the widget as is, the figure is a bokeh/panel plot of real+virtual spectra.

New additive behavior: `LineInfoWidget.from_simulation(sim, show_sdec=True, sdec_kwargs=dict(packets_mode='virtual', observed_spectrum=..., ...))` -> creates the widget with a bokeh/panel figure of SDEC plot, where `sdec_kwargs` are simply arguments accepted by `SDECPlotter.generate_plot_bk()` method.

---

This design prevents you/us from maintaining a duplicate shadow API of SDEC plot in LineInfoWidget and simplifies the use for end-user since they don't need to learn new set of arguments for plotting a SDEC plot in Line Info widget.

You can see the plotting options available for SDECPlotter in [generate_plot_mpl](https://tardis-sn.github.io/tardis/analyzing_tardis/visualization/how_to_sdec_plot.html#Additional-plotting-options) and [generate_plot_ply](https://tardis-sn.github.io/tardis/analyzing_tardis/visualization/how_to_sdec_plot.html#id3) and how symmetric they are except a few library specific arguments. You can add bokeh version to it as `generate_plot_bk` and expose all of its argument into `sdec_kwargs` in `LineInfoWidget`, which will take effect only when `show_sdec=True`.

When you figure out the implementation details to split the code, a good mental model is: `SDECPlotter` should own the bokeh figure content whereas the `LineInfoWidget` should own the Panel widget composition.

SS-9098:
I have restructured the code, and removed toggles.

image

andrewfullard (dismissed):

tardis/visualization/widgets/line_info.py
```diff
@@ -471,41 +543,7 @@ def plot_spectrum(
...
-        ]
-        self._wavelength_data = (
-            wavelength  # Store wavelength data for callback access
-        )
-
-        # Connect selection callback
-        source.selected.on_change("indices", self._selection_callback)
-
-        return pn.pane.Bokeh(p)
+        return self._setup_selection(p, wavelength, luminosity_density_lambda)
```
jaladh-singhal, line 546: Good modularization.

tardis/visualization/widgets/line_info.py
jaladh-singhal: This return type needs updating too.

tardis/visualization/tools/sdec_plot.py
```diff
@@ -1076,6 +1076,195 @@ def _make_colorbar_colors(self):
...
+        self.fig.varea(x=self.plot_wavelength.value, y1=lower_level, y2=upper_level, fill_color="#4C4C4C", legend_label="No interaction")
+
+        lower_level = upper_level
+        upper_level = lower_level + self.emission_luminosities_df[("escatter", "")].to_numpy()
+        self.fig.varea(x=self.plot_wavelength.value, y1=lower_level, y2=upper_level, fill_color="#8F8F8F", legend_label="Electron Scatter Only")
+
+        if "other" in self.emission_luminosities_df.keys():
+            lower_level = upper_level
+            upper_level = lower_level + self.emission_luminosities_df[("other", "")].to_numpy()
+            self.fig.varea(x=self.plot_wavelength.value, y1=lower_level, y2=upper_level, fill_color="#C2C2C2", legend_label="Other elements")
```
jaladh-singhal: [Nitpick/Optional] Instead of hardcoding colors and labels, you can define them in a dict and use in mpl, bk, ply functions - it will ensure consistency if someone decides to change them.

jaladh-singhal (dismissed):
Thanks for addressing my comments @SS-9098 - the code looks clean and much more modular now (hence easy to maintain).

The screenshot of widget looks great but can you add this exact example to relevant file generating https://tardis-sn.github.io/tardis/analyzing_tardis/visualization/how_to_generating_widgets.html#Line-Info-Widget? You can create a subheading in that section to indicate "With SDEC Plot" or something better.

tardis/visualization/tools/sdec_plot.py
SS-9098: I noticed there already was a dictionary called `_predefined_traces` which had the color and labels, so I have used that.
It was being used only in ply functions. I also updated the mpl functions to use it

jaladh-singhal (approved):
Thanks for the updates - looks good to me!

jvshields (approved):

Issue #2917, opened 2024-12-30
Cannot disable convergence plots


atharva-2001:
**Describe the bug**

Cannot disable convergence plots in Standard Workflows notebook

```py
UnboundLocalError                         Traceback (most recent call last)
Cell In[4], line 1
----> 1 workflow.run()

File ~/tardis-main/tardis/tardis/workflows/standard_tardis_workflow.py:231, in StandardTARDISWorkflow.run(self)
    222 opacity_states = self.solve_opacity()
    224 transport_state, virtual_packet_energies = self.solve_montecarlo(
    225     opacity_states, self.real_packet_count
    226 )
    228 (
    229     estimated_values,
    230     estimated_radfield_properties,
--> 231 ) = self.get_convergence_estimates(transport_state)
    233 if self.convergence_plots is not None:
    234     self.convergence_plots.update()

File ~/tardis-main/tardis/tardis/workflows/standard_tardis_workflow.py:169, in StandardTARDISWorkflow.get_convergence_estimates(self, transport_state)
    159 if self.convergence_plots is not None:
    160     plot_data = {
    161         "t_inner": [self.simulation_state.t_inner.value, "value"],
    162         "t_rad": [self.simulation_state.t_radiative, "iterable"],
   (...)
    167         "Requested": [self.luminosity_requested.value, "value"],
    168     }
--> 169 self.update_convergence_plot_data(plot_data)
    171 logger.info(
    172     f"\n\tLuminosity emitted   = {emitted_luminosity:.3e}\n"
    173     f"\tLuminosity absorbed  = {absorbed_luminosity:.3e}\n"
    174     f"\tLuminosity requested = {self.luminosity_requested:.3e}\n"
    175 )
    177 self.log_plasma_state(
    178     self.simulation_state.t_radiative,
    179     self.simulation_state.dilution_factor,
   (...)
    183     estimated_t_inner,
    184 )

UnboundLocalError: cannot access local variable 'plot_data' where it is not associated with a value
```

**To Reproduce**

**Screenshots**

**System**
 - OS:
    - [ ] GNU/Linux
    - [ ] macOS

 - Environment (`conda list`):

**Additional context**

karthik11135:
We can only update the convergence plots when we have `plot_data`.  So the function has to be called inside the if statement. (line 169)

atharva-2001:
That's right. I will fix this issue on another pull request. Opening a pull request just to solve this issue wouldn't be nice.

[...]

PR #3409, opened 2026-02-04
Fix UnboundLocalError when convergence plots are disabled (#2917)

PritamHazra2708:
Fix: Disable convergence plots safely in StandardTARDISWorkflow

When convergence plots are disabled (self.convergence_plots = None), StandardTARDISWorkflow.run() still calls get_convergence_estimates() during the final iteration (around line ~265) inside a plot-only block.

get_convergence_estimates() performs physics computations and logging in addition to plotting, and its return values are discarded in this context. This leads to redundant computations and inconsistent behavior when convergence plots are disabled.

This PR avoids calling get_convergence_estimates() solely for plotting, ensuring convergence plots can be disabled cleanly without unintended side effects.


connor-mcclellan:
Hi @PritamHazra2708 , are you applying to Google Summer of Code?

PritamHazra2708:
Yes Sir

PritamHazra2708:
@connor-mcclellan Yes Sir

connor-mcclellan:
Please address the bot comments and follow the PR description template

tardis/workflows/standard_tardis_workflow.py
```diff
@@ -262,7 +262,7 @@ def run(self):
         )
         self.reshape_plasma_state_store(self.completed_iterations)
         if self.convergence_plots is not None:
-            self.get_convergence_estimates()
+            _, _ = self.get_convergence_estimates()
```
connor-mcclellan, line 266: The function is still being called - its return values are now just being assigned to an ignored variable. This may prevent an error assigning the return values to ignored variables, but the real question is why is it there in the first place. Investigate how this function is actually used in the code and what it should be doing here, if anything.

PR #3410, opened 2026-02-06
Excise transport_state initializiation

wkerzendorf:
As described based on CSVY PR for now

### DIFF
--- a/tardis/transport/montecarlo/mc_transport_state/base.py
+++ b/tardis/transport/montecarlo/mc_transport_state/base.py
@@ -1,0 +1,371 @@
+import warnings
+
+import pandas as pd
+from astropy import units as u
+
+from tardis.io.hdf_writer_mixin import HDFWriterMixin
+from tardis.model.geometry.radial1d import (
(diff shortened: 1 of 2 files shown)


tardis/transport/montecarlo/mc_transport_state/base.py
```diff
@@ -45,19 +115,41 @@ class MonteCarloTransportState(HDFWriterMixin):
...
-        tracker_full_df=None,
-        tracker_last_interaction_df=None,
-        vpacket_tracker=None,
-    ):
+        packet_collection: PacketCollection,
+        radfield_mc_estimators: RadiationFieldMCEstimators,
+        geometry_state: NumbaRadial1DGeometry,
+        opacity_state: OpacityStateNumba,
+        time_explosion: u.Quantity,
+        tracker_full_df: pd.DataFrame | None = None,
```
connor-mcclellan, line 121: Note - this will confuse LSPs since it has two potential types. We should look into how to avoid this and do it the right way

PR #3436, opened 2026-02-17
Add iip opacity state and use absorbing markov chain transition probabilities

jvshields:
This PR adds the functionality to do two step packet redistribution sampling, using interaction handler transition probabilities that have been processed by an absorbing markov chain methodology. When the interaction handler activates, it takes a single step to a new place, and then chooses a deactivation transition. This includes new opacity states for iip because two transition arrays need to be passed in to the interaction handler instead of one. 

UPDATE: 
This no longer adds all of the opacity state construction because that's not quite ready to go yet. It currently only provides enough for me to construct my own iip opacity state and run packet transport with it. As a result, it breaks the workflow for now.

### DIFF
--- a/tardis/transport/montecarlo/interaction_event_callers.py
+++ b/tardis/transport/montecarlo/interaction_event_callers.py
@@ -214,6 +214,10 @@ def line_scatter_event(
             enable_full_relativity,
         )
     else:  # includes both macro atom and downbranch - encoded in the transition probabilities
+        comov_nu = r_packet.nu * old_doppler_factor  # Is this necessary?
+        r_packet.nu = (
+            comov_nu * inverse_new_doppler_factor
+        )  # NOTE - this should get overwritten by macro_atom_event, but makes tests fail potentially from different rng
         activation_level_id = opacity_state.line2macro_level_upper[
             r_packet.next_line_id
         ]
--- a/tardis/tests/test_tardis_full_formal_integral.py
+++ b/tardis/tests/test_tardis_full_formal_integral.py
@@ -88,7 +88,7 @@ class TestTransportSimpleFormalIntegral:
         expected = regression_data.sync_ndarray(luminosity.cgs.value)
         expected = u.Quantity(expected, "erg /s")
         assert_quantity_allclose(
-            luminosity, expected, rtol=1.1e-11, atol=0 * u.erg / u.s
+            luminosity, expected, rtol=1e-11, atol=0 * u.erg / u.s
         )
 
     def test_spectrum_integrated(self, simulation, request):


tardis/transport/montecarlo/configuration/constants.py
```diff
@@ -1,7 +1,7 @@
 from tardis import constants as const
 
 SIGMA_THOMSON = const.sigma_T.to("cm^2").value
-CLOSE_LINE_THRESHOLD = 1e-14
+CLOSE_LINE_THRESHOLD = 1e-7
```
andrewfullard: This is a pretty major change! Why 1e-7?

tardis/transport/montecarlo/interaction_event_callers.py
```diff
@@ -19,31 +21,38 @@
...
 def macro_atom_event(
-    destination_level_idx,
-    r_packet,
-    time_explosion,
-    opacity_state,
-    enable_full_relativity,
+    destination_level_idx: int,
+    r_packet: RPacket,
+    time_explosion: float,
+    opacity_state,  # this currently either accepts the normal numba or the iip numba opacity state
```
andrewfullard: `: OpacityStateNumba | OpacityStateNumbaIIP`

tardis/transport/montecarlo/macro_atom.py
```diff
@@ -79,3 +87,96 @@ def macro_atom_interaction(
...
+                ]
+            ):
+                absorbing_state_probability += state_probability
+
+                if absorbing_state_probability > probability_event:
+                    absorbing_activation_level_idx = to_state_index
+                    break
+            else:
+                raise MacroAtomError(
+                    "MacroAtom failed to select an absorbing state. The absorbing Markov"  # noqa: S608
```
andrewfullard: Why the `noqa` comment?

jvshields: It signals ruff to ignore the comment. Otherwise, ruff complains. 
"Possible SQL injection vector through string-based query constructionRuff[S608](https://docs.astral.sh/ruff/rules/hardcoded-sql-expression)"
To me this sounded like something we didn't need to worry about. Am I supposed to avoid using f strings in error statements?

tardis/transport/montecarlo/configuration/constants.py
jvshields: To match ctardis. It has made ~no change on the spectrum but admittedly that's with pure hydrogen which has very separated lines. Happy to revert this one though. It was just along the lines of debugging discrepancies.

tardis/transport/montecarlo/macro_atom.py
jvshields: I guess the worry is that somebody could build their own macro atom data, and then inject some nefarious code in the error statement.

tardis/transport/montecarlo/configuration/constants.py
andrewfullard: It should be something we check with regression data comparisons I think,

tardis/transport/montecarlo/macro_atom.py
andrewfullard: I suppose so. Weird.

tardis/transport/montecarlo/configuration/constants.py
jvshields: Tests fail with the new threshold, so I did just revert it.

PR #3444, opened 2026-02-18
GSoC :fix: clip negative gamma_corr to zero in CorrPhotoIonRateCoeff

### LINKED ISSUE #3355: Continuum transitions don't work with current tardis plasma because of CorrectedPhotoIonizationRateCoefficient
When trying to run the tardis plasma with continuum transitions the simulation fails after the first iteration when it tries to incorporate estimators. It fails with the following error
PlasmaException: Negative values in CorrPhotoIonRateCoeff. Try raising the number of montecarlo packets.

which traces back to https://github.com/tardis-sn/tardis/blob/6025bed1d586c2a5c48571c78597dfd61f2ab9be/tardis/plasma/properties/continuum_processes/rates.py#L315

In the short term we are ignoring this. When we clean up the IIp workflow, this will be one of the things we need to fix or understand.

sakirr05:
### :pencil: Description

**Type:** :beetle: bugfix

Closes #3355

Continuum transitions were broken since `CorrPhotoIonRateCoeff` was introduced in #3318. Monte Carlo estimator noise can drive `gamma_corr` (= γ - α_stim × n_k/n_i × n_e) slightly below zero, which previously triggered a hard `PlasmaException` and stopped the simulation.

The corrected photo-ionization rate is physically bounded below by zero, so slightly-negative values are a statistical artifact of limited packet counts, not a physical inconsistency. The fix replaces the hard exception with a `logger.warning` and clips `gamma_corr` to zero with `DataFrame.clip(lower=0)`.

Also removes the now-unused `PlasmaException` import.

### :pushpin: Resources

- Source under fix: `tardis/plasma/properties/continuum_processes/rates.py`
- New regression tests: `tardis/plasma/properties/continuum_processes/tests/test_corr_photo_ion_rate_coeff.py`

- [x] Testing pipeline
- [ ] Other method (describe)

All 5 unit tests pass locally:
- `test_positive_values_unchanged`
- `test_slightly_negative_values_clipped_to_zero`
- `test_no_plasma_exception_raised`
- `test_warning_emitted_for_negative_values`
- `test_zero_inputs_return_zero`

### AI Disclosure

I used AI tooling to investigate the root cause (tracing the estimator pipeline from MC transport through `CorrPhotoIonRateCoeff`), draft the fix, and generate the regression tests. The fix logic and test assertions were verified manually against the source code.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -309,3 +309,5 @@ Riddhi Gangbhoj <riddhigangbhoj76@gmail.com>
 Riddhi Gangbhoj <riddhigangbhoj76@gmail.com> riddhigangbhoj <riddhigangbhoj76@gmail.com>
 
 Connor McClellan <b.connor.mcc@gmail.com>
+
+Sakir Ahmed <sakirahmed75531@gmail.com>
--- a/.orcid.csv
+++ b/.orcid.csv
@@ -28,3 +28,4 @@ erin.visser1@gmail.com,0009-0001-8470-275X
 lujingeve158@gmail.com,0000-0002-3900-1452
 b.connor.mcc@gmail.com,0000-0002-6040-8281
 aaryandadu5157@gmail.com,0009-0005-3670-8777
+sakirahmed75531@gmail.com,0009-0009-7323-2076


andrewfullard:
Thanks for your contribution. This is not an appropriate fix for this issue. Furthermore, we are unlikely to be using this class in the future due to changes in our plasma architecture that are being worked on.

PR #3458, opened 2026-02-23
fix: broken benchmark ci

Palakb273:
### :pencil: Description

**Type:** :beetle: `bugfix` | :vertical_traffic_light: `testing`

This pull request fixes a Python 3.14 compatibility bug in atomic HDF loading that caused benchmark setup failures.

In `AtomData.from_hdf`, loading tables with `store.select(name)` could fail with:

`TypeError: a bytes-like object is required, not 'str'`

The root cause is a metadata type mismatch in some HDF nodes where `pandas_type` is stored as `bytes` instead of `str`, which breaks pandas storer detection on newer runtimes.

#### Changes included

1. Added a metadata normalization step in `tardis/io/atom_data/base.py` before selecting each HDF key.
2. Introduced `_normalize_hdf_pandas_type(store, key)` to decode `bytes` values of `pandas_type` into ASCII strings.
3. Added a targeted regression test in `tardis/io/tests/test_atomic.py`:
   `test_normalize_hdf_pandas_type_decodes_bytes`.

#### Why this approach

The fix is minimal and local to HDF read-time metadata handling. It preserves existing behavior for valid files while making the loader robust to mixed metadata encodings seen in Python 3.14 environments.

Fixes #<issue-number>

### :pushpin: Resources

1. Error context from ASV benchmark setup (`benchmarks/opacities_opacity_state.py`) showing failure path via `AtomData.from_hdf`.
2. Updated files:
   `tardis/io/atom_data/base.py`
   `tardis/io/tests/test_atomic.py`
3. Relevant stack trace entry:
   `pandas/io/pytables.py::_create_storer` around `"table" not in pt`.

- [ ] Testing pipeline
- [x] Other method (describe)

Manual/local test performed:

`pytest -q tardis/io/tests/test_atomic.py -k normalize_hdf_pandas_type_decodes_bytes`

Result: `1 passed`

### DIFF
--- a/tardis/io/tests/test_atomic.py
+++ b/tardis/io/tests/test_atomic.py
@@ -95,3 +95,86 @@ def test_normalize_hdf_pandas_type_decodes_bytes():
 
     assert isinstance(store.get_node("/atom_data")._v_attrs.pandas_type, str)
     assert store.get_node("/atom_data")._v_attrs.pandas_type == "frame_table"
+
+
+def test_normalize_hdf_pandas_type_string_unchanged():
+    class DummyAttrs:
+        pandas_type = "frame_table"
+
+    class DummyNode:
+        _v_attrs = DummyAttrs()
+
+    class DummyStore:
+        def __init__(self):
+            self.node = DummyNode()
+
+        def __contains__(self, key):
+            return key == "atom_data"
+
+        def get_node(self, key):
+            assert key == "/atom_data"
+            return self.node
+
+    store = DummyStore()
+
+    AtomData._normalize_hdf_pandas_type(store, "atom_data")
+
+    assert isinstance(store.get_node("/atom_data")._v_attrs.pandas_type, str)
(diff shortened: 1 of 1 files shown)


tardis/io/atom_data/base.py
```diff
@@ -267,6 +268,17 @@ def from_hdf(cls, fname=None):
 
         return atom_data
 
+    @staticmethod
+    def _normalize_hdf_pandas_type(store, key):
+        """Normalize bytes-valued pandas_type metadata for newer Python runtimes."""
+        if key not in store:
+            return
+
```
atharva-2001, line 276: can you run CI locally and test it and share logs.

atharva-2001: In AtomData.from_hdf, loading tables with store.select(name) could fail with:
-

is this pr created to fix a bug that doesn't even exist yet?

connor-mcclellan:
This  PR does not adhere to the guidelines and you have not addressed reviewer comments. Closing.

PR #3468, opened 2026-02-27
fix FileNotFoundError in regression tests

### LINKED ISSUE #3354: FileNotFoundError in regression tests
**Describe the bug**

While running the full test suite with the regression data, two equilibrium plasma tests fail due to filename mismatch between the expected files and the files present in the tardis-regression-data repository
The failing tests are: tardis/plasma/equilibrium/tests/test_level_populations.py and tardis/plasma/equilibrium/tests/test_rate_matrix.py

possible fix: The FileNotFoundError is due to file name mismatch **test_rate_matrix_solver__collisional_rate_solver0-radiative_transitions0__.h5** is the file name in data while after renaming it to **test_rate_matrix_solver__radiative_transitions0-collisional_rate_solver0__.h5** the test passes, similar result for test_solve__collisional_rate_solver0-radiative_transitions0__.h5

**To Reproduce**
1. Clone tardis-regression-data `git clone https://github.com/tardis-sn/tardis-regression-data`
2. `git lfs install`
3. `git lfs pull`
4. run the test suite

**Screenshots**

Image

---

Image

(After Renaming)

**System**
 - OS:
    - [x] GNU/Linux
    - [ ] macOS

**Additional context**

Aaryan-Dadu:
### :pencil: Description

**Type:** :beetle: `bugfix`

While running the full test suite with the regression data, two equilibrium plasma tests fail due to filename mismatch between the expected files and the files present in the tardis-regression-data repository
- In test_level_populations.py::test_solve the file name is manually specified
- In test_rate_matrix.py::test_rate_matrix_solver the file name for regression data is manually specifed

It happened due to change in order of parameters in the auto generated file name by regression_data.sync_dataframe()

**Before:**
 Screenshot From 2026-02-27 12-38-12

**Now:**
Screenshot From 2026-02-27 14-42-51

Resolves #3354 

### :pushpin: Resources

- [x] Testing pipeline
- [ ] Other method (describe)

### AI Usage
I have read the checklist and the [AI usage policy](https://tardis-sn.github.io/summer_of_code/ai_usage_policy)
No AI or LLM tools were used in any part of the contribution.

### DIFF
--- a/tardis/plasma/equilibrium/tests/test_level_populations.py
+++ b/tardis/plasma/equilibrium/tests/test_level_populations.py
@@ -68,7 +68,11 @@ class TestLevelPopulationSolver:
 
     def test_solve(self, regression_data):
         """Test the solve method."""
-        expected_populations = False
         result = self.solver.solve()
-        expected_populations = regression_data.sync_dataframe(result)
+
+        # The order of parameters in the auto-generated filename has changed.
+        # We manually specify the path to the existing regression file.
+        expected_fname = "test_solve__collisional_rate_solver0-radiative_transitions0__.h5"
+        expected_fpath = regression_data.fpath.parent / expected_fname
+        expected_populations = pd.read_hdf(expected_fpath)
         pdt.assert_frame_equal(result, expected_populations, atol=0, rtol=1e-15)
--- a/tardis/plasma/equilibrium/tests/test_rate_matrix.py
+++ b/tardis/plasma/equilibrium/tests/test_rate_matrix.py
@@ -33,7 +33,11 @@ def test_rate_matrix_solver(
 
     actual = rate_matrix_solver.solve(rad_field, electron_dist)
 
-    expected = regression_data.sync_dataframe(actual)
+    # The order of parameters in the auto-generated filename has changed.
+    # We manually specify the path to the existing regression file.
+    expected_fname = "test_rate_matrix_solver__collisional_rate_solver0-radiative_transitions0__.h5"
+    expected_fpath = regression_data.fpath.parent / expected_fname
+    expected = pd.read_hdf(expected_fpath)
 
     pdt.assert_frame_equal(actual, expected, atol=0, rtol=1e-15)
 


tardis/plasma/equilibrium/tests/test_level_populations.py
```diff
@@ -68,7 +68,11 @@ def test_calculate_level_population_zeros(self):
 
     def test_solve(self, regression_data):
         """Test the solve method."""
-        expected_populations = False
         result = self.solver.solve()
-        expected_populations = regression_data.sync_dataframe(result)
+
+        # The order of parameters in the auto-generated filename has changed.
```
connor-mcclellan: I think our preference would be to fix the auto-generated filename rather than hardcoding the path. We still want to use `sync_dataframe`, so look into what changes would be necessary to make it consistent with elsewhere in the code. We can rename regression data files and change how they're named when they're produced, if that's the issue.

Aaryan-Dadu: Hi @connor-mcclellan ,
I looked about it, the thing is that when pytest parameterizes a test, it creates names like `test_abc[xyz_def]` the part inside `[]` comes from the parameters used in that test. The order of the pieces inside `[]` is arbitrary and depends on the order pytest collects the fixtures. So on one run pytest might build `test_abc__xyz-def__.h5` or `test_abc__def-xyz__.h5`. So to make everything consistent and deterministic we would need to change `fname_prefix` function of [regression_data.py of tardisbase](https://github.com/tardis-sn/tardisbase/blob/master/tardisbase/testing/regression_data/regression_data.py) to keep the parameters in sorted order or something and same to be done for regression data files so as to ensure both follow the same convention.
This is the actual fix, but as the fail was for two tests only so I considered to patch it by hardcoding.
So which way would be better?

connor-mcclellan: Gotcha, I'll take a look myself and let you know. Thanks for looking into this!

RishmitaR: I'm not able to recreate the error on my machine, every time I run these tests pytest uses the fixtures in the way that the regression data expects. I found this stack overflow thread about manipulating the order fixtures are used - could you try a solution in this vein and see if it works?

https://stackoverflow.com/questions/25660064/in-which-order-are-pytest-fixtures-executed

RishmitaR (changes requested):

tardis/plasma/equilibrium/tests/test_level_populations.py
Aaryan-Dadu: @RishmitaR yeah, that thread highlights the exact issue I mentioned to Connor above. Pytest fixture order is arbitrary, which is why it works for you but fails for me.

I could try forcing the fixture order like the post suggests to see if it works, but previously I mentioned that the actual problem isn't the order in which Pytest runs the tests, the problem is that the function generating the filenames `fname_prefix` is non-deterministic. Fixing that directly (by sorting the inputs) solves the root cause.

RishmitaR: @Aaryan-Dadu I maybe should have clarified, I understand that my suggestion is another quick fix, but if the idea I suggested works it's a more preferable quick fix to hard coding the regression data path. We're going to open the underlying problem as an issue in the tardis repo.

Aaryan-Dadu: > @Aaryan-Dadu I maybe should have clarified, I understand that my suggestion is another quick fix, but if the idea I suggested works it's a more preferable quick fix to hard coding the regression data path. We're going to open the underlying problem as an issue in the tardisbase repo.

Okay I get it now, thanks for clarifying.
I have tried the fix and it works.

connor-mcclellan:
Thank you for your contribution! We're going to look into a more robust fix for this, but thank you for bringing it to our attention.

PR #3480, opened 2026-03-04
[GSOC] Rewrite the Custom Abundance Widget in Panel First Objective

SS-9098:
### :pencil: Description

**Type:**  :rocket: `feature` | :memo: `documentation`

Shell Info Widget and Line Info Widget are now interactive in the documentation.

Demo:

https://github.com/user-attachments/assets/165f0f69-e5f5-4cbe-b439-63000f17bb8f

#### Changes-
- **Uses Pyodide to construct widgets in docs** 
The shell info and line info rely heavily on python callbacks. rather than updating the tables by just putting values from a fixed data, they do some processing for each update. This makes it impossible to make them interactive in static embeddings. While we can make different javascript and html files for each widget and insert them individually, it is impractical. Pyodide allows the widgets to be interactive with just the help of a setup script and it can easily be extended to support more widgets.
- **Added nbsite.pyodide extension** 
There are other ways to run python in browser like using Thebe, Binder and JupyterLite. But they are slow to load and heavyweight. `nbsite.pyodide` is relatively lightweight and it gives the option to run and display the output cell in the page itself, like in a jupyter notebook. It is highly compatible with panel and also [suggested by them for sphinx embeddings](https://panel.holoviz.org/how_to/wasm/sphinx.html).
- **Changed recommanmark to myst_parser** 
It is currently not possible to use pyodide directly in a .ipynb file. We need to specify its directive in a .md file. Recommonmark does not allow for this and is also [deprecated in favour of myst_parser](https://github.com/readthedocs/recommonmark/issues/221). This change does not affect the existing documentation as we use .rst files to construct it.
- **Added mock tardis modules in pyodide's configuration** Pyodide cannot run the whole tardis package, nor do we need to. By setting up a mock tardis package in pyodide's files system imports like from tardis.visualization works and we can add files the widgets need instead of the entire tardis repository.
- **Uses pickled data objects** 
As we cannot run a whole tardis simulation in the browser, we need another way to get the data needed to construct the widget. Converting the data objects to JSON files will be quite difficult as we have complex data types such as Astropy.quantity and multiarrays. We may loose some info during reconstruction and also require new functions to construct the widget from JSON files. Pickling the data object allows for easy reconstruction without the need for  a new function.
- **Restructured line_info.py**
This widget had a single class and it was difficult to separate the data object we need to pickle from the widget. I separated the data functions in a new Data Info class following a similar structure to that of `shell_info.py.` Functionally, the widget will behave the same, but this refactoring allows us to isolate the data info object for pickling.
- **Added a new condition during from_simulation method**
If the code is running in browser using pyodide, they use pickled object instead of simulation object. The reason I have added a condition here rather than a new function, is so that will not be need to use a different function in docs page and the widget can be constructed with the same function we would normally use
- **Added new functions to return data info object** 
These functions provide access to the data object for storage purposes.

### :pushpin: Resources

https://pyodide.org/en/stable/usage/loading-custom-python-code.html for pyodide setup
https://panel.holoviz.org/how_to/wasm/index.html for nbsite.pyodide configuration
https://github.com/holoviz-dev/nbsite for reference

- [ ] Testing pipeline
- [X] Other method (manual)

### AI and LLM Usage
Github copilot was used for help in analyzing and understanding nbsite.pyodide's various abilities and limitations from [its repository](https://github.com/holoviz-dev/nbsite).
It was also used for help in restructuring line_info.py.
I confirm that I fully understand and can explain all AI assisted contribution.


tardis/visualization/widgets/shell_info.py
```diff
@@ -1,3 +1,4 @@
+import cloudpickle
```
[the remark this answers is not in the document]
SS-9098: Should I move import cloudpickle to inside the if statement?

docs/conf.py
```diff
@@ -315,9 +316,37 @@
 # -- Options for Panel -------------------------------------------
 from ipywidgets.embed import DEFAULT_EMBED_REQUIREJS_URL
 
+# Temporary patch as docs build is currently failing in the master branch
+nbsphinx_allow_errors = True
```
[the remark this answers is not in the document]
SS-9098: I included this so that build docs workflow would succeed. Can be removed later.

andrewfullard:
Thanks for your 2 first objective PRs! Please write and submit your draft proposal.

SS-9098:
Sure, I am working on it!

docs/_static/pyodide_setup.py
```diff
@@ -0,0 +1,217 @@
...
+
+write_package('tardis/util/environment.py', """\
+class Environment:
+    @staticmethod
+    def allows_widget_display():
+        return True
+
+    @classmethod
+    def get_current_environment(cls):
+        return 'pyodide'
```
atharva-2001, line 174: so in CI this overwrites the files- is there a better way? I am guessing this would impact the other widgets

docs/_static/pyodide_setup.py
```diff
@@ -0,0 +1,217 @@
...
+import pandas as pd
+import re
+from collections import OrderedDict
+
+ATOMIC_SYMBOLS_DATA = (
+    pd.read_csv(
+        "/data/atomic_symbols.dat",
+        sep=r"\s+",
+        names=["atomic_number", "symbol"],
+    )
```
atharva-2001, line 76: this is basically entire files being stripped down and copied to get a similar data file? this would be too hard to maintain- I guess this doesn't run the simulation as well- just sort of mimics it?

docs/_static/pyodide_setup.py
```diff
@@ -0,0 +1,217 @@
...
+class InteractionType(IntEnum):
+    NO_INTERACTION = -1
+    BOUNDARY = 1
+    LINE = 2
+    ESCATTERING = 4
+    CONTINUUM_PROCESS = 8
+""")
+
+write_package('tardis/configuration/sorting_globals.py', """\
+SORTING_ALGORITHM = "stable"
```
atharva-2001, line 188: why?

docs/conf.py
```diff
@@ -92,8 +92,9 @@
     "matplotlib.sphinxext.plot_directive",
     "sphinxcontrib.jquery",
     "nbsphinx",
+    "nbsite.pyodide",
     "numpydoc",
-    "recommonmark",
+    "myst_parser",
```
atharva-2001, line 97: why?

atharva-2001 (commented):
This is interesting work, but it will be hard to maintain. Do you have any other idea/method to do this? Does pyodide work with environment YAMLS?

docs/_static/pyodide_setup.py
SS-9098: This should not trigger during CI at all. This whole file will only be executed in client-side browser when they run a pyodide cell. 
This will only modify pyodide's virtual file system, so there will be no impact on other widgets during build.

docs/_static/pyodide_setup.py
SS-9098: This method contatins copies of a few functions from `util/base.py` required to construct the widget. The original `base.py` reads `data/atomic_symbols.dat` for atomic data, and this is doing the same from a copy of the file
Ideally, I would have liked to fetch `base.py` file directly from tardis master branch, but that will not work as base.py contains a few functions using radioactivedecay(which is not available in pyodide). Those functions are not used in the visualization modules so this method makes a stub of `base.py` containing only the necessary functions.

docs/_static/pyodide_setup.py
SS-9098: line_info.py imports `SORTING_ALGORITHM` from `tardis.configuration.sorting_globals`. These stubs are so imports like that works.

docs/conf.py
SS-9098: recommonmark does not support pyodide directives and is also deprecated in favour of myst_parser.
I have explained this in more detail in my pr description

SS-9098:
> This is interesting work, but it will be hard to maintain. Do you have any other idea/method to do this? Does pyodide work with environment YAMLS?

Thanks for the review!

I considered many alternatives such as thebe, jupyterlite and also not using a kernel at all. Basically, all of them will require a seperate file which will contruct the widget in a different way to be interactive in the docs.
I went with this solution because
- it is considerably more lightweight and faster to run in browser.
- We do not need a separate function for the widget.
- It uses most of the files directly from the tardis master branch.
- The only noticeable file which is not fetched this way is `util/base.py` which contains some helper functions such as converting atomic symbols to numbers. This is because it has some functions which use radioactivedecay library that we do not need for the widgets. This library is not available in pyodide, but it should be possible to still fetch the whole file and make the widgets work by adding a stub of radioactivedecay. Will this solve the maintainence concern?

Also, I am intentionally avoiding running a full TARDIS simulation in the browser. If the goal is simply to view a pre-generated interactive widget, requiring the entire simulation to execute seems too excessive.

PR #3483, opened 2026-03-07
[GSoC] Add ASV benchmark for macro atom solver

Deshan-5:
📝 Description
Type: 🚦 testing

This PR adds an ASV benchmark for `BoundBoundMacroAtomSolver.solve()` in TARDIS, as part of the benchmark optimisation GSoC 2026 project.

The original PR benchmarked individual distance calculation functions (`calculate_distance_boundary`, `calculate_distance_electron`, `calculate_distance_line`), but these completed in nanoseconds and weren't meaningful for regression tracking. This update replaces them entirely.

📌 Resources

* Macro atom solver module: `tardis.opacities.macro_atom.macroatom_solver`
* TARDIS benchmarking framework (ASV)
* ASV documentation: __https://asv.readthedocs.io/__

🚦 Testing
- [ ] Testing pipeline
- [x] Other method (describe)

Benchmarks were tested locally:
* Verified benchmark runs successfully without failures.
* Confirmed stable timing ~39 ms across repeated runs.
* Ensured proper integration with existing ASV benchmark discovery

☑️ Checklist

Note: If you are not allowed to perform any of these actions, ping (@) a contributor.

🤖 AI Usage
used Github Co pilot and gemini to help debugging and solve ASV run issues .


andrewfullard:
These benchmarks complete very quickly (some in nanoseconds). Do you think these are meaningful functions to benchmark?

Deshan-5:
> These benchmarks complete very quickly (some in nanoseconds). Do you think these are meaningful functions to benchmark?

You're right, some are too fast to be meaningful. Should I replace these with a benchmark for trace_packet in transport/montecarlo/modes/classic/rad_packet_transport.py? It calls all three distance calculations together as a single packet trace.

benchmarks/montecarlo_transport.py
```diff
@@ -0,0 +1,57 @@
...
+
+@parameterize({"Interaction type": ["scatter", "macroatom"]})
+class BenchmarkMontecarloTransport(BenchmarkBase):
+    """
+    Class to benchmark core Monte Carlo transport distance calculations.
+    """
+
+    repeat = 2
+
+    def setup(self, interaction_type):
```
hmperk18: please add some docstrings to these benchmarks

hmperk18 (changes requested):

benchmarks/montecarlo_transport.py
Deshan-5: added docstrings.

Deshan-5:
> > These benchmarks complete very quickly (some in nanoseconds). Do you think these are meaningful functions to benchmark?
> 
> You're right, some are too fast to be meaningful. Should I replace these with a benchmark for trace_packet in transport/montecarlo/modes/classic/rad_packet_transport.py? It calls all three distance calculations together as a single packet trace.



> These benchmarks complete very quickly (some in nanoseconds). Do you think these are meaningful functions to benchmark?

hi andrew after looking at [trace_packet](https://github.com/tardis-sn/tardis/blob/7dae5afb091f882896454a594350dbc2aa1f781b/tardis/transport/montecarlo/modes/classic/rad_packet_transport.py#L34) its complex than i thought ,the Numba-typed struct arguments  means basically rebuilding a full simulation in setup() . The individual distance fns  are too fast to be meaningful. which defeats the purpose of isolated benchmarks, i think i picked up the wrong scope here. 

Would redirecting this to [macro_atom_interaction ](https://github.com/tardis-sn/tardis/blob/master/tardis/transport/montecarlo/macro_atom.py#L37)  make sense? Unlike the distance fns it has a while loop that walks transition probabilities until emission, variable iterations, so regressions in the transition prob    data structures would actually show up. I tested it locally with [nb_simulation_verysimple ](https://github.com/tardis-sn/tardis/blob/7dae5afb091f882896454a594350dbc2aa1f781b/benchmarks/benchmark_base.py#L89)and it runs at 0.44 μs per call  meaningful to track. Happy to push it today if you give me the green light.

andrewfullard:
Thank you for your contribution, please try GSOC again next year!

PR #3532, opened 2026-03-27
Match continuum transition probs to ctardis

jvshields:
This one is ready to go now. The tardis macroatom solver gives the same unnormalized transition values for each transition, both on first solve and on subsequent iterations. The follow throughs are not tested here (i.e., what happens when you roll an emission probability), but that'll come later. 

I'm not super happy with the tests that are mostly ai written unit tests. Those should really do regression data comparisons, but I'm not sure these are completely settled yet so I figured it's easier to worry about that once the entire probability and metadata dataframes won't be tweaked anymore.

### DIFF
--- a/tardis/workflows/type_iip_workflow.py
+++ b/tardis/workflows/type_iip_workflow.py
@@ -746,9 +746,12 @@ class TypeIIPWorkflow(WorkflowLogging):
             recombination_rate = (
                 recomb_rates_solver._calculate_rate_coefficient()
             )
-
-            macro_atom_state = self.macro_atom_solver.solve(
+            j_blues_df = pd.DataFrame(
                 self.plasma_solver.j_blues,
+                index=self.plasma_solver.lines.index,
+            )
+            macro_atom_state = self.macro_atom_solver.solve(
+                j_blues_df,
                 opacity_state.beta_sobolev,
                 self.plasma_solver.stimulated_emission_factor,
                 photoion_rate,


tardis/transport/montecarlo/macro_atom.py
```diff
@@ -17,20 +17,31 @@ class MacroAtomError(ValueError):
...
     FF_EMISSION = -3
     ADIABATIC_COOLING = -4
     BF_COOLING = -5  # TODO: Maybe merge this with BF_EMISSION - Yes this is taken care of by BF_EMISSION
     TWO_PHOTON = -6
-    COLL_DOWN_TO_K_PACKET = 9
-    COLL_DOWN_INTERNAL = 10
-    COLL_EXC_COOL_TO_MACRO = 11
-    COLL_ION_COOL_TO_MACRO = 12
+    # Photoionization block
+    PHOTOIONIZATION_INTERNAL = 2
```
andrewfullard, line 31: There are a lot of types here. This should be a priority for cleanup later. Either more TODO or an issue (or both)

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -80,110 +77,80 @@ def continuum_transition_recombination_internal(
...
     Calculate unnormalized probabilities of radiative recombination emission transitions.
 
     Parameters
     ----------
-    spontaneous_recombination_coeff : pd.DataFrame
+    spontaneous_recombination_coeff
         Rate coefficient for spontaneous recombination from `k` to level `i`.
-    photoionization_data_frequencies : pd.Series
+    energies_diff_bound_free
         Ionization threshold frequencies for the levels.
```
andrewfullard: Is that still correct?

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -192,25 +159,23 @@ def probability_photoionization(
...
     """
-    p_photoionization = (
-        stim_recomb_corrected_photoionization_rate_coeff.multiply(
-            photoionization_data_level_energies, axis=0
-        )
+    p_photoion_int = stim_recomb_corrected_photoionization_rate_coeff.multiply(
+        photoionization_data_level_energies, axis=0
     )
-    return p_photoionization
+    return p_photoion_int
```
andrewfullard: int is ambiguous, use internal

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -739,3 +780,313 @@ def collisional_transition_excitation_cool(
...
+    Calculate collisional ionization internal probabilities.
+
+    Parameters
+    ----------
+    coll_ion_coeff
+        Collisional ionization coefficients.
+    electron_densities
+        Electron number densities.
+    energies_coll_ion
+        Ionization energy terms.
```
andrewfullard: Bit vague

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -739,3 +780,313 @@ def collisional_transition_excitation_cool(
...
+    Calculate collisional ionization emission probabilities.
+
+    Parameters
+    ----------
+    coll_ion_coeff
+        Collisional ionization coefficients.
+    electron_densities
+        Electron number densities.
+    energies_diff_bound_free
+        Bound-free energy differences.
```
andrewfullard: Differences between what?

andrewfullard (changes requested):
Generally good. Needs some clarity on a few variable names and docstrings to make it easier to understand. A lot of these comments apply to multiple instances of the same issue.

tardis/transport/montecarlo/macro_atom.py
jvshields: I agree it's a lot of types, but to some extent that's intentional. Before processing, I figure you want to know how each probability was calculated, and what physical transition it traces back to. After macro atom processing, all of the internal types will vanish, and the deactivation types will ultimately just correspond to emission paths you follow.

andrewfullard (dismissed):

connor-mcclellan (approved):
Looks sensible to me!

PR #3540, opened 2026-04-03
Use cooling transitions

jvshields:
Last transition PR didn't include BF or FF cooling. Here's the start to making those work correctly. Not sure this is done yet.


tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -809,35 +789,39 @@ def collisional_transition_excitation_cool(
 
     Returns
     -------
-    p_coll_excitation_cool
+    p_coll_excitation_to_k_packet
         Unnormalized collisional excitation cooling rates.
-    coll_excitation_cool_metadata
+    coll_excitation_to_k_metadata
```
connor-mcclellan: `coll_excitation_to_k_packet_metadata`, for consistency with the other names, maybe? I don't mind the long variable name in exchange for name symmetry with the other cases

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -616,19 +624,22 @@ def create_source_and_destination_idx_columns(
...
         Parameters
         ----------
-        macro_atom_transition_metadata : pandas.DataFrame
+        macro_atom_transition_metadata
             DataFrame containing metadata for macro atom transitions.
 
         Returns
         -------
-        pandas.Series
+        pd.Series
```
connor-mcclellan, line 636: Is this right? I'm not sure if code style specifies whether type descriptors use the base package name or the most commonly import-as name, but I'm used to seeing "this returns a pandas.Series" or "this returns a numpy.ndarray"

If you do `type(my_var)` I think you always get the full name. I recall readthedocs requiring that style to properly link to external packages but maybe that's changed

Issue #3545, opened 2026-04-08
[Docs] How To Plasma Graph Notebook shows no plasma graphs


RishmitaR:
**Describe the bug**

On the documentation page with examples of how to generate plasma graphs that show the relationships of variables used to compute the plasma state, the code blocks that display the images of the generated latex drawings instead show error messages indicating that these images do not exist. 

**To Reproduce**

To see the published documentation: 
- Go to https://tardis-sn.github.io/tardis/how_to/output/how_to_plasma_graph.html and see the error messages.

To view in local development environment:
-  Go to tardis/docs/how_to/output/how_to_plasma_graph.ipynb and see that the  images trying to be displayed do not exist

**Screenshots**

<img width="963" height="556" alt="Image" src="https://github.com/user-attachments/assets/a7fbf08e-bcc7-43e7-86e9-39b076d1c624" />

**System**
 - OS:
    - [x] GNU/Linux
    - [x] macOS

**Additional context**

I'm about to open a PR for this issue that simply just adds these images to docs/how_to/output folder. I ran the code to generate the tex code, pasted the tex code into overleaf, took screenshots of the drawings, and I added the screenshots to the docs/how_to/output folder. I would really appreciate if someone could review the new documentation to see if the images should be formatted different (I have questions about resizing mainly). I was also wondering if the image generation is supposed to be automated, and if thats a feature worth working on?

andrewfullard:
Note: the solution in #3546 pre-runs the notebook, and this will NOT change the output if the plasma graph changes. We will need to do that manually.

connor-mcclellan:
Marking this as resolved, but I also opened #3572 since we may still want a more permanent solution later down the line

PR #3546, opened 2026-04-08
Fix for issue #3545 How To Plasma Graph Notebook shows no plasma graphs

RishmitaR:
### :pencil: Description

This is a PR to adress issue #3545 which I raised as a location to facilitate further discussion on the issue if it's needed. This PR is a potentially rudimentary fix. On the documentation page with examples of how to generate plasma graphs that show the relationships of variables used to compute the plasma state, the code blocks that display the images of the generated latex drawings instead show error messages indicating that these images do not exist.  I ran the cells which generate the tex code, pasted the tex code into overleaf, took screenshots of the drawings, and I added the screenshots to the docs/how_to/output folder. As I discussed in the issue, I would really appreciate if someone could review the new documentation to see if the images should be formatted different.  

A potential feature that could fix this issue is have the base_plasma class have a write_to_png function that generates a image directly.

### :pushpin: Resources

 https://tardis-sn.github.io/tardis/how_to/output/how_to_plasma_graph.html 

- [ ] Testing pipeline
- [x] Other method (describe)

I built the documentation with make html to make sure the notebook looked correct


connor-mcclellan:
Nice, good catch! We looked into it and I think we can render the LaTeX directly here rather than having to load in a pre-rendered png (though your fix is a sensible approach to solving this problem).

See here: https://stackoverflow.com/a/61584424
IPython can import a LaTeX renderer, so you would just pass the plasma graph TeX output to `display(Latex())`. This is great because we won't have to add any additional dependencies to TARDIS.

RishmitaR:
Hi @connor-mcclellan thanks for the feedback! I looked into saving the output of the tex file into a string (both a regular string and a triple quotation string) and it seems like the Ipython Latex parser only supports math mode Latex renderings. Trying to render full documents didn't work. The only solutions I found to render a full latex document in a notebook involved saving the output to a pdf and either rendering the pdf inside the notebook or converting the pdf to an image.

Here are some of the Latex to PDF solutions I found online
1. https://stackoverflow.com/questions/69856790/how-to-render-a-full-latex-file-in-jupyter-notebook - this requires new dependencies to be installed
2. https://hatchjs.com/how-to-use-latex-in-jupyter-notebook/ - this solution just didn't work - writing to IOByte was not valid.

In my opinion, if the tardis team wanted to go through the hassle of adding new dependencies in tardis to generate the latex as a pdf and then rendering it as image which then gets embedded in the notebook would warrant it's own function (something like write_to_pdf or write_to_png in the plasma module) rather than just doing it in one notebook. 

Thanks again for the PR review, just let me know if I should look more into a Latex to image solution or if the original fix is good enough.

andrewfullard (approved):

jvshields (approved):

PR #3547, opened 2026-04-08
Adding adaptive damping method for convergence

DeekshaMohanty:
### :pencil: Description

**Type:** :rocket: `feature` 

This PR introduces the `adaptive_damped` convergence strategy in TARDIS.

It includes:
- the implementation in `convergence.py`
- tests in `test_convergence.py`
- a quickstart notebook `quickstart_adaptive_damped.ipynb` demonstrating usage and comparison with the standard `damped` method
- an example configuration `tardis_adaptive_damped_example.yml` for running simulations with the adaptive strategy

The approach is based on adaptive damping developed during this associated [master's thesis](https://d.lib.msu.edu/etd/52612?q=Mentor).

- [ ] Testing pipeline

### DIFF
--- a/tardis/io/configuration/config_reader.py
+++ b/tardis/io/configuration/config_reader.py
@@ -413,13 +413,21 @@ class Configuration(ConfigurationNameSpace, ConfigWriterMixin):
             ] = Configuration.parse_convergence_section(
                 montecarlo_section["convergence_strategy"]
             )
+        
+        elif montecarlo_section["convergence_strategy"]["type"] == "adaptive_damped":
+            montecarlo_section[
+                "convergence_strategy"
+            ] = Configuration.parse_convergence_section(
+                montecarlo_section["convergence_strategy"]
+            )
+
         elif montecarlo_section["convergence_strategy"]["type"] == "custom":
             raise NotImplementedError(
                 'convergence_strategy is set to "custom"; '
                 "you need to implement your specific convergence treatment"
             )
         else:
-            raise ValueError('convergence_strategy is not "damped" or "custom"')
+            raise ValueError('convergence_strategy is not "damped", "adaptive_damped" or "custom"')
 
     @staticmethod
     def parse_convergence_section(convergence_section_dict):
--- a/tardis/io/configuration/config_reader.py
+++ b/tardis/io/configuration/config_reader.py
@@ -407,20 +407,12 @@ class Configuration(ConfigurationNameSpace, ConfigWriterMixin):
         ----------
         montecarlo_section : dict
         """
-        if montecarlo_section["convergence_strategy"]["type"] == "damped":
+        if montecarlo_section["convergence_strategy"]["type"] in ["damped", "adaptive_damped"]:
             montecarlo_section[
                 "convergence_strategy"
             ] = Configuration.parse_convergence_section(
                 montecarlo_section["convergence_strategy"]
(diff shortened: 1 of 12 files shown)


andrewfullard (dismissed):
Very happy to see this cleaned up and in

andrewfullard (approved):

andrewfullard (changes requested):
I checked the docs and 1) the new notebook isn't in any of the side bar entries 2) the convergence plots aren't showing up, and they are probably the most important thing to show.

DeekshaMohanty:
Got it, thanks! I was initially planning to add the docs updates in a separate PR since I wasn’t fully sure where to place the notebook, but I suppose I could edit the quickstart to include two convergence method options instead of having a separate notebook for the new method?

andrewfullard:
> Got it, thanks! I was initially planning to add the docs updates in a separate PR since I wasn’t fully sure where to place the notebook, but I suppose I could edit the quickstart to include two convergence method options instead of having a separate notebook for the new method?

Don't edit the existing quickstart. Add an entry to the How to guides perhaps?

DeekshaMohanty:
> > Got it, thanks! I was initially planning to add the docs updates in a separate PR since I wasn’t fully sure where to place the notebook, but I suppose I could edit the quickstart to include two convergence method options instead of having a separate notebook for the new method?
> 
> Don't edit the existing quickstart. Add an entry to the How to guides perhaps?

Sounds good! I'll do that

tardis/io/configuration/config_reader.py
```diff
@@ -413,13 +413,21 @@ def validate_montecarlo_section(montecarlo_section):
...
                 montecarlo_section["convergence_strategy"]
             )
+        
+        elif montecarlo_section["convergence_strategy"]["type"] == "adaptive_damped":
+            montecarlo_section[
+                "convergence_strategy"
+            ] = Configuration.parse_convergence_section(
+                montecarlo_section["convergence_strategy"]
+            )
+
```
connor-mcclellan: The logic here is identical, so you don't need an `elif` at all

Do something like
```
if montecarlo_section["convergence_strategy"]["type"] in ["damped", "adaptive_damped"]:
```
instead

tardis/simulation/convergence.py
```diff
@@ -52,6 +59,50 @@ def damped_converge(self, value, estimated_value):
...
+        for lam in candidates:
+            x_new = value + lam * (estimated_value - value)
+            res = np.mean(np.abs((estimated_value - x_new) / (estimated_value)))
+            if best_residual is None or res < best_residual:
+                best_residual = res
+                best_lambda = lam
+
+        self.damping_factor = best_lambda
+
+        return value + best_lambda * (estimated_value - value)
```
connor-mcclellan: This is nitpicky and probably not a huge deal in terms of performance for a Python code anyway, but I want to point out that you've already computed this value above so you're doing unnecessary repeated math operations in the `return` line. Once you loop over the candidates, you already know that once you have your `best_residual` and `best_lambda`, whichever `x_new` you've calculated just above is, necessarily, the `best_x_new`. You might as well assign that value to a variable just below `best_lambda` and then return `best_x_new` instead of repeating the calculation.

connor-mcclellan (approved):

andrewfullard (approved):

PR #3554, opened 2026-04-15
Add absorbing markov chain processing and Hydrogen complete workflow

jvshields:
This PR only works for pure Hydrogen simulations, but is an important step in getting IIPs working. In this commit, you can insert the same inputs into the MC step and get the same answer as CTARDIS. 

There's also some changes to make the whole workflow better agree with CTARDIS. You can ask to restrict the detailed j_blues to a window between 2500 AA and 10,000 AA because that's what CTARDIS does. The workflow does do this by default. There are many other changes here to make this work but I don't want to list them all in the description. 

The point is, this is the big PR that now gives us the CTARDIS answer, as best as we can manage.


tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -0,0 +1,99 @@
...
+            Array of shape (num_cells, num_states, num_states) containing
+            absorption probabilities from each state for each cell.
+        - deactivating_probs : pd.DataFrame
+            DataFrame of deactivation transition probabilities from
+            absorption states.
+        - deactivating_metadata : pd.DataFrame
+            Metadata for deactivation transitions.
+    """
+    num_cells = transition_probabilities.shape[1]
+    probs = transition_probabilities.copy()
```
andrewfullard, line 50: This could be a problem as we move away from pure Hydrogen.

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Parameters
+    ----------
+    transition_probabilities : pd.DataFrame
+        DataFrame with transition probabilities between states for each cell.
+        Rows represent transitions, columns represent cells.
+    metadata : pd.DataFrame
+        Metadata about transitions including source_level_idx,
+        destination_level_idx, and transition_type. Values >= 0 for
+        transition_type indicate internal transitions; negative values
```
andrewfullard: What exactly is an "internal transition" in this scenario?

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -63,13 +62,8 @@ def continuum_transition_recombination_internal(
         (int(first), int(second), int(third))
         for first, second, third in p_recomb_internal.index.values
     ]
+    sources = [("i", -99, -99)] * len(p_recomb_internal.index)
```
andrewfullard, line 64: Why these values? See also, other similar lines

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -248,84 +234,79 @@ def continuum_transition_photoionization_internal(
...
-            "transition_type": MacroAtomTransitionType.PHOTOIONIZATION_TO_K_PACKET,
-            "transition_line_idx": -99,
-            "photoionization_key_idx": range(len(energies_bound_free_lower)),
-            "collision_key_idx": -99,
-        },
-        index=p_photoion_to_k_packet.index,
-    )
-
-    return p_photoion_to_k_packet, photoion_to_k_packet_metadata
+# def probability_photoionization_to_k_packet(
```
andrewfullard: Why is this commented out? Presumably it's not needed because of the Markov chain?

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -1133,33 +1123,204 @@ def collisional_transition_recombination_emission(
...
+        Shape is (n_zones, n_bound_levels).
+
+    Returns
+    -------
+    pd.DataFrame
+        Metadata DataFrame describing the free-bound cooling transitions for each bound level.
+    """
+    destinations = [
+        (1, 0, i) for i in range(fb_cool_probs_arr.shape[1])
+    ]  # hard coded for hydrogen
```
andrewfullard, line 881: Maybe add TODO to this so it's easier to find. This is important to modify in the (near?) future.

tardis/opacities/macro_atom/macroatom_line_transitions.py
```diff
@@ -84,7 +84,6 @@ def line_transition_internal_up(
         },
         index=p_internal_up.index,
     )
-    p_internal_up["source"] = sources
```
andrewfullard, line 87: Why is this removed?

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -1133,33 +1123,204 @@ def collisional_transition_recombination_emission(
...
+                    -99,
+                    -99,
+                    -99,
+                )
+            ],  # k-packet destruction to ff emission
+            "transition_type": MacroAtomTransitionType.FF_COOLING,
+            "transition_line_idx": -99,
+            "photoionization_key_idx": -99,
+            "collision_key_idx": -99,
+            "destination_level_idx": -99.0,
```
connor-mcclellan: Why is this a float?

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -896,6 +915,20 @@ def solve(
             Electron number densities for each cell.
         delta_E_yg
             Energy differences for transitions.
+        coll_exc_cool_rate
```
andrewfullard, line 918: Confusing- why is there both a "rate per cell" and an "array per transition"?

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
```diff
@@ -1133,33 +1123,204 @@ def collisional_transition_recombination_emission(
...
+    pd.DataFrame
+        Metadata DataFrame describing the free-free cooling transition.
+    """
+    free_free_cooling_metadata = pd.DataFrame(
+        {
+            "transition_line_id": -99,
+            "source": [("k", -99, -99)],
+            "destination": [
+                (
+                    -99,
```
connor-mcclellan, line 843: This is the only placeholder value I've seen so far for this field --- will it need to type match with the (string, int, int) dtype, or is it fine as-is?

tardis/opacities/macro_atom/macroatom_state.py
```diff
@@ -110,26 +111,38 @@ def __init__(
...
+        transition_metadata
             Metadata for the macro atom, including atomic number, ion number, level numbers for the transition, destination, and source.
-        line2macro_level_upper : pd.Series
+        line2macro_level_upper
             Mapping from lines to the upper levels of the macro atom transitions.
-        macro_block_references : pd.Series, optional
+        macro_block_references
             Index references to the Macro Atom blocks. Default is None.
+        normalized_deactivating_probs
+            Dataframe containing emission probabilities from a chosen absorbing state
```
andrewfullard, line 132: Normalized to sum to 1 I assume?

tardis/opacities/opacity_state.py
```diff
@@ -214,7 +211,58 @@ def to_numba(
...
         else:
+            # Not continuum
+            transition_probabilities = np.ascontiguousarray(
+                macro_atom_state.transition_probabilities.values.copy(),
+                dtype=np.float64,
+            )
+            line2macro_level_upper = (
+                macro_atom_state.line2macro_level_upper.values
+            )
+            # TODO: Fix setting of block references for non-continuum mode
```
andrewfullard, line 265: This breaks non-continuum mode?

tardis/transport/montecarlo/macro_atom.py
```diff
@@ -121,72 +131,59 @@ def macro_atom_interaction_iip(
...
+            absorbing_activation_level_idx = to_state_index
+            break
+    else:
+        raise MacroAtomError(
+            "MacroAtom failed to select an absorbing state. The absorbing Markov"
+            "chain probabilities matrix may not be normalized or may contain only zeros. \n"
+            f"Attempted to activate from {activation_level_idx}."
+        )
+
+    # Handle second prob call for emission process from that state
```
andrewfullard, line 149: So there is still a second check to determine emission type? But presumably k-packets are skipped over as part of the Markov chain?

connor-mcclellan (dismissed):

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -1219,24 +1286,58 @@ def _solve_first_macroatom_iteration(
...
-        )
+
         macro_block_references = self.create_macro_block_references(
-            macro_atom_transition_metadata
+            deactivating_metadata
         )
 
+        # getting all sorts of downstream problems because we trim the (1,0,0) because
+        # it does not have deactivation transitions so gets trimmed by the internal drop
+        # but trimming it screws up indexing
```
andrewfullard, line 1334: This is unfortunate

tardis/opacities/macro_atom/macroatom_solver.py
```diff
@@ -1544,21 +1621,48 @@ def _solve_next_macroatom_iteration(
...
             coll_recomb_coeff.iloc[collisional_recomb_emission_idxs],
             electron_densities,
             self._delta_E_yg_ionization.iloc[collisional_recomb_emission_idxs],
         ).to_numpy()
 
-        probabilities_df["source"] = (
-            macro_atom_transition_metadata.source.values
-        )  # Normalize by source in the next line, so need source column.
+        probabilities_df, macro_atom_transition_metadata = (
+            self.normalize_cooling_block_by_rate_fractions(
```
andrewfullard, line 1632: Two normalization steps?

andrewfullard (commented):
A few questions

tardis/opacities/macro_atom/absorbing_markov_chain.py
jvshields: What do you mean? As far as I can think about it, the absorbing markov chain stuff itself should be pretty agnostic. I would think it should even be fine having certain states inaccessible from continuum states, since it would just be accompanied by a 0 transition probability from x state to y unconnected state.

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
jvshields: If you have a better nomenclature in mind, I'm all ears. I'm mostly trying to make it clear that this is going to ionization energy. The tuple usually specifies atomic number, ionization number, excitation level number. For ionization energy that doesn't belong to an atom, I figured this was clearest.

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
jvshields: This is commented out because it's not actually a used transition. I double checked with christian and confirmed that you only get "deactivation transitions" from downwards transitions. So recombination or deexcitation. Probabilities for upwards transitions, like ionization, are calculated in ctardis but never used, which is what the template for this block is. I had it commented out because I wanted to double check with christian, but I think at this point it's safe to delete.

jvshields: It is unrelated to the absorbing markov chain stuff. Any transition that goes into that stuff still needs to be calculated on the front end before that part reduces out the internal transitions.

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
jvshields: I mentioned this in another comment, but the nomenclature is really inherited from (atomic_number, ionization_number, excitation_level_number), so the cleanest thing to do from a datatype perspective is have them all always be integers. The mixing types is bad, but I think better than having it be unclear by using some integer to point to k packet or r packet states. Ultimately before it goes into numba, the source and destination levels get converted to "idx" integers, so the gross mixed type tuples are relatively safe. 

Also, the placeholder here is pointing out that this is a k-packet doing free-free cooling, which isn't really associated with any active interaction handler state. Maybe it would be clearer as something like ("free-free", -99, -99), but that to me somehow implies that "free-free" is a state of the interaction handler, when it's not really.

tardis/opacities/macro_atom/macroatom_line_transitions.py
jvshields: Removed because it was used to do something hacky in the normalization part. You need to know the source block to normalize appropriately, but now I just grab the sources from the transition metadata rather than relying on it being carried along inconsistently from the metadata calculation function.

tardis/opacities/macro_atom/macroatom_solver.py
jvshields: Because the "rate" inputs are just shape [cells] and in ctardis are selected and then broken down further by process if that process can happen in multiple ways. E.g., you select that you're going to bound-free cool, and then after that you select which excitation state of hydrogen (or whatever) you bound-free recombine to for the reemission process. That's all to say, they have different shapes. I have made it clearer that the _arr inputs are 2d arrays by transition and cell.

tardis/opacities/macro_atom/macroatom_solver.py
jvshields: yes

tardis/opacities/macro_atom/macroatom_solver.py
jvshields: Yes and I think it's necessary. You need to normalize the before absorbing markov chain stuff because you do all the linear transformations on the internal jump matrix which is calculated assuming you have X chance of internally transiting, and 1-X chance to absorb, so you need to be normalized going in. Then after you trim everything, you normalize again because from your absorbing state, you have probability 1 of deactivating in some way. Without the second normalization, you would still have deactivation transition probabilities from states that sum to less than 1.

tardis/opacities/macro_atom/macroatom_state.py
jvshields: Yup.

tardis/opacities/opacity_state.py
jvshields: Will double check this, but it shouldn't, no. This part should be the same code and behavior as was happening before, but I restructured this part to have the conditional be

if scatter:
elif continuum:
else (classic mode):

as opposed to before where it was just 

if scatter:
else:

I needed to do this to be able to return the iip opacity state and this seemed to be the cleanest to do it. Again, should be the same behavior as before, just moved into the final else block.

tardis/transport/montecarlo/macro_atom.py
jvshields: Err, yes and no. There is a second check to determine emission type, yes. k-packets are only skipped over in the same way that any other internal state is skipped over, in that if you would do some string of internal transitions, you don't see that. You just know that you activate in a state, take a single jump to the state you will deactivate from, then do the second check to see which deactivating channel you actually do. Because k-packets have deactivations transitions (bf emission and ff emission), in the same way that you have a chance from going like (1, 0, 2) -> (1, 0, 4) and then deactivate via (1, 0, 4) to something below it like (1, 0, 3), you can also jump (1, 0, 2) -> (k, -99, -99) and then decide to do (k, -99, -99) -> ff emission. 

So in that sense, there's still a k-packet state here, you just only go to it if you deactivate from it, because if you would go to it and then do another internal transition away from it, that history gets compressed.

tardis/opacities/macro_atom/absorbing_markov_chain.py
andrewfullard: I wasn't clear. Copying the entire probabilities dataframe may become a memory issue.

tardis/opacities/macro_atom/macroatom_continuum_transitions.py
andrewfullard: That's fine, just make sure it is documented each time

tardis/transport/montecarlo/macro_atom.py
andrewfullard: Okay. Seems like something that could use documentation.

tardis/opacities/macro_atom/absorbing_markov_chain.py
jvshields: That's fair. I didn't want to modify it inplace because I had to stick the source column back on there. If you tried to do continuum with a large number of heavy elements the probability dataframe could probably get quite bulky, though for hydrogen we're still only about ~2000 or so. The copied dataframe should also get garbage collected once you leave this function, so I'm hoping that's fine enough.

tardis/workflows/tests/test_iip_workflow.py
```diff
@@ -720,7 +720,7 @@ def test_solve_continuum_state_after_nlte_init(
     )
 
 
-@pytest.mark.skip  # TODO: Re-enable when the workflow uses the two step probabilities
+# @pytest.mark.skip  # TODO: Re-enable when the workflow uses the two step probabilities
```
andrewfullard: Remove commented line entirely

connor-mcclellan (dismissed):

andrewfullard (approved):

connor-mcclellan (approved):

PR #3563, opened 2026-04-30
Added notebook that compares 2026 TARDIS to 2014 TARDIS

ClydeME:
### :pencil: Description

Adds a notebook that compares the spectrum generated by the 2026 version of TARDIS to the spectrum data found in the 2014 simulation setups folder. The notebook finds the residuals between the values of the two spectra over three different wavelength windows, then it divides by the standard deviation of one of the spectra to standardize the residuals. It then creates a histogram for each window showing what percentage of the residuals are at different distances (measured in Standard Deviations) from the mean value of the spectra in that window. It overlays a Normal Distribution on those histograms, because we would expect the histograms to follow a Normal Distribution.

- [ ] Testing pipeline
- [ ] Other method (describe)
- [X] My changes can't be tested (explain why)

These changes only add documentation, they do not affect or add to the functionality of TARDIS.


docs/Comparing_TARDISs/tardis_example_spec.dat
connor-mcclellan, line 1: These spectra files should be moved to regression data and loaded into the notebook that way rather than including them in the source code - no reason to version control a data file

docs/Comparing_TARDISs/tardis_example_spec_virt.dat
connor-mcclellan, line 1: These spectra files should be moved to regression data and loaded into the notebook that way rather than including them in the source code - no reason to version control a data file

PR #3582, opened 2026-05-22
Add multiple element ejecta support to IIP workflow

jvshields:
Currently opening this for @DeekshaMohanty to play around with.

Fixing bugs to allow adding multiple elements in the IIP workflow.

### DIFF
--- a/tardis/opacities/macro_atom/absorbing_markov_chain.py
+++ b/tardis/opacities/macro_atom/absorbing_markov_chain.py
@@ -23,6 +23,11 @@ def create_absorbing_probs(
     The fundamental matrix is computed to determine absorption probabilities
     and expected number of steps to absorption from each state.
 
+    WARNING: This is currently slower than it needs to be for many element sims.
+    You should be able to slice out and process a subset of the probabilities
+    based on some source state filtering. We have a draft PR open for this
+    but it has not been finished.
+
     Parameters
     ----------
     transition_probabilities : pd.DataFrame
@@ -37,7 +42,7 @@ def create_absorbing_probs(
 
     Returns
     -------
-    tuple[np.ndarray, pd.DataFrame, pd.DataFrame]
+    tuple[np.ndarray, pd.DataFrame]
         - absorbing_probability_matrix : np.ndarray
             Array of shape (num_cells, num_states, num_states) containing
             absorption probabilities from each state for each cell.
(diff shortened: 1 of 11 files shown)


jvshields:
Needs https://github.com/tardis-sn/tardis-regression-data/pull/97

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -36,65 +42,62 @@ def create_absorbing_probs(
...
+        identity_minus_Q = (
+            scipy.sparse.identity(internal_jump_matrix.shape[0])
+            - internal_jump_matrix
         )
-        fundamental_matrix_N = np.linalg.inv(inv_matrix_fundamental_N)
+        csc_N = identity_minus_Q.tocsc()
 
-        expected_steps = fundamental_matrix_N.sum(axis=1)
-        expected_steps_in_cells_from_states[cell] = expected_steps
+        # expected_steps = np.asarray(
```
andrewfullard, line 82: Still useful for something?

tardis/workflows/tests/test_iip_workflow.py
```diff
@@ -671,6 +671,7 @@ def test_thermal_balance_solver(
     )
 
 
+@pytest.mark.xfail  # JOSH: This test fails because I disabled diagonalize_ma() in the BaseContinuum object to handle multi-element sims
```
andrewfullard, line 674: Is this still the case with new regression data?

andrewfullard (changes requested):

tardis/opacities/macro_atom/absorbing_markov_chain.py
jvshields: There's a comment in the file already. 

    # Josh: The expected steps calculation is another linear algebra solve. We don't need
    # it for the MC calculation, so I'm removing it, but leaving it commented in case it is
    # useful for diagnostic purposes in the future.
    # expected_steps_in_cells_from_states = np.zeros((num_cells, num_states))

tardis/workflows/tests/test_iip_workflow.py
jvshields: Yes, unless you wanted to change the hardcoded file paths which I can't regenerate because they're not hooked up to the regression data sync stuff.  

This test compares to "continuum_recomb_transition_prob_init_nlte.h5"
but the recombination transition probabilities are different now. We could either compare to ctardis also not doing the diagonalization, or we could remove this test. In any case it's not worth getting diagonalize_ma() to work in the base_continuum object because we're trying not to touch the ported ctardis code.

andrewfullard: Ah, I see. That's unfortunate but will have to do I suppose.

andrewfullard (approved):
Will merge pending tests

PR #3607, opened 2026-06-09
Full test of the simple workflow

andrewfullard:
### :pencil: Description

The env upgrade revealed a problem with our test coverage of workflows. This adds a test that runs a full simple workflow with iteration and compares to the existing output of run tardis.

- [x] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/workflows/tests/test_workflows.py
+++ b/tardis/workflows/tests/test_workflows.py
@@ -2,7 +2,9 @@ import numpy as np
 import pandas as pd
 import pytest
 from astropy import units as u
+from astropy.tests.helper import assert_quantity_allclose
 
+from tardis.io.configuration.config_reader import Configuration
 from tardis.workflows.simple_tardis_workflow import SimpleTARDISWorkflow
 from tardis.workflows.standard_tardis_workflow import StandardTARDISWorkflow
 from tardis.workflows.v_inner_solver import InnerVelocitySolverWorkflow
@@ -35,6 +37,17 @@ def v_inner_workflow(v_inner_config):
 @pytest.fixture(scope="module")
 def simple_workflow_one_loop(config_verysimple_for_simulation_one_loop):
     workflow = SimpleTARDISWorkflow(config_verysimple_for_simulation_one_loop)
+    workflow.run()
+    return workflow
+
+
+@pytest.fixture(scope="module")
+def simple_workflow_verysimple(atomic_data_fname, example_configuration_dir):
+    config = Configuration.from_yaml(
+        str(example_configuration_dir / "tardis_configv1_verysimple.yml")
+    )
+    config["atom_data"] = atomic_data_fname
+    workflow = SimpleTARDISWorkflow(config)
     workflow.run()
     return workflow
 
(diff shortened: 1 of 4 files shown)


tardis/workflows/tests/test_workflows.py
```diff
@@ -133,6 +146,52 @@ def test_simple_tardis_workflow_against_standard_workflow(
     )
 
 
+def test_simple_tardis_workflow_against_test_tardis_full_regression(
+    simple_workflow_verysimple, regression_data
+):
+    ref_file = (
```
jvshields: Are these LLM generated tests that don't understand how our regression data works? Why aren't these just a bunch of sync_dataframes calls?

jvshields: I guess because we want to compare them to the regression data generated by the classic tardis tests. I get a bit scared by working outside of our framework in this way though because if something changes we'll get test failures here that are actually related to tests that generate the TestTransportSimple.h5 file, which changes the scope of the test in a weird way. 

Is there a fixture from the classic tardis runs that we can compare against instead? I would much rather do some syncs for internal testing, and comparing against fixtures for the workflow -> classic_run comparison.

andrewfullard: That's reasonable. We do already have tests like this for workflows though, that point explicitly at regression data files. The ol' LLM didn't make it up wholesale. But since in most test runs we will create all the fixtures anyway it may as well compare to that.

jvshields (dismissed):

connor-mcclellan:
Tests failing:
```
FAILED tardis/workflows/tests/test_workflows.py::test_simple_tardis_workflow_against_test_tardis_full_regression - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 487956 / 584480 (83.5%)
First 5 mismatches are at indices:
 [2020, 11]: 3.3945662126384763e-22 (ACTUAL), 0.0 (DESIRED)
 [2021, 11]: 3.3877456693209765e-22 (ACTUAL), 0.0 (DESIRED)
 [2022, 11]: 3.3877456693209765e-22 (ACTUAL), 0.0 (DESIRED)
 [2023, 11]: 3.387351206641851e-22 (ACTUAL), 0.0 (DESIRED)
 [2024, 11]: 3.3854928335010723e-22 (ACTUAL), 0.0 (DESIRED)
Max absolute difference among violations: 1.23348514e-17
Max relative difference among violations: 8.467167
 ACTUAL: array([[0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],...
 DESIRED: array([[0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],...
= 1 failed, 1723 passed, 480 skipped, 32 xfailed, 3 xpassed, 84 warnings in 825.67s (0:13:45)
```

This is just a case of relative tolerance being unsuitable for stuff extremely close to exactly zero, though. Use absolute tolerance for this one?

andrewfullard:
Probably need to rebase/merge to fix the benchmarks

jvshields (approved):

connor-mcclellan (approved):

PR #3608, opened 2026-06-09
Monte Carlo refactoring tests

andrewfullard:
### :pencil: Description

These tests are intended to characterize the MC part of TARDIS so that it can be refactored safely. Lots more tests for gamma-ray transport, including a bug catch that we may want to fix before the refactor (see comment)

Requires https://github.com/tardis-sn/tardis-regression-data/pull/99

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/energy_input/transport/tests/test_gamma_transport.py
```diff
@@ -0,0 +1,598 @@
...
+    distance, shell_change = calculate_distance_radial(
+        packet,
+        5.0e13,
+        2.0e14,
+    )
+
+    npt.assert_allclose(distance, 5.0e13)
+    # Current implementation returns +1 here because the shell-change test
+    # compares against ``inner_1 or inner_2``.
+    assert shell_change == 1
```
andrewfullard, line 99: This is a bug, but probably hasn't affected things because gamma-rays are very unlikely to scatter inwards.

tardis/energy_input/transport/tests/test_gamma_transport_characterization.py
```diff
@@ -0,0 +1,615 @@
...
+    get_compton_fraction_artis,
+    pair_creation_packet,
+    scatter_type,
+)
+from tardis.energy_input.transport.GXPacket import GXPacket, GXPacketStatus
+from tardis.energy_input.util import ELECTRON_MASS_ENERGY_KEV, H_CGS_KEV
+
+
+@njit
+def _seed_numba_random(seed: int) -> None:
```
jvshields: What is this about? How is this different from replacing np.random.seed with _seed_numba_random, which is pretty comparable character-wise. Is the njit important for some reason?

tardis/energy_input/transport/tests/test_gamma_transport_characterization.py
```diff
@@ -0,0 +1,615 @@
...
+    regression_data,
+) -> None:
+    _skip_unless_python_gamma_loop()
+    packet = gamma_packet
+    distances = [
+        (1.0, 20.0, 10.0, 0),
+        (20.0, 1.0, 10.0, 1),
+    ]
+
+    def fake_distance_trace(*args):
```
jvshields: What's all this about? I'm not familiar with why you'd want to do function factory stuff, but this looks scary and very unintuitive to me.

tardis/transport/montecarlo/tests/test_single_packet_loop.py
```diff
@@ -1,10 +1,418 @@
...
+        source_rpacket_index=0,
+        spectrum_frequency_grid=np.array([1.0e14, 2.0e14]),
+        number_of_vpackets=0,
+        v_packet_spawn_start_frequency=0.0,
+        v_packet_spawn_end_frequency=np.inf,
+        temporary_v_packet_bins=0,
+    )
+
+
+def _patch_common_classic_hooks(monkeypatch, module, interactions) -> None:
```
jvshields: More of the same questions from my last comment. It's possible I'm just not familiar with what's going on here, but working a bunch with monkeypatch attribute setting I think is a pretty big departure from how we do things currently so I'd love to have a conversation about if this is something we want to do going forward or not and why

tardis/energy_input/transport/tests/test_gamma_transport_characterization.py
andrewfullard: Yep, you need a jitted numpy seed to apply to jitted code.

jvshields: Right, but it looked to me like you were calling this in jitted functions anyway. If you just set the numpy seed the normal way in the jitted function, does that not work?

tardis/transport/montecarlo/tests/test_single_packet_loop.py
andrewfullard: What it is doing is setting specific attributes in a class so that they can be parametrized more easily. I agree it's complex and I can look for alternative methods if you want.

tardis/transport/montecarlo/tests/test_transport_characterization.py
```diff
@@ -0,0 +1,716 @@
...
+        nu=4.0e14,
+        energy=0.9,
+        seed=1963,
+        index=0,
+    )
+    packet.current_shell_id = 1
+    return packet
+
+
+def _make_rpacket(
```
jvshields: This one has to be kinda weird right. There's an underscore method to make a packet that always initializes the packet with the same r, mu, nu, and energy, but lets you put in a variable shell id? And it doesn't interact at all with the rpacket fixture defined directly above.

tardis/energy_input/transport/tests/test_gamma_transport_characterization.py
andrewfullard: I agree this test is pretty complex. I'm okay cutting it and seeing what that does to coverage, or add descriptive comments.

tardis/transport/montecarlo/tests/test_transport_characterization.py
```diff
@@ -0,0 +1,716 @@
...
+
+    return NumbaNonhomologousRadial1DGeometry(
+        np.array([7.0e14, 8.0e14]),
+        np.array([r_outer_first_shell, 3.0e16]),
+        v_inner,
+        v_outer,
+    )
+
+
+def _make_line_estimators(
```
jvshields: Also flagging this as a weird helper function that seems to save very little text when it's actually called, compared to initalizing the EstimatorsLine class there directly.

tardis/transport/montecarlo/tests/conftest.py
```diff
@@ -131,16 +145,122 @@ def static_packet():
...
+        np.array([r_outer_first_shell, 3.0e16]),
+        np.array([-1.0, -1.0]),
+        np.array([-1.0, -1.0]),
+    )
+
+
+@pytest.fixture
+def nonhomologous_geometry(request) -> NumbaNonhomologousRadial1DGeometry:
+    params = getattr(request, "param", {})
+    if params.get("negative_velocity_gradient", False):
```
connor-mcclellan, line 213: I think these are wrong, right? You have two shells, so three interfaces at v=1.0e9, 1.5e9, 2.0e9

If the velocity gradient is positive:
v_inner should be [1.0e9, 1.5e9]
v_outer should be [1.5e9, 2.0e9]

If the velocity gradient is negative, the interfaces are in the opposite order: v=2.0e9, 1.5e9, 1.0e9
v_inner should be [2.0e9, 1.5e9]
v_outer should be [1.5e9, 1.0e9]

connor-mcclellan:
I don't quite get the point of the name "characterization" for these tests. Why not just call `test_gamma_transport_characterization` -> `test_gamma_transport`
`test_homologous_move_r_packet_characterization` -> `test_homologous_move_r_packet`, etc?

andrewfullard:
> I don't quite get the point of the name "characterization" for these tests. Why not just call `test_gamma_transport_characterization` -> `test_gamma_transport` `test_homologous_move_r_packet_characterization` -> `test_homologous_move_r_packet`, etc?

Yeah I've considered just removing that. Since you agree I'll do so.

tardis/transport/montecarlo/tests/conftest.py
andrewfullard: Yeah I don't think the LLM understood the mathematics here

PR #3622, opened 2026-06-23
Remove Unbuilt Plasma Attributes from Tested Plasma Properties

RishmitaR:
.### :pencil: Description

**Type:**  :vertical_traffic_light: `testing`

The test_plasma_properties test currently looks at specific plasmas built from simulation states stored in hdf5 files and checks to see if properties exist by seeing if the attribute exists as a key in the file. If it doesn't exist as a key in the file then the plasma is assumed to not have that attribute and pytest skips testing the attribute. This is a little opaque because  it's hard to discern if the tests are being skipped because the attribute does not exist in the plasma or if it's being skipped because attributes don't exists as keys even when the plasma does have the attribute and the regression data is being read incorrectly. 

This PR checks if a plasma has attribute and skips the test for the attribute if the plasma does not have the attribute. That way the only checks to regression data being made are for attributes that are 100% known to exist. If the property is still not found then an explicit warning is sent. This PR also gives a more clear message on why the skips occur. 

I've also went ahead and removed properties in the combined properties list that were always being skipped and are properties associated with no plasmas. 

This PR now also modifies the test file so that all tests use the regression data sync functions rather than using pd.read_hdf. Modified regression data is in this PR: https://github.com/tardis-sn/tardis-regression-data/pull/102 

### :pushpin: Resources

- [x] Testing pipeline
- [ ] Other method (describe)


andrewfullard (dismissed):

tardis/plasma/tests/test_complete_plasmas.py
```diff
@@ -177,28 +174,24 @@ def plasma(
...
-            if attr == "selected_atoms":
-                actual = actual.to_frame(index=False)
-                pdt.assert_frame_equal(actual, expected)
-            elif actual.ndim == 1:
-                actual = pd.Series(actual)
-                pdt.assert_series_equal(actual, expected)
-            else:
-                actual = pd.DataFrame(actual)
-                pdt.assert_frame_equal(actual, expected)
+        expected = pd.read_hdf(self.regression_data.fpath, key)
```
jvshields: Can we change this to a regression_data.sync_dataframe call? It's much more maintainable, and I think should be an easy swap, though maybe the path or naming will change. If nothing else I'd like you to try it out and see if the test passes.

RishmitaR: Sadly it's not a easy swap. I think new regression data would have to be generated to make it work and a lot of the changes would undo a lot of Andrew's latest PR https://github.com/tardis-sn/tardis/pull/3616

andrewfullard:
@Rodot- given the test failures I'm in favour of not merging this and closing the PR, unless you have time to fix it.

RishmitaR:
> @Rodot- given the test failures I'm in favour of not merging this and closing the PR, unless you have time to fix it.

I think I can fix this within this week, I'll let you know if not

andrewfullard:
> > @Rodot- given the test failures I'm in favour of not merging this and closing the PR, unless you have time to fix it.
> 
> I think I can fix this within this week, I'll let you know if not

It's been a week, if you're too busy that's fine.

connor-mcclellan:
Closing for now, but if we want to come back to fix this in the future we can always re-open

RishmitaR:
@andrewfullard both this pr and regression data pr are now updated with respective master and main

jvshields (approved):

jvshields:
It looks like a lot of the tests are failing because of the high energy data changes, so that will probably need another rebase. 

I really like this PR. It should help safeguard us against some of the plasma changes that have slipped by in the past.

PR #3624, opened 2026-06-25
Improve performance of stimulated emission factor calculation with NLTE species

andrewfullard:
### :pencil: Description

**Type:** :beetle: `bugfix` 

The NLTE mask calculation for stimulated emission factor uses a lambda function acting on rows of a dataframe. This is extremely slow. Instead here is a vectorized version.

Also includes various other performance improvements across the IIP plasma and the macroatom. Needs regression data.

### :pushpin: Resources

[nlte_line_mask_benchmark.pdf](https://github.com/user-attachments/files/29320046/nlte_line_mask_benchmark.pdf)

https://github.com/tardis-sn/tardis-regression-data/pull/103

- [x] Testing pipeline
- [ ] Other method (describe)


jvshields (dismissed):

tardis/iip_plasma/properties/radiative_properties.py
```diff
@@ -105,15 +105,16 @@ def calculate(
...
-                    in self.nlte_species,
-                    axis=1,
+            atomic_numbers = lines.index.get_level_values(
+                "atomic_number"
+            ).to_numpy()
+            ion_numbers = lines.index.get_level_values("ion_number").to_numpy()
+
+            nlte_lines_mask = np.zeros(len(lines), dtype=bool)
+            for atomic_number, ion_number in self.nlte_species:
+                nlte_lines_mask |= (atomic_numbers == atomic_number) & (
```
connor-mcclellan: Took me awhile to understand what this is doing, but I guess it's okay. I might prefer

```
species_match = (atomic_numbers == atomic_number) & (ion_numbers == ion_number)
nlte_lines_mask = nlte_lines_mask | species_match
```

andrewfullard:
Needs https://github.com/tardis-sn/tardis-regression-data/pull/103

andrewfullard:
Note that benchmarks are failing because the new benchmarks don't exist at the head of the branch.

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -4,17 +4,27 @@
...
-References:
+References
+----------
     Absorbing Markov chain theory: https://en.wikipedia.org/wiki/Absorbing_Markov_chain
 """
 
+from __future__ import annotations
+
+from concurrent.futures import ThreadPoolExecutor
+from typing import TYPE_CHECKING
```
jvshields, line 13: I was under the impression we just always import modules and do type checking the simple way, rather than do the conditional import stuff like on line 20.

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -64,19 +74,18 @@ def create_absorbing_probs(
 
     rows = metadata[internal_mask].source_level_idx.values
     cols = metadata[internal_mask].destination_level_idx.values
-    for cell in range(num_cells):
+    identity_matrix = scipy.sparse.identity(num_states, format="csc")
+
+    def solve_cell(cell: int) -> tuple[int, np.ndarray, np.ndarray]:
```
jvshields: I have no idea why you need to define another function inside the scope of the original function, but that's probably a gap in my programming knowledge. Can this be explained in some way or is there a resource I can read about it?

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -87,15 +96,51 @@ def create_absorbing_probs(
         deactivation_row = np.asarray(
             internal_jump_matrix.sum(axis=1)
         ).flatten()
-        # Solve (I - Q) * X = diag(1 - deactivation_row) directly
-        # instead of computing inv(I - Q) explicitly
-        rhs = np.diag(1 - deactivation_row)
-        absorbing_probability_matrix[cell] = scipy.sparse.linalg.spsolve(
-            csc_N, rhs
+        rhs_diagonal = 1 - deactivation_row
+        active_rhs_columns = np.flatnonzero(rhs_diagonal)
```
jvshields: It's very confusing to me that you need to do this to be honest. I was under the impression that the coordinate matrix machinery already lets the linear algebra solution only concern itself with nonzero values.

tardis/opacities/macro_atom/absorbing_markov_chain.py
andrewfullard: It's just a way to avoid redefining variables that are in-scope in the parent function. If you break it out, you have to pass those variables in to the child function.

tardis/opacities/macro_atom/absorbing_markov_chain.py
andrewfullard: Yeah I want to clean that up. If nothing else, this is pre-3.14 behaviour

tardis/opacities/macro_atom/absorbing_markov_chain.py
andrewfullard: Apparently not, but I can check.

tardis/opacities/macro_atom/absorbing_markov_chain.py
andrewfullard: https://docs.astral.sh/ruff/rules/typing-only-third-party-import/

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -87,15 +96,51 @@ def create_absorbing_probs(
...
+                    solved_columns
+                )
+    else:
+        for cell in range(num_cells):
+            _, active_rhs_columns, solved_columns = solve_cell(cell)
+            absorbing_probability_matrix[cell][:, active_rhs_columns] = (
+                solved_columns
+            )
+
+    deactivating_probs = transition_probabilities.__class__(
```
jvshields: This is a somewhat insane way of building a simple dataframe.

tardis/opacities/macro_atom/absorbing_markov_chain.py
jvshields: And it doesn't add overhead from having python redefine the child function every time the parent function is called?

tardis/opacities/macro_atom/absorbing_markov_chain.py
jvshields: Also not quite sure why it chose to reverse the behavior and make a dataframe of zeros and the write in the deactivating probabilites, rather than before where we copied the dataframe over and then zeroed out the internal transitions. I guess I don't have a strong preference and I expect the number of deactivating and internal transitions to be roughly equal so it should be pretty either way. I would also expect that it doesn't matter at all performance wise so I guess I don't care enough to request it changes.

andrewfullard: Yeah I agree haha

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -667,6 +721,90 @@ def _caculate_beta_sobolevs(self, level_number_density):
...
+            ).to_numpy()
+            ion_numbers = pl.lines.index.get_level_values(
+                "ion_number"
+            ).to_numpy()
+            nlte_lines_mask = np.zeros(len(pl.lines), dtype=bool)
+            for atomic_number, ion_number in self.plasma_parent.nlte_species:
+                nlte_lines_mask |= (atomic_numbers == atomic_number) & (
+                    ion_numbers == ion_number
+                )
+            tau_sobolev_property = self.plasma_parent.plasma_properties_dict[
```
jvshields: I think this function needs to be re-examined and looks sloppy. I'm not sure totally why it needs to exist in the first place, but if it does, this is accessing a property of self.plasma_parents which is already popped out on line 739. 

I'd at least be more comfortable with the function if some docstrings explained why this is a thing we need in the first place.

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -667,6 +721,90 @@ def _caculate_beta_sobolevs(self, level_number_density):
...
+                    * pl.wavelength_cm.to_numpy()
+                    * time_explosion
+                ),
+            }
+        return self._beta_sobolev_inputs
+
+    def _calculate_beta_sobolevs_from_values(
+        self,
+        level_density_values: np.ndarray,
+        line_indices: np.ndarray | slice | None = None,
```
jvshields, line 829: This looks like some very modular defensive coding that we try to avoid.

tardis/iip_plasma/properties/partition_function.py
jvshields: Thinking more about it, I think this is a weird thing I've noticed LLMs like to do, preparing a set of inputs and calling it an "input" object. I think this is bad and confusing because we don't have abstract "input" objects since it's not clear what they're inputs for, and we think much more about our python objects as physical things rather than coding objects. Like, a beta sobolev calculator function should just be fed the physical things it needs directly.

tardis/iip_plasma/tests/test_base_plasma.py
```diff
@@ -0,0 +1,37 @@
...
+
+class InputB(Input):
+    outputs = ("b",)
+
+
+class SumAB(ProcessingPlasmaProperty):
+    outputs = ("sum_ab",)
+
+    def calculate(self, a: float, b: float) -> float:
+        return a + b
```
jvshields, line 19: No way this is a reasonable test

tardis/iip_plasma/tests/test_base_plasma.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class TwiceSum(ProcessingPlasmaProperty):
+    outputs = ("twice_sum",)
+
+    def calculate(self, sum_ab: float) -> float:
+        return 2.0 * sum_ab
```
jvshields, line 26: This also seems insane. I don't think we're interested in testing whether the plasma builds correctly are we?

jvshields: If the point is to test mutation maybe the class name should mention that that's the point of it. But generally I didn't think we were interested in testing this.

tardis/iip_plasma/properties/partition_function.py
andrewfullard: The intent for this is to cache beta sobolev inputs that are static between iterations, because the beta sobolev calculation is called thousands of times. It should probably be renamed to reflect that.

tardis/iip_plasma/tests/test_base_plasma.py
andrewfullard: The point of this file and its tests is to validate that caching the construction of a networkx plasma works. I'm not sure how significant the performance gains of said caching are though!

tardis/iip_plasma/base.py
```diff
@@ -49,7 +58,7 @@ def __dir__(self):
 
     @property
     def plasma_properties_dict(self):
-        return {item.name: item for item in self.plasma_properties}
+        return self._plasma_properties_dict
```
jvshields, line 61: This looks bad to me. Why even have the property at this point? Seems like the init should just set self.plasma_properties_dict appropriately in the first place.

tardis/iip_plasma/properties/partition_function.py
andrewfullard: It's so that it can be compared to the previous dataframe version in test_iip_workflow

tardis/iip_plasma/base.py
andrewfullard: It's the typical LLM scenario of wanting to keep APIs static

tardis/iip_plasma/properties/continuum.py
```diff
@@ -59,6 +59,19 @@ def calculate_rate_coefficient_from_estimator(
 def prepare_level_group_integration(
     nu: pd.Series,
 ) -> tuple[tuple[np.ndarray, ...], pd.MultiIndex, np.ndarray]:
+    """Prepare reusable level-group metadata for continuum integration.
+
+    Parameters
+    ----------
+    nu : pandas.Series
```
jvshields, line 66: Type hints in function definitions, not in docstrings

andrewfullard: numpydoc still requires types to be included in docstrings, we should consider changing our style guide if that's what you want

jvshields: Hmm. I could have sworn that was a Wolfgang hard line at one point. I'd personally rather adhere to numpydoc standards.

tardis/iip_plasma/properties/partition_function.py
jvshields: Ok but at the very least in this line shouldn't this change be 
```suggestion
            tau_sobolev_property = pl.plasma_properties_dict[
```

jvshields (dismissed):

connor-mcclellan (dismissed):

andrewfullard:
Note to self: fix rtol on failing tests

tardis/workflows/type_iip_workflow.py
```diff
@@ -932,9 +933,9 @@ def run(self):
             )
```
jvshields, line 934:
```suggestion
logger.info("Beginning opacity solve.")
```

jvshields: Earlier point of discussion when I put in the logger statements at first. Should be safe enough to apply.

tardis/workflows/tests/test_iip_workflow.py
```diff
@@ -19,11 +25,142 @@ def _max_rel_diff(actual, expected):
     return float(np.nanmax(relative_difference))
 
 
+def _as_regression_dataframe(value: Any) -> pd.DataFrame:
```
jvshields, line 28: This PR has gotten big and I'll sign off on this here, but I want to point out for future LLM generated tests, we should be careful about doing generic stuff like this in specific test files. If this is something we want to be able to do, it should go in a much higher level of tardis testing structure. 

I also hate (value: Any) on principle.

andrewfullard:
Closing this PR in favour of #3649

PR #3626, opened 2026-06-25
Handle TARDIS config validator schema paths differently

connor-mcclellan:
### :pencil: Description

**Type:** :beetle: `bugfix`

The TARDIS config validator has hardcoded paths for the schemas it compares against. STARDIS currently uses the TARDIS config validator for its own schemas which are in its own directory, so these hardcoded paths can't be overridden.

This is a small change that moves the `SCHEMA_DIR` from a hardcoded path to an argument of the `_create_schema_registry` method. This will let STARDIS use the config validator from TARDIS rather than having to duplicate its own version of a config validator, while still allowing it to manage its own config schemas.

- [X] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/io/configuration/config_validator.py
+++ b/tardis/io/configuration/config_validator.py
@@ -80,7 +80,7 @@ def extend_with_default(
 DefaultDraft7Validator = extend_with_default(Draft7Validator)
 
 
-def _create_schema_registry(schema_dir: Path) -> Registry:
+def _create_schema_registry(schema_dir: Path = SCHEMA_DIR) -> Registry:
     """
     Create a registry containing all schema files for reference resolution.
 
@@ -143,7 +143,7 @@ def is_quantity(checker: Any, instance: Any) -> bool:
 
 def validate_dict(
     config_dict: dict[str, Any],
-    schemapath: Path = CONFIG_SCHEMA_FNAME,
+    schemapath: str | Path = CONFIG_SCHEMA_FNAME,
     validator: type[Draft7Validator] = DefaultDraft7Validator,
 ) -> dict[str, Any]:
     """
@@ -158,7 +158,7 @@ def validate_dict(
     ----------
     config_dict : dict[str, Any]
         The configuration dictionary to validate.
-    schemapath : Path, optional
+    schemapath : str or Path, optional
         Path to the main schema file. Defaults to CONFIG_SCHEMA_FNAME.
     validator : type[Draft7Validator], optional
         The validator class to use. Defaults to DefaultDraft7Validator.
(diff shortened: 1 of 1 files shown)


andrewfullard:
Is this testable?

tardis/io/configuration/config_validator.py
```diff
@@ -102,7 +107,7 @@ def _create_schema_registry() -> Registry:
     registry = Registry()
 
     # Load all schema files in the schemas directory
-    for schema_file in SCHEMA_DIR.glob("*.yml"):
+    for schema_file in schema_dir.glob("*.yml"):
```
jvshields, line 110: As far as I can tell, this file now doesn't use the SCHEMA_DIR constant is now orphaned. Should it be removed?

Also, does that path need to be provided now in a default case wherever _create_schema_registry is called? Maybe the correct thing to do is like
```def _create_schema_registry(schema_dir: Path = SCHEMA_DIR) -> Registry:```

**Comment by connor-mcclellan** [comment]
> Is this testable?

This is the exact same as what it was doing before (when the path was hardcoded), just flexible enough to accept something else instead of the hardcoded value. I'm not sure what a test would accomplish.

Codex suggests something like this, but I don't really think it's necessary. All it's doing is confirming that you can pass either a Path or a string and have it successfully find parameters from that schema. 

```
def test_validate_dict_accepts_string_schemapath(tmp_path):
    """Test validation with a custom schema path passed as a string."""
    main_schema = tmp_path / "base.yml"
    referenced_schema = tmp_path / "section.yml"

    main_schema.write_text(
        """
type: object
properties:
  section:
    $ref: section.yml
required:
  - section
"""
    )
    referenced_schema.write_text(
        """
type: object
properties:
  value:
    type: number
    default: 3
"""
    )

    validated_config = validate_dict({"section": {}}, schemapath=str(main_schema))

    assert validated_config["section"]["value"] == 3
```

File: tardis/io/configuration/config_validator.py (hunk 1 above)
Comment (connor-mcclellan) [reply tardis/io/configuration/config_validator.py:110]: Not quite orphaned, just slightly redundant. We could collapse this to a single path to `base.yml` now.

```
CONFIGURATION_DIR = Path(__file__).resolve().parent
SCHEMA_DIR = CONFIGURATION_DIR / "schemas"
CONFIG_SCHEMA_FNAME = SCHEMA_DIR / "base.yml"
```

I'm not opposed to setting the default to `SCHEMA_DIR`---good idea

jvshields (approved):

andrewfullard (approved):

PR #3630, opened 2026-06-29
Fix stardis-tests action

### LINKED ISSUE #3381: Skip stardis tests in our development pipeline, or make an expected failur

connor-mcclellan:
### :pencil: Description

**Type:** :beetle: `bugfix`| :vertical_traffic_light: `testing`

Fixes #3381.

When https://github.com/tardis-sn/stardis/pull/288 is merged into STARDIS, this action will again function as a check to make sure a TARDIS PR doesn't break STARDIS.

The specific change in this PR to acquire atom data from the LFS cache (like TARDIS does) instead of trying to download the atom data from a URL. The previous `download_from_url` function no longer works since we use LFS storage now.

- [X] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/.github/workflows/stardis-tests.yml
+++ b/.github/workflows/stardis-tests.yml
@@ -22,7 +22,15 @@ env:
   CACHE_NUMBER: 0  # increase to reset cache manually
 
 jobs:
+  test-cache:
+    uses: ./.github/workflows/lfs-cache.yml
+    with:
+      atom-data-sparse: true
+      regression-data-repo: tardis-sn/tardis-regression-data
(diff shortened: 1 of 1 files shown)


andrewfullard (approved):
Hooray!

jvshields (approved):

PR #3631, opened 2026-06-29
Rework first time contributor github action

connor-mcclellan:
Very quick Codex rewrite of the `check-first-time-committer` utility action to not use `BOT_TOKEN`, which was causing failures. Not sure if it works yet, but we'll see!

### DIFF
--- a/.github/workflows/util.yml
+++ b/.github/workflows/util.yml
@@ -42,23 +42,55 @@ jobs:
             });
 
   check-first-time-committer:
+    if: >
+      github.event.action == 'opened' &&
+      contains(fromJSON('["FIRST_TIMER", "FIRST_TIME_CONTRIBUTOR"]'), github.event.pull_request.author_association)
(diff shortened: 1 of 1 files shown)


andrewfullard:
"Skipping...Not an Opened Event" is reported by the action. I don't think that's the correct reason!

connor-mcclellan:
> "Skipping...Not an Opened Event" is reported by the action. I don't think that's the correct reason!

I think you were looking at the output from the second commit's action. The first looks like this:

```
Run actions/first-interaction@v3
Running actions/first-interaction!
Skipping...Not First Contribution
```

It'll only check if you're a first-time contributor for the initial opening of the PR, so that's intended behavior. Every commit to the PR after this, it should skip because it isn't the opening event.

andrewfullard (approved):

jvshields (approved):

PR #3636, opened 2026-07-02
Fix artis model read isotopic behavior

jvshields:
Our ARTIS model reader wasn't actually hooked up when called from the config, and would instead just fall back to a default ascii model reader instead. Also, the ARTIS model reader didn't pass along the isotopic information necessary for high energy workflow things. Finally, when hooking those things up correctly, you get incorrect slicing behavior that causes tardis to crash when it tries to calculate cell masses because of misaligned density and mass fraction dataframes in the composition object. This PR fixes those problems and makes it so that you can run a tardis sim from an artis model including isotope information. 

Also includes some tests to verify that it's working correctly. I'm not entirely done proofing this PR but it MIGHT be fine to merge as is so I'm not marking it as draft. 

Oh and there's also some more RUFF application, which is always nice.


tardis/io/model/artis/data.py
```diff
@@ -1,10 +1,15 @@
...
+
 from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
 
-import numpy as np
 import pandas as pd
-from astropy import units as u
 
-from tardis.model.geometry.radial1d import HomologousRadial1DGeometry
+if TYPE_CHECKING:
```
andrewfullard: Feel free to drop this block

tardis/io/model/parse_geometry_configuration.py
```diff
@@ -63,7 +63,7 @@ def parse_structure_from_config(config: Configuration):
             electron_densities,
             temperature,
         ) = read_density_file(structure_config_fname, structure_config.filetype)
-        density = density.insert(0, 0)
+        # density = density.insert(0, 0) # Commented out from previous inclusion. This is explicitly checked later.
```
andrewfullard: Possibly delete? This is potentially a breaking change that we don't test

jvshields: Delete the commenting out? Or delete the now comment?

andrewfullard: Take your pick! Either we stick with this change in which case delete the line, or you keep the line and leave the comment noting that it's a duplicate operation.

connor-mcclellan:
Merge master in to fix the test tolerances and then we can see if these pass

tardis/io/model/artis/tests/test_artis_readers.py
```diff
@@ -25,17 +37,18 @@ def artis_density_fname(artis_data_dir):
...
-    )
+    time_model, velocity, mean_density = read_artis_density(artis_density_fname)
     # Check that time is recognized as time
     assert time_model.unit.physical_type == "time"
     # Check velocity unit is cm/s
     assert velocity.unit == u.Unit("cm/s")
-    # Dummy check for mean_density value
-    npt.assert_allclose(mean_density.value[0], 10**-5.449497e-05)
+    assert len(mean_density) == len(velocity) - 1
+    # The first ARTIS row is the unused inner shell.
```
andrewfullard, line 46: We should think about the fencepost problem as usual, and how artis records velocity, density etc

tardis/io/model/artis/data.py
```diff
@@ -25,6 +26,8 @@ def to_geometry(self) -> HomologousRadial1DGeometry:
         tardis.model.geometry.radial1d.HomologousRadial1DGeometry
             The geometry object constructed from the ARTIS data.
         """
+        from tardis.model.geometry.radial1d import HomologousRadial1DGeometry
```
andrewfullard: Not a fan of internal imports like this

tardis/io/model/artis/readers.py
andrewfullard, line 1: Missing full numpy docstrings for new functions

tardis/io/model/artis/data.py
jvshields: Yeah that one is particularly weird since HomologousRadial1DGeometry is imported at the top of the file anyway. Surprised ruff didn't complain about a redundant import like this.

andrewfullard (approved):

tardis/io/model/artis/tests/test_artis_readers.py
```diff
@@ -84,3 +111,104 @@ def test_simple_legacy_read_artis_density(artis_density_fname: str):
     )
     assert len(mean_density) == 69
     assert len(velocity) == len(mean_density) + 1
+
+
+def test_artis_generic_readers_shell_slicing(artis_data_dir):
```
connor-mcclellan, line 116: Are we moving to "cell" instead of "shell"?

jvshields: Yes, though I figured that if artis is shells then using that word might be more correct here. Since it's an artis reader it'd make sense to go by what they use rather than what we have now decided.

That said, I'm also realizing I am making that assumption but don't know it's true that they are strictly shell-based coordinates.

tardis/io/model/artis/readers.py
```diff
@@ -7,99 +6,131 @@
...
+    downstream composition validation.
     """
+    isotope_element_mass_fractions = isotope_mass_fractions.groupby(
+        level="atomic_number"
+    ).sum()
+    residual_elemental_mass_fractions = elemental_mass_fractions.sub(
+        isotope_element_mass_fractions, fill_value=0.0
+    )
+
+    roundoff_negative = (residual_elemental_mass_fractions < 0.0) & (
```
connor-mcclellan, line 66: Machine precision can bias the result in either direction of zero, right?

(1.0 ± e) - 1.0 = ± e

I think you might want to mask out everything within e from 0, but I don't think it'll make a huge difference if you leave it as-is either. It would just mean you don't end up masking things that do actually subtract to zero, giving you +e instead. Not sure how much that matters for this case and can definitely see that negative values would be the bigger problem

tardis/io/model/artis/readers.py
```diff
@@ -131,16 +162,67 @@ def read_artis_mass_fractions(
...
+            mass_fractions_df.sum(axis=1), axis=0
+        )
     mass_fractions_df = mass_fractions_df.T
     mass_fractions_df.index.name = "atomic_number"
     mass_fractions_df.columns.name = "cell_index"
 
     return mass_fractions_df
 
 
+def _read_artis_abundances(
```
connor-mcclellan, line 178: Not sure why this is a private function?

tardis/io/model/artis/readers.py
```diff
@@ -7,99 +6,131 @@
...
+        isotope_element_mass_fractions, fill_value=0.0
+    )
+
+    roundoff_negative = (residual_elemental_mass_fractions < 0.0) & (
+        residual_elemental_mass_fractions >= -np.finfo(np.float64).eps
+    )
+    return residual_elemental_mass_fractions.mask(roundoff_negative, 0.0)
+
+
+def _read_artis_structure(fname: str | Path) -> ArtisData:
```
connor-mcclellan, line 72: Not sure why this is a private function?

jvshields: This is helpful. This one is private in theory because it's an underlying method that doesn't yield the useful processed objects you'd actually want to work with, but I think what I should really do is change each of these names to be clear what each of them really do. This one returns the ArtisData dataclass, which is important and should be more explicit.

connor-mcclellan (approved):

PR #3642, opened 2026-07-07
Restructure/index rename

wkerzendorf:


tardis/energy_input/transport/gamma_packet_loop.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 from numba import njit
+from numpy.typing import NDArray  # noqa: TC002
```
andrewfullard, line 3: Guess we need to skip TC002 in the ruff toml

tardis/opacities/opacity_state_numba_iip.py
```diff
@@ -16,7 +16,7 @@ class OpacityStateNumbaIIP:
     tau_sobolev: nb.float64[:, :]  # type: ignore[misc]
     transition_probabilities: nb.float64[:, :]  # type: ignore[misc]
     line2macro_level_upper: nb.int64[:]  # type: ignore[misc]
-    macro_block_references: nb.int64[:]  # type: ignore[misc]
```
jvshields, line 19: Change these similarly in the opacity_state_numba.py file, and I believe the tests will pass.

jvshields (approved):

andrewfullard (approved):

PR #3646, opened 2026-07-09
Incorporate homology into more general non-homologous expansion

jvshields:
Codex driven attempt at removing the homologous geometry. We decided to incorporate homology into the more general case and have the homologous geometry handle specific case conversions. 

Required due to fixture rename:

https://github.com/tardis-sn/tardis-regression-data/pull/110


tardis/transport/montecarlo/modes/homologous_rad_packet_transport.py
```diff
@@ -21,7 +21,9 @@
 )
 
 if TYPE_CHECKING:
```
connor-mcclellan: Didn't we want to remove these blocks and just import the stuff needed for type checking?

connor-mcclellan (commented):
I like the approach and think it makes sense to keep the actual homologous geometry structures around as a special case of nonhomology. Makes sense to do that for consistency with existing tests. Approved pending the regression data merge

tardis/simulation/base.py
```diff
@@ -781,13 +783,14 @@ def from_config(
...
                 packet_source=simulation_state.packet_source,
                 enable_virtual_packet_logging=virtual_packet_logging,
             )
         if opacity is None:
-            opacity = OpacitySolver(
+            opacity = NonhomologousOpacitySolver(
+                simulation_state.geometry.velocity_gradient,
                 config.plasma.line_interaction_type,
                 config.plasma.disable_line_scattering,
             )
```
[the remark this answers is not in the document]
connor-mcclellan: ~~^ this is worth looking into and I'm not sure if this case is covered by our tests~~

outdated comment

tardis/workflows/tests/test_workflows.py
```diff
@@ -3,6 +3,9 @@
 import pytest
 from astropy import units as u
 
+from tardis.transport.montecarlo.modes.nonhomologous.solver import (
```
jvshields: Is this import stranded? I don't see an added usage of MCTransportSolverNonhomologous

tardis/io/model/artis/data.py
andrewfullard, line 1: My remaining concern about this change is that we are changing the default way external models are imported. I am not sure how this interacts with running tardis since by default it will prefer the homologous geometry.

jvshields:
Codex also points out this. I might leave it as an issue but I would figure that worrying about infall right now isn't worth it for this PR. 

[P1] Velocity-to-radius inversion assumes a globally monotonic velocity profile. [radial1d.py (line 117)](/Users/joshuashields/codes/tardis/tardis/tardis/model/geometry/radial1d.py:117) uses searchsorted, but Radial1DGeometry permits gradients to change sign between shells. For a profile with velocities [10, 20, 15], setting the outer boundary to 17 maps it to radius 1.7 cm on the inner branch instead of 2.6 cm on the outer branch. This silently selects the wrong active volume. The inversion needs to be shell/boundary-aware, with tests for non-monotonic and zero-gradient profiles.

andrewfullard (approved):

connor-mcclellan (approved):

PR #3649, opened 2026-07-13
IIP Plasma Performance improvements

andrewfullard:
### :pencil: Description

**Type:** :beetle: `bugfix`

This PR is intended to test that performance improvements pass all tests reliably.

Based on #3651 

- [x] Testing pipeline
- [ ] Other method (describe)


andrewfullard:
Benchmark failures are expected, because this PR adds new benchmarks for new functions. Happy to discuss removing the benchmarks if they don't seem useful.

benchmarks/iip_plasma.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+    def time_cumulative_integrate_array_by_blocks(self):
+        cumulative_integrate_array_by_blocks(
+            self.alpha_sp_energy,
+            self.frequency,
+            self.block_references,
+        )
+
+    def time_normalize_transition_probabilities(self):
```
jvshields, line 84: Not sure about the other ones but this does seem like a weird check. I'll point out that I disabled the diagonalize_ma call on the base continuum object because we don't use the ctardis processed probabilities, which I think is the only place this is called on anything substantial. Otherwise I think it's called in one or two of the individual transitions, but I can't imagine those matter much.

tardis/iip_plasma/properties/partition_function.py
```diff
@@ -26,6 +26,77 @@
 ]
 
 
+@njit(error_model="numpy")
```
jvshields, line 29: Slightly worried about the error model. I would hope it either consistently works compiled, or it doesn't. Fallback cases are scary because they lead to silently inconsistent behavior.

andrewfullard: This is the error model we use in our default `njit` configuration. This just avoids fastmath.

tardis/opacities/macro_atom/tests/test_absorbing_markov_chain.py
```diff
@@ -0,0 +1,104 @@
...
+import pandas as pd
+import scipy
+
+from tardis.opacities.macro_atom.absorbing_markov_chain import (
+    create_absorbing_probs,
+)
+
+
+def _create_absorbing_probs_reference(
+    transition_probabilities: pd.DataFrame, metadata: pd.DataFrame
```
jvshields, line 11: This looks exactly like the code in tardis/opacities/macro_atom/absorbing_markov_chain.py. I'm generally not a fan of exporting the code to a test function and testing that code against the actual function. But maybe this is to test against the changed parallelized method implemented?

tardis/opacities/macro_atom/absorbing_markov_chain.py
```diff
@@ -52,9 +57,9 @@ def create_absorbing_probs(
     """
     num_cells = transition_probabilities.shape[1]
 
-    num_states = len(metadata.source.unique())
     internal_mask = metadata.transition_type >= 0
     internal_jump_probs = transition_probabilities[internal_mask]
+    num_states = int(metadata.source_level_idx.max()) + 1
```
jvshields: I have remembered that the reason I didn't do this originally is if you ever drop any of the source blocks, this logic falls apart. Currently I don't believe this is something we ever do, but if you ever wanted to do the partial reduction on some subset of elements this doesn't work anymore.

tardis/opacities/macro_atom/tests/test_absorbing_markov_chain.py
andrewfullard: Correct. This is the previous implementation as reference for the new one to make sure it does the same thing.

tardis/workflows/type_iip_workflow.py
```diff
@@ -932,9 +933,9 @@ def run(self):
             )
```
jvshields:
```suggestion
logger.info("Opacity solve started.")
```

Looks like this got dropped from the last PR and was a point of discussion.

jvshields:
Oops looks like the github suggested change to include the logger message didn't include indentation, so now everything breaks.

tardis/opacities/macro_atom/absorbing_markov_chain.py
jvshields: I think I'm fine with all the other changes, but I would like a note to be left here in a comment or something pointing out what I said above. I wouldn't be that surprised if we want to do the partial stuff at some point, and it would be helpful to point out that this is a potential failure point.

jvshields (approved):

connor-mcclellan (approved):

PR #3651, opened 2026-07-13
Additional IIP workflow tests

andrewfullard:
### :pencil: Description

**Type:**  :vertical_traffic_light: `testing`

Checking the tests vs Mac and Linux runners

### :pushpin: Resources

- [x] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/workflows/tests/test_iip_workflow.py
+++ b/tardis/workflows/tests/test_iip_workflow.py
@@ -65,7 +65,6 @@ PLASMA_SOLVER_REGRESSION_OUTPUTS = (
     "link_t_rad_t_electron",
     "p_fb_deactivation",
     "chi_bf",
-    "fractional_heating",
     "sp_fb_cooling_rates",
     "stimulated_emission_factor",
     "b",
@@ -90,13 +89,10 @@ def iip_regression_path(tardis_regression_path):
 
 
 @pytest.fixture
-def thermal_balance_guess() -> Callable[
-    [Any], tuple[np.ndarray, np.ndarray]
-]:
+def thermal_balance_guess() -> Callable[[Any], tuple[np.ndarray, np.ndarray]]:
     def build_guess(plasma_solver: Any) -> tuple[np.ndarray, np.ndarray]:
         max_electron_number_density = (
-            plasma_solver.number_density
-            .multiply(
+            plasma_solver.number_density.multiply(
                 plasma_solver.number_density.index.values,
                 axis=0,
             )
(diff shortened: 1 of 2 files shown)


tardis/workflows/tests/test_iip_workflow.py
```diff
@@ -1,12 +1,20 @@
...
 import pytest
 
 from tardis.iip_plasma.standard_plasmas import LegacyPlasmaArray
-from tardis.io.atom_data import AtomData
 from tardis.io.configuration.config_reader import Configuration
 from tardis.workflows.type_iip_workflow import TypeIIPWorkflow
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Any
```
jvshields: Any way we can be more strict on typing so we don't need either of these imports?

tardis/workflows/tests/test_iip_workflow.py
```diff
@@ -1,12 +1,20 @@
+from __future__ import annotations
```
connor-mcclellan: This is almost always added by codex and never used anywhere, right? I've seen this a bunch of times, still no idea what it does

connor-mcclellan (dismissed):

tardis/workflows/tests/test_iip_workflow.py
jvshields: It's something you needed in an earlier version of python to do the conditional type checking import stuff. I think we decided we don't want to do the conditional import anyway, but even if we did you don't need this anymore on python 3.14.

tardis/workflows/tests/test_iip_workflow.py
```diff
@@ -19,11 +27,91 @@ def _max_rel_diff(actual, expected):
...
+    if isinstance(value, pd.Series):
+        return value.to_frame("value")
+
+    values = np.asarray(value)
+    if values.ndim == 1:
+        return pd.DataFrame({"value": values})
+    return pd.DataFrame(values)
+
+
+def _assert_regression_dataframe(
```
jvshields: Still kinda think that if this and the previous method are things we want, they should go in tardisbase and not here, so they're more coupled to the regression data itself.

tardis/workflows/tests/test_iip_workflow.py
andrewfullard: Yep, has to become a regular function though.

tardis/workflows/tests/test_iip_workflow.py
andrewfullard: Not exactly conditional type checking, rather things like `var: int | float`. The OR operator for type checking.

tardis/workflows/tests/test_iip_workflow.py
andrewfullard: I agree. But I didn't want to have to depend on a new tardisbase for this.

andrewfullard: I've merged them together at least to simplify it and make it clearer. There's no actual defined type for regression_data (see https://github.com/tardis-sn/tardisbase/blob/master/tardisbase/testing/regression_data/regression_data.py ) so the type hint here is not entirely correct.

andrewfullard: Never mind, merging them is a bad idea due to early returns.

andrewfullard: Put them into conftest which seems more appropriate. As with many things, we could use a cleanup pass on test helpers and fixtures (not just because of LLMs)

jvshields (dismissed):

jvshields (approved):

connor-mcclellan (approved):

PR #3652, opened 2026-07-13
bump rtol of stim emission factor test for linux github runner

jvshields:
A repeat of this run should work now https://github.com/tardis-sn/tardis/actions/runs/29276448895/job/86906816472?pr=3636

Bumped rtol from 2e-14 to 1e-12. Failure was a diff of 4e-13, so a factor of ~2 below.

### DIFF
--- a/tardis/plasma/properties/tests/test_basic_properties.py
+++ b/tardis/plasma/properties/tests/test_basic_properties.py
@@ -4,7 +4,6 @@ import pandas as pd
 import pandas.testing as pdt
 import pytest
 
-from tardis import constants as const
 from tardis.plasma.base import BasePlasma
 from tardis.plasma.exceptions import IncompleteAtomicData
 from tardis.plasma.properties.atomic import IonizationData, Levels, Lines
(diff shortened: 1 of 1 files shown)


tardis/plasma/properties/tests/test_basic_properties.py
```diff
@@ -124,16 +122,16 @@ def test_stimulated_emission_factor_lte_values_not_above_one(
 def test_stimulated_emission_factor_regression(
     stimulated_emission_factor, regression_data
 ):
-    expected = regression_data.sync_ndarray(
-        stimulated_emission_factor
-    )
+    expected = regression_data.sync_ndarray(stimulated_emission_factor)
     npt.assert_allclose(
-        stimulated_emission_factor, expected, atol=0, rtol=2e-14
+        stimulated_emission_factor, expected, atol=0, rtol=1e-13
```
andrewfullard: This will still fail- it should be 4e-13 for the failing case to pass

andrewfullard (changes requested):
The ruff changes are odd, check the line width

tardis/plasma/properties/tests/test_basic_properties.py
jvshields: Ah woops, good catch. I'll adjust it to 1e-12 so we can be a factor of ~2 different.

andrewfullard (approved):

connor-mcclellan (approved):

PR #3657, opened 2026-07-16
Migrate Grotrian Widget to Panel

SS-9098:
### :pencil: Description

Replaces Jupyter Widgets in grotrian widget with Panel. The functionality and behaviour of the widget remains the same.

### :pushpin: Resources

- [ ] Testing pipeline
- [ ] Other method (describe)


SS-9098:
New UI:

image

tardis/visualization/widgets/grotrian.py
```diff
@@ -1120,96 +1120,86 @@ def _change_handler(self, attribute, value):
...
 
         if min_wavelength is None or max_wavelength is None:
-            self.wavelength_range_selector.layout.visibility = "hidden"
+            self.wavelength_range_selector.visible = False
+            self.wavelength_range_selector.disabled = True
             return
 
-        elif min_wavelength == max_wavelength:
-            self.wavelength_range_selector.layout.visibility = "visible"
+        if f"{min_wavelength:.1e}" == f"{max_wavelength:.1e}":
```
andrewfullard, line 1168: Add a comment about the rounding choice here

andrewfullard (dismissed):

hmperk18 (approved):

tardis/visualization/widgets/grotrian.py
```diff
@@ -1120,96 +1120,87 @@ def _change_handler(self, attribute, value):
...
-        index = self.fig.children.index(self.plot.fig)
+        atomic_number, ion_number = species_string_to_tuple(change.new)
         self.plot.set_ion(atomic_number, ion_number)
-
-        # Set the updated plot in the figure
-        children_list = list(self.fig.children)
-        children_list[index] = self.plot.display()
-        self.fig.children = tuple(children_list)
-        # self._wavelength_resetter()
+        self._wavelength_resetter()
```
jaladh-singhal, line 1138: I notice that `self._wavelength_resetter()` is being called here and in init and get_species(). Earlier it wasn't being called here but in observe/watch functions (which is now removed). Is there any advantage of doing this post panel migration?

tardis/visualization/widgets/grotrian.py
```diff
@@ -1019,79 +1020,78 @@ def __init__(self, plot, num_shells, **kwargs):
...
+            value=(self.plot.min_wavelength, self.plot.max_wavelength),
+            start=self.plot.min_wavelength,
+            end=self.plot.max_wavelength,
             step=0.1,
-            description="Wavelength",
-            layout=ipw.Layout(width="605px"),
-            readout_format=".1e",
+            format=PrintfTickFormatter(format='%0.1e'),
+            margin=(5, 10, 5, 65),
+            width= 605,
```
jaladh-singhal: I think this is too wide for Panel widget - it was working for ipywidget. It's causing horizontal scrollbars (note the bottom most scrollbar in [the docs build](https://tardis-sn.github.io/tardis/pull/3657/analyzing_tardis/visualization/how_to_generating_widgets.html#Energy-Level-Diagram)) which is not wrong but less ideal:

Image

jaladh-singhal (commented):
@SS-9098 this is looking good. Thanks for maintaining attention to detail regarding layout this time - it looks clean.

I have one cosmetic suggestion about extra wideness - see if you can tackle that. Otherwise it's fine as is.

Also, can you confirm the interactivity of wavelength slider is working as before since there are some changes in the logic?

tardis/visualization/widgets/grotrian.py
SS-9098: The widgets this was previously attached to already have another watch callback that invokes a different function, so they end up with two event listeners attached. I think it would be better to call this function from within the function that the existing listener invokes instead, so each widget has one event listener. This also avoids potential race conditions where wavelength_resetter() could be called first, which we do not want.

tardis/visualization/widgets/grotrian.py
SS-9098: This looks fine in notebooks, but is too wide in docs for some reason. I tried reducing all the width values, but that doesn't seem to be effective.

jaladh-singhal (approved):
Thanks for trying to address the width comment, it's fine if it's not working just for docs.

Good to merge!

andrewfullard (approved):

PR #3654, opened 2026-07-15
Remove old continuum/NLTE implementation

andrewfullard:
### :pencil: Description

**Type:** :biohazard: `breaking change` 

This completely removes the previously implemented continuum processes and NLTE solvers because:

1. Currently all continuum NLTE solutions are handled as part of the iip_plasma
2. The equilibrium plasma module should be replacing this soon(tm)

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #3661, opened 2026-07-20
Remove continuum matrix from workflows

andrewfullard:
### :pencil: Description

We don't use the continuum marker for tests any more (and definitely not when #3654 is merged.


wkerzendorf (approved):
I just want to make sure that if there is tests that are now no longer run, we should make sure they are removed also from regression data

PR #3668, opened 2026-07-24
Fix docs publish workflow bugs

andrewfullard:
### :pencil: Description

**Type:** :beetle: `bugfix` | :roller_coaster: `infrastructure`

There are still a couple of bugs with the docs publishing workflow (which handles PR-specific docs builds and comments)

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -102,15 +102,11 @@ jobs:
           edit-mode: replace
           body: |
             <!-- tardis:docs-preview -->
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -104,26 +106,11 @@ jobs:
 
             Hi, human.
```
connor-mcclellan: can we strip the bot of its personality. I hate it

connor-mcclellan (approved):

PR #3671, opened 2026-07-27
Stricter Ruff configuration

andrewfullard:
### :pencil: Description

Ruff can handle a lot of our design requirements for code, and at least notify developers about them automatically.

- Add type checking block ignore because we don't care about the performance overhead
- Remove static typing ignores. We want to include type hints.
- Remove blank except ignores. We don't want blank except clauses.
- Remove undocumented class/function/etc ignores. We want all our definitions to have docstrings.
- Remove docstring formatting ignores. We want docstrings to be clearly formatted and valid.
- Remove "fixme" and "hack" ignores. I don't think we use those anywhere, but we also don't want to use them anywhere in the future
- Remove PEP8 naming ignores. This handles variable, class, function naming style (but not content)
- Remove Path lib ignores. We want to use the Path library.

### :pushpin: Resources

https://docs.astral.sh/ruff/rules/


.ruff.toml
```diff
@@ -5,13 +5,6 @@ lint.ignore = [
     # NOTE: to find a good code to fix, run:
     # ruff --select="ALL" --statistics astropy/<subpackage>
```
wkerzendorf: should we remove astropy?

.ruff.toml
```diff
@@ -131,27 +90,6 @@ lint.ignore = [
     "PT022",   # pytest-useless-yield-fixture
     "PT023",   # pytest-incorrect-mark-parentheses-style
 
-    # flake8-use-pathlib (PTH)
-    "PTH100",  # os-path-abspath
```
wkerzendorf, line 133: don't we want that flagged? we want to use path lib

.ruff.toml
```diff
@@ -81,23 +57,6 @@ lint.ignore = [
     "FBT002",  # boolean-default-value-in-function-definition
     "FBT003",  # boolean-positional-value-in-function-call
 
-    # flake8-fixme (FIX)
-    "FIX001",  # Line contains FIXME.  this should be fixed or at least FIXME replaced with TODO
-    "FIX004",  # Line contains HACK. replace HACK with NOTE.
-
-    # pep8-naming (N)
```
wkerzendorf, line 86: we want that no?

wkerzendorf (dismissed):
looks good - misunderstood at first

connor-mcclellan (dismissed):

PR #3675, opened 2026-07-27
Opacity and Transport state disambiguation

### LINKED ISSUE #3206: Rename simulation.transport.transport_state.opacity_state
This opacity state is the numba version of the opacity_state, which should be renamed into something else more informative, such as `simulation.transport.transport_state.opacity_state_numba`.

In case when we need to access information within the opacity state, this should not be the one that we use, see relevant PR #3205 .

andrewfullard:
### :pencil: Description

Removes unused opacity state init functions and renames properties of the transport state to more accurately reflect their type. Also improves docstrings and type hinting for opacity states so they are easier to use.

Fixes #3206 

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/opacities/opacity_state.py
```diff
@@ -35,26 +41,36 @@
...
-        line_list_nu,
-        tau_sobolev,
-        beta_sobolev,
-        continuum_state,
-    ):
+        electron_density: pd.DataFrame,
+        t_electrons: npt.NDArray[np.float64],
+        line_list_nu: pd.Series,
+        tau_sobolev: pd.DataFrame,
+        beta_sobolev: pd.DataFrame | None,
```
connor-mcclellan, line 55: I vaguely remember @jvshields mentioning he doesn't like this but I don't remember why. Have you considered whether `| None` is the right thing to do here? If I recall correctly it may throw off the VSCode snippets or something similar because it can't get attributes of `None`, but I haven't checked that carefully

tardis/opacities/opacity_state_numba.py
```diff
@@ -38,31 +45,31 @@ class OpacityStateNumba:
...
-        x_sect: np.ndarray,
-        phot_nus: np.ndarray,
-        ff_opacity_factor: np.ndarray,
-        emissivities: np.ndarray,
-        photo_ion_activation_idx: np.ndarray,
+        electron_density: npt.NDArray[np.float64],
+        t_electrons: npt.NDArray[np.float64],
+        line_list_nu: npt.NDArray[np.float64],
+        tau_sobolev: npt.NDArray[np.float64],
+        transition_probabilities: npt.NDArray[np.float64],
```
connor-mcclellan, line 54: Do we really want to get `ndarray` from numpy testing? I don't understand this preference unless it's Codex trying to avoid importing numpy

andrewfullard: It's numpy typing, not numpy testing

connor-mcclellan: Oh wait, npt is numpy.typing, not numpy testing

tardis/transport/montecarlo/montecarlo_transport_state.py
```diff
@@ -1,8 +1,17 @@
+from __future__ import annotations
```
connor-mcclellan: Codex scraps. No idea why it always adds this

tardis/opacities/opacity_state.py
```diff
@@ -1,11 +1,17 @@
+from __future__ import annotations
```
connor-mcclellan: Codex leftovers

connor-mcclellan: After looking this up it's backward compatibility for type hinting.
https://stackoverflow.com/questions/61544854/from-future-import-annotations

> Since this change breaks compatibility, the new behavior needs to be enabled on a per-module basis in Python 3.7 using a `__future__` import:
> 
> `from __future__ import annotations`

We certainly don't need to support anything as old as Python 3.7

tardis/transport/montecarlo/montecarlo_transport_state.py
connor-mcclellan: After looking this up it's backward compatibility for type hinting.
https://stackoverflow.com/questions/61544854/from-future-import-annotations

> Since this change breaks compatibility, the new behavior needs to be enabled on a per-module basis in Python 3.7 using a `__future__` import:
> 
> `from __future__ import annotations`

We certainly don't need to support anything as old as Python 3.7

tardis/transport/montecarlo/tests/test_macro_atom.py
```diff
@@ -5,7 +5,7 @@
 
 @pytest.mark.parametrize(
     ["seed", "expected"],
-    [(1963, 10015), (1, 9993), (2111963, 17296), (10000, 9993)],
+    [(1963, 10015), (1, 9993), (2111963, 13494), (10000, 9993)],
```
connor-mcclellan, line 8: This is scary

connor-mcclellan (changes requested):

tardis/transport/montecarlo/tests/test_macro_atom.py
andrewfullard: The test previously used the plasma ordering of line2macro_level_upper in the opacity state. The to_numba() method that builds the opacity state for the updated test uses the MacroAtomState ordering, which removes empty blocks. This changes the transition block offset and thus the transition order when the probabilities are summed up across blocks. Currently, tardis uses to_numba() everywhere anyway, and the plasma ordering is not used. So, the test value was just outdated.

connor-mcclellan (approved):

PR #3682, opened 2026-08-03
IIP segfault fix

andrewfullard:
Also adds supporting tests

### :pencil: Description

**Type:** :beetle: `bugfix` 

Fixes segfaults caused by NaN/inf values of stimulated recombination number densities. Uses an alternate calculation path that doesn't use a division, so div/0 isn't possible. The calculation is numerically equivalent.

Also handles zero or negative values in the cumulative integral.

- [ ] Testing pipeline
- [ ] Other method (describe)


andrewfullard:
Codecov failure is because the changes are behind Numba

tardis/plasma/array_util.py
```diff
@@ -112,6 +112,10 @@ def cumulative_integrate_array_by_blocks(f, x, block_references):
                     (x[k] - x[k - 1]) * (f[k, i] + f[k - 1, i]) / 2.0
                 )
                 integrated[k, i] = cumulative_integral
+            if cumulative_integral <= 0.0 or not np.isfinite(
+                cumulative_integral
+            ):
+                continue
```
connor-mcclellan: What is the intended outcome for accumulations where the integral becomes negative? If I'm reading this right, this `continue` breaks out of the `j` loop if the loop over `k` produces something less than zero or infinite, but it doesn't overwrite or zero out the value-- it just stops accumulating. So you could still end up with negative results in `integrated[k, i]`. I'm not super familiar with this part of the code but this seems strange, and the PR Review codex skill also flagged this:

> Zero-rate recombination channels in Monte Carlo sampling
>
> The changes replace undefined normalizations with all-zero probability columns or unnormalized cumulative integrals. Those arrays are finite for the tested zero case, but they are not valid probability distributions; if transport selects such a channel, its deactivation choice or emitted frequency is undefined.
>
> Does macro-atom transition selection guarantee that zero-total recombination channels can never reach bound-free cooling?

andrewfullard: Yeah this is definitely the change I am most unsure about, and want to talk to Josh about as well, probably

andrewfullard: Implemented a nicer fix. CDF still zeros out but not arbitrarily in the integrator.

jvshields:
I think that this is fine if slightly wrong and might be worrying if this ends up getting applied to nebular things ever. I am specifically worried about casting infinites to 0s, since you could imagine something just being the correct exit from a state but killing that exit path instead. Probably the only physical places this comes up for us currently is in low density environments where interactions don't happen often, so you don't care that you're doing the wrong thing. 

I would guess to do it right you'd have to build in some limit comparison step or other approximation to safeguard against the multiple infinite probability case though. 

I figure this would get worse in the nebular case where you're seeing more information from low-density regions. But maybe we're not trying to use this plasma by the point nebular matters.

andrewfullard:
The current fix does not zero out infinity, just NaN caused by 0/0 division in the CDF integration. It's definitely a low-density problem though and something we should investigate for nebular-phase plasma.

tardis/iip_plasma/properties/continuum.py
```diff
@@ -1461,6 +1497,9 @@ def calculate(
             cool_rate_fb.sum(axis=0), "bf"
         )
         p_fb_deactivation = cool_rate_fb / cool_rate_fb_tot.values
+        p_fb_deactivation = p_fb_deactivation.where(
```
andrewfullard, line 1500: Work out what this does

andrewfullard: `df.where` replaces entries where the condition is **false**. Using `fillna` is probably better.

jvshields (approved):

connor-mcclellan (approved):

PR #3688, opened 2026-08-12
Better distance to line calculation and Sobolev optical depth for nonhomology

connor-mcclellan:
### :pencil: Description

**Type:** :beetle: `bugfix`

A fix for an oversimplification in the original implementation of non-homologous transport. This enables the distance-to-line calculation and corresponding sobolev optical depth calculations to use the projected velocity gradient along the packet trajectory instead of presuming that a direct replacement of `time_explosion` with `1 / velocity_gradient` produces the same result (it doesn't, except in the case of homology).

~~Will need to be rebased on https://github.com/tardis-sn/tardis/pull/3646 when it is merged.~~ Master has been merged in after the MC restructure

Depends on https://github.com/tardis-sn/tardis-regression-data/pull/112 for regression tests


tardis/opacities/opacity_state.py
```diff
@@ -75,6 +82,7 @@ def __init__(
         self.line_list_nu = line_list_nu
 
         self.tau_sobolev = tau_sobolev
+        self.sobolev_optical_depth_coefficient = None
```
andrewfullard: So does this attribute only get set completely externally to the init of the class? I think it would be better as an optional init arg.

tardis/opacities/opacity_state_numba.py
```diff
@@ -194,3 +203,7 @@ def __getitem__(self, i: slice) -> Self:
             self.photo_ion_activation_idx,
             self.k_packet_idx,
         )
+        opacity_state.sobolev_optical_depth_coefficient = (
+            self.sobolev_optical_depth_coefficient[:, i]
```
andrewfullard, line 207: Why the second slice action here?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,17 +120,36 @@ def calculate_distance_line(
...
+    maximum_distance : float, optional
+        Inclusive upper bound for candidate resonance distances.
 
     Returns
     -------
-    distance (cm)
+    float
+        Distance to the line resonance in centimeters.
     """
     #TODO: unit check / handling here?
```
andrewfullard, line 154: Still TODO?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -146,54 +171,222 @@ def calculate_distance_line_nonhomologous(
     # Characteristic scales for non-dimensionalization
     r0 = r_outer - r_inner
```
andrewfullard, line 172: Use better names here. r_diff? r_scale?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -146,54 +171,222 @@ def calculate_distance_line_nonhomologous(
...
+        if q != 0.0 and n * n < q * q:
+            x_squared = (
+                n * n * impact_parameter_squared / (q * q - n * n)
+            )
+            x_root = math.copysign(math.sqrt(x_squared), n / q) / r0
+            x = (x_root, math.nan, math.nan, math.nan)
+        else:
+            x = (math.nan, math.nan, math.nan, math.nan)
+    else:
+        # Dimensionless quantities to use in the quartic solver - improves floating point accuracy
```
andrewfullard, line 186: Are there more meaningful names than single or double chars here?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -63,31 +65,36 @@ def calculate_distance_boundary(r, mu, r_inner, r_outer):
...
 @njit(**njit_dict_no_parallel)
-def calculate_distance_line(
-    r_packet,
-    comov_nu,
-    is_last_line,
-    nu_line,
-    time_explosion,
-    enable_full_relativity,
-):
+def calculate_distance_line_homologous(
```
andrewfullard, line 68: This rename is not consistent, since we have other functions further down that specify non-homology

tardis/transport/montecarlo/modes/nonhomologous/opacity_solver.py
```diff
@@ -66,7 +70,26 @@ def legacy_solve(self, plasma) -> OpacityState:
...
 
+        sobolev_optical_depth_coefficient = (
+            calculate_sobolev_optical_depth_coefficient(
+                plasma.atomic_data.lines,
+                plasma.level_number_density,
+                plasma.stimulated_emission_factor,
+            )
+        )
+        if self.disable_line_scattering:
+            sobolev_optical_depth_coefficient.iloc[:, :] = 0.0
```
andrewfullard, line 81: Could just init as zero?

tardis/transport/montecarlo/modes/nonhomologous/rad_packet_transport.py
```diff
@@ -72,127 +101,169 @@ def trace_packet(
...
+                    comov_nu_end,
+                )
+                return (
+                    distance_boundary,
+                    InteractionType.BOUNDARY,
+                    delta_shell,
+                )
+
+            # connor-mcclellan: some hardcoded overrides here until
+            # update_estimators_line is generalized for nonhomologous geometry
```
andrewfullard, line 183: Raise an issue

tardis/transport/montecarlo/modes/nonhomologous/tau_sobolev.py
```diff
@@ -113,6 +164,69 @@ def calculate_beta_sobolev(tau_sobolevs):
...
+        Radial velocity derivative in each shell [s^-1].
+    velocity_over_radius : astropy.units.Quantity
+        Ratio of radial velocity to radius in each shell [s^-1].
+
+    Returns
+    -------
+    pandas.DataFrame
+        Angle-averaged Sobolev escape probabilities for each line and shell.
+    """
+    mus, weights = np.polynomial.legendre.leggauss(20)
```
andrewfullard, line 188: Does this determine the angle-average resolution? Should this be a global/configurable?

tardis/transport/montecarlo/modes/nonhomologous/virtual_packet.py
```diff
@@ -134,54 +136,98 @@ def trace_vpacket_within_shell(
...
+                interval_start,
+                interval_end,
+            )
+            if (
+                distance_trace_line > interval_end
+                or distance_boundary <= distance_trace_line
+            ):
+                continue
+
+            new_r = math.sqrt(
```
andrewfullard, line 191: Some duplicated code here from the line distance stuff, perhaps related to the potential issue earlier that is just a comment?

tardis/transport/montecarlo/tests/test_transport.py
```diff
@@ -478,9 +478,9 @@ def test_iip_trace_packet_no_line_fallthrough(
             1.0e-20,
             NONHOMOLOGOUS_LINE_OPACITY,
             {"next_line_id": 0, "prev_line_id": 0},
-            InteractionType.BOUNDARY, # end of line list
-            1,
+            InteractionType.LINE,
```
andrewfullard, line 481: I assume this is related to the regression data change

andrewfullard (changes requested):

tardis/transport/geometry/calculate_distances.py
```diff
@@ -146,54 +171,222 @@ def calculate_distance_line_nonhomologous(
...
+    distance_tolerance = CLOSE_LINE_THRESHOLD * r0
     if v0 == 0.0:
-        # Velocity gradient is zero - packet cannot shift into a line in this shell
-        return MISS_DISTANCE
+        impact_parameter_squared = r * r * p
+        if q != 0.0 and n * n < q * q:
+            x_squared = (
+                n * n * impact_parameter_squared / (q * q - n * n)
+            )
+            x_root = math.copysign(math.sqrt(x_squared), n / q) / r0
```
jvshields, line 181: Any particular reason to use the math library instead of numpy?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -146,54 +171,222 @@ def calculate_distance_line_nonhomologous(
...
+            continue
+
+        new_mu = (r * mu + d_candidate) / new_r
+        new_v = v_inner + m * (new_r - r_inner)
+        comov_nu = nu_rest * (
+            1.0 - new_v / C_SPEED_OF_LIGHT * new_mu
+        )
+        if (
+            not math.isfinite(comov_nu)
+            or abs(comov_nu - nu_line) > 1.0e-7 * nu_line
```
jvshields, line 235: Maybe assign the 1.0e-7 magic number to a constant with a a descriptive name?

tardis/transport/montecarlo/modes/nonhomologous/tau_sobolev.py
connor-mcclellan: This is potentially a much deeper conversation! Having angle-dependent beta sobolevs at all is a huge step up in complexity compared with homology.

Math summary from Codex:

> Yes: 20 is the angular quadrature order and therefore controls the numerical resolution of the angle-averaged Sobolev escape probability.
> 
> ## What the code computes
> 
> For a spherical radial velocity field, the velocity gradient projected along a ray with direction cosine $\mu$ is
> 
> $$\
> Q(\mu)=
> \mu^2\frac{dv}{dr}
> +
> (1-\mu^2)\frac{v}{r}.
> $$
> 
> For line $\ell$, the directional Sobolev optical depth is
> 
> $$
> \tau_\ell(\mu)=\frac{K_\ell}{|Q(\mu)|},
> $$
> 
> where $K_\ell$ is the stored velocity-gradient-independent coefficient. The directional escape probability is
> 
> $$
> \beta_\ell(\mu)=
> \frac{1-\exp[-\tau_\ell(\mu)]}{\tau_\ell(\mu)}.
> $$
> 
> The function computes its angular average:
> 
> $$
> \bar{\beta}_\ell=
> \frac{1}{2}\int_{-1}^{1}\beta_\ell(\mu)\,d\mu.
> $$
> 
> The 20-point Gauss–Legendre rule supplies the $\mu$ values and weights used to approximate this integral. The factor of `0.5` at [line 221](https://github.com/tardis-sn/tardis/blob/fix/nonhomology-transport/tardis/transport/montecarlo/modes/nonhomologous/tau_sobolev.py#L221) is the normalization above.
> 
> In homologous expansion,
> 
> $$
> \frac{dv}{dr}=\frac{v}{r}=\frac{1}{t},
> $$
> 
> so $Q=1/t$ for every angle. The integrand is constant and any correctly normalized quadrature rule returns the exact result. Angular resolution only becomes important in nonhomology.
> 
> ## Impact on Monte Carlo transport
> 
> This angular average is separate from the packet’s trajectory-specific Sobolev optical depth.
> 
> For a real or virtual packet, TARDIS evaluates $Q$ at the actual resonance position and packet direction, then divides the coefficient by $|Q|$, as seen in [rad_packet_transport.py](https://github.com/tardis-sn/tardis/blob/fix/nonhomology-transport/tardis/transport/montecarlo/modes/nonhomologous/rad_packet_transport.py#L193). That calculation decides whether the packet interacts with the line.
> 
> The averaged $\bar{\beta}$, on the other hand, is passed into the macro-atom solver by [simple_tardis_workflow.py](https://github.com/tardis-sn/tardis/blob/fix/nonhomology-transport/tardis/workflows/simple_tardis_workflow.py#L379). It multiplies the macro-atom emission and internal-transition weights, for example in [macroatom_line_transitions.py](https://github.com/tardis-sn/tardis/blob/fix/nonhomology-transport/tardis/opacities/macro_atom/macroatom_line_transitions.py#L324).
> 
> Consequently:
> 
> - In pure `scatter` mode, this angular integration has no direct effect on packet propagation because no macro-atom state is constructed.
> 
> - In `downbranch` and `macroatom` modes, an error in $\bar{\beta}$ changes the normalized branching probabilities after a line absorption.
> 
> - That can change the re-emitted line, packet frequency redistribution, and ultimately the emergent spectrum.
> 
> - It does not directly change the distance to the next line resonance; that uses the packet-direction calculation.
> 
> ## Projected-gradient zero crossings
> 
> Writing
> 
> $$
> A=\frac{v}{r},\qquad B=\frac{dv}{dr},
> $$
> 
> gives
> 
> $$
> Q(\mu)=A+(B-A)\mu^2.
> $$
> 
> An interior zero exists when $A$ and $B$ have opposite signs:
> 
> $$
> \mu_0=
> \sqrt{\frac{-A}{B-A}}=
> \sqrt{\frac{A}{A-B}},
> \qquad 0<\mu_0<1.
> $$
> 
> Because the integrand depends only on $\mu^2$, zeros occur symmetrically at $\pm\mu_0$. It is sufficient to integrate over $0\leq\mu\leq1$, where there is at most one interior zero.
> 
> The important cases are:
> 
> - $A$ and $B$ have the same nonzero sign: no zero crossing.
> 
> - $A$ and $B$ have opposite signs: one zero in $0<\mu<1$.
> 
> - $A=0$: a zero at the tangential direction, $\mu=0$.
> 
> - $B=0$: a zero at the radial direction, $\mu=1$.
> 
> - $A=B=0$: $Q=0$ in every direction.
> 
> - $K=0$: the line has no opacity, so $\tau=0$ and $\beta=1$, including at a gradient zero.
> 
> - A negative $K$, possible for a population inversion, is not handled correctly by the current thin/thick branching. That is an existing, separate assumption.
> 
> ## Why the zeros are numerically and physically difficult
> 
> As $Q\rightarrow0$,
> 
> $$
> \tau\rightarrow\infty,
> \qquad
> \beta\sim\frac{1}{\tau}
> =\frac{|Q|}{K}\rightarrow0.
> $$
> 
> The integral therefore does not diverge. Near a simple zero,
> 
> $$
> \beta(\mu)
> \sim
> \frac{|Q'(\mu_0)|}{K}|\mu-\mu_0|.
> $$
> 
> This is a finite V-shaped cusp. Gauss–Legendre quadrature converges very rapidly for smooth functions, but much less rapidly when a cusp lies inside the interval. A fixed set of nodes may either sample very close to the dip or largely miss it, making the error sensitive to the precise zero location.
> 
> For weak lines, the integrand can be almost one everywhere except for a very narrow dip around $\mu_0$. Even after splitting, a fixed-order rule may under-resolve this boundary layer when the zero is extremely close to 0 or 1.
> 
> There is also a physical limitation: $Q=0$ means the comoving frequency is stationary to first order along the ray. The Sobolev resonance length formally becomes infinite, and the local Sobolev approximation ceases to be valid. A sign change can also produce multiple spatial resonances with the same line. Splitting the angular integral improves the numerical evaluation of the local formula; it does not repair that underlying physical approximation.
> 
> ## Splitting the integral
> 
> Using symmetry,
> 
> $$
> \bar{\beta}=\int_0^1\beta(\mu)\,d\mu.
> $$
> 
> For a shell with an interior zero, replace this with
> 
> $$
> \bar{\beta}=
> \int_0^{\mu_0}\beta(\mu)\,d\mu
> +
> \int_{\mu_0}^{1}\beta(\mu)\,d\mu.
> $$
> 
> Each subinterval can be evaluated by mapping the usual Gauss–Legendre nodes $x_i\in[-1,1]$ onto $[a,b]$:
> 
> $$
> \mu_i=\frac{a+b}{2}+\frac{b-a}{2}x_i,
> \qquad
> w_i'=\frac{b-a}{2}w_i.
> $$
> 
> This moves the cusp to an interval endpoint. Within each open interval, $Q$ has a fixed sign, so $|Q|$ no longer introduces an interior kink.
> 
> Implementation would require:
> 
> 1. Compute the zero once per shell from `velocity_gradient` and `velocity_over_radius`. It does not depend on the line.
> 
> 2. Integrate $[0,1]$ normally for shells without an interior zero.
> 
> 3. Integrate $[0,\mu_0]$ and $[\mu_0,1]$ separately for shells with one.
> 
> 4. Vectorize over all lines within a shell because every line shares the same interval boundaries.
> 
> 5. Handle the all-zero-gradient and zero-opacity cases explicitly.
> 
> 6. Make the quadrature order a named default or keyword argument so convergence can be tested.
> 
> 7. Add numerical tests for homology, opposite-sign gradients, zeros near both endpoints, and the all-zero case.
> 
> An adaptive integral per line would be prohibitively expensive because this function processes the full line-by-shell matrix. Fixed split Gaussian quadrature is the practical approach.
> 
> In a synthetic sweep over root positions and optical-depth scales, the current unsplit 20-point rule reached about $2.7\times10^{-2}$ maximum absolute error. Using 10 points on each side—still 20 evaluations after exploiting symmetry—reduced that to about $2.0\times10^{-3}$; 20 points on each side reduced it to roughly $4.7\times10^{-4}$. These are illustrative results, not yet an acceptance criterion; the order should be selected against the range of gradients and line strengths expected in TARDIS models.
> 
> My recommendation is to make the quadrature order an internal named default rather than a user-facing configuration option initially, split only shells with interior zeros, and establish a numerical tolerance through convergence tests. A YAML setting would add substantial configuration and documentation surface for what is currently an implementation-accuracy parameter.

docs/io/configuration/components/models/index.rst
```diff
@@ -125,9 +125,18 @@ We present an example of the above schema:
                 start: 1000 km/s
                 stop: 2000 km/s
                 num: 15
+            radius:
+                start: 1.0e14 cm
+                stop: 3.0e14 cm
             density:
                 type: branch85_w7 #see density schemas below for all options in the density section
 
+The ``radius`` section is optional for homologous workflows. For a
```
jvshields, line 134: I think this needs to say what happens if you provide both. Will the radius section be ignored? What happens if r != vt

docs/io/configuration/components/models/index.rst
```diff
@@ -253,17 +262,21 @@ as shown in the schema below:
...
 .. literalinclude:: csvy_full_rad.csvy
 
 Notice that for each column that is used in the CSV section of the file, there is a corresponding field under
 ``datatype`` in the YAML section of the file. In our example, each of the fields under ``datatype`` has a brief
 description to go along with it. While the description is not necessary for any of the fields, the unit section
-is required for ``velocity``, ``density``, and ``t_rad``.
+is required for ``radius``, ``velocity``, ``density``, and ``t_rad``.
+
+When both ``radius`` and ``velocity`` are present, their entries specify the same shell boundaries and TARDIS
+retains them as independent quantities. In particular, it does not replace the supplied radii with radii derived
```
jvshields, line 278: I guess it says here, but I'd also like it up above

tardis/io/model/parse_geometry_configuration.py
```diff
@@ -86,13 +85,7 @@ def parse_geometry_from_config(config: Configuration, time_explosion):
...
         The parsed homologous geometry.
     """
-    (
-        density_time,
-        velocity,
-        density,
-        electron_densities,
-        temperature,
-    ) = parse_structure_from_config(config)
+    _, velocity, _, _, _ = parse_structure_from_config(config)
```
jvshields, line 88: Um, doesn't this contradict what was said above? It seems to me like the radius section will be ignored if supplied in the homologous case. The radius isn't being passed along.

tardis/io/model/parse_geometry_configuration.py
```diff
@@ -103,61 +96,157 @@ def parse_geometry_from_config(config: Configuration, time_explosion):
...
+    ----------
+    config : Configuration
+        Configuration containing explicit velocity and radius boundaries.
+
+    Returns
+    -------
+    Radial1DGeometry
+        Geometry retaining independent radius and velocity boundaries.
+    """
+    _, velocity, _, _, _ = parse_structure_from_config(config)
```
jvshields, line 114: Hate this code block. The parse_structure_from_config function doesn't actually parse structure? Because config.model.structure is looked at immediately following. So somehow this is implying that the structure is completely different from what the parse_structure_from_config file returns and is used for.

tardis/model/base.py
```diff
@@ -195,8 +198,9 @@ def isotopic_number_density(self):
         return isotopic_number_density
 
     @property
-    def radius(self):
-        return self.time_explosion * self.velocity
+    def radius(self) -> u.Quantity:
+        """Return the active shell-boundary radii."""
+        return self.r_outer.insert(0, self.r_inner[0])
```
andrewfullard, line 203: Might want to rebase now the slicing fixes are in, might affect some of these changes

tardis/opacities/opacity_state_numba.py
connor-mcclellan: This has the same shape as tau_sobolev, so it's (n_lines, n_shells). So this particular slice is all lines for a single shell

tardis/transport/geometry/calculate_distances.py
connor-mcclellan: Agreed, I reverted that and we can save it for the MC refactor

tardis/transport/montecarlo/modes/nonhomologous/rad_packet_transport.py
connor-mcclellan: https://github.com/tardis-sn/tardis/issues/3745

tardis/transport/montecarlo/modes/nonhomologous/virtual_packet.py
connor-mcclellan: Moved the shared resonance position, direction cosine, velocity, and projected gradient calculations into `calculate_packet_velocity_properties`. The distance solver and both real and virtual packet transport now use that helper.

tardis/transport/geometry/calculate_distances.py
connor-mcclellan: There was a loose precedent in some of the surrounding code to use `math` for scalar operations and reserve `numpy` for array operations. There's not a real reason other than that- it's not faster or anything.

docs/io/configuration/components/models/index.rst
connor-mcclellan: So, the good news is this was a leftover diff from #3689 and is no longer a part of this PR after the rebase. The bad news is it was already merged as-is. I'll raise an issue and open a new PR to clear this up.

tardis/io/model/parse_geometry_configuration.py
connor-mcclellan: yep, docs are inaccurate but aren't in this PR anymore after rebasing. I'll do a follow-up PR to fix the docs

tardis/io/model/parse_geometry_configuration.py
connor-mcclellan: yep, I hate this too. After rebasing, this code belongs to #3689, so I'll fix this in a follow-up PR

tardis/model/base.py
connor-mcclellan: Done! Rebased onto current master after the slicing fixes. The upstream geometry-based radius implementation is retained, and model/base.py is no longer part of this PR’s diff.

tardis/transport/montecarlo/modes/nonhomologous/tau_sobolev.py
```diff
@@ -17,15 +19,16 @@
     .value
 )
 
+DEFAULT_BETA_SOBOLEV_QUADRATURE_ORDER = 20
```
connor-mcclellan, line 22: The argument for this staying local to this file is that it's too nitty-gritty to be exposed to the user, but having a default that we can override still allows us to do proper convergence testing. It's passed to `calculate_beta_sobolev_directional` as the argument `quadrature_order`.
Open to opinions if anyone disagrees

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,91 +120,350 @@ def calculate_distance_line(
...
-    if (comov_nu - nu_line > 1e-14*nu_line and m > 0.0) or (comov_nu - nu_line < -1e-14*nu_line and m < 0.0):
-        # Obtain roots of the quartic polynomial for x (= d_line + r_i \mu_i)
-        x = depressed_quartic(a, b, c, d, e)
-        # Convert each root x_i to a candidate distance: d = r0*x_i - r*mu
-        # Select the smallest positive, finite distance among all four roots.
-        distance = MISS_DISTANCE
-        for x_root in x:
-            if math.isnan(x_root):
+        # Obtain roots of the quartic polynomial for the dimensionless
+        # x = (d_line + r_i \mu_i) / r0.
```
andrewfullard, line 261: Could you define the vars in this?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,91 +120,350 @@ def calculate_distance_line(
...
+        projected_position_roots = solve_resonance_quartic(
+            scaled_target_projected_velocity,
+            scaled_impact_parameter_squared,
+            scaled_velocity_intercept,
+        )
+
+    # Convert each dimensionless root to a distance: d = r0*x_i - r*mu.
```
andrewfullard, line 268: Could you define the vars in this?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,91 +120,350 @@ def calculate_distance_line(
...
+    # Select the nearest root that satisfies the original, unsquared resonance equation.
+    distance = MISS_DISTANCE
+    for projected_position_root in projected_position_roots:
+        if not math.isfinite(projected_position_root):
+            continue
+        if shell_velocity_difference != 0.0:
+            # The quartic comes from squaring the resonance equation. In the
+            # original equation, x - N and Q*x must have opposite signs.
```
andrewfullard, line 276: Could you define the vars in this?

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,91 +120,350 @@ def calculate_distance_line(
...
-            if d_candidate > 0.0 and d_candidate < distance:
-                distance = d_candidate
-    else:
-        distance = MISS_DISTANCE
+        candidate_distance = (
+            shell_width * projected_position_root - radius * direction_cosine
+        )
+        if candidate_distance <= minimum_distance + distance_tolerance:
+            continue
+        if candidate_distance > maximum_distance + distance_tolerance:
```
andrewfullard, line 295: Either this should be an `or`, or the previous `or` should be split like this (the split is nice for clarity potentially)

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,91 +120,350 @@ def calculate_distance_line(
...
+    rpacket: RPacket,
+    geometry: NumbaRadial1DGeometry,
+    distance_boundary: float,
+) -> tuple[float, float, int]:
+    """Calculate forward projected-gradient zeros in the current shell.
+
+    Returns up to two distances where the derivative of the projected fluid
+    velocity along the packet trajectory changes sign. Missing distances are
+    returned as ``MISS_DISTANCE``.
+    """
```
andrewfullard, line 328: Missing parameters

tardis/transport/montecarlo/tests/test_nonhomologous.py
```diff
@@ -45,22 +56,374 @@
...
+    )
+
+    npt.assert_allclose(
+        distance,
+        0.14167578672128847 * shell_width,
+        rtol=1.0e-9,
+    )
+
+
+def test_nonhomologous_distance_rejects_sign_reversed_root() -> None:
```
connor-mcclellan, line 373: All tests except this one are pure Codex test-driven development, so they all have the typical compare-against-magic-numbers approach we've gotten used to seeing.

The only difference here is I told it I didn't like comparing against a magic number in the test, so it changed it to derive the magic number here rather than hardcoding it. I'm not sure which is better for our purposes, but if we like this, it could be a good example of a preference we could set in an agents.md.

tardis/transport/geometry/calculate_distances.py
```diff
@@ -113,91 +120,350 @@ def calculate_distance_line(
...
+        + distance * distance
+        + 2.0 * rpacket.r * distance * rpacket.mu
+    )
+    new_mu = (rpacket.r * rpacket.mu + distance) / new_r
+    new_v = geometry.get_velocity(new_r, rpacket.current_shell_id)
+    return rpacket.nu * (1.0 - new_v / C_SPEED_OF_LIGHT * new_mu)
+
+
+@njit(**njit_dict_no_parallel)
+def get_line_id_range_nonhomologous(
```
andrewfullard, line 405: Is this intended to reduce the line search size?

tardis/transport/montecarlo/modes/nonhomologous/mc_rad_field_solver.py
```diff
@@ -118,7 +118,6 @@ def estimate_jblues(
     ) -> np.ndarray:
         j_blues_norm_factor = (
             const.c.cgs
-            / velocity_gradient
             / (4 * np.pi * time_of_simulation * volume)
```
andrewfullard, line 121: Why is this gone from the estimator?

tardis/transport/montecarlo/nonhomologous_grid.py
```diff
@@ -1,86 +1,538 @@
...
 from numba import njit
 
-from tardis.transport.montecarlo import (
-    njit_dict_no_parallel,
-)
+_FLOAT64_EPSILON = np.finfo(np.float64).eps
+_FLOAT64_TINY = np.finfo(np.float64).tiny
+_RECIPROCAL_ROOT_THRESHOLD = math.sqrt(_FLOAT64_EPSILON)
+_QUARTIC_NJIT_OPTIONS = {
+    "fastmath": False,
```
connor-mcclellan, line 10: Disabling fastmath evidently improves floating point precision, so that's the motivation for disabling it for the quartic solver since it's so sensitive.

tardis/transport/montecarlo/modes/nonhomologous/rad_packet_transport.py
```diff
@@ -73,126 +101,156 @@ def trace_packet(
...
+                    r_packet,
+                    opacity_state.line_list_nu,
+                    comov_nu_end,
+                )
+                return (
+                    distance_boundary,
+                    InteractionType.BOUNDARY,
+                    delta_shell,
+                )
+
```
andrewfullard, line 176: Would be nice to have some annotation of the transport step here since it's a bunch of relatively long if blocks, e.g. "Line interaction"

tardis/transport/montecarlo/modes/nonhomologous/tau_sobolev.py
```diff
@@ -113,8 +169,98 @@ def calculate_beta_sobolev(tau_sobolevs):
...
 
 
-class TauSobolev(ProcessingPlasmaProperty):
+def calculate_beta_sobolev_directional(
+    sobolev_optical_depth_coefficient: pd.DataFrame,
+    velocity_gradient: u.Quantity,
+    velocity_over_radius: u.Quantity,
+    quadrature_order: int = DEFAULT_BETA_SOBOLEV_QUADRATURE_ORDER,
+) -> pd.DataFrame:
+    """Calculate the angle-averaged Sobolev escape probability.
```
andrewfullard, line 178: Would be nice to have more info about the mathematics behind this

tardis/transport/montecarlo/tests/test_nonhomologous.py
```diff
@@ -45,22 +56,374 @@
     [
         # x^4 - 14x^3 + 71x^2 - 154x + 120 = 0
         # roots 2, 3, 4, 5
-        (1.0, -14.0, 71.0, -154.0, 120.0, [5.0, 4.0, 3.0, 2.0]),
+        (1.0, -14.0, 71.0, -154.0, 120.0, [2.0, 3.0, 4.0, 5.0]),
```
andrewfullard, line 59: Are these changes demonstrating the previously incorrect maths?

tardis/transport/montecarlo/nonhomologous_grid.py
```diff
@@ -1,86 +1,538 @@
+import math
+
 import numpy as np
 from numba import njit
 
-from tardis.transport.montecarlo import (
-    njit_dict_no_parallel,
-)
+_FLOAT64_EPSILON = np.finfo(np.float64).eps
```
andrewfullard, line 6: Not sure why these are _ at the start. tbh we will probably want to make these globals somewhere at the top of the tardis package eventually.

tardis/transport/montecarlo/nonhomologous_grid.py
```diff
@@ -1,86 +1,538 @@
...
+    """Solve a monic quadratic without cancellation between its roots."""
+    discriminant = (
+        linear_coefficient * linear_coefficient - 4.0 * constant_coefficient
+    )
+    discriminant_scale = max(
+        abs(linear_coefficient * linear_coefficient),
+        abs(4.0 * constant_coefficient),
+        _FLOAT64_TINY,
+    )
+    discriminant_tolerance = 8.0 * _FLOAT64_EPSILON * discriminant_scale
```
andrewfullard, line 29: Why 8.0?

tardis/transport/montecarlo/nonhomologous_grid.py
```diff
@@ -1,86 +1,538 @@
...
+    quadratic_coefficient: float,
+    constant_coefficient: float,
+    root_shift: float,
+) -> tuple[float, float, float, float]:
+    """Solve a depressed quartic with no linear term."""
+    first_squared_root, second_squared_root = _solve_real_quadratic(
+        quadratic_coefficient, constant_coefficient
+    )
+    squared_root_tolerance = (
+        8.0
```
andrewfullard, line 64: Why 8.0?

tardis/transport/montecarlo/nonhomologous_grid.py
```diff
@@ -1,86 +1,538 @@
...
             )
+            if recovered_squared_factor > 0.0:
+                squared_factor = recovered_squared_factor
+
+    return squared_factor
+
+
+@njit(**_QUARTIC_NJIT_OPTIONS)
+def _solve_depressed_quartic(
+    quadratic_coefficient: float,
```
andrewfullard, line 229: I think descriptions of these variables would help make this more understandable

tardis/transport/montecarlo/nonhomologous_grid.py
```diff
@@ -1,86 +1,538 @@
...
+    impact_parameter_squared: float,
+    velocity_intercept: float,
+) -> tuple[float, float, float, float]:
+    """Solve the structured quartic for a line-resonance distance.
+
+    The polynomial is
+
+    ``(x - N)**2 * (x**2 + H**2) - Q**2 * x**2``,
+
+    where the arguments are ``N``, ``H**2``, and ``Q`` respectively.
```
andrewfullard, line 376: And what are the arguments physically?

andrewfullard (changes requested):

PR #3689, opened 2026-08-12
Add radius and velocity CSVY geometry support

connor-mcclellan:
### :pencil: Description

**Type:** :rocket: `feature`

Adds support for creating geometry objects based on a CSVY model that has both radius and velocity data, not presuming homology. Will need a substantial refactor / rebase onto https://github.com/tardis-sn/tardis/pull/3646 when it is merged.


tardis/model/base.py
```diff
@@ -195,8 +196,9 @@ def isotopic_number_density(self):
         return isotopic_number_density
 
     @property
-    def radius(self):
-        return self.time_explosion * self.velocity
+    def radius(self) -> u.Quantity:
+        radius = self.geometry.r_outer_active.copy()
```
andrewfullard: Better to use self.r_outer and self.r_inner for this?

tardis/workflows/nonhomologous_tardis_workflow.py
```diff
@@ -64,18 +61,18 @@ def __init__(
...
-        self.simulation_state.geometry = Radial1DGeometry(
-            r_inner=geometry.v_inner * t_exp,
-            r_outer=geometry.v_outer * t_exp,
-            v_inner=geometry.v_inner,
-            v_outer=geometry.v_outer,
-            r_inner_boundary=geometry.v_inner_boundary * t_exp,
-            r_outer_boundary=geometry.v_outer_boundary * t_exp,
-            v_inner_boundary=geometry.v_inner_boundary,
-            v_outer_boundary=geometry.v_outer_boundary,
+        if isinstance(geometry, HomologousRadial1DGeometry):
```
jvshields, line 77: I would rather just crash if somehow you end up with a homologous geometry in the non homologous workflow. Is there a reason you would ever want to do something like this?

tardis/io/model/parse_geometry_configuration.py
```diff
@@ -108,7 +107,7 @@ def parse_geometry_from_csvy(
     csvy_model_config: Configuration,
     csvy_model_data: pd.DataFrame | None,
     time_explosion: u.Quantity,
-) -> HomologousRadial1DGeometry:
+) -> HomologousRadial1DGeometry | Radial1DGeometry:
```
jvshields: I'd like @andrewfullard to weigh in on this too, but I think I prefer unique methods for reading the different objects to growing existing methods to handle different cases. That way each workflow, in which you already know what type of geometry you're going to make, you can just call the appropriate method and we have fewer branching codepaths.

andrewfullard: Agree that is probably safer here.

tardis/model/base.py
connor-mcclellan: Potentially - this was patterned directly after the same function for `velocity(self)`, which also grabs `self.geometry.v_outer_active.copy()`. I haven't looked critically at that though and I recall that there was some uncertainty about how robust the active cells vs not active cells assignment was anyway

tardis/workflows/nonhomologous_tardis_workflow.py
connor-mcclellan: The reason would be to test nonhomologous transport on a homologous grid. This is a stand-in for an actual conversion utility between the two, which we will probably want at some point.

The defense for keeping it is that you don't lose any information by creating a nonhomologous grid using r = vt, whereas it should *definitely* crash if you try to go the other way. If we do want to keep this, it would be probably be appropriate to add a warning to make sure the user is aware they've provided a homologous geometry.

andrewfullard:
Can you sneak a fix for the csvy-related docs build issue?

tardis/model/base.py
connor-mcclellan: Fixed

tardis/io/model/csvy/data.py
```diff
@@ -1,4 +1,5 @@
 from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
```
andrewfullard, line 2: Codex sneaked this in

connor-mcclellan: Yeah, the issue is that that block is the only thing preventing the circular import. So we can choose between having a type checking block or removing the type hints for the `to_geometry` and `to_nonhomologous_geometry` methods

connor-mcclellan: Sorry, I removed the actual block but forgot to remove the import. But actually using a type checking block in this case would fix the issue

andrewfullard: Sometimes it's useful I guess! A comment to that end on the block so we know not to delete it.

connor-mcclellan:
Whew, that ended up being a bigger fix than I thought. The circular import wasn't the end of the story - after I fixed that, then it actually failed with the error I specifically put in there to catch the case where you pass a HomologousRadial1DGeometry to a nonhomologous workflow.

We don't always want to require a CSVY model for nonhomology mode. There are plenty of cases in testing where we would want to be able to specify a nonhomologous velocity profile just from a YAML input, essentially just *also* specifying a radius start and stop alongside a velocity start and stop (plus number of bins) where the end result is something not homologous.

A first pass at adding support for that has been added in the most recent commit

tardis/io/model/csvy/data.py
```diff
@@ -1,15 +1,18 @@
...
 import pandas as pd
 from astropy import units as u
 
 from tardis.io.configuration.config_reader import Configuration
-from tardis.model.geometry.radial1d import Radial1DGeometry
-from tardis.model.geometry.radial1d_homologous import (
-    HomologousRadial1DGeometry,
-)
+
+if TYPE_CHECKING:
```
andrewfullard, line 12: A comment noting why this is needed would be useful

tardis/io/model/csvy/data.py
```diff
@@ -58,6 +61,10 @@ def to_geometry(
         tardis.model.geometry.radial1d_homologous.HomologousRadial1DGeometry
             The geometry object constructed from the CSVY data.
         """
+        from tardis.model.geometry.radial1d_homologous import (  # noqa: PLC0415
```
andrewfullard, line 67: again, comment why we are ignoring the ruff requirement here

tardis/workflows/nonhomologous_tardis_workflow.py
```diff
@@ -64,19 +74,16 @@ def __init__(
...
-        )
+        if isinstance(geometry, HomologousRadial1DGeometry):
+            geometry_error = (
+                "NonhomologousTARDISWorkflow requires explicit radius and "
+                "velocity boundaries. For a specific YAML model, provide "
+                "model.structure.radius start and stop values; for CSVY, "
+                "provide both radius and velocity columns."
+            )
+            warnings.warn(geometry_error, UserWarning, stacklevel=2)
+            raise TypeError(geometry_error)
```
andrewfullard, line 84: Why is there a warning plus exception?

connor-mcclellan: good point. Removing the warning

andrewfullard (dismissed):

tardis/io/model/parse_geometry_configuration.py
```diff
@@ -103,61 +96,157 @@ def parse_geometry_from_config(config: Configuration, time_explosion):
...
+        Configuration containing explicit velocity and radius boundaries.
+
+    Returns
+    -------
+    Radial1DGeometry
+        Geometry retaining independent radius and velocity boundaries.
+    """
+    _, velocity, _, _, _ = parse_structure_from_config(config)
+    structure_config = config.model.structure
+    radius = quantity_linspace(
```
jvshields, line 116: For parsing methods like this, I would prefer explicitly using  .to(u.cm) rather than just setting it to cgs. Easier to catch unit problems before it breaks somewhere else downstream, like if a unit is forgotten in the config when it's required. 

If somehow elsewhere an enforcement that the structure radius is already a length unit, ignore this comment.

tardis/io/model/parse_geometry_configuration.py
```diff
@@ -103,61 +96,157 @@ def parse_geometry_from_config(config: Configuration, time_explosion):
...
+    v_outer_boundary = getattr(model, "v_outer_boundary", None)
+    velocity = parse_velocity_from_csvy(csvy_model_config, csvy_model_data)
+
+    radius_field_idx = [
+        field["name"] for field in csvy_model_config.datatype.fields
+    ].index("radius")
+    radius_unit = u.Unit(
+        csvy_model_config.datatype.fields[radius_field_idx]["unit"]
+    )
+    radius = (csvy_model_data["radius"].values * radius_unit).to("cm")
```
jvshields, line 237: See previous comment, it's enforced here but not in the config parser. That's a bit weird.

jvshields:
One small complaint but otherwise looks good

jvshields (dismissed):

andrewfullard (dismissed):

jvshields (dismissed):

andrewfullard (dismissed):

jvshields (approved):

andrewfullard (approved):
tut tut, units etc

PR #3692, opened 2026-08-14
fix naming collision rate name inversion

jvshields:
The equation in the rate solver seems correct, but the names are inverted because the class defined degeneracies are dropping the wrong ones. This small patch is just a rename.


andrewfullard (approved):
Woops! This is just a human error :P

connor-mcclellan (approved):

PR #3693, opened 2026-08-15
Fix shell velocities used for CSVY header density profiles

### LINKED ISSUE #3621: Issue with density profiles defined within a CSVY model
**Describe the bug**
If one sets up a CSVY model with a power law density profile in the standard way, the innermost shell of the model will have incorrect density. It seems that shell 0 will have the same density as the defined rho_0 value, unscaled (so in this case, rho in shell 0 == 2e-13 g/cm^3). Then shell 1 has what shell 0 density should be, and 2-->1 etc. (see attached plot).

```
    density:
      type: power_law
      rho_0: 2e-13 g/cm^3
      v_0: 10000 km/s
      exponent: -6
```

[density_profile_comparisons.pdf](https://github.com/user-attachments/files/29255010/density_profile_comparisons.pdf)

**To Reproduce**
Run any model with a power law density profile defined in the .yml file, then an equivalent model set up in a .csvy file. Here I attach some example files to reproduce the bug.

**Screenshots**

**System**
 - OS:
    - [ ] GNU/Linux
    - [x] macOS

 - Environment (`conda list`):
I am using TARDIS v2024.01.08, but the issue persists in newer versions (e.g., v2025.05.25.

**Additional context**
NOTE. I added the suffix ```.txt``` to the csvy to allow it to upload - remove before running.

[model.yml](https://github.com/user-attachments/files/29254943/model.yml)

[epoch1.yml](https://github.com/user-attachments/files/29254947/epoch1.yml)

[epoch1.csvy.txt](https://github.com/user-attachments/files/29254965/epoch1.csvy.txt)

NoiceHax:
`parse_density_section_csvy` inserted a fake zero velocity in front of the shell interfaces before working out the shell mid velocities. A density profile declared in the CSVY header was therefore evaluated at half the inner velocity for the first shell, and the function returned one density more than there are shells. `SimulationState` slices that array down to the active shells, so every density came out shifted by one. For `powerlaw_csvy.csvy` the first shell is 4.3 times too dense and the rest are off by about 11 percent.

The fix takes the mid velocities straight from the interfaces, matching what the config path produces after `parse_density_from_config` drops its leading value. `test_compare_models` missed this because `assert_array_almost_equal` on densities near 1e-9 has an absolute tolerance that hides it. The new test checks the power law and exponential models against analytic values with a relative tolerance, and needs no atom data or regression data.

The else branch of the same function also looks broken (`field.name`, `csvy_model_config.velocity.values`), but I left it alone to keep this small.

Fixes #3621.

I used an AI assistant on this change. I have gone through it myself and can explain it.

### DIFF
--- a/tardis/io/model/parse_density_configuration.py
+++ b/tardis/io/model/parse_density_configuration.py
@@ -103,9 +103,7 @@ def parse_density_section_csvy(
         )
         velocity = csvy_model_config.velocity.values * velocity_unit
 
-    adjusted_velocity = velocity.insert(0, 0)
-    v_middle = adjusted_velocity[1:] * 0.5 + adjusted_velocity[:-1] * 0.5
-    no_of_shells = len(adjusted_velocity) - 1
+    v_middle = velocity[1:] * 0.5 + velocity[:-1] * 0.5
 
     if hasattr(csvy_model_config, "density"):
         density_0, time_0 = parse_density_section_config(
--- a/tardis/model/tests/test_csvy_model.py
+++ b/tardis/model/tests/test_csvy_model.py
@@ -83,6 +83,48 @@ def test_compare_models(model_config_fnames, atomic_dataset):
     )
 
 
+@pytest.mark.parametrize(
+    ("csvy_model_name", "density_profile"),
+    [
+        ("powerlaw", lambda velocity_ratio: velocity_ratio**-2),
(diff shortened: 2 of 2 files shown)


connor-mcclellan:
Thanks for your contribution. We're still figuring how we want to approach this and it's part of a larger devops discussion, so we will close this for now until we've decided on an approach.

PR #3228, opened 2025-08-07
Starting of TARDIS HE parallel

Knights-Templars:
This PR aims at making TARDIS-HE optimized and also run in parallel.

PR #3706, opened 2026-08-25
TARDIS HE performance improvements

andrewfullard:
### :pencil: Description

Based on ideas in #3228, moves the gamma-ray transport closer to classic TARDIS. Note that this requires seeding packets differently, so it breaks regression data. However, the actual spectra are very similar, relative difference notwithstanding..

Requires https://github.com/tardis-sn/tardis-regression-data/pull/115 now

### :pushpin: Resources

image

image

- [ ] Testing pipeline
- [ ] Other method (describe)

### DIFF
--- a/tardis/energy_input/util.py
+++ b/tardis/energy_input/util.py
@@ -97,7 +97,7 @@ def doppler_factor_3d(direction_vector, position_vector, time):
 
 
 @njit(**njit_dict_no_parallel)
-def doppler_factor_3D_all_packets(
+def doppler_factor_3d_all_packets(
     direction_vectors: npt.NDArray[np.float64],
     position_vectors: npt.NDArray[np.float64],
     times: npt.NDArray[np.float64],
@@ -115,17 +115,12 @@ def doppler_factor_3D_all_packets(
     array
         Doppler factors
     """
-    no_of_packets = times.shape[0]
-    doppler_factors = np.empty(no_of_packets, dtype=np.float64)
-    for packet_idx in range(no_of_packets):
-        velocity_dot_direction = 0.0
-        for dimension_idx in range(3):
-            velocity_dot_direction += (
-                position_vectors[dimension_idx, packet_idx]
-                / times[packet_idx]
-                * direction_vectors[dimension_idx, packet_idx]
-            )
-        doppler_factors[packet_idx] = 1 - velocity_dot_direction / C_CGS
+    doppler_factors = position_vectors[0] * direction_vectors[0]
+    doppler_factors += position_vectors[1] * direction_vectors[1]
+    doppler_factors += position_vectors[2] * direction_vectors[2]
+    doppler_factors /= times
+    doppler_factors *= -1.0 / C_CGS
+    doppler_factors += 1.0
 
     return doppler_factors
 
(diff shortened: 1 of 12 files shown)


tardis/transport/montecarlo/packet_source/high_energy.py
```diff
@@ -428,6 +433,10 @@ def create_packets(
         nus_rf = np.zeros(number_of_packets)
         nus_cmf = np.zeros(number_of_packets)
         statuses = np.ones(number_of_packets, dtype=np.int64) * 3
+        self._reseed(self.base_seed + 3)
```
jvshields, line 439: What's this about? I don't think we do any sort of reseeding thing for the other packets.

andrewfullard: It's the same as the re-seeding for direction and velocity. See https://github.com/andrewfullard/tardis/blob/554b936a1a829ebd1924724a2a6078a9610738c5/tardis/transport/montecarlo/packet_source/high_energy.py#L482 and https://github.com/andrewfullard/tardis/blob/554b936a1a829ebd1924724a2a6078a9610738c5/tardis/transport/montecarlo/packet_source/high_energy.py#L134

jvshields: Aha. You learn something new about tardis every day.

jvshields:
I'm pretty happy with this PR. Everything looks like sensible targeted improvements. Happy to approve pending regression data and tests passing.

jvshields (dismissed):

jvshields (dismissed):

tardis/energy_input/util.py
```diff
@@ -110,9 +115,12 @@ def doppler_factor_3D_all_packets(direction_vectors, position_vectors, times):
...
     """
-    velocity_vector = position_vectors / times
-    vel_mul_dir = np.multiply(velocity_vector, direction_vectors)
-    doppler_factors = 1 - (np.sum(vel_mul_dir, axis=0) / C_CGS)
+    doppler_factors = position_vectors[0] * direction_vectors[0]
+    doppler_factors += position_vectors[1] * direction_vectors[1]
+    doppler_factors += position_vectors[2] * direction_vectors[2]
+    doppler_factors /= times
+    doppler_factors *= -1.0 / C_CGS
+    doppler_factors += 1.0
```
connor-mcclellan: This seems a little strange. I much prefer the old notation unless there's a specific reason for this rewrite

connor-mcclellan (changes requested):

tardis/energy_input/util.py
andrewfullard: This layout is more performant, but I agree it's not the most readable. I'll do a timing comparison to see if it's worth it.

andrewfullard: It was significantly faster, but since it is only called once upon packet creation, the previous version is still fast enough (42ms for 1e6 packets). Brought back the cleaner maths.

jvshields (approved):
Approving but voicing my disapproval of importing callable. I see that the test is pre-existing kind of, so it won't stop the approval.

connor-mcclellan (approved):

PR #3707, opened 2026-08-25
Example refactor of packet_propagation with wrappers

andrewfullard:
### :pencil: Description

For inspection and understanding of a possible path towards refactoring the MC further

### DIFF
--- a/tardis/transport/montecarlo/modes/homologous_packet_propagation.py
+++ b/tardis/transport/montecarlo/modes/homologous_packet_propagation.py
@@ -1,0 +1,197 @@
+"""Shared packet propagation for homologous transport modes."""
+
+from collections.abc import Callable
+
+from numba import njit
+
+from tardis import constants as const
+from tardis.model.geometry.radial1d_homologous import (
+    NumbaHomologousRadial1DGeometry,
+)
+from tardis.opacities.opacity_state_numba import OpacityStateNumba
+from tardis.opacities.opacity_state_numba_iip import OpacityStateNumbaIIP
+from tardis.transport.frame_transformations import (
+    get_inverse_doppler_factor,
+)
+from tardis.transport.montecarlo.configuration.base import (
+    MonteCarloConfiguration,
+)
+from tardis.transport.montecarlo.estimators.estimators_bulk import (
+    EstimatorsBulk,
+)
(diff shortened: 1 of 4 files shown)


tardis/transport/montecarlo/modes/homologous_packet_propagation.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+        rpacket_tracker.track_escattering_interaction_before(r_packet)
+        thomson_scatter(
+            r_packet,
+            time_explosion,
+            enable_full_relativity,
+        )
+        rpacket_tracker.track_escattering_interaction_after(r_packet)
+    else:
+        return False
```
jvshields, line 162: For the sake of discussion, I would at least want a function like this to be more clear why we're returning this bool and what it's meant to be used for downstream. Like a comment saying "Packet interaction type not in common processes. Will be handled elsewhere."

But now that I type that out, I feel like I would rather do the check on a higher level outside of the handler function. You would do the type check in the caller function, and only call "handle_common_interaction" if you're one of the common types, so you don't have to do the bool return.

tardis/transport/montecarlo/modes/homologous_packet_propagation.py
```diff
@@ -0,0 +1,197 @@
...
+    RPacket,
+)
+from tardis.transport.montecarlo.packets.trackers.tracker_full import (
+    TrackerFull,
+)
+from tardis.transport.montecarlo.packets.trackers.tracker_last_interaction import (
+    TrackerLastInteraction,
+)
+
+HomologousOpacityState = OpacityStateNumba | OpacityStateNumbaIIP
```
jvshields, line 51: This is certainly a choice I haven't seen before. Is it mostly for the type hinting? Do any of these variables ever get set by something beyond the or?

andrewfullard: It's just shorthand for the type hints, yeah

jvshields: I guess I don't hate it, but I'd probably want it to be used in multiple places or it's just space inefficient.

andrewfullard:
Closing, we are not going with this approach

PR #3708, opened 2026-08-26
TARDIS HE bug fixes

andrewfullard:
### :pencil: Description

**Type:** :beetle: `bugfix` 

There were a number of subtle bugs in the HE transport module. This PR addresses them:

1. Electron density calculation was ignoring elements defined without their atomic number in a composition, underestimating the electron density for those shells. Fix: the supplied atomic data is applied to all elements and isotopes as needed.
2. Shell boundary calculations did not correctly check the inner shell boundary, and packets were deleted if they crossed the inner simulation boundary. Fix: packets can cross the inner simulation boundary and exit out the other side. Most HE models should not have an inner simulation boundary anyway.
3. Incorrect frame transformations of total opacity, leading to reduced pair creation probability. Fixed.
5. Incorrect frame transformation of Compton scattering angle. Fixed.
6. ARTIS pair creation opacity used the wrong energy scale. We don't usually use this, but now it is correct.
7. Positronium sampling should have used E dN/dE instead of just dN/dE. Fixed.

Based on #3706 

Requires https://github.com/tardis-sn/tardis-regression-data/pull/116

Minimal summed escaping packet changes despite all these fixes. The comparison is based on the regression data from #3706.

image

- [ ] Testing pipeline
- [ ] Other method (describe)


Knights-Templars:
Do these bugs affect the gamma-ray spectrum at a particular epoch? If so, how much is that difference?

tardis/energy_input/samplers.py
```diff
@@ -144,17 +144,19 @@ def sample_decay_time(
...
+
+    def __init__(self, n_grid: int = 10000) -> None:
+        """Initialize the energy-weighted sampling grid.
+
         Parameters
         ----------
         n_grid : int, optional
-            Number of grid points for the CDF, by default 1000
+            Number of grid points for the CDF, by default 10000
         """
```
Knights-Templars, line 156: The Ore Powell spectrum is itself an energy spectrum. So, in the individual packet scheme, it should be sampled from F(E) and not another E.

Knights-Templars: F(E) \equiv EdN/dE ?

andrewfullard: Yeah, because F(dN/dE) is not equal to F(E)

tardis/energy_input/transport/gamma_ray_interactions.py
```diff
@@ -231,17 +233,19 @@ def compton_scatter(photon, compton_angle):
...
+    assert np.abs(norm_theta - np.cos(compton_angle)) < 1e-8, (
+        "Error, difference between new vector angle and Compton angle is more than 0!"
+    )
 
     # Calculate the angle aberration of the final direction
     final_direction = angle_aberration_gamma(
-        final_compton_scattered_vector, photon.location, photon.time_start
+        final_compton_scattered_vector,
+        packet.location,
+        -1 * packet.time_start,
```
jvshields: Where did this -1 come from? Should it be handled in the function if it was wrong before? A bit weird and unintuitive for this to appear in the function call.

andrewfullard: It creates the inverse angle aberration. We could add it as an option, or a duplicate function similarly to the Doppler factor, but this does the same thing.

tardis/energy_input/samplers.py
andrewfullard: Ore and Powell’s spectrum describes the one-photon number distribution for ortho-positronium decay,
image

Our packets represent photons equal to 1/E, so we need to multiply the spectrum by E.

andrewfullard:
> Do these bugs affect the gamma-ray spectrum at a particular epoch? If so, how much is that difference?

I don't think the effect is very strong given the summed result. Here's the epoch with the most signal for the tests:

image
image

It seems to mostly be noise. This is the binned version:

image

tardis/energy_input/transport/gamma_ray_interactions.py
andrewfullard: Added the option since Anirban pointed it out as well

jvshields (dismissed):

andrewfullard:
High-resolution simulation. There is an increase in total escaped energy towards the lower energies.

image

image

Knights-Templars:
If we multiply by another `E`, then the spectrum with fp=1.0 will be harder in the lower energy region. 

An interesting test would be the spectral changes with fp=0.0, if that is not already tested.

Knights-Templars:
> High-resolution simulation. There is an increase in total escaped energy towards the lower energies.
> 
> image image

Why does the figure title says summed over 499 epochs? 
Is it the spectrum generated at 499 days since explosion?

andrewfullard:
> Why does the figure title says summed over 499 epochs? Is it the spectrum generated at 499 days since explosion?

It's the sum of all escaping packets over the entire time (499 time steps), not the spectrum from a specific time, to better show the difference between before/after.

> If we multiply by another E, then the spectrum with fp=1.0 will be harder in the lower energy region.
>
> An interesting test would be the spectral changes with fp=0.0, if that is not already tested.

Yes, there is a small change in the low-energy continuum because of this. I will generate a fp=0.0 spectrum to show the difference.

andrewfullard:
The high resolution spectrum earlier was actually the fp=0.0 case. This is fp=1.0:
image
image

Knights-Templars:
So, whatever changes have been made in the PR is mostly effecting the low energy part of the spectrum upto like 10% while the high energy part remains mostly unchanged despite all the bug fixes.

andrewfullard:
> So, whatever changes have been made in the PR is mostly effecting the low energy part of the spectrum upto like 10% while the high energy part remains mostly unchanged despite all the bug fixes.

Yep! The positronium spectrum is the big change. The rest of the fixes will only affect specific models which we haven't tested.

andrewfullard:
@Knights-Templars are you happy with these changes?

Knights-Templars:
I am happy with the changes. 

Thanks for the work. 
It will help my current work.

PR #3710, opened 2026-08-27
TARDIS HE estimators

andrewfullard:
### :pencil: Description

**Type:** :rocket: `feature` 

This adds the Kasen et al. 2006 deposition rate estimators as an output of the gamma-ray transport, and fixes the Kasen photoabsorption opacity approximation so that it can be used if desired. 

Requires https://github.com/tardis-sn/tardis-regression-data/pull/120

- [x] Testing pipeline
- [ ] Other method (describe)


docs/new_developers/nebular_phase/positronium.rst
```diff
@@ -4,7 +4,7 @@ Positronium
...
 
-The mean lifetime of Ps is very short, on the order of nanoseconds. It is most likely to produce 2 or 3 gamma-ray photons depending on if it is para-Ps (2 photons of 511 keV each) or ortho-Ps (3 photons). The Ps types are dependent on the spin state of the electron and so occur in a 1/4 (para-Ps, S = 0, Ms = 0) to 3/4 (ortho-Ps, S = 1, Ms = −1, 0, 1) ratio. The triplet emission produces a continuum of photon energies described in :cite:`Ore1949`.
+The mean lifetime of Ps is very short, on the order of nanoseconds. It is most likely to produce 2 or 3 gamma-ray photons depending on if it is para-Ps (2 photons of 511 keV each) or ortho-Ps (3 photons). The Ps types are dependent on the spin state of the electron and so occur in a 1/4 (para-Ps, S = 0, Ms = 0) to 3/4 (ortho-Ps, S = 1, Ms = −1, 0, 1) ratio. The triplet emission produces a continuum of photon energies described in :cite:`Ore1949`. Because TARDIS uses equal-energy packets, it samples this continuum from the energy distribution :math:`E\,dN/dE`, rather than directly from the photon-number distribution :math:`dN/dE`.
```
Knights-Templars, line 7: OrePowell 1949 call it an `energy_spectrum` F(x) - or the probability of a quanta to have energy between k to k+dk (where `k=hnu`). If I understand it correctly - indivisible packet method of Lucy needs an energy spectrum [where an energy spectrum is the amount of energy radiated in a particular energy bin] or the probability distribution.

tardis/energy_input/transport/GXPacket.py
```diff
@@ -68,7 +64,10 @@ def __init__(
         self.time_start = time_start
         self.time_idx = time_idx
         # TODO: rename to tau_event
-        self.tau = -np.log(np.random.random())
+        if initialize_tau:
```
Knights-Templars, line 67: At `make_gx_packet` and also on the transport we are getting an `tau_event` so, why this hack is necessary?

tardis/energy_input/transport/gamma_packet_loop.py
```diff
@@ -26,17 +32,75 @@
...
     photoabsorption_opacity_calculation,
+    photoabsorption_opacity_calculation_kasen,
+)
+from tardis.transport.montecarlo import njit_dict, njit_dict_no_parallel
+from tardis.transport.montecarlo.modes.montecarlo_transport import (
+    update_packet_progress,
 )
-from tardis.transport.montecarlo import njit_dict_no_parallel
+
+COMPTON_QUADRATURE_MU, COMPTON_QUADRATURE_WEIGHT = (
```
Knights-Templars, line 42: Why `64` and not 32 or 128? Is it a reasonable value?

tardis/energy_input/transport/gamma_packet_loop.py
```diff
@@ -212,12 +350,30 @@ def gamma_packet_loop(
...
+            )
+
+            # Kasen et al. (2006), equation A10. Pair creation is omitted as
+            # in the original estimator; it can be generalized separately.
+            deposition_opacity = (
+                compton_opacity
+                * get_compton_energy_loss_fraction(comoving_energy)
+                + photoabsorption_opacity
+            )
+            energy_deposition_estimator_thread[
```
Knights-Templars, line 369: This is in ergs.

tardis/energy_input/transport/gamma_packet_loop.py
```diff
@@ -212,12 +350,30 @@ def gamma_packet_loop(
...
             )
 
-            distance = min(distance_interaction, distance_boundary, distance_time)
+            distance = min(
+                distance_interaction, distance_boundary, distance_time
+            )
+
+            # Kasen et al. (2006), equation A10. Pair creation is omitted as
+            # in the original estimator; it can be generalized separately.
+            deposition_opacity = (
```
Knights-Templars, line 364: Since, we are not comparing directly against Kasen's results, how about the pair creation opacity? 
Maybe part of a different PR.

tardis/energy_input/transport/gamma_ray_interactions.py
```diff
@@ -231,17 +233,19 @@ def compton_scatter(photon, compton_angle):
...
+    )
 
-    assert (
-        np.abs(norm_theta - np.cos(compton_angle)) < 1e-8
-    ), "Error, difference between new vector angle and Compton angle is more than 0!"
+    assert np.abs(norm_theta - np.cos(compton_angle)) < 1e-8, (
+        "Error, difference between new vector angle and Compton angle is more than 0!"
+    )
 
     # Calculate the angle aberration of the final direction
```
Knights-Templars, line 244: A clearer code for understanding would be use an argument inverse=True inside the angle_aberration function, and not reversing time explicitly. Since, this is giving impression the packet is traveling backward in `time`, while we want -v

tardis/energy_input/transport/gamma_ray_grid.py
```diff
@@ -76,24 +90,25 @@ def calculate_distance_radial(photon, r_inner, r_outer):
     shortest = min(distance_list)
     shell_change = 1
```
Knights-Templars, line 91: shortest = min(distance_list)

if shortest in (inner_1, inner_2):
    shell_change = -1
elif shortest in (outer_1, outer_2):
    shell_change = 1
else:
    raise some error!!

Is it possible to make this explicit?
I am not sure if the PR version is better optimized, but it will be great for understanding maybe?

tardis/energy_input/main_gamma_ray_loop.py
```diff
@@ -270,64 +261,27 @@ def run_gamma_ray_loop(
...
     isotope_positron_fraction = legacy_calculate_positron_fraction(
         legacy_isotope_decacy_df,
         packet_collection.source_isotopes,
         number_of_packets,
     )
-    for i, p in enumerate(packets):
-        total_energy[p.shell, p.time_idx] += (
-            isotope_positron_fraction[i] * energy_per_packet
-        )
+    for i in range(number_of_packets):
```
Knights-Templars, line 281: Since, this function is not jit decorated, this is a bottleneck, so can this binned using a numpy function using `isotope_positron_fraction * energy_per_packet` as weights? 
This is not a new line, but the PR is big enough that maybe this can be accommodated here? 

I wrote something similar before to make this faster locally - but have forgotten since.

Knights-Templars (commented):
Thanks for putting the effort.

tardis/energy_input/transport/gamma_ray_interactions.py
andrewfullard: Addressed in #3708

tardis/energy_input/transport/gamma_ray_grid.py
andrewfullard: Addressed in #3708

tardis/energy_input/transport/gamma_packet_loop.py
andrewfullard: Different PR I think. Pair creation is typically a very small contributor.

tardis/energy_input/transport/gamma_packet_loop.py
andrewfullard: Yes

tardis/energy_input/transport/gamma_packet_loop.py
andrewfullard: 64 gets to within 2e-14 relative error for the angular distribution compared to 256 points, so it's basically equivalent to that higher value. 32 is 5e-9 relative error. Since it is precomputed, I don't think we need to go lower, but we also don't get benefit from higher.

PR #3711, opened 2026-08-28
fix(visualization): align density array length in CustomAbundanceWidgetData.from_yml

### LINKED ISSUE #3666: Error when adding new shells in custom abundance widget
**Describe the bug**

At some point(possibly due to changes in the way density is parsed from config files), we lost the last density value in `CustomAbundanceWidget.data.density`. In the plot of a custom abundance widget generated from a yaml file, the last density point is missing. This also causes the following error in the interpolation function within `on_btn_add_shell()` method when new shells are added to the widget- `ValueError: fp and xp are not of the same length.` . 
This does not happen when the widget is constructed from a csvy file.

**To Reproduce**

When custom abundance widget is created using `tardis_example.yml`-
`len(CustomAbundanceWidget.data.density)`=20 (should be 21)
`len(CustomAbundanceWidget.data.velocity)`=21
number of shells=20

In the resulting graph the last density point is missing-
Image

**Screenshots**

How the plot is supposed to look-

Image

  
**Additional context**

Example notebook- [how_to_abundance_widget.ipynb](https://github.com/user-attachments/files/30305023/how_to_abundance_widget.ipynb)

yml file- [tardis_example.yml](https://github.com/user-attachments/files/30305037/tardis_example.yml)

sundeep8967:
### Problem
When constructing `CustomAbundanceWidget` from a YAML file via `from_yml()`, `simulation_state.density` contains $ shell values whereas `simulation_state.velocity` has +1$ boundaries. In `CustomAbundanceWidget.from_csvy()`, the density array is sized to match velocity, but `from_yml()` passed the raw 569Xlength density array. As a result:
1. The last density point was missing in the plot.
2. Clicking **Add Shell** in `on_btn_add_shell()` triggered `ValueError: fp and xp are not of the same length` during interpolation.

### Solution
In `CustomAbundanceWidgetData.from_yml()`, check if `len(density) == len(velocity) - 1` and append `density[-1]` to ensure equal lengths across `velocity` and `density`, matching the expected contract across all widget loaders.

Fixes #3666

### DIFF
--- a/tardis/visualization/widgets/custom_abundance.py
+++ b/tardis/visualization/widgets/custom_abundance.py
@@ -215,6 +215,8 @@ class CustomAbundanceWidgetData:
         velocity = simulation_state.velocity
         density_t_0 = simulation_state.time_explosion
         density = simulation_state.density
+        if len(density) == len(velocity) - 1:
+            density = np.insert(density, 0, density[0]) * density.unit
         abundance = simulation_state.abundance
         isotopic_mass_fraction = (
             simulation_state.composition.isotopic_mass_fraction
@@ -293,6 +295,8 @@ class CustomAbundanceWidgetData:
         velocity = sim.simulation_state.velocity
         density_t_0 = sim.simulation_state.time_explosion
         density = sim.simulation_state.density
+        if len(density) == len(velocity) - 1:
+            density = np.insert(density, 0, density[0]) * density.unit
 
         return cls(
             density_t_0=density_t_0,


connor-mcclellan:
Thanks for your contribution. Please let us know if you're planning to apply for Google Summer of Code and structure your PR according to our guidelines.

`SimulationState.density` contains one value per physical shell, so for 20 shells it has 20 values. `SimulationState.velocity` contains 21 shell boundaries. The widget was written for CSVY data containing an unused zeroth density value plus 20 shell values, so it expects 21 density values but intentionally skips the first one. Appending density[-1] just duplicates the outer-shell density, but since the _zeroth_ index is skipped, this fix just effectively shifts all shell values by one position: shell 1 displays shell 2, and the final point is repeated.

sundeep8967:
thanks for the explanation! understood clearly now regarding the skipped zeroth index in the widget. updated the implementation to insert a placeholder at index 0 via `np.insert(density, 0, density[0])` so all shell indices align properly without any shift.

connor-mcclellan:
We don't have the capacity to review Google Summer of Code pull requests at the moment, so we are closing this. Please feel free to re-open it when we announce our participation in GSoC.

sundeep8967:
Hi @connor-mcclellan, thanks for letting me know!

Just wanted to clarify, I am not applying for GSoC. I am a beginner trying to learn the codebase and saw the issue in #3666 while trying out the project, so I wanted to see if I could fix it.

Totally understand if you don't have time to review this right now. If there are any other smaller or beginner-friendly issues that you need help with instead, please let me know and I'd be happy to work on them!

PR #3714, opened 2026-08-31
Example refactor of transport with a physics-driven approach

andrewfullard:
Compare and contrast with #3707. This approach splits transport up by physics approach. There are still wrappers involved, but they tend to be a bit more obvious. This probably makes it easier to mix and match options.


andrewfullard:
Closing since we are going with this approach

PR #3727, opened 2026-09-08
Equilibrium solver cleanup

andrewfullard:
### :pencil: Description

**Type:**  :roller_coaster: `infrastructure`

Based on #3704, this PR cleans up a lot of excess code and performs some refactoring.

### :vertical_traffic_light: Testing

How did you test these changes?

- [ ] Testing pipeline
- [ ] Other method (describe)
- [ ] My changes can't be tested (explain why)

### :ballot_box_with_check: Checklist

- [ ] I requested two reviewers for this pull request
- [ ] I updated the documentation according to my changes
- [ ] I built the documentation by applying the `build_docs` label

> **Note:** If you are not allowed to perform any of these actions, ping (@) a contributor.

PR #3728, opened 2026-09-08
Opacity state refactor

andrewfullard:
### :pencil: Description

Based on #3727. Splits up some of the large data classes, the opacity state, and refactors the equilibrium solution further.

- [ ] Testing pipeline
- [ ] Other method (describe)


tardis/plasma/equilibrium/evaluator.py
```diff
@@ -240,8 +240,8 @@ def calculate_nlte_level_population_residual(
 
 
 @dataclass(frozen=True)
-class _LevelSolveState:
-    """Intermediate level results shared by the optional final stages."""
+class _CandidateThermalState:
```
jvshields: I think classes being demarcated as private with underscores is a bridge too far for me to stomach. I can sort of see trying to signal that this is an intermediate piece for the evaluator, but I think if that's the case the name should just say that. Like "IntermediateEvaluatorFrozenThermalState" which is long but makes it clear what the purpose of it is.

tardis/plasma/equilibrium/evaluator.py
```diff
@@ -250,6 +250,12 @@ class _LevelSolveState:
     continuum_coefficients: ContinuumCoefficientState
     thermal_partition_function: pd.DataFrame
     thermal_level_boltzmann_factor: pd.DataFrame
+
+
+@dataclass(frozen=True)
+class _LevelSolveState:
```
jvshields: Similar comment here, though I think the name is better than the last one. LevelSolvedEvaluatorState maybe.

tardis/plasma/properties/hydrogen_continuum.py
jvshields, line 1: Love to see this one go.

tardis/transport/montecarlo/modes/iip/interaction_event_callers.py
```diff
@@ -0,0 +1,191 @@
...
+        velocity, r_packet.mu, enable_full_relativity
+    )
+    r_packet.mu = get_random_mu()
+    inverse_new_doppler_factor = get_inverse_doppler_factor(
+        velocity, r_packet.mu, enable_full_relativity
+    )
+    comov_energy = r_packet.energy * old_doppler_factor
+    r_packet.energy = comov_energy * inverse_new_doppler_factor
+
+    if line_interaction_type == LineInteractionType.SCATTER:
```
jvshields: Double check this, but I think this code block might be able to be removed. If memory serves, the SCATTER flag is only for when you bypass the macroatom entirely from config setup. I would argue that makes absolutely no sense in the IIP mode. Nobody should ever be using the line scattering approximation at the same time as continuum stuff, so we shouldn't handle the case. Maybe raise an error here if necessary, but you might be able to just get rid of the if block entirely.

tardis/transport/montecarlo/modes/iip/solver.py
```diff
@@ -126,6 +124,10 @@ def initialize_transport_state(
         opacity_state_numba = opacity_state.to_numba(
             macro_atom_state,
             self.line_interaction_type,
+            continuum_processes_enabled=True,
+        )
+        continuum_state_numba = continuum_state.to_numba(
+            opacity_state.t_electrons, macro_atom_state
         )
         # opacity_state_numba = opacity_state_numba[
```
jvshields: This comment block can just be removed now, assuming the slicing fix PR covers it and we don't need to keep it?

tardis/transport/montecarlo/modes/nonhomologous/interaction_event_callers.py
```diff
@@ -1,7 +1,5 @@
-import numpy as np
```
jvshields, line 1: Being that, at some point we probably want to actively support nonhomology with continuum, I'd at least leave a comment pointing to the IIP mode interaction_event_callers.py file saying that, to add continuum interactions, take a look at that file.

tardis/transport/montecarlo/interaction_event_callers.py
```diff
@@ -222,11 +83,9 @@ def line_scatter_event(
             opacity_state,
             enable_full_relativity,
         )
-    else:  # includes both macro atom and downbranch - encoded in the transition probabilities
-        comov_nu = r_packet.nu * old_doppler_factor  # Is this necessary?
-        r_packet.nu = (
-            comov_nu * inverse_new_doppler_factor
-        )  # NOTE - this should get overwritten by macro_atom_event, but makes tests fail potentially from different rng
```
jvshields, line 229: Not sure I want this comment to get quietly erased since we never got to the bottom of it. It's pretty weird and worth investigating.

tardis/transport/montecarlo/interaction_event_callers.py
jvshields, line 1: So is this file implicitly just handling classic montecarlo? I vaguely wonder if it should be in a modes/classic subdirectory or something, but I'm just bringing that up as food for thought. Doesn't have to be done in this PR and I'm not sold on the idea.

andrewfullard: Yeah after revisiting this PR, there are a lot of MC changes that probably aren't relevant.

andrewfullard: Never mind, these changes are necessary due to the opacity state split (things start breaking when I bring them back)

tardis/transport/montecarlo/interaction_events.py
```diff
@@ -107,7 +114,7 @@ def bf_cooling(r_packet, time_explosion, opacity_state, enable_full_relativity):
     # Josh: I don't think we need to do this - BF cooling already picks one
```
connor-mcclellan, line 114: Ooh, an ancient Josh comment? What's the word on this?

connor-mcclellan (approved):
Looks good, pending responses to Josh's comments. The PR review skill didn't identify any areas needing scientific attention either.

tardis/transport/montecarlo/interaction_events.py
jvshields: Ancient is strong. I still think this is true and would be nice to change but hasn't been revisited.

jvshields (approved):

PR #3641, opened 2026-07-07
Convert .values to .to_numpy() in energy_source.py

Angel-Gupta037:
## What does this PR do?
Converts pandas` .values` calls to `.to_numpy()` in energy_source.py

## Why?
As recommended by pandas documentation, `.to_numpy()` is more explicit 
about the type returned and is the preferred method over `.values`

## Changes
- line 63: .values →` .to_numpy()`
- line 64: .values → `.to_numpy()  `
- line 192: .values[0] →` .to_numpy()[0]`

see #3603

Issue #3603, opened 2026-06-09
Convert all dataframe .values to .to_numpy


andrewfullard:
This is the recommended method from Pandas and is more explicit about the type returned.

ali-john:
I would like to work on this.

krish-kedia:
hi, I'd like to work on this issue as my first open-source contribution. I'll go through the codebase, find all instances of .values used on pandas DataFrames/Series, and replace them with .to_numpy(), making sure the behavior stays the same at each call site.

Angel-Gupta037:
Hi! I'd like to work on this issue. I'll start converting the `.values` calls to `.to_numpy()` and submit a PR soon.

Angel-Gupta037:
Hi! I've submitted PR #3641 which converts `.values` to `.to_numpy()` in energy_source.py as a start. Happy to continue with other files based on feedback!

PR #3739, opened 2026-09-16
Use .to_numpy() instead of .values in grid and composition

Phantom8420:
### Description

Small piece of #3603 (convert `.values` to `.to_numpy()`). Covers two files that weren't touched by the earlier merged PR (#3641):

- `tardis/grid/base.py`
- `tardis/model/matter/composition.py`

`.values` is legacy pandas API; `.to_numpy()` is the recommended replacement and behaves identically here.

Fixes part of #3603

### Testing

No local conda/tardis environment available in this session, so I did not run the test suite. The change is a like-for-like API swap with no behavior difference for these call sites.


connor-mcclellan:
We don't have the capacity to review Google Summer of Code pull requests at the moment, so we are closing this. Please feel free to re-open it when we announce our participation in GSoC.

PR #3740, opened 2026-09-16
Correct ordering of optical depth calculation for v_inner workflow

andrewfullard:
### :pencil: Description

**Type:** :beetle: `bugfix`

The integrated optical depth was only being calculated with the previous iteration's grid and that misaligned when the grid changed shape. This adds specific re-calculation in and a test to catch it.

Needs https://github.com/tardis-sn/tardis-regression-data/pull/121

- [x] Testing pipeline
- [ ] Other method (describe)


tardis/workflows/v_inner_solver.py
```diff
@@ -110,7 +108,6 @@ def store_plasma_state(
         electron_densities,
         t_inner,
         v_inner_boundary,
-        tau_integ,
```
jvshields, line 113: Maybe worth checking upstream in the workflow in front of where store_plasma_state() is called to make sure tau_integ isn't still be calculated and unused now that it doesn't have to be fed into this function.

andrewfullard: It's still calculated intentionally in estimate_v_inner so that the previous grid can be used to find the new inner velocity.

jvshields (approved):

connor-mcclellan (approved):

connor-mcclellan:
Just restarted jobs after the regression data merge but it failed to grab the updated data from Moria (possibly related to the power outage?)

